PR #2, opened 2022-03-04
PPM protocol messages with serialization/deserialization support.

BranLwyd:
This matches what is in the draft-interop-target branch of the PPM
specification at d8c6677, and was heavily inspired by the message
definitions throughout the ppm-prototype.

### DIFF
--- a/janus_server/src/message.rs
+++ b/janus_server/src/message.rs
@@ -1,0 +1,1378 @@
+//! PPM protocol message definitions with serialization/deserialization support.
+
+use anyhow::anyhow;
+use num_enum::TryFromPrimitive;
+use prio::codec::{
+    decode_u16_items, encode_u16_items, CodecError, Decode, Encode, ParameterizedDecode,
+    ParameterizedEncode,
+};
+use ring::{
+    digest::SHA256_OUTPUT_LEN,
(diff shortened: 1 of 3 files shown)


janus_server/src/message.rs
```diff
@@ -0,0 +1,1378 @@
...
+    pub encapsulated_context: Vec<u8>,
+    /// An HPKE ciphertext.
+    pub payload: Vec<u8>,
+}
+
+impl Encode for HpkeCiphertext {
+    fn encode(&self, bytes: &mut Vec<u8>) {
+        self.config_id.encode(bytes);
+        encode_u16_items(bytes, &(), &self.encapsulated_context);
+        encode_u16_items(bytes, &(), &self.payload);
```
tgeoghegan, line 155: Ah, we should probably introduce variants of `encode_$SIZE_items` generic over `Encode` that omit the decoding parameter. That's not needed right here though.

tgeoghegan (approved):

janus_server/src/message.rs
BranLwyd: Indeed: I left those functions for later with the PR to split parameterized from non-parameterized encoding/decoding, but with this code change it seems like it would be useful.

PR #5, opened 2022-03-07
Generate "test vectors" for all PPM messages.

BranLwyd:
These vectors encode byte-for-byte what each test message should encode
to. To mitigate against the encoded messages reflecting bugs in our
existing code or dependencies, I generated these encodings mostly by
hand (using the PPM spec & a hex converter). The exception is the HMAC
tags, which were computed using the same ring library as used by the
real code; I think the risk of a bug in ring here is acceptable.

### DIFF
--- a/janus_server/Cargo.toml
+++ b/janus_server/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1"
+hex = "0.4.3"
 lazy_static = "1"
 num_enum = "0.5.6"
 prio = { git = "https://github.com/abetterinternet/libprio-rs", rev = "7551ba5352794d5ff89d70f886f73cd5020d3fe9" } # TODO: use a specific version number once this is possible
--- a/janus_server/src/message.rs
+++ b/janus_server/src/message.rs
@@ -852,27 +852,29 @@ mod tests {
         );
     }
 
-    // TODO(brandon): create "test vectors" for all test cases
-
-    fn roundtrip_encoding<T>(vals: &[T])
+    fn roundtrip_encoding<T>(vals_and_encodings: &[(T, &str)])
     where
         T: Encode + Decode + core::fmt::Debug + Eq,
     {
-        for val in vals {
+        for (val, hex_encoding) in vals_and_encodings {
             let mut encoded_val = Vec::new();
             val.encode(&mut encoded_val);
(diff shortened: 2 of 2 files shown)


BranLwyd:
I'm curious what folks think about how I lay out the test encodings; right now each encoding is on its own line, often scrolling off to the right, and very opaque to would-be readers. I'm curious what folks think would be best:

1. Kept as-is. (I guess this is a vote that everything else is just as opaque as the "easy" way.)
2. Escaped onto multiple lines at some sane column width. This keeps things from scrolling off to the right, but consumes more vertical space with opaque data.
3. Escaped onto multiple lines, in some semantically-meaningful fashion? The "obvious" choice would be to do each field on its own line, indenting into substructures. But vectors sorta break this: it's not clear how their lengths should indent. And sadly, multiline strings can't contain comments, so it'd still be up to the reader to interpret these encodings.

I'm happy to do whatever folks think is best, but I wanted to get opinions before I spent a bunch of time formatting.

divergentdave:
If we want to do option 3, we could use concat!(), and then interleave string literals and comments. Plus, we could nest concat!() calls to get different indentations in a rustfmt-compliant way. I could take it from here and make those formatting changes, if you like.

tgeoghegan:
Another choice would be to encode the test vectors in a JSON file and have the tests load them. This would make it easy to then share the test vectors with other projects that want to verify the correctness of their message handling. Plus, if the PPM spec ever publishes message test vectors, they're likely to be in JSON (see for instance the [prio3 VDAF test vectors](https://github.com/abetterinternet/libprio-rs/tree/main/src/vdaf/testdata) which correspond to the [test vectors in the VDAF spec](https://github.com/cjpatton/vdaf/blob/main/draft-patton-cfrg-vdaf.md#test-vectors-test-vectors)).

tgeoghegan (approved):
I think this change is good as-is, though I have a slight preference for wrapping the test vectors at 80 or 100 columns.

BranLwyd:
I like the concatenation-based approach; it makes things more readable, and more importantly, much more maintainable (adding a new field will no longer require regenerating the whole test vector, or parsing the existing test vector well enough to figure out where the new field's bytes should be added).

I'll leave a JSON-based solution for later; I suppose I don't want to introduce a second serialization format for these messages, but nothing would stop us from doing so in the future.

PR #7, opened 2022-03-08
terraform: deploy echo container into kind cluster

tgeoghegan:
This commit adds a basic Terraform module that deploys a Kubernetes
service running `hashicorp/http-echo` into a Kubernetes cluster. The
intent is to support deploying into a `kind` cluster using local
Terraform state storage for dev and CI, and clusters hosted in GKE with
remote Terraform state storage in GCS for staging and prod environments.
We crib enough of the `Makefile` from `prio-server` to achieve this,
with some modifications to `chdir` into the appropriate directory to use
the `kind` or cloud specific `main.tf`.

### DIFF
--- a/terraform/.gitignore
+++ b/terraform/.gitignore
@@ -1,0 +1,4 @@
+# Terraform
+terraform.tfstate*
+.terraform/
+crds.yml
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -1,0 +1,61 @@
+# Janus Terraform module
+
+We use these Terraform modules to manage deployments of Janus to cloud and local dev environments. We provide a distinct main module for each so that we can use different [Terraform backends](https://www.terraform.io/language/settings/backends) and configure other modules separately.
+
+A Janus environment `foo` is defined by a file `variables/foo.tfvars` which contains configuration for that environment. Among other behaviors, the variables files specify whether the environment is deployed to Google Kubernetes Engine (GKE) or to a local `kind` cluster. To see what Terraform wants to do to environment foo, run:
+
+    ENV=foo make plan
+
+To apply those changes, run:
+
+    ENV=foo make apply
+
+You may need additional environment variables or other context to interact with an environment via Terraform. See [Authentication for Terraform](https://github.com/abetterinternet/isrg-cloud-bootstrap#authentication-for-terraform) for details, as well as the sections below relevant to the environment you are working with.
+
+## Google Kubernetes Engine
+
+Staging and production environments of Janus are deployed into [Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine). This uses `main_gcp/main.tf`. Variables files for GKE environments should include the variable `state_bucket`, whose value should be the name of a GCS bucket to use with the [`gcs` Terraform backend](https://www.terraform.io/language/settings/backends/gcs).
+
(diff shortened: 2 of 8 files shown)


tgeoghegan:
This PR raises an important question: do we want to keep Terraform modules and variables files in the same repository as the implementation of Janus itself? We could also create a separate repository called `janus-ops` and keep all that there, which we could then keep private.

I wanted to stop before the PR got too big but obviously there's a lot of missing pieces. I wrote up #7 to track what I want to do next.

terraform/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Terraform
+terraform.tfstate*
```
divergentdave, line 2: When I ran this locally, I got `main_local/local-cluster.tfstate` and `main_local/local-cluster.tfstate.backup`.
```suggestion
*.tfstate
*.tfstate.backup
```

divergentdave (approved):
LGTM, modulo nits. Could you generate and check in `main_local/.terraform.lock.hcl` and `main_gcp/.terraform.lock.hcl` as well?

I have an idea for how to make invoking Terraform from outside of the Makefile safer; I'll write it up separately in an issue.

terraform/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+[TODO: adapt prio-server/cluster-bootstrap to bringup a GKE cluster for Janus and document how to do that here]
+
+## Local development with Kind
+
+To enable local development and testing in CI, we use [`kind`](https://kind.sigs.k8s.io/) (Kubernetes IN Docker). Kind uses Docker containers to run a single-node Kubernetes control plane and optionally additional worker nodes. This uses `main_local/main.tf`. Variables files for local environments may include the `tfstate_path` variable, whose value is used as the `path` value in the [`local` Terraform backend](https://www.terraform.io/language/settings/backends/local).
+
+You will need the `kind` to manage clusters. See [the `kind` documentation](https://kind.sigs.k8s.io/docs/user/quick-start/#installation) for instructions on installing. The quick start guide also has instructions on [creating a cluster](https://kind.sigs.k8s.io/docs/user/quick-start/#creating-a-cluster). By default `kind` creates a cluster where a single node hosts both the control plane and workers. If you want to create a more complex cluster topology, you can use a configuration file like this:
```
BranLwyd, line 25:
```suggestion
You will need to install `kind` to manage clusters. See [the `kind` documentation](https://kind.sigs.k8s.io/docs/user/quick-start/#installation) for instructions on installing. The quick start guide also has instructions on [creating a cluster](https://kind.sigs.k8s.io/docs/user/quick-start/#creating-a-cluster). By default `kind` creates a cluster where a single node hosts both the control plane and workers. If you want to create a more complex cluster topology, you can use a configuration file like this:
```

BranLwyd (changes requested):
I would like to keep our operational configuration separate from our code, and private, IMO. I don't want to get into a situation similar to ENPA where other orgs attempt to use ISRG's Terraform directly (and then configurations diverge, since TF's backwards-compatibility story hasn't always been great, and we don't always think about it since it's "our" TF). And we can more freely list e.g. team-member emails etc without needing to worry as much about privacy concerns, as you've noted.

Count this as a vote for a `janus_ops` repository.

terraform/README.md
tgeoghegan: > Count this as a vote for a `janus_ops` repository.

SGTM! I will close this PR and will create a new, private repo for Terraform and config.

Issue #12, opened 2022-03-10
HPKE support


tgeoghegan:
I'm going to borrow code from ppm-prototype to get HPKE message exchange working.

- [x] Construct HPKE sender and receiver in memory and transmit a message in a unit test (#14)
- [x] Put together an aggregator binary that exposes an `hpke-config` endpoint compliant with the PPM spec (#15)
- [x] Once we have figured out configuration (#20), provide HPKE configurations and the address/port to listen on to `aggregator` instead of hard-coding them. (#37)

tgeoghegan:
After some discussion in #37, we've agreed to provide HPKE private keys in the database, encrypted under a key that we will provide to running containers. This intersects with the question of whether the same HPKE config should be used for all tasks (my 2c: no), which means we need to address ppm-specification/#146.

- [ ] hash out https://github.com/abetterinternet/ppm-specification/issues/146
- [ ] Wire up secret encryption key in k8s secrets and use it to obtain HPKE private keys from database

tgeoghegan:
I don't think there's anything outstanding to do here.

PR #14, opened 2022-03-10
janus_server: Add basic HPKE support

tgeoghegan:
Adds an `hpke` module that can seal messages to a public key and open
them with the corresponding private key. The code is largely cribbed
from `ppm-prototype`, but I stripped out the support for loading HPKE
config from JSON files because we still need to discuss how we want
Janus to handle this. I also made a point of distinguishing the structs
that are responsible for encrypting and decrypting from the wire
messages defined in `mod message`. We are currently using crate `hpke`,
which is awkward to use when you are using dynamic configuration to
figure out what set of HPKE algorithms to use. In this PR, I've wired up
support for two sets of algorithms to illustrate how we might go about
plumbing all the HPKE setups we need.

Part of #12

### DIFF
--- a/janus_server/src/hpke.rs
+++ b/janus_server/src/hpke.rs
@@ -1,0 +1,382 @@
+//! Encryption and decryption of messages using HPKE (RFC 9180).
+
+use crate::message::{HpkeAeadId, HpkeCiphertext, HpkeConfig, HpkeKdfId, HpkeKemId, Role, TaskId};
+use hpke::{
+    aead::{Aead, AesGcm256, ChaCha20Poly1305},
+    kdf::{HkdfSha256, HkdfSha512, Kdf},
+    kem::{DhP256HkdfSha256, X25519HkdfSha256},
+    setup_receiver, setup_sender, Deserializable, HpkeError, Kem, OpModeR, OpModeS, Serializable,
+};
+use rand::thread_rng;
+
+#[derive(Debug, thiserror::Error)]
+enum Error {
+    /// Wrapper around errors from crate hpke. See [`hpke::HpkeError`] for more
+    /// details on possible variants.
+    #[error("HPKE error")]
+    Hpke(#[from] HpkeError),
+    #[error("invalid HPKE configuration: {0}")]
+    InvalidConfiguration(&'static str),
+}
+
+/// Labels incorporated into HPKE application info string
+#[derive(Clone, Copy, Debug)]
+enum Label {
+    InputShare,
+    AggregateShare,
+}
+
+impl Label {
+    fn as_bytes(&self) -> &'static [u8] {
+        match self {
+            Self::InputShare => b"ppm input share",
+            Self::AggregateShare => b"ppm aggregate share",
+        }
+    }
+}
+
+/// PPM protocol message representing an HPKE private key, serialized using the
+/// `SerializePrivateKey` function as described in RFC 9180, §4 and §7.1.2.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct HpkePrivateKey(pub(crate) Vec<u8>);
+
+impl HpkePrivateKey {
+    pub(crate) fn new(bytes: Vec<u8>) -> Self {
+        Self(bytes)
+    }
+}
+
+impl AsRef<[u8]> for HpkePrivateKey {
+    fn as_ref(&self) -> &[u8] {
+        self.0.as_ref()
+    }
+}
+
+impl From<Vec<u8>> for HpkePrivateKey {
+    fn from(v: Vec<u8>) -> Self {
+        Self::new(v)
+    }
+}
+
+/// Application info used in HPKE context construction
+#[derive(Clone, Debug)]
+struct HpkeApplicationInfo(Vec<u8>);
+
+impl HpkeApplicationInfo {
+    /// Construct HPKE application info from the provided PPM task ID, label and
+    /// participant roles.
+    fn new(task_id: TaskId, label: Label, sender_role: Role, recipient_role: Role) -> Self {
+        Self(
+            [
+                task_id.as_bytes(),
+                label.as_bytes(),
+                &[sender_role as u8],
+                &[recipient_role as u8],
+            ]
+            .concat(),
+        )
+    }
+}
+
+/// A one-shot HPKE sender that encrypts messages to the public key in
+/// `recipient_config` using the AEAD, key derivation and key encapsulation
+/// mechanisms specified in `recipient_config`, and using `label`, `sender_role`
+/// and `recipient_role` to derive application info.
+//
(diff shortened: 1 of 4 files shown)


janus_server/src/hpke.rs
```diff
@@ -0,0 +1,379 @@
...
+enum Label {
+    InputShare,
+    AggregateShare,
+}
+
+impl Label {
+    fn as_bytes(&self) -> &'static [u8] {
+        match self {
+            Self::InputShare => "ppm input share",
+            Self::AggregateShare => "ppm aggregate share",
```
divergentdave, line 36: nit: these could be `b""` strings, to skip the need for `str::as_bytes()`.

janus_server/src/hpke.rs
```diff
@@ -0,0 +1,379 @@
...
+    sender_role: Role,
+    recipient_role: Role,
+    recipient_private_key: HpkePrivateKey,
+}
+
+impl HpkeRecipient {
+    /// Generate a new X25519HkdfSha256 keypair and construct an HPKE recipient
+    /// using the private key, with KEM = X25519HkdfSha256, KDF = HkdfSha512 and
+    /// AEAD = ChaCha20Poly1305, and the specified label, and roles.
+    fn generate(label: Label, sender_role: Role, recipient_role: Role) -> Self {
```
divergentdave, line 190: I presume this method is just for testing purposes, right? (based on the random task_id, zero HpkeConfig.id, and ephemeral keypairs)

janus_server/src/hpke.rs
```diff
@@ -0,0 +1,379 @@
...
+            recipient_role,
+            recipient_private_key: HpkePrivateKey(private_key.to_bytes().as_slice().to_vec()),
+        }
+    }
+
+    /// Decrypt `ciphertext` and return the plaintext.
+    ///
+    /// In PPM, an HPKE context can only be used once (we have no means of
+    /// ensuring that sender and recipient "increment" nonces in lockstep), so
+    /// this method consumes self.
```
divergentdave, line 219: In the bare function `open()` below, we create a context, use it once, and drop it, which should be sufficient to ensure we only use an HPKE context once (with its initial nonce only). As `HpkeRecipient` contains an `HpkeConfig`, private key, and AAD-related fields, it should be fine for it to be reused, and moreover we may want to do so when handling a series of aggregate requests, rather than construct this struct repeatedly, copying both sides of the keypair each time.

janus_server/src/hpke.rs
```diff
@@ -0,0 +1,379 @@
...
+    task_id: TaskId,
+    recipient_config: HpkeConfig,
+    label: Label,
+    sender_role: Role,
+    recipient_role: Role,
+}
+
+impl HpkeSender {
+    /// Encrypt `plaintext` and return the HPKE ciphertext.
+    fn seal(self, plaintext: &[u8], associated_data: &[u8]) -> Result<HpkeCiphertext, Error> {
```
divergentdave, line 102: As with `HpkeReceiver::open()`, this could take self by reference, as the bare function `seal()` ensures we don't use a context more than once.

janus_server/src/hpke.rs
```diff
@@ -0,0 +1,379 @@
...
+    ciphertext: &HpkeCiphertext,
+    associated_data: &[u8],
+    serialized_recipient_private_key: &HpkePrivateKey,
+) -> Result<Vec<u8>, Error> {
+    // Deserialize recipient priv into the appropriate PrivateKey type for
+    // the KEM
+    let recipient_private_key =
+        Encapsulate::PrivateKey::from_bytes(&serialized_recipient_private_key.0)?;
+
+    // Deserialize sender encapsulated pub into the appropriate EncappedKey
```
divergentdave, line 269: nit: this would be an encapsulated context/encapsulated key/encapsulated KEM secret.
```suggestion
    // Deserialize sender encapsulated key into the appropriate EncappedKey
```

janus_server/src/hpke.rs
BranLwyd: I think so; if so, I suggest moving this to the `tests` module, or otherwise arranging for this to be `#[cfg(test)]`.

janus_server/src/message.rs
```diff
@@ -188,6 +189,23 @@ impl Decode for TaskId {
     }
 }
 
+impl TaskId {
+    /// Get a reference to the task ID as a byte slice
+    pub(crate) fn as_bytes(&self) -> &[u8] {
```
BranLwyd, line 194: nit: since the internal `[u8; 32]` is marked `pub`, this method isn't really necessary: anyone could just do `&task_id.0` themselves. Of course, this comes at an understandability cost to folks not already familiar with how `TaskId` is laid out -- so `as_bytes` is better, but why have more than one way to do it?

Maybe we should un-`pub` the internal byte arrays for types encoding a single value, like `TaskId`, `AggregationJobId`, etc, and provide an `as_bytes` method as needed? In this proposal we'd leave structures with multiple values alone. (if you agree, no need to do in this PR, I can pick it up next time I touch this file)

janus_server/src/hpke.rs
tgeoghegan: We will need this method outside of test targets in an upcoming PR, which will create an `aggregator` binary that uses this to implement an `hpke_config` endpoint. Past that we'll have to figure out how to load HPKE keys and configurations from config, but we eventually need to generate keys which might use a modified version of this method.

All that being said, it makes sense to restrict this method to `cfg(test)` in this PR.

janus_server/src/hpke.rs
tgeoghegan: Done!

janus_server/src/hpke.rs
tgeoghegan: Done!

janus_server/src/message.rs
tgeoghegan: Caveat: all that follows is subjective. I can't point to authoritative Rust style guides or anything.

I like the tuple struct pattern for newtypes, but I dislike the `.0`s that crop up in code that uses the newtype, because when I encounter that, I get no hints as to either the type of `.0` nor what it semantically represents. To me, a method like `as_bytes` reads more clearly. However we could achieve the same thing the method does by defining `pub struct TaskId { pub bytes: [u8; 32] }` and then accessing `task_id.bytes`.

I agree with you that establishing a consistent style for newtypes and only having one way to do the thing is a good idea. In the interest of reducing the amount of boilerplate we have to write, I vote for using "regular" structs (i.e., structs with labeled fields) for newtypes.

divergentdave: One alternative is we could have such structs Deref into a byte slice, which would mean we could take advantage of automatic dereferencing to reduce boilerplate, but it's a matter of taste whether that's too clever.

divergentdave (approved):

PR #23, opened 2022-03-16
`janus_server`: `/upload` endpoint on aggregator

tgeoghegan:
Adds an implementation of an upload endpoint to `src/aggregator.rs`
wired up through a `warp` filter. We check report nonces and HPKE config
IDs per the PPM specification. We also decrypt the leader input share
while handling upload, primarily to exercise HPKE message decryption. We
might choose to not try to do that until the prepare sub-protocol begins
because there's no way to alert the client that decryption failed.

To test error conditions around report nonces, I also introduced a
`time` module which implements a simple clock interface.

We also extend the `aggregator` binary target so that it can be
instructed to run either a helper or a leader by passing the role name
as a positional argument (i.e., `./aggregator leader` or `./aggregator
helper`).

This commit does not implement the client side of upload (except
inasmuch as was needed to unit test `mod aggregator`; that will arrive
in a later PR.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -1,21 +1,190 @@
 //! Common functionality for PPM aggregators
 use crate::{
-    hpke::{HpkeRecipient, Label},
-    message::{Role, TaskId},
+    hpke::HpkeRecipient,
+    message::{HpkeConfigId, Nonce, Report, Role},
+    time::Clock,
 };
+use bytes::Bytes;
+use chrono::Duration;
 use http::{header::CACHE_CONTROL, StatusCode};
-use prio::codec::Encode;
-use std::{future::Future, net::SocketAddr};
-use warp::{filters::BoxedFilter, reply, trace, Filter, Reply};
+use prio::codec::{Decode, Encode};
+use std::{
+    collections::HashMap,
+    convert::Infallible,
+    future::Future,
+    net::SocketAddr,
+    ops::Sub,
+    sync::{Arc, Mutex},
+};
+use tracing::warn;
+use warp::{filters::BoxedFilter, reply, trace, Filter, Rejection, Reply};
 
-/// Constructs a Warp filter with an aggregator's endpoints.
-fn aggregator_filter(task_id: TaskId) -> BoxedFilter<(impl Reply,)> {
-    let hpke_recipient =
-        HpkeRecipient::generate(task_id, Label::InputShare, Role::Client, Role::Leader);
+/// Errors returned by functions and methods in this module
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    /// An invalid configuration was passed.
+    #[error("Invalid configuration: {0}")]
+    InvalidConfiguration(&'static str),
+    /// Error decoding an incoming message.
+    #[error("Message decoding failed: {0}")]
+    MessageDecode(#[from] prio::codec::CodecError),
+    /// Corresponds to `staleReport`, §3.1
+    #[error("Stale report: {0}")]
+    StaleReport(Nonce),
+    /// Corresponds to `unrecognizedMessage`, §3.1
+    #[error("Unrecognized message: {0}")]
+    UnrecognizedMessage(&'static str),
+    /// Corresponds to `outdatedHpkeConfig`, §3.1
+    #[error("Outdated HPKE config: {0}")]
+    OutdatedHpkeConfig(HpkeConfigId),
+    /// A report was rejected becuase the timestamp is too far in the future,
+    /// §4.3.4.
+    // TODO(timg): define an error type in §3.1 and clarify language on
+    // rejecting future reports
+    #[error("Report from the future: {0}")]
+    ReportFromTheFuture(Nonce),
+}
+
+// This impl allows use of [`Error`] in [`warp::reject::Rejection`]
+impl warp::reject::Reject for Error {}
+
+/// A PPM aggregator
+#[derive(Clone, Debug)]
+pub struct Aggregator<C> {
+    /// This aggregator's perception of what time it is
+    clock: C,
+    /// How much clock skew to allow between client and aggregator. Reports from
+    /// farther than this duration into the future will be rejected.
+    tolerable_clock_skew: Duration,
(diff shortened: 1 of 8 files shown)


tgeoghegan:
In this change, we use `warp::reject::custom` and generally use `warp` `Rejection`s to indicate errors. However, we don't render those into HTTP responses with the correct error code or body. The way to do that is to use [`warp::fiter::recover`](https://docs.rs/warp/latest/warp/trait.Filter.html#method.recover) ([example](https://github.com/seanmonstar/warp/blob/master/examples/rejections.rs)). If we merge this change, then we should file an issue about wiring up rejection handlers and crafting correct responses.

janus_server/src/aggregator.rs
```diff
@@ -1,21 +1,189 @@
...
+impl<C: Clock> Aggregator<C> {
+    /// Create a new aggregator. `report_recipient` is used to decrypt reports
+    /// received by this aggregator.
+    fn new(
+        clock: C,
+        tolerable_clock_skew: Duration,
+        role: Role,
+        report_recipient: HpkeRecipient,
+    ) -> Result<Self, Error> {
+        if tolerable_clock_skew < Duration::zero() {
```
divergentdave, line 74: nit: this is unreachable, as `Duration` can only represent positive durations

divergentdave: I take it back, this is `chrono::Duration`, not `std::time::Duration`

BranLwyd (changes requested):

janus_server/src/aggregator.rs
```diff
@@ -1,21 +1,189 @@
...
+    // rejecting future reports
+    #[error("Report from the future: {0}")]
+    ReportFromTheFuture(Nonce),
+}
+
+// This impl allows use of [`Error`] in [`warp::reject::Rejection`]
+impl warp::reject::Reject for Error {}
+
+/// A PPM aggregator
+#[derive(Clone, Debug)]
```
divergentdave, line 48: nit: Note that this Clone implementation is currently unused.

janus_server/src/aggregator.rs
```diff
@@ -24,39 +192,135 @@ fn aggregator_filter(task_id: TaskId) -> BoxedFilter<(impl Reply,)> {
...
+    clock: C,
+    tolerable_clock_skew: Duration,
+    role: Role,
+    hpke_recipient: HpkeRecipient,
     listen_address: SocketAddr,
-) -> (SocketAddr, impl Future<Output = ()> + 'static) {
-    let routes = aggregator_filter(task_id).with(trace::request());
+) -> Result<(SocketAddr, impl Future<Output = ()> + 'static), Error> {
+    let routes = aggregator_filter(clock, tolerable_clock_skew, role, hpke_recipient)?
+        .with(trace::request());
```
divergentdave, line 234: We are doubling up on warp tracing filters, with one here, and separate `trace::named` filters below. This results in the following output.

```
  2022-03-17T17:17:36.432990Z  INFO warp::filters::trace: finished processing with success, status: 200
    at /home/david/.cargo/registry/src/github.com-1ecc6299db9ec823/warp-0.3.2/src/filters/trace.rs:243 on ThreadId(2)
    in warp::filters::trace::context with hpke_config
    in warp::filters::trace::request with method: GET, path: /hpke_config, version: HTTP/1.1, remote.addr: 127.0.0.1:39378

  2022-03-17T17:17:36.433033Z  INFO warp::filters::trace: finished processing with success, status: 200
    at /home/david/.cargo/registry/src/github.com-1ecc6299db9ec823/warp-0.3.2/src/filters/trace.rs:243 on ThreadId(2)
    in warp::filters::trace::request with method: GET, path: /hpke_config, version: HTTP/1.1, remote.addr: 127.0.0.1:39378
```

janus_server/src/aggregator.rs
```diff
@@ -67,9 +327,239 @@ mod tests {
...
+    #[test]
+    fn upload_wrong_hpke_config_id() {
+        INSTALL_TRACE_SUBSCRIBER.call_once(|| install_subscriber().unwrap());
+
+        let skew = Duration::minutes(10);
+        let (aggregator, mut report) = setup_upload_test(skew);
+
+        report.encrypted_input_shares[0].config_id = HpkeConfigId(101);
+
+        // should reject duplicate reports
```
divergentdave, line 503: nit: copy-paste error

divergentdave (approved):

janus_server/src/aggregator.rs
```diff
@@ -24,39 +193,130 @@ fn aggregator_filter(task_id: TaskId) -> BoxedFilter<(impl Reply,)> {
...
         })
-        .with(trace::named("hpke_config"))
-        .boxed()
+        .with(trace::named("hpke_config"));
+
+    let upload_endpoint = warp::path("upload")
+        .and(warp::post())
+        .and(with_shared_value(aggregator))
+        .and(with_decoded_message())
+        .and_then(|aggregator: Aggregator<C>, report: Report| async move {
```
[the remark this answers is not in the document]
tgeoghegan: Yup, I left myself a task to file an issue about rejection recovery. I have done this before in ppm-prototype but (1) I'm not sure what I did there is the best way to go about it and (2) it would have been a bunch more code in an already-large PR.

janus_server/src/time.rs
```diff
@@ -0,0 +1,53 @@
...
+    fn now(&self) -> NaiveDateTime;
+}
+
+/// A real clock returns the current time relative to the Unix epoch.
+#[derive(Clone, Default)]
+pub struct RealClock {}
+
+impl Clock for RealClock {
+    fn now(&self) -> NaiveDateTime {
+        NaiveDateTime::from_timestamp(Utc::now().timestamp(), 0)
```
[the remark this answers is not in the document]
tgeoghegan: Rounding to the second makes sense to me since that is the resolution supported by PPM `Time`, however the reason I used this form is because I didn't notice the one you suggested. This works so I'll leave it as is but I won't object to someone rewriting this module to make it simpler or more idiomatic.

BranLwyd (approved):

PR #24, opened 2022-03-17
Initial Datastore implementation.

BranLwyd:
Datastore is backed by a Postgres database. Tests are implemented via
the testcontainers crate, which allows spinning up a real instance of
Postgres. All transactions are currently fully serializable; I think we
should try very hard to maintain this, relaxing this constraint only if
we encounter a performance problem that cannot be solved through other
means.

I found a solution to the lifetimes issue[1], which is "have the callback return a boxed/pinned dynamic future". TBH, I'm not sure why this works to resolve the lifetime issues I was seeing (why is the compiler now confident that the future doesn't outlive the Postgres client?), but resolve the issue it does & with a minimum of per-tx boilerplate code. See the tests for how run_tx is expected to be used.

Using testcontainers for a Postgres DB works well. The containers get cleaned up even on test failure. I found that Ctrl+C'ing the tests while running will leave a container behind, but this is easy enough to clean up. (I use an invocation like `docker container rm -f $(docker container ls -q)`, but be aware this will attempt to remove _all_ containers running on your system.) The turnup/turndown is very fast--the tests only take ~2s to run on my workstation. I propose we go ahead and plan to use testcontainers-based Datastores for all unit testing, rather than attempting to implement an in-memory fake Datastore, both because this approach will require less work & because the test Datastores will be much more likely to be functionally equivalent to real Datastores.

I left a number of important TODOs, which I will be cleaning up next, starting with connection pooling to allow for concurrent transaction processing. I wanted to send this as soon as possible to keep from blocking others' work -- apologies, this took much longer than expected due to the lifetime issue.

[1] https://www.reddit.com/r/rust/comments/kymau7/how_to_perform_async_transaction_retry_with_tokio/


janus_server/src/datastore.rs
```diff
@@ -0,0 +1,318 @@
+use prio::codec::{decode_u16_items, encode_u16_items, CodecError, Decode};
+use tokio_postgres::{error::SqlState, IsolationLevel, Row};
+
```
tgeoghegan, line 3: nit: delete the blank line so rustfmt can keep the `use`s in order

janus_server/src/datastore.rs
```diff
@@ -0,0 +1,318 @@
...
+    use std::str;
+    use testcontainers::{clients, images::postgres::Postgres, Container, Docker};
+    use tokio_postgres::NoTls;
+
+    lazy_static! {
+        static ref SCHEMA: &'static str = {
+            let schema_bytes = include_bytes!("../../db/schema.sql");
+            str::from_utf8(schema_bytes).unwrap()
+        };
+    }
```
divergentdave, line 220: (will also require dropping a * below)
```suggestion
    static SCHEMA: &'static str = include_str!("../../db/schema.sql");
```

janus_server/src/datastore.rs
```diff
@@ -0,0 +1,318 @@
...
+    /// An arbitrary error returned from the user callback; unrelated to DB internals. This error
+    /// will never be generated by the datastore library itself.
+    #[error(transparent)]
+    User(#[from] Box<dyn std::error::Error + Send + Sync>),
+}
+
+impl Error {
+    // is_serialization_failure determines if a given error corresponds to a Postgres
+    // "serialization" failure, which requires the entire transaction to be aborted & retried from
+    // the beginning per https://www.postgresql.org/docs/9.5/transaction-iso.html.
```
tgeoghegan, line 190: In the test below, we use PostgreSQL version 14, but this links to docs for 9.5 which is unsupported.

janus_server/src/datastore.rs
```diff
@@ -0,0 +1,318 @@
...
+    pub async fn run_tx<F, T>(&self, f: F) -> Result<T, Error>
+    where
+        for<'a> F: Fn(&'a Transaction) -> Pin<Box<dyn Future<Output = Result<T, Error>> + 'a>>,
+    {
+        loop {
+            let rslt = self.run_tx_once(&f).await;
+            if let Some(err) = rslt.as_ref().err() {
+                if err.is_serialization_failure() {
+                    continue;
+                }
```
tgeoghegan, line 41: Besides the case of serialization failures, what's our story for retries in the face of network problems? Does `tokio-postgres` do something sensible with exponential backoffs and so on or are we on the hook for implementing that based on CloudSQL documentation?

divergentdave (approved):

janus_server/src/datastore.rs
```diff
@@ -0,0 +1,318 @@
...
+        let tx = Transaction {
+            tx: guard
+                .build_transaction()
+                .isolation_level(IsolationLevel::Serializable)
+                .start()
+                .await?,
+        };
+
+        // Run user-provided function with the transaction.
+        // TODO(brandon): can we box/pin the return value, rather than requiring `f` to return a pre-boxed/pinned future?
```
tgeoghegan, line 62: I doubt that you can `Box::new` the returned value because then you'd have `F: Fn(&'a Transaction) -> dyn Future<...>`, and that can't work because the compiler needs to know the size of returned values.

I hesitate to comment on the usage of `Pin` until I have spent more time [contemplating it on the tree of woe](https://www.youtube.com/watch?v=p48daZLNYFs).

janus_server/src/datastore.rs
```diff
@@ -0,0 +1,318 @@
...
+            .unwrap();
+        tokio::spawn(async move { conn.await.unwrap() });
+
+        // Run our schema on the new database.
+        postgres_client.batch_execute(*SCHEMA).await.unwrap();
+        (Datastore::new(postgres_client), db_container)
+    }
+
+    #[tokio::test]
+    async fn roundtrip_report() {
```
tgeoghegan, line 250: I suggest installing a `tracing` subscriber in tests or you won't ever see `info!`, `err!` etc. output. The obnoxious thing is that you can only install a single subscriber per process, so you have to guard it with an `std::sync::Once`, as I did in the `src/aggregator.rs` tests. This might be a good time to place the test subscriber installation logic into a shared `tests` module and use that here. However we could also leave that for a future PR; a TODO here will suffice.

janus_server/src/message.rs
```diff
@@ -375,8 +386,8 @@ impl Decode for Report {
 /// PPM protocol message representing an arbitrary extension included in a client report.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Extension {
-    extension_type: ExtensionType,
-    extension_data: Vec<u8>,
+    pub extension_type: ExtensionType,
+    pub extension_data: Vec<u8>,
```
tgeoghegan, line 390: nit:
```suggestion
    pub(crate) extension_type: ExtensionType,
    pub(crate) extension_data: Vec<u8>,
```
It's debatable if being disciplined about this really matters for a binary crate, though we might soon publish a `janus-client` crate.

janus_server/Cargo.toml
```diff
@@ -17,11 +18,14 @@ rand = "0.8"
...
+tokio-postgres = { version = "0.7.5", features = ["with-chrono-0_4"] }
 tracing = "0.1.32"
 tracing-subscriber = { version = "0.3.9", features = ["std", "env-filter"] }
 url = "2.2.2"
 warp = { version = "^0.3", features = ["tls"] }
 
 [dev-dependencies]
+assert_matches = "1"
 hyper = "0.14.17"
+testcontainers = "0.12.0"
```
tgeoghegan, line 31: nit: newline
```suggestion
testcontainers = "0.12.0"

```

janus_server/src/datastore.rs
```diff
@@ -0,0 +1,318 @@
...
+
+/// Datastore represents a datastore for Janus, with support for transactional reads and writes.
+/// In practice, Datastore instances are currently backed by a PostgreSQL database.
+pub struct Datastore {
+    db: Mutex<tokio_postgres::Client>,
+}
+
+impl Datastore {
+    /// new creates a new Datastore using the given Client for backing storage. It is assumed that
+    /// the Client is connected to a database with a compatible version of the Janus database schema.
```
tgeoghegan, line 19: We should figure out DB schema versioning so that we can validate at startup that the DB we connect to has the expected schema. Of course that would probably be a feature of the DB migration tool we adopt so perhaps we shouldn't solve that ourselves.

janus_server/src/datastore.rs
```diff
@@ -0,0 +1,318 @@
...
+    use super::*;
+    use lazy_static::lazy_static;
+    use std::str;
+    use testcontainers::{clients, images::postgres::Postgres, Container, Docker};
+    use tokio_postgres::NoTls;
+
+    lazy_static! {
+        static ref SCHEMA: &'static str = {
+            let schema_bytes = include_bytes!("../../db/schema.sql");
+            str::from_utf8(schema_bytes).unwrap()
```
tgeoghegan, line 218: I think [`include_str!`](https://doc.rust-lang.org/std/macro.include_str.html) might save you a line or two here.

tgeoghegan (approved):
Just nits and suggestions, nothing I think is blocking.

janus_server/src/datastore.rs
```diff
@@ -0,0 +1,318 @@
...
+
+    async fn run_tx_once<F, T>(&self, f: &F) -> Result<T, Error>
+    where
+        for<'a> F: Fn(&'a Transaction) -> Pin<Box<dyn Future<Output = Result<T, Error>> + 'a>>,
+    {
+        // Open transaction.
+        let mut guard = self.db.lock().await;
+        let tx = Transaction {
+            tx: guard
+                .build_transaction()
```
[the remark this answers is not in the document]
BranLwyd: Agreed, but I'm going to leave this for future right now. I think we may want to expose two different types (`ReadOnlyTransaction` vs `Transaction`) and I don't want to include that complexity in this PR.

janus_server/src/datastore.rs
BranLwyd: tokio_postgres doesn't document anything in terms of automatic retries, nor does it document what happens if it encounters a network failure. We might want to do something better here, but we'll need to see what the errors actually look like before we can discriminate based on them.

Due to the stateful nature of a DB connection (as opposed to cloud APIs which are typically stateless from request to request in a connection), I strongly suspect tokio_postgres doesn't e.g. automatically retry a failed `query` call. I think we'd want to notice a network error at the level of `run_tx`, drop the client/connection in use, and grab a new connection from the pool to try again.

janus_server/src/datastore.rs
BranLwyd: That's a good idea, but we do need to know how we're going to version database schemas first.

We'd also need to be careful to permit live DB schema upgrades -- if we're too restrictive, upgrading to version N+1 would cause running processes expecting version N very unhappy.

The best solution might be to set expectations that all DB schema migrations are done in a way that supports no-downtime migration (i.e. we arrange releases so that we can live-migrate to the new schema while running jobs can work with the DB), but especially in early development we're likely to want to make non-backwards-compatible changes to the schema and doing the release dances to allow these live is probably too much work.

janus_server/src/message.rs
BranLwyd: I'm currently touching messages.rs in a different PR that is making a few changes, I'll update this there since ~all of the messages should be modified for consistency.

janus_server/src/datastore.rs
tgeoghegan: OK. We should write a TODO and/or an issue about retries then.

Issue #18, opened 2022-03-11
Proposed crate/directory layout


divergentdave:
Now that we are building client, server, and common PPM components, I thought it would be good to reorganize them into separate crates. How does this look?

Before:
```
├── .github ...
├── .gitignore
├── janus_server
│   ├── Cargo.lock
│   ├── Cargo.toml
│   ├── src
│   │   ├── aggregator.rs
│   │   ├── bin
│   │   │   └── aggregator.rs
│   │   ├── client.rs
│   │   ├── hpke.rs
│   │   ├── lib.rs
│   │   └── message.rs
│   └── tests
│       └── integration_test.rs
├── LICENSE
└── README.md
```

After:
```
├── .github ...
├── .gitignore
├── Cargo.lock
├── Cargo.toml (workspace only)
├── LICENSE
├── README.md
├── janus
│   ├── Cargo.toml
│   ├── src
│   │   ├── lib.rs
│   │   ├── hpke.rs
│   │   └── message.rs
│   └── tests
├── janus_client
│   ├── Cargo.toml
│   ├── src
│   │   └── lib.rs
│   └── tests
│       └── integration_test.rs
└── janus_server
    ├── Cargo.toml
    ├── src
    │   ├── aggregator.rs
    │   ├── bin
    │   │   └── aggregator.rs
    │   └── lib.rs
    └── tests
        └── integration_test.rs
```

BranLwyd:
Sure, this layout seems reasonable.

Can we get away with doing this lazily? e.g. for now, write code into `janus_server` specifically. Then as we move into the client implementation, factor the things we need out of `janus_server` to a common `janus` crate. I only suggest this because factorings like this tend to work out better if they are based on a concrete need rather than our best guesses as to what will need to be factored where, though admittedly `hpke.rs` and `message.rs` will pretty clearly be needed in both a client & server.

tgeoghegan:
In #37, I add dependencies on `structopt` and `serde_yaml` to `janus_server` that I think aren't necessarily needed in the library crate. The former is only needed in the `aggregator` bin and the latter could be in `dev-dependencies` but for the binary target. If we split out the binaries into their own crate in the workspace proposed here, I think we could distinguish between things that `janus_server` needs and that its binaries need. Or is Rust dead code detection already smart enough to deal with this for us?

divergentdave:
[...]

PR #29, opened 2022-03-22
/upload endpoint: store reports to a Datastore.

BranLwyd:
This replaces the previous temporary solution of storing them in-memory
only. This commit also includes some refactoring of the ephemeral
datastore test functionality to make it easier to use & ease later
refactorings.

I'm pretty unhappy about having to make the `datastore::test_util` module
compile under `debug_assertions`; this means I have to upgrade several
test-only dependencies from `dev-dependencies` to `depdendencies`.
Unfortunately, I think this is required because integration tests do not
run in `#[cfg(test)]`. (Better approaches very welcome.)

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -1,5 +1,6 @@
 //! Common functionality for PPM aggregators
 use crate::{
+    datastore::{self, Datastore},
     hpke::HpkeRecipient,
     message::{HpkeConfigId, Nonce, Report, Role},
     time::Clock,
@@ -8,14 +9,7 @@ use bytes::Bytes;
 use chrono::Duration;
 use http::{header::CACHE_CONTROL, StatusCode};
 use prio::codec::{Decode, Encode};
-use std::{
-    collections::HashMap,
-    convert::Infallible,
-    future::Future,
-    net::SocketAddr,
-    ops::Sub,
-    sync::{Arc, Mutex},
-};
+use std::{convert::Infallible, future::Future, net::SocketAddr, ops::Sub, sync::Arc};
 use tracing::warn;
 use warp::{filters::BoxedFilter, reply, trace, Filter, Rejection, Reply};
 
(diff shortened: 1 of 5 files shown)


janus_server/src/aggregator.rs
```diff
@@ -8,68 +9,66 @@ use bytes::Bytes;
...
     OutdatedHpkeConfig(HpkeConfigId),
     /// A report was rejected becuase the timestamp is too far in the future,
     /// §4.3.4.
     // TODO(timg): define an error type in §3.1 and clarify language on
     // rejecting future reports
-    #[error("Report from the future: {0}")]
+    #[error("report from the future: {0}")]
     ReportFromTheFuture(Nonce),
+    #[error("datastore error: {0}")]
+    Datastore(#[from] datastore::Error),
```
divergentdave, line 41: We shouldn't allow this conversion to happen automatically, as it would be easy to miss a call to `downcast_to_aggregator_error()`, and wind up with `crate::Error::Datastore(datastore::Error::User(crate::Error::...)`. In fact, we could turn `downcast_to_aggregator_error()` into our own `From` implementation.
```suggestion
    Datastore(datastore::Error),
```

janus_server/src/datastore.rs
```diff
@@ -208,25 +254,35 @@ impl Error {
     }
 }
 
-#[cfg(test)]
-pub(crate) mod test_util {
+// debug_assertions allows this module to be used in integration tests.
+// XXX: if the debug_assertions strategy doesn't make it through code review, move things back to dev-dependencies
+#[cfg(any(test, debug_assertions))]
+pub mod test_util {
```
divergentdave, line 260: I'd prefer we avoid conditional compilation on `--cfg debug_assertions`, note that it breaks `cargo test --release`. Two alternative solutions:

1. Make the module `#[doc(hidden)]` and accept taking testcontainers and lazy_static as runtime dependencies.
2. Move the contents of this module to a separate file, and include it in the module tree of both the library crate (behind `#[cfg(test)]` and integration crate. This will require using `#[path="..."]` on one of the two module declarations. (note that the contents of `test_util` only depend on the public API of the crate, particularly, `Datastore::new()`, so it can be made to work in both contexts)

divergentdave (approved):

janus_server/src/aggregator.rs
```diff
@@ -8,68 +9,66 @@ use bytes::Bytes;
...
 impl warp::reject::Reject for Error {}
 
 /// A PPM aggregator
-#[derive(Clone, Debug)]
+#[derive(Clone, derivative::Derivative)]
+#[derivative(Debug)]
 pub struct Aggregator<C> {
-    /// This aggregator's perception of what time it is
+    #[derivative(Debug = "ignore")]
+    /// The datstore used for durable storage.
```
tgeoghegan, line 52:
```suggestion
    /// The datastore used for durable storage.
```

janus_server/src/datastore.rs
tgeoghegan: The other difficulty is pulling in certain dependencies only in the integration test target. In `libprio-rs`, I made a workspace and put a distinct `Cargo.toml` for the binaries so that I could enable features in crate `prio` only when building binary `generate-test-vector`:

https://github.com/divviup/libprio-rs/blob/main/binaries/Cargo.toml
```
prio = { path = "..", features = ["test-vector"] }
```

Then `test-vector` conditionally enables dependencies: https://github.com/divviup/libprio-rs/blob/main/Cargo.toml

However if we want to do that, we could wait until we adopt @divergentdave's proposed crate layout in #18.

tgeoghegan (approved):

janus_server/src/datastore.rs
BranLwyd: I went with `#[doc(hidden)]` for now. I'm not happy with it, but it's better than the other solutions IMO.

TBH, I think this is a wart on Rust. I understand the motivation of wanting integration tests to use the public API; but this feels like the wrong lever, as I'm only a moderately-experienced Rust developer and this is already the first time I've been caught off-guard by integration tests not being compiled under `#[cfg(test)]`.

Issue #25, opened 2022-03-17
`janus_server`: set up Warp rejection handlers to send correct responses to HTTP clients in error cases


tgeoghegan:
We should use `warp::filter::recover` to do something with `Rejection`s from filters to massage them into responses with appropriate HTTP status codes and bodies.

See comment on PR #23: https://github.com/abetterinternet/janus/pull/23#issuecomment-1069552558

divergentdave:
To reiterate a [comment on #23](https://github.com/abetterinternet/janus/pull/23#discussion_r829381568), I think we should not pass `crate::Error` through rejection and recovery, as it is non-idiomatic to handle application errors with this mechanism. Instead, we should make each filter with aggregator business logic return `Result<impl Reply, Error>`, and then chain a filter after our router that will take that result as an input, turn errors into problem documents, and return `impl Reply`.

PR #31, opened 2022-03-22
Problem details

divergentdave:
This is a work-in-progress branch to implement #25. Thus far I've got all the types and signatures lined up, tomorrow I'll read through RFC 7807 and flesh out the error response.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -28,25 +28,25 @@ pub enum Error {
     #[error("message decoding failed: {0}")]
     MessageDecode(#[from] prio::codec::CodecError),
     /// Corresponds to `staleReport`, §3.1
-    #[error("stale report: {0}")]
+    #[error("stale report: {0} {1:?}")]
     StaleReport(Nonce, TaskId),
     /// Corresponds to `unrecognizedMessage`, §3.1
-    #[error("unrecognized message: {0}")]
+    #[error("unrecognized message: {0} {1:?}")]
     UnrecognizedMessage(&'static str, TaskId),
     /// Corresponds to `unrecognizedTask`, §3.1
-    #[error("unrecognized task")]
+    #[error("unrecognized task: {0:?}")]
     UnrecognizedTask(TaskId),
     /// Corresponds to `outdatedHpkeConfig`, §3.1
-    #[error("outdated HPKE config: {0}")]
+    #[error("outdated HPKE config: {0} {1:?}")]
     OutdatedHpkeConfig(HpkeConfigId, TaskId),
     /// A report was rejected becuase the timestamp is too far in the future,
     /// §4.3.4.
     // TODO(timg): define an error type in §3.1 and clarify language on
     // rejecting future reports
-    #[error("report from the future: {0}")]
+    #[error("report from the future: {0} {1:?}")]
     ReportFromTheFuture(Nonce, TaskId),
     /// Corresponds to `invalidHmac`, §3.1
-    #[error("invalid HMAC tag")]
+    #[error("invalid HMAC tag: {0:?}")]
     InvalidHmac(TaskId),
     /// An error from the datastore.
     #[error("datastore error: {0}")]
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -11,7 +11,6 @@ use http::{header::CACHE_CONTROL, StatusCode};
 use prio::codec::{Decode, Encode};
 use std::{convert::Infallible, future::Future, net::SocketAddr, ops::Sub, sync::Arc};
 use tracing::warn;
-use url::Url;
 use warp::{
     filters::BoxedFilter,
     reply::{self, Response},
@@ -87,8 +86,6 @@ pub struct Aggregator<C> {
     // TODO: Aggregators should have multiple generations of HPKE config
     // available to decrypt tardy reports
     report_recipient: HpkeRecipient,
-    /// The aggregator's own HTTPS endpoint, to be advertised in error responses.
-    endpoint: Url,
 }
 
 impl<C: Clock> Aggregator<C> {
@@ -100,7 +97,6 @@ impl<C: Clock> Aggregator<C> {
         tolerable_clock_skew: Duration,
         role: Role,
         report_recipient: HpkeRecipient,
-        endpoint: Url,
     ) -> Result<Self, Error> {
         if tolerable_clock_skew < Duration::zero() {
             return Err(Error::InvalidConfiguration(
@@ -114,7 +110,6 @@ impl<C: Clock> Aggregator<C> {
             tolerable_clock_skew,
             role,
             report_recipient,
-            endpoint,
         })
     }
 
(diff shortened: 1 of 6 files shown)


janus_server/src/aggregator.rs
```diff
@@ -185,6 +198,16 @@ impl<C: Clock> Aggregator<C> {
...
 
+impl<C> Aggregator<C> {
+    /// Returns the HTTPS URL of this aggregator's own endpoint.
+    fn own_endpoint(&self) -> Url {
+        // TODO (issue #20): determine this URL endpoint from configuration
+        // TODO (issue abetterinternet/ppm-specification#209): This may no longer be needed if the
+        // requirements for the "instance" problem details member change.
+        Url::parse("https://example.com/ppm_aggregator").unwrap()
+    }
+}
```
tgeoghegan, line 209: I think this should be re-worked on top of #37 so that the aggregator's own endpoint is obtained from task parameters.

janus_server/src/aggregator.rs
```diff
@@ -194,10 +217,119 @@ fn with_cloned_value<T: Clone + Sync + Send>(
...
+        ),
+        status,
+    )
+    .into_response()
+}
+
+/// Produces a closure that will transform applicable errors into a problem details JSON object.
+/// (See RFC 7807) The returned closure is meant to be used in a warp `map` filter.
+fn error_handler<C, R>(
+    aggregator: &Aggregator<C>,
```
tgeoghegan, line 301: nit: all you need here is the aggregator's endpoint, so you could remove some generic complexity by doing `fn error_handler(endpoint: Url) -> ...`

janus_server/src/aggregator.rs
```diff
@@ -230,20 +364,22 @@ fn aggregator_filter<C: 'static + Clock>(
...
-
-            Ok(reply::with_status(warp::reply(), StatusCode::OK)) as Result<_, Rejection>
+            Ok(aggregator)
         })
+        .and(with_decoded_message())
+        .then(
+            |aggregator: Aggregator<C>, report_res: Result<Report, Error>| async move {
+                aggregator.handle_upload(&report_res?).await?;
+
+                Ok(warp::reply()) as Result<_, Error>
```
tgeoghegan, line 379: I prefer to explicitly set status HTTP 200 rather than rely on the implicit behavior of `warp::reply`, because then we can point to an obvious place where we implement the RFC requirement. Plus it'll make this consistent with other endpoints that return HTTP statuses besides OK.

tgeoghegan (changes requested):

janus_server/src/aggregator.rs
```diff
@@ -117,6 +125,7 @@ impl<C: Clock> Aggregator<C> {
             );
             return Err(Error::UnrecognizedMessage(
                 "unexpected number of encrypted shares in report",
+                report.task_id,
```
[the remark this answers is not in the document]
divergentdave: Generally I agree on using `tracing_error`, but in this case the task ID is needed in the JSON document. I think it's more foolproof to pass it as part of the error enum variant, because we can't forget the ID when constructing an error. I would rather not fetch from a tracing span to assemble the response.

janus_server/src/aggregator.rs
```diff
@@ -195,12 +221,114 @@ fn with_decoded_message<T: Decode + Send + Sync>(
...
+            PpmProblemType::StaleReport => "urn:ietf:params:ppm:error:staleReport",
+            PpmProblemType::InvalidHmac => "urn:ietf:params:ppm:error:invalidHmac",
+        }
+    }
+
+    /// Returns a human-readable summary of a problem type.
+    fn description(&self) -> &'static str {
+        match self {
+            PpmProblemType::UnrecognizedMessage => {
+                "The message type for a response was incorrect or the payload was malformed"
```
BranLwyd, line 249:
```suggestion
                "The message type for a response was incorrect or the payload was malformed."
```

(nit, typo)

janus_server/src/aggregator.rs
```diff
@@ -195,12 +221,114 @@ fn with_decoded_message<T: Decode + Send + Sync>(
...
+            http::header::CONTENT_TYPE,
+            PROBLEM_DETAILS_JSON_MEDIA_TYPE,
+        ),
+        status,
+    )
+    .into_response()
+}
+
+/// Produces a closure that will transform applicable errors into a problem details JSON object.
+/// (See RFC 7807) The returned closure is meant to be used in a warp `map` filter.
```
BranLwyd, line 299: QQ: why not a `recover` filter? that seems to be the dedicated "I am handling an error" filter, and the [docs](https://docs.rs/warp/0.3.2/warp/reject/fn.custom.html) seem to indicate it's preferred -- but I'll admit I'm not a `warp` expert just yet.

BranLwyd (approved):

janus_server/src/aggregator.rs
divergentdave: [Elsewhere](https://docs.rs/warp/0.3.2/warp/trait.Filter.html#method.and_then) the docs say that rejections are for route-related issues, and not application-level errors. So, `warp::path` rejects with a "not found" reason when a request's path doesn't match its argument, `warp::get()` rejects with a "method not allowed" reason when a request's method doesn't match, and the `or` combinator will try evaluating its second filter if the first filter rejects a request. As multiple chains of filters may be tried while servicing a request, rejections are stored in a linked list of trait objects. Filters that recover from rejection walk the stack, trying to cast each rejection to a candidate type. For cases where we encounter a PPM-level error, passing a `Result` between two of our filters correctly encodes the behavior we want, because we don't need the filters to unwind and try the other possible routes we have. Plus, once the types are worked out, it's more ergonomic than using `Rejection::find()` on the stack.

janus_server/src/config.rs
```diff
@@ -38,6 +39,8 @@ pub struct AggregatorConfig {
     pub listen_address: SocketAddr,
     /// The aggregator's database configuration.
     pub database: DbConfig,
+    /// The aggregator's public HTTPS endpoint, which will be advertised in error responses.
+    pub https_endpoint: Url,
```
tgeoghegan, line 43: Shouldn't this come from the task parameters, since HPKE config and the rest of the API has to be advertised relative to the aggregator endpoint there?

tgeoghegan (changes requested):

janus_server/src/config.rs
divergentdave: Based on the current text of the interop draft, I don't think so, as one HTTPS endpoint may service multiple tasks. It's true that this endpoint should match either the first or second endpoint in each task's configuration, but I think it still makes sense to provide this as part of the server-wide configuration, before the first task is loaded in.

tgeoghegan: I don't agree -- this just introduces the possibility that the URL in the config will disagree with the URL in the task parameters. In any case, [RFC 7807 section 3.1](https://datatracker.ietf.org/doc/html/rfc7807#section-3.1) makes it clear we can provide a relative URI here, which means we can avoid putting absolute URIs here at all. But we can revisit this as protocol text evolves.

tgeoghegan (approved):

janus_server/src/config.rs
divergentdave: Good point! I like that

PR #32, opened 2022-03-23
janus_server: implement upload in client

tgeoghegan:
Adds an implementation of `upload` to `client::Client`, which is now
generic over `prio::vdaf::Client`.

### DIFF
--- a/janus_server/src/client.rs
+++ b/janus_server/src/client.rs
@@ -2,10 +2,14 @@
 
 use crate::{
     hpke::{HpkeSender, Label},
-    message::{HpkeConfig, Role, TaskId},
+    message::{HpkeCiphertext, HpkeConfig, Nonce, Report, Role, TaskId, Time},
+    time::Clock,
 };
 use http::StatusCode;
-use prio::codec::Decode;
+use prio::{
+    codec::{Decode, Encode},
+    vdaf::Client as VdafClient,
+};
 use std::io::Cursor;
 use url::Url;
 
@@ -13,12 +17,16 @@ use url::Url;
 pub enum Error {
     #[error("HTTP client error: {0}")]
     HttpClient(#[from] reqwest::Error),
-    #[error("Codec error")]
+    #[error("Codec error: {0}")]
     Codec(#[from] prio::codec::CodecError),
     #[error("HTTP response status {0}")]
     Http(StatusCode),
     #[error("URL parse: {0}")]
     Url(#[from] url::ParseError),
+    #[error("VDAF error: {0}")]
+    Vdaf(#[from] prio::vdaf::VdafError),
+    #[error("HPKE error: {0}")]
+    Hpke(#[from] crate::hpke::Error),
 }
 
 static CLIENT_USER_AGENT: &str = concat!(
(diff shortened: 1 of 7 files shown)


janus_server/src/client.rs
```diff
@@ -29,63 +37,215 @@ static CLIENT_USER_AGENT: &str = concat!(
     "client"
 );
 
+pub async fn aggregator_hpke_sender(
```
tgeoghegan, line 40: This function and `default_http_client` used to be associated functions on `struct Client`. I moved them up to the module so that they could be used without worrying about the generic type parameters on `Client`, since none of those are relevant to these functions.

janus_server/src/datastore.rs
```diff
@@ -76,7 +76,8 @@ pub struct Transaction<'a> {
 impl Transaction<'_> {
     // TODO(brandon): implement basic getters/putters for all types
 
-    #[cfg(test)]
+    // This is pub to be used in integration tests
+    #[doc(hidden)]
```
tgeoghegan, line 80: This is bad, as we discussed in another PR. I'm happy to hold off on this change until we do #18, and then I can rebase on that.

janus_server/src/client.rs
```diff
@@ -29,63 +37,215 @@ static CLIENT_USER_AGENT: &str = concat!(
...
+            time: Time::from_naive_date_time(self.clock.now()),
+            rand: rand::random(),
+        };
+
+        // No extensions supported yet
+        let extensions = vec![];
+
+        let associated_data = Report::associated_data(nonce, &extensions);
+
+        // Leader's share MUST be the first in the report. We assume/guess that
```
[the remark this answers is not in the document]
BranLwyd: (though fwiw, I couldn't find verbiage in the VDAF spec that for vectored data, data at the same index has to be sent to the same participant in all cases -- `run_vdaf` maintains this, but I'm not sure if it's normative)

janus_server/src/client.rs
```diff
@@ -29,63 +37,215 @@ static CLIENT_USER_AGENT: &str = concat!(
...
+    use prio::vdaf::prio3::{Prio3Aes128Count, Prio3Aes128Sum};
 
-        Ok(HpkeSender {
+    fn setup_client<V: VdafClient>(
+        vdaf_client: V,
+        public_parameter: V::PublicParam,
+    ) -> Client<V, V::PublicParam, MockClock> {
+        let task_id = TaskId::random();
+        let clock = MockClock::default();
+        let leader_hpke_recipient =
```
BranLwyd, line 182: nit: it looks like there are a number of parameters shared between the `HpkeRecipient` and the `HpkeSender` that I suppose need to match up. Could we refactor creation of the `HpkeSender` so that it's created from a `TaskId` & `HpkeRecipient`? e.g. `HpkeSender::new(task_id, recipient)`, or perhaps `HpkeSender::to` if that name isn't too generic.

janus_server/src/datastore.rs
BranLwyd: ah well. this will eventually be expanded into a real `pub` method, anyway...

janus_server/src/client.rs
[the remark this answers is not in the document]
tgeoghegan: >In PPM, I think we need to care about keeping the order of the vectors generated by the underlying VDAF consistent, because the VDAF operations expect them to be consistent.

I agree with this, but we don't have text in either the PPM or VDAF specifications that declares this requirement.

Let's say that you call `Vdaf.measurement_to_shares` and it returns `[uncompressed_share, compressed_share]`. I don't think that there's anything in the text of VDAF or PPM currently that stops an implementation from sending `compressed_share` to the leader and `uncompressed_share` to the helper. This will fail, because the leader is expecting uncompressed shares and will reject the compressed shares for being the wrong size (same deal on the helper side). But the implementation that does this still complies with the specification.

To be clear, I don't think this is a problem in `janus` -- our implementation is doing the right thing. I want the protocol text to rule out this problem, hence the issue I filed on VDAF.

janus_server/src/client.rs
tgeoghegan: I've added `HpkeSender::from_recipient` and `HpkeSender::new`. The former does what you described (though it doesn't take `TaskId` because that can be copied from the `HpkeRecipient`). I didn't implement it as trait `From` because AFAICT we don't yet have any use case for generically transforming an `HpkeRecipient` into an `HpkeSender`, and this way callers don't ever have to worry about disambiguating which `From` implementation they want. My intuition and taste on when to implement traits like `From` is still evolving.

`HpkeSender::new` is a pretty dopey function, but it allowed me to make all the fields on `HpkeSender` private, which I think is defensible. I kinda wish Rust had a `write-only` sort of visibility modifier, so that struct definitions could be used to create trivial constructors but then forbid later mutation or reading of fields.

BranLwyd: I'm totally OK with not doing a From implementation; I think a From implementation on `HpkeRecipient` would consume the recipient, which isn't what we want in this case. (though I think you could do `From<&HpkeRecipient>`? but that seems pretty unnatural to me, though my own intuition is also not very strong)

+1 to making fields private, even with the additional overhead of needing to write a static constructor method & possibly accessors. If nothing else, it might ease refactorings later, and will make it possible to enforce invariants on the internal structures. I should probably go through and apply this everywhere... we've got a lot of structs that currently make their internal state public without a good reason (other than it's easy).

BranLwyd (approved):

Issue #22, opened 2022-03-15
SQL schema: reconsider type of `report_aggregations.error_code`.


tgeoghegan:
Recording changes to the storage schema I think we will need.

- [x] enum `VDAF_IDENTIFIER` needs more entries to align with VDAFs instantiated in https://datatracker.ietf.org/doc/draft-patton-cfrg-vdaf/
```
-- Identifies a particular VDAF.
CREATE TYPE VDAF_IDENTIFIER AS ENUM(
    'PRIO3AES128COUNT',
    'PRIO3AES128SUM',
    'PRIO3AES128HISTOGRAM',
    'POPLAR1'
);
```
- [x] VDAF verification parameters are secret (https://datatracker.ietf.org/doc/draft-patton-cfrg-vdaf/ section 4.1) and also have to be rotated at some interval. Therefore `vdaf_verify_param` should not be in table `tasks`.
- [x] `tasks.min_batch_duration` is an `INTERVAL`, which is awkward and can be surprising ([discussion](https://github.com/divviup/janus/pull/33#discussion_r833787897)). Consider storing that as a BIGINT (better matches how we will use it in code anyway).

divergentdave:
`report_aggregations.error_code` could stand to be a nullable `smallint` or nullable `ENUM`, rather than a nullable `bigint`.

divergentdave:
I think we should add a `UNIQUE(task_id)` constraint on the tasks table.

tgeoghegan:
I think we should lump this in with the other [draft-ietf-ppm-dap-02 changes](https://github.com/divviup/janus/milestone/1) since this means changing the DB schema, which is a breaking change until we have a DB migrations story.

PR #33, opened 2022-03-23
janus_server: introduce `struct TaskParameters`

tgeoghegan:
Define a struct for PPM task parameters, and use it to create
`client::Client` instead of a handful of different parameters. This also
updates `datastore` to use a `TaskParameter` instead of a `TaskId`,
though I got stuck figuring out how to render all the fields of
`TaskParameters` into a prepared statement.

### DIFF
--- a/janus_server/src/datastore.rs
+++ b/janus_server/src/datastore.rs
@@ -1,6 +1,9 @@
 //! Janus datastore (durable storage) implementation.
 
-use crate::message::{Extension, HpkeCiphertext, Nonce, Report, TaskId, Time};
+use crate::{
+    message::{Extension, HpkeCiphertext, Nonce, Report, TaskId, Time},
+    task::TaskParameters,
+};
 use prio::codec::{decode_u16_items, encode_u16_items, CodecError, Decode, Encode};
 use std::{future::Future, io::Cursor, pin::Pin};
 use tokio_postgres::{error::SqlState, IsolationLevel, Row};
@@ -78,7 +81,8 @@ impl Transaction<'_> {
 
     // This is pub to be used in integration tests
     #[doc(hidden)]
-    pub async fn put_task(&self, task_id: TaskId) -> Result<(), Error> {
+    pub async fn put_task(&self, task: &TaskParameters) -> Result<(), Error> {
+        // TODO: interpolate values from `task` into prepared statement
         let stmt = self
             .tx
             .prepare_cached(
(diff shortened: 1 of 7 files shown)


janus_server/src/datastore.rs
```diff
@@ -78,7 +81,8 @@ impl Transaction<'_> {
 
     // This is pub to be used in integration tests
     #[doc(hidden)]
-    pub async fn put_task(&self, task_id: TaskId) -> Result<(), Error> {
+    pub async fn put_task(&self, task: &TaskParameters) -> Result<(), Error> {
+        // TODO: interpolate values from `task` into prepared statement
```
tgeoghegan, line 85: I had trouble doing this because I couldn't figure out how to render a `task::Vdaf` into the `VDAF_IDENTIFIER` SQL enum, as well as creating a Postgres `INTERVAL` from `Duration`.

divergentdave: For the enum, we likely want to use [ToSql/FromSql derive macros](https://docs.rs/postgres-types/latest/postgres_types/#enums) along with [some renaming attribute macros](https://docs.rs/postgres-types/latest/postgres_types/#naming) to handle different case conventions. I don't see any conversions for `INTERVAL`, so we may have to implement `ToSql`/`FromSql` on a newtype of Duration ourselves, for `Type::INTERVAL`.

divergentdave: Ah, looking at the rust-postgres issue tracker, we actually may want to change what we're using in the database. `INTERVAL` has multiple separate fields (it takes up 16 bytes total) and it can represent ambiguous things like "one year plus one month". Maybe we write `ToSql`/`FromSql` for our own purposes to be strict enough that every field for months and upwards must always be zero, or maybe we make a DOMAIN  of BIGINT to represent a duration in seconds, and use that.

tgeoghegan: Yes, and at that point we have to pull in crate `postgres_types`. I decided to "cut my losses" and stop this PR from getting too big.

BranLwyd: https://www.postgresql.org/docs/current/datatype-datetime.html:
> interval internally stores only three integer units (months, days, microseconds), fractional units must be spilled to smaller units.

!! I support changing `min_batch_duration` to e.g. a BIGINT counting a number of seconds.

BranLwyd: Approving this PR, I suppose the datastore changes don't need to go here.

BranLwyd (approved):

janus_server/src/datastore.rs
tgeoghegan: I made a note about `min_batch_duration` in #22.

PR #37, opened 2022-03-24
janus_server: load aggregator config from YAML

### LINKED ISSUE #20: `janus_server`: configuration
We need to decide how we will provide configuration values to `janus_server` workers.

tgeoghegan:
Adds aggregator configuration struct with ser/de from YAML, and plumbs
it into `src/bin/aggregator.rs`. Adds derivations of serde traits on
various existing structs and aliases needed to construct
`AggregatorConfig`. A sample aggregator config file is provided through
doccomment on `config::AggregatorConfig`.

Fixes #20


tgeoghegan:
Some sample output from running the `aggregator` binary target:

Help text, including version:
```
$> ./target/debug/aggregator --help
janus-aggregator 0.1.0
PPM aggregator server

USAGE:
    aggregator --config-file <config-file> --hpke-private-key <HEX>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --config-file <config-file>    path to configuration file [env: CONFIG_FILE=]
        --hpke-private-key <HEX>       hex encoded private key for decrypting reports [env: HPKE_PRIVATE_KEY]

```

Invalid config file:
```
$> HPKE_PRIVATE_KEY=ffffffffff ./target/debug/aggregator --config-file /no/such/file.yaml
Error: failed to open config file: "/no/such/file.yaml"

Caused by:
    No such file or directory (os error 2)
```

Starting up successfully:
```
$> RUST_LOG=info HPKE_PRIVATE_KEY=ffffffffff ./target/debug/aggregator --config-file /tmp/config.yaml 
  2022-03-24T20:56:48.715377Z  INFO aggregator: starting aggregator, options: Options { config_file: "/tmp/config.yaml", hpke_private_key: "redacted" }
    at src/bin/aggregator.rs:72 on ThreadId(1)

  2022-03-24T20:56:48.715841Z  INFO aggregator: configuration, config: AggregatorConfig { listen_address: 0.0.0.0:8080, database: DbConfig { url: Url { scheme: "postgres", cannot_be_a_base: false, username: "postgres", password: Some("postgres"), host: Some(Domain("localhost")), port: Some(5432), path: "/postgres", query: None, fragment: None } }, task_id: 41524c07214502416f56b1258d97521b083483a0fbb082464e0741ace1403a2e, role: Leader, tolerable_clock_skew_seconds: Duration(10), hpke_config: HpkeConfig { id: HpkeConfigId(0), kem_id: X25519HkdfSha256, kdf_id: HkdfSha512, aead_id: ChaCha20Poly1305, public_key: a6f620ccc9d35164d92a4073727645f20329e1e1fe6bac7400c75a627c64783d } }
    at src/bin/aggregator.rs:81 on ThreadId(1)

  2022-03-24T20:56:48.730952Z  INFO aggregator: running aggregator, config.task_id: 41524c07214502416f56b1258d97521b083483a0fbb082464e0741ace1403a2e, bound_address: 0.0.0.0:8080
    at src/bin/aggregator.rs:113 on ThreadId(1)
```

janus_server/src/config.rs
```diff
@@ -0,0 +1,110 @@
...
+/// "#;
+///
+/// let _decoded: AggregatorConfig = serde_yaml::from_str(yaml_config).unwrap();
+/// ```
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub struct AggregatorConfig {
+    /// Address on which this server should listen for connections and serve its
+    /// API endpoints.
+    // TODO: Options for terminating TLS, unless that gets handled in a load
+    // balancer?
```
divergentdave, line 47: FWIW, GCP's HTTP(S) load balancers do terminate TLS. Traffic from the GFE to the VM can be over TLS, but it seems this is mainly intended for compliance reasons and HTTP/2 compatibility. GFE does no certificate validation on backend TLS connections. They mention there is "automatic network-level encryption" transparently applied between GFE and the VM. https://cloud.google.com/load-balancing/docs/ssl-certificates/encryption-to-the-backends#encryption-to-backends

janus_server/src/config.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub listen_address: SocketAddr,
+    /// The aggregator's database configuration.
+    pub database: DbConfig,
+    /// The PPM task this aggregator is participating in. Parameters for this
+    /// task must exist in the aggregator's database.
+    // TODO: if we want a single aggregator instance to participate in multiple
+    // tasks, then we should provide a list or map of per-task parameters
+    pub task_id: TaskId,
```
divergentdave, line 55: This is fine for now, but I imagine the server may eventually get its list of task IDs from the database.

divergentdave (approved):

janus_server/src/bin/aggregator.rs
```diff
@@ -1,60 +1,116 @@
...
+    // The bulk of configuration options are in the config file.
+    let config_file = File::open(&options.config_file)
+        .with_context(|| format!("failed to open config file: {:?}", options.config_file))?;
+
+    let config: AggregatorConfig = serde_yaml::from_reader(&config_file)
+        .with_context(|| format!("failed to parse config file: {:?}", options.config_file))?;
+
+    info!(?config, "configuration");
+
+    let cfg = Config::from_str(config.database.url.as_str()).with_context(|| {
```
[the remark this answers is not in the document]
tgeoghegan: Done, and also I removed the import of `tokio_postgres::Config`. The fully qualified symbol path makes this read more clearly.

janus_server/src/config.rs
tgeoghegan: That's an interesting thought. You envision that the aggregator would launch, discover all the tasks that are defined in the DB, and start serving requests for all of them? That makes a lot of sense.

janus_server/src/config.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// The PPM [`Role`] this aggregator plays. Should be either
+    /// [`Role::Leader`] or [`Role::Helper`].
+    pub role: Role,
```
BranLwyd, line 58: Following the above discussion about `task_id`, I think `role` should not be in the config either since it is a piece of task-specific configuration that should instead live in the DB.

janus_server/src/config.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// How much clock skew to allow between client and aggregator. Reports from
+    /// farther than this duration into the future will be rejected.
+    pub tolerable_clock_skew_seconds: Duration,
```
BranLwyd, line 61: [no changes requested] We haven't yet decided if clock skew is a global or task-specific setting, but if we decide it's task-specific we should move it out of this config too. I'm fine with keeping it right now since it's not clear what level of granularity is appropriate, and even if we decide to make this task-specific the global setting could still be useful as a default.

BranLwyd (changes requested):

tgeoghegan:
In light of what has been discussed in this PR and since our storage story is getting pretty robust, I'm going to extend this a fair bit to teach aggregators to fetch quite a bit of task related parameters from the database.

At startup, aggregators will query the database to get all the rows in table `tasks`, and then create an `Aggregator` for each. I'm also going to add a table `hpke_configs` where we store the public HPKE config for aggregators, with an FK to `tasks` (there can be multiple HPKE configs per task due to key rotation).

I'm still mulling what to do about per-task secrets (like HPKE private keys or VDAF verification parameters).

BranLwyd:
> In light of what has been discussed in this PR and since our storage story is getting pretty robust, I'm going to extend this a fair bit to teach aggregators to fetch quite a bit of task related parameters from the database.

This is a great idea -- consider doing it in a separate PR? (this request may come too late depending on how far you've gotten by the time you read it, no biggie if so :-))


> At startup, aggregators will query the database to get all the rows in table `tasks`, and then create an `Aggregator` for each. I'm also going to add a table `hpke_configs` where we store the public HPKE config for aggregators, with an FK to `tasks` (there can be multiple HPKE configs per task due to key rotation).

A few thoughts:
1. We will eventually need to think through how to deal with tasks that come into existence after the process starts--just loading tasks once at startup won't work for that. One solution would be to read the task ID from the request, then load the task information from the DB (probably with caching to minimize impact on the DB?). I guess I don't object to load-everything-once-on-startup for now, but if there's a way for your implementation to easily allow extension to loading tasks from the DB on-the-fly, please take it to avoid later churn. [for example, the `/aggregate`-init PR looks up HMAC key by task ID, which would allow eventually looking up a task-specific secret, even though for now the "lookup" implementation just returns the one hardcoded HMAC key in use]
1. In the current PPM spec, HPKE configs are aggregator-specific rather than (aggregator, task)-specific. This is because the client's `/hpke_config` request doesn't specify a task -- so aggregators would have no way to determine which config to return. If that's not intentional, we may want to update the spec.


> I'm still mulling what to do about per-task secrets (like HPKE private keys or VDAF verification parameters).

Indeed. All of my thinking so far is that we'll use GCP's Secret Store, but perhaps we can have a constant number of "internal" secrets and use those to encrypt/decrypt application-level secrets like the HPKE private keys/VDAF verification parameters you mention. e.g. GCP secret store has an AEAD key, we use the AEAD key to encrypt/decrypt other secrets, storing the ciphertext in our normal DB.

tgeoghegan:
Given our recent discussion here and in Slack, I've re-scoped this PR so that we load a config file that just contains a listen address and a database connection URL. Everything else remains either randomly generated (HPKE config and task ID) or read from a command line arg (I added a `--role` flag which will go away when we start reading task parameters from the database).

divergentdave (approved):

BranLwyd (approved):

PR #38, opened 2022-03-24
janus_server: trace event output options

tgeoghegan:
Adds configuration options to the `tracing` subscriber installed by
`trace::install_subscriber`, allowing output to be forced to JSON and
the installation of a test writer that lets the test harness capture
event output.

WIP because I can't figure out how to optionally configure a layer.


divergentdave:
I may have confused warp filters with tracing-subscriber layers yesterday, but at any rate, `Option<impl Layer>` also implements `Layer`, which should allow constructing one big layer chain with a fixed type, and then swapping out None/Some at runtime. https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/index.html#runtime-configuration-with-layers

janus_server/src/trace.rs
```diff
@@ -1,40 +1,77 @@
...
+
+    if config.use_test_writer {
+        fmt_layer = fmt_layer.with_test_writer();
+    }
+
+    // If stderr is not a tty or if forced by config, output logs as JSON
+    // structures
+    if atty::isnt(Stream::Stdout) || config.force_json_output {
+        fmt_layer = fmt_layer.json();
+    };
```
tgeoghegan, line 44: I am having trouble getting this part to work and could use some help. You can't do this because `fmt_layer` and `fmt_layer.with_test_writer()` don't have the same type. Normally I would deal with this by boxing the values ([as I did with `slog::Drain`s in `prio-server`](https://github.com/divviup/prio-server/blob/9faa6e2566bb9c5b590e85d223f08e34b8e1d95a/facilitator/src/logging.rs#L130)) but that doesn't work here.

For one thing, even if I could box these into `Box<dyn Layer<...>>`, [`trait Layer`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.Layer.html#) doesn't have the same methods as [`struct Layer`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#) that would let me chain on a call to `.json()`. 

I must be missing something about how to compose layers together. I also tried creating multiple layers and layering them together into [`Layered`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/struct.Layered.html), but the result is that you get events going to multiple places -- stdout with `pretty` and then also stdout with `json`.

janus_server/src/trace.rs
```diff
@@ -1,40 +1,88 @@
...
+pub fn install_subscriber(config: &TraceConfiguration) -> Result<(), Error> {
+    // If stdout is not a tty or if forced by config, output logs as JSON
+    // structures
+    let output_json = atty::isnt(Stream::Stdout) || config.force_json_output;
+
+    let (pretty_layer, json_layer, test_layer) = match (output_json, config.use_test_writer) {
+        (true, false) => (None, Some(base_layer().json()), None),
+        (false, false) => (Some(base_layer().pretty()), None, None),
+        (_, true) => (None, None, Some(base_layer().pretty().with_test_writer())),
+    };
```
tgeoghegan, line 53: It's not especially elegant, but I was able to get this working without repeating too much code.

janus_server/src/trace.rs
```diff
@@ -1,40 +1,88 @@
...
+}
+
+/// Configuration for the tracing subscriber.
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct TraceConfiguration {
+    /// If true, uses a [`tracing_subscriber::fmt::TestWriter`] to capture trace
+    /// events when running tests
+    #[serde(default)]
+    pub use_test_writer: bool,
+    /// If true OR if stdout is a tty, trace events are output in JSON format by
```
divergentdave, line 24:
```suggestion
    /// If true OR if stdout is not a tty, trace events are output in JSON format by
```

divergentdave (approved):

divergentdave:
Ah, this was merged into `timg/config` after #37 was already merged into main and closed

PR #41, opened 2022-03-28
Implement SQL mapping for VDAF enum

divergentdave:
This is a quick follow-up to #33 to implement `ToSql` and `FromSql` for our `Vdaf` enum. I added variants to the schema's enum to reflect the multiple instantiations of prio3.

Note that the attribute macro setting the database name of the enum uses lowercase, `"vdaf_identifier"`. This is what the derive macro insisted on, I take it that SQL enum names are case-insensitive like most other objects. The crate must normalize the name to lowercase, and then the generated implementations do their type checks against that.

### DIFF
--- a/janus_server/src/datastore.rs
+++ b/janus_server/src/datastore.rs
@@ -95,8 +95,8 @@ impl Transaction<'_> {
             .execute(
                 &stmt,
                 &[
-                    &&task.id.0[..],         // id
-                    &Vdaf::Prio3Aes128Count, // vdaf
+                    &&task.id.0[..], // id
+                    &task.vdaf,      // vdaf
                 ],
             )
             .await?;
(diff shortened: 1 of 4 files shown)


janus_server/src/datastore.rs
```diff
@@ -88,11 +88,17 @@ impl Transaction<'_> {
...
+                VALUES ($1, 0, '{}', $2, '', 0, 0, INTERVAL '0', '')",
             )
             .await?;
         self.tx
-            .execute(&stmt, &[/* task_id */ &&task.id.0[..]])
+            .execute(
+                &stmt,
+                &[
+                    &&task.id.0[..],         // id
+                    &Vdaf::Prio3Aes128Count, // vdaf
```
tgeoghegan, line 99: Surely this should use `task._vdaf` (and of course take the opportunity to rename the field to `vdaf` since it will now be read). Also the tests should be updated to verify that the expected `vdaf_identifier` value is written.

tgeoghegan (changes requested):

janus_server/src/datastore.rs
```diff
@@ -210,6 +216,19 @@ impl Transaction<'_> {
             .await?;
         Ok(row.get("id"))
     }
+
+    /// This is a test-only method that is used to test round-tripping of values.
+    /// TODO: remove this once the tasks table is finalized, and there's a method to retrieve an
+    /// entire `TaskParameters` from the database.
+    #[allow(unused)]
```
tgeoghegan, line 223: Does `#[cfg(test)]` not work here?

tgeoghegan (approved):

janus_server/src/datastore.rs
divergentdave: That works, with some use juggling.

PR #45, opened 2022-03-30
tokio-console support

divergentdave:
This adds a configuration option to enable tokio-console support. Note that I didn't add `--cfg tokio_unstable` to RUSTFLAGS anywhere, so you'll need to rebuild with that, in addition to setting a variable in the configuration, to use it.

This PR will look funny because I am basing it off of #38, which was not merged into main. It appears there's a conflict as well.

### DIFF
--- a/janus_server/src/trace.rs
+++ b/janus_server/src/trace.rs
@@ -48,6 +48,16 @@ fn base_layer<S>() -> tracing_subscriber::fmt::Layer<S> {
         .with_line_number(true)
 }
 
+/// Filter factory to create per-layer filters for stdout layers. The returned
+/// filter will prevent verbose runtime-related events from being printed to
+/// stdout.
+fn runtime_tracing_filter<S>() -> tracing_subscriber::filter::combinator::Not<Targets, S> {
+    Targets::new()
+        .with_target("tokio", tracing::Level::TRACE)
+        .with_target("runtime", tracing::Level::TRACE)
+        .not()
+}
+
 /// Configures and installs a tracing subscriber, to capture events logged with
 /// [`tracing::info`] and the like. Captured events are written to stdout, with
 /// formatting affected by the provided [`TraceConfiguration`].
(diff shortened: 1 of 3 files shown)


janus_server/src/trace.rs
```diff
@@ -47,19 +56,73 @@ pub fn install_subscriber(config: &TraceConfiguration) -> Result<(), Error> {
...
+
+    // Configure filters with RUST_LOG env var. Format discussed at
+    // https://docs.rs/tracing-subscriber/latest/tracing_subscriber/struct.EnvFilter.html
+    let mut global_filter = EnvFilter::from_default_env();
+
+    let console_layer = match &config.tokio_console {
+        true => {
+            global_filter = global_filter.add_directive("tokio=trace".parse().unwrap());
+            global_filter = global_filter.add_directive("runtime=trace".parse().unwrap());
+            Some(console_subscriber::spawn())
```
tgeoghegan, line 115: I think we're going to want to be able to customize the address that the console subscriber listens on so that we can do the corresponding Kubernetes network config. But that can land in a later PR.

janus_server/src/trace.rs
```diff
@@ -47,19 +56,73 @@ pub fn install_subscriber(config: &TraceConfiguration) -> Result<(), Error> {
...
+        .with_target("runtime", tracing::Level::TRACE)
+        .not();
+    let inverted_tokio_console_filter_2 = Targets::new()
+        .with_target("tokio", tracing::Level::TRACE)
+        .with_target("runtime", tracing::Level::TRACE)
+        .not();
+    let inverted_tokio_console_filter_3 = Targets::new()
+        .with_target("tokio", tracing::Level::TRACE)
+        .with_target("runtime", tracing::Level::TRACE)
+        .not();
```
tgeoghegan, line 74: nit: I think you could factor this into a function and then call `.with_filter(make_console_target())` in the `match` block below.

tgeoghegan (approved):
Looks good, though of course it needs to be rebased. I also think this feature should be written up a bit in a README, especially since it requires building with a special flag.

BranLwyd (approved):

Issue #48, opened 2022-03-31
Secure storage of tasks and related values in database


tgeoghegan:
We've been discussing how to handle discovering PPM tasks and storing secret values in a variety of venues, so I'm gathering up what I believe to be our consensus plan in this issue, and I will use it to track the phases of implementation.

- [ ] Aggregators should discover defined tasks in the database at startup and set up handlers for each. 

- [ ] A task's HPKE config, HPKE private key, VDAF verify parameter and aggregator auth key can be rotated independently of the task. Those should be stored in a separate table to allow for a many:1 relationship.

- [ ] Secret values (HPKE private key, VDAF verify parameter and aggregator auth key) should be stored encrypted under a key provided to the aggregator in its environment (likely from a Kubernetes secret). We need to devise a scheme for protecting secrets in the database in a way that binds them to a task.

tgeoghegan:
We have to decide how a single aggregator instance will serve requests for multiple PPM tasks. PPM defines endpoints like `[aggregator]/hpke_config` or `[leader]/upload`, which leaves us free to construct `[aggregator]` or `[leader]` however want, potentially incorporating the task ID. I think we should use the base64 encoded task ID as a path parameter in endpoints, such that if we have our API at `api.divviup.org/something/or/other`, then for task ID `b"12345678901234567890123456789012"`, we have `api.divviup.org/something/or/other/MTIzNDU2Nzg5MDEyMzQ1Njc4OTAxMjM0NTY3ODkwMTIK/upload` and for task ID `b"09876543210987654321098765432109"`, we have `api.divviup.org/something/or/other/MDk4NzY1NDMyMTA5ODc2NTQzMjEwOTg3NjU0MzIxMDkK/upload`.

BranLwyd:
Why not have the client send a "request" message including the task ID? This is consistent with all other request/response patterns, and this would also allow the task ID to be sent w/o base64 encoding (probably negligible savings).

Another way of looking at it: if we want to put the task ID in the request path, we should drop it from all request messages (but I don't think that's a palatable change).

BranLwyd:
Also, one nit:
> We need to devise a scheme for protecting secrets in the database in a way that binds them to a task.

IMO, we should aim for better than task-level binding: otherwise, malicious storage could swap encrypted values between different "secret" values for the same task, e.g. it could swap an `agg_auth_key` with a `verify_param`. This is _probably_ unexploitable, but why take the chance?

[...]

PR #50, opened 2022-04-01
janus_server: write TaskParameters to database

tgeoghegan:
Adds support for writing all the fields of a `struct TaskParameters`
into the `tasks` table in the database, with new fields/columns for the
aggregator auth key, HPKE config and HPKE private key. `tasks.ord` is
also now `role` and is an enumerated type with cases for the various PPM
protocol roles.

Part of #48

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -9,19 +9,30 @@ CREATE TYPE VDAF_IDENTIFIER AS ENUM(
     'POPLAR1'
 );
 
+-- Identifies which aggregator role is being played for this task.
+CREATE TYPE AGGREGATOR_ROLE AS ENUM(
+    'LEADER',
+    'HELPER'
+);
+
 -- Corresponds to a PPM task, containing static data associated with the task.
 CREATE TABLE tasks(
     id                     BYTEA PRIMARY KEY,         -- 32-byte TaskID as defined by the PPM specification
-    ord                    BIGINT NOT NULL,           -- the order of this aggregator for this task; 0 is leader, 1 or larger is helper
+    aggregator_role        AGGREGATOR_ROLE NOT NULL,  -- the role of this aggregator for this task
     aggregator_endpoints   TEXT[] NOT NULL,           -- aggregator HTTPS endpoints, leader first
     vdaf                   VDAF_IDENTIFIER NOT NULL,  -- the VDAF in use for this task
     vdaf_verify_param      BYTEA NOT NULL,            -- the VDAF verify parameter (opaque VDAF message)
     max_batch_lifetime     BIGINT NOT NULL,           -- the maximum number of times a given batch may be collected
     min_batch_size         BIGINT NOT NULL,           -- the minimum number of reports in a batch to allow it to be collected
-    min_batch_duration     INTERVAL NOT NULL,         -- the duration of a single batch interval
-    collector_hpke_config  BYTEA NOT NULL             -- the HPKE config of the helper (encoded HpkeConfig message)
+    min_batch_duration     BIGINT NOT NULL,           -- the minimum duration in seconds of a single batch interval
+    collector_hpke_config  BYTEA NOT NULL,            -- the HPKE config of the collector (encoded HpkeConfig message)
+    agg_auth_key           BYTEA NOT NULL,            -- HMAC key used by this aggregator to authenticate messages to/from the other aggregator
+    hpke_config            BYTEA NOT NULL,            -- the HPKE config of this aggregator (encoded HpkeConfig message)
+    hpke_private_key       BYTEA NOT NULL             -- private key corresponding to hpke_config (hpke::HpkePrivateKey)
 
-    -- TODO(brandon): include agg_auth_key if we decide it shouldn't go in a secret store
+    -- TODO(timg): move vdaf_verify_param, agg_auth_key, hpke_config, hpke_private_key to new
+    -- tables with many:1 relationships to tasks to allow for rotation of secrets
+    -- TODO(timg): vdaf_verify_params, agg_auth_key and hpke_private_key should be encrypted
 );
 
 -- Individual reports received from clients.
@@ -29,7 +40,7 @@ CREATE TABLE client_reports(
     id            BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY, -- artificial ID, internal-only
     task_id       BYTEA NOT NULL,      -- task ID the report is associated with
     nonce_time    TIMESTAMP NOT NULL,  -- timestamp from nonce
-    nonce_rand    BIGINT NOT NULL,     -- random value from nonce
+    nonce_rand    BYTEA NOT NULL,      -- random value from nonce
     extensions    BYTEA,               -- encoded sequence of Extension messages (populated for leader only)
     input_shares  BYTEA,               -- encoded sequence of HpkeCiphertext messages (populated for leader only)
 
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -175,7 +175,7 @@ where
         let leader_report = &report.encrypted_input_shares[0];
 
         // §4.2.2: verify that the report's HPKE config ID is known
-        if leader_report.config_id != self.report_recipient.config.id {
+        if leader_report.config_id != self.report_recipient.config().id {
             warn!(
                 config_id = ?leader_report.config_id,
                 "unknown HPKE config ID"
@@ -621,7 +621,7 @@ where
         return Err(Error::InvalidConfiguration("role is not an aggregator"));
     }
 
-    let hpke_config_encoded = hpke_recipient.config.get_encoded();
+    let hpke_config_encoded = hpke_recipient.config().get_encoded();
 
     let aggregator = Arc::new(Aggregator::new(
         vdaf,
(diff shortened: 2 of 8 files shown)


janus_server/src/task.rs
```diff
@@ -48,17 +89,24 @@ pub struct TaskParameters {
...
-    /// HPKE configuration for the collector
-    _collector_hpke_config: HpkeConfig,
+    pub(crate) min_batch_duration: Duration,
+    /// HPKE configuration for the collector.
+    pub(crate) collector_hpke_config: HpkeConfig,
+    /// Key used to authenticate messages sent to or received from the other
+    /// aggregators.
+    pub(crate) agg_auth_key: AggregatorAuthKey,
+    /// HPKE recipient used by this aggregator to decrypt client reports
+    pub(crate) hpke_recipient: HpkeRecipient,
```
tgeoghegan, line 109: This is kind of awkward, because `TaskParameters` now contains several values that not only not useful to the client but should be secret from the client, and yet `client::Client` has a field `TaskParameters`. I think the answer might be to define multiple structs, each representing a different protocol participant's view of the task parameters, and a further one in `datastore::models` representing a row in table `tasks`. However this PR is already large and I wanted to get opinions on how to proceed before doing that.

janus_server/src/aggregator.rs
```diff
@@ -1784,8 +1798,13 @@ mod tests {
         datastore
             .run_tx(|tx| {
                 Box::pin(async move {
-                    tx.put_task(&TaskParameters::new_dummy(task_id, Vec::new()))
-                        .await
+                    tx.put_task(&TaskParameters::new_dummy(
+                        task_id,
+                        Vec::new(),
+                        Vdaf::Prio3Aes128Count,
```
tgeoghegan, line 1804: I considered adding a `Fake1` variant to `task::Vdaf` to be used in tests like this that use `aggregator::test_util::fake::Vdaf`. But then you have to add a corresponding value to the enum definition in `db/schema.sql`, and there's no (obvious) way to only do that in test setups.

divergentdave: How about we add a second SQL script, an overlay just for testing environments, that runs `ALTER TYPE vdaf ADD VALUE 'FAKE1';`?

janus_server/src/task.rs
BranLwyd: +1 to defining per-participant task parameters e.g. `client::TaskParameters`, `aggregator::TaskParameters`, etc.

janus_server/src/aggregator.rs
BranLwyd: That would be a pretty good way to do things (though since the fake VDAF is configurable, we'd probably need e.g. `FAKE_SUCCESS`, `FAKE_FAILS_PREP_INIT`, `FAKE_FALES_PREP_STEP`, ... based on the needs of the tests).

But I think we should wait to do this until we have a concrete test case that needs it. Our eventual end-to-end integration test should ideally use a real VDAF IMO.

db/schema.sql
```diff
@@ -9,19 +9,31 @@ CREATE TYPE VDAF_IDENTIFIER AS ENUM(
...
+    'CLIENT',
+    'LEADER',
+    'HELPER'
+);
+
 -- Corresponds to a PPM task, containing static data associated with the task.
 CREATE TABLE tasks(
     id                     BYTEA PRIMARY KEY,         -- 32-byte TaskID as defined by the PPM specification
-    ord                    BIGINT NOT NULL,           -- the order of this aggregator for this task; 0 is leader, 1 or larger is helper
+    role                   ROLE NOT NULL,             -- the role of this aggregator for this task
```
BranLwyd, line 22: I used an integer to allow for eventual expansion to multi-helper scenarios. That's probably preparing for a future that will never be, so this change makes sense.

But can we constrain this role value to be either `LEADER` or `HELPER`? These are specifically task parameters for `janus`, an aggregator, so the other role values do not make sense here.

janus_server/src/datastore.rs
```diff
@@ -91,38 +98,101 @@ impl Transaction<'_> {
...
-    /// This is a test-only method that is used to test round-tripping of values.
-    /// TODO: remove this once the tasks table is finalized, and there's a method to retrieve an
-    /// entire `TaskParameters` from the database.
+    /// Fetch the task parameters corresponing to the provided `task_id`.
+    //
+    // Only available in test configs for now, but will soon be used by
+    // aggregators to discover tasks from the database.
     #[cfg(test)]
-    async fn get_task_vdaf_by_id(&self, task_id: &TaskId) -> Result<crate::task::Vdaf, Error> {
+    pub(crate) async fn get_task_by_id(&self, task_id: &TaskId) -> Result<TaskParameters, Error> {
```
BranLwyd, line 147: nit: apologies, I missed this before: `TaskId` values should be passed around by-value rather than by-reference -- they implement `Copy`. (as do other value-type identifiers, like `AggregationJobId`)

janus_server/src/datastore.rs
```diff
@@ -461,6 +531,27 @@ fn single_row(rows: Vec<Row>) -> Result<Row, Error> {
...
+        I: RowIndex + Display;
+}
+
+impl RowExt for Row {
+    fn get_bigint_as_u64<I>(&self, idx: I) -> Result<u64, Error>
+    where
+        I: RowIndex + Display,
+    {
+        let bigint: i64 = self.try_get(idx)?;
+        Ok(u64::try_from(bigint)?)
```
BranLwyd, line 551: I think we really do want to do `nonce_rand as u64` rather than `u64::try_from(nonce_rand)`, when converting `nonce_rand` specifically. (this may not hold for other `u64` values that are stored as `BIGINT` in the DB.)

`u64::try_from(some_i64)` will fail if `some_i64` has a numeric value that is outside the range of what can be represented as `u64`. OTOH, `some_i64 as u64` simply reinterprets the bytes (and so may not give the same numeric value).

Since this part of the nonce is random, about half of the possible `u64` values we might generate will not be representable as an `i64`. Prior to this change, we'd just convert back & forth by reinterpreting the bytes, since we ultimately don't care about the specific numeric value, only that we get a consistent result & can check equality. With this change, we'll error out whenever we happen to generate a random part of the nonce that is outside the numeric range of `i64`.

All that said: this is perhaps too confusing/error-prone. (If nothing else, for a given `nonce_rand` value we'd need to refer to it as one numeric value in Rust, and a different numeric value in Postgres.) Perhaps we should instead represent the random part of the nonce as e.g. `[u8; 8]` in Rust & as a `BYTEA` in the DB? This typing is also semantically closer to what `nonce_rand` actually is, i.e. an opaque, uninterpreted random value.

Rust playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d1465a9869f0b23c6fe62b267c0b35ad

janus_server/src/datastore.rs
```diff
@@ -485,6 +576,12 @@ pub enum Error {
     /// will never be generated by the datastore library itself.
     #[error(transparent)]
     User(#[from] Box<dyn std::error::Error + Send + Sync>),
+    #[error("URL parse error")]
+    Url(#[from] url::ParseError),
+    #[error("Invalid task parameters: {0}")]
+    TaskParameters(#[from] crate::task::Error),
+    #[error("Integer conversion failed: {0}")]
+    TryFromInt(#[from] std::num::TryFromIntError),
```
BranLwyd, line 584: (very pedantic, sorry)

```suggestion
    #[error("invalid task parameters: {0}")]
    TaskParameters(#[from] crate::task::Error),
    #[error("integer conversion failed: {0}")]
    TryFromInt(#[from] std::num::TryFromIntError),
```

The usual practice with error messages I've seen is to not capitalize them, so you get error log lines that read like "Couldn't frob the widget: couldn't find widget frobber: couldn't access frob database: connection reset by peer" rather than "Couldn't frob the widget: Couldn't find widget frobber: Couldn't access frob database: connection reset by peer." Separately, log lines _should_ start capitalized.

BranLwyd (changes requested):

janus_server/src/datastore.rs
tgeoghegan: You have pointed this out to me three times now, and I promise that this time I will remember it. Thanks!

db/schema.sql
tgeoghegan: Sure, I removed the `COLLECTOR` and `CLIENT` values from the SQL enum. However now the SQL enum doesn't line up with `messages::Role`, so to reduce the risk of someone in the future assuming that the 0th value of the SQL enum is the same as the 0th value of `messages::Role`, I renamed the SQL enum to `AGGREGATOR_ROLE` and introduced a corresponding type in `datastore::models`.

janus_server/src/datastore.rs
tgeoghegan: I like that idea, because TBH all these signed-to-unsigned conversions make me nervous due to past trauma induced by [`CFNumberGetValue`](https://developer.apple.com/documentation/corefoundation/1543114-cfnumbergetvalue?language=objc).

janus_server/src/task.rs
tgeoghegan: I added `client::ClientParameters`, but stopped short of creating a new `aggregator::TaskParameters` or `datastore::model::TaskParameters`, because for the moment they'd be identical to `task::TaskParameters`. I think we can revisit that as the current `tasks` table gets normalized into `hpke_configs` and `aggregator_auth_keys` tables.

BranLwyd (approved):

janus_server/src/datastore.rs
```diff
@@ -485,6 +601,12 @@ pub enum Error {
     /// will never be generated by the datastore library itself.
     #[error(transparent)]
     User(#[from] Box<dyn std::error::Error + Send + Sync>),
+    #[error("URL parse error")]
```
divergentdave, line 604: nit: add `{0}` to the display format string

divergentdave (approved):

PR #51, opened 2022-04-01
Add basic Dockerfile

divergentdave:
This adds a simple Dockerfile that just builds the aggregator, and stores the schema file in the image as well.


Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+WORKDIR /src/janus_server
+COPY janus_server/Cargo.toml /src/janus_server/Cargo.toml
+COPY janus_server/Cargo.lock /src/janus_server/Cargo.lock
+COPY janus_server/src /src/janus_server/src
+COPY db/schema.sql /src/db/schema.sql
+RUN cargo build --release --bin aggregator
+
+FROM alpine:3.15.0
+RUN apk add libgcc
```
tgeoghegan, line 12:
```suggestion
RUN apk add libgcc && apk update
```

Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM rust:1.59.0-alpine as builder
+RUN apk add libc-dev protoc
```
tgeoghegan, line 2: I'm curious, what do we need a protobuf compiler for?

tgeoghegan, line 2:
```suggestion
RUN apk add libc-dev protoc && apk update
```

tgeoghegan (changes requested):
We should also add a GitHub Action that builds the container, even if it doesn't push it anywhere yet.

Dockerfile
divergentdave: console-subscriber pulls in prost, which requires protoc with its default features.

Dockerfile
divergentdave: `apk update` fetches an updated repository index, but this would go unused twice over, because there are no subsequent `apk` commands, and the docs say:
> This command is not needed in normal operation as all applets requiring indexes will automatically refresh them after caching time expires.
Was the intent to do `apk upgrade` maybe?

tgeoghegan: You are correct, `apk upgrade` is what we want here, or we could choose to stick with whatever is in `alpine 3.15`. Either is fine by me.

tgeoghegan (approved):

Dockerfile
divergentdave: :+1: I think I'll pass on `apk upgrade` for the time being. Dependabot will keep us up with base layer updates (speaking of, just noticed a mistake in the configuration) and this way we'll have predictable container layer sizes.

PR #53, opened 2022-04-04
Hide primary keys from users of the datastore library.

BranLwyd:
As part of this, I refactor all tables to have an "artificial" primary
key, which is a BIGINT; internally the datastore code manages the
necessary joins. Code using the datastore uses the "natural" primary
keys--the BIGINT primary keys are not exposed at all.

This simplifies the client code, while keeping join indexes small (which
may help performance) and allowing flexibility in case the "natrual"
primary key for a given entity changes with future changes to the PPM
spec.


tgeoghegan (approved):
LGTM, modulo rebasing on main

PR #56, opened 2022-04-06
`janus_server`: storage of task parameters

tgeoghegan:
With these changes we now store the tolerable clock skew in the tasks table and add functions for fetching task parameters from the database, to be used by aggregators to dynamically discover the defined tasks.

### DIFF
--- a/janus_server/src/datastore.rs
+++ b/janus_server/src/datastore.rs
@@ -108,14 +108,16 @@ impl Transaction<'_> {
         let max_batch_lifetime = i64::try_from(task.max_batch_lifetime)?;
         let min_batch_size = i64::try_from(task.min_batch_size)?;
         let min_batch_duration = i64::try_from(task.min_batch_duration.0)?;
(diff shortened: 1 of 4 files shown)


janus_server/src/datastore.rs
```diff
@@ -140,22 +143,21 @@ impl Transaction<'_> {
         Ok(())
     }
 
-    /// Fetch the task parameters corresponing to the provided `task_id`.
-    //
-    // Only available in test configs for now, but will soon be used by
-    // aggregators to discover tasks from the database.
+    /// Construct a [`TaskParameters`] from the contents of the provided `Row`.
+    /// If `task_id` is `None`, it is used. Otherwise the task ID is read from
+    /// the row.
```
divergentdave, line 148:
```suggestion
    /// Construct a [`TaskParameters`] from the contents of the provided `Row`.
    /// If `task_id` is not `None`, it is used. Otherwise the task ID is read from
    /// the row.
```

divergentdave (approved):

BranLwyd (approved):

PR #55, opened 2022-04-06
Multiple tasks in aggregator

tgeoghegan:
This PR is too darn big and I'm going to break it into smaller, reviewable pieces. I'm creating this draft PR so I can check that CI passes and so that people reviewing the constituent changes can see where this is all headed.

PR #57, opened 2022-04-06
Initial /aggregate implementation, for continue requests.

BranLwyd:
The line count looks huge, but almost all of the lines are in tests.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -25,8 +25,8 @@ use prio::{
 };
 use ring::hmac;
 use std::{
-    collections::HashSet, convert::Infallible, future::Future, net::SocketAddr, ops::Sub, pin::Pin,
-    sync::Arc,
+    collections::HashSet, convert::Infallible, future::Future, io::Cursor, net::SocketAddr,
+    ops::Sub, pin::Pin, sync::Arc,
 };
 use tracing::warn;
 use warp::{
@@ -177,7 +177,7 @@ where
         let leader_report = &report.encrypted_input_shares[0];
 
         // §4.2.2: verify that the report's HPKE config ID is known
-        if leader_report.config_id != self.report_recipient.config.id {
+        if leader_report.config_id != self.report_recipient.config().id {
             warn!(
                 config_id = ?leader_report.config_id,
                 "unknown HPKE config ID"
@@ -213,10 +213,7 @@ where
                 let report = report.clone();
                 Box::pin(async move {
                     // §4.2.2 and 4.3.2.2: reject reports whose nonce has been seen before
-                    match tx
-                        .get_client_report_by_task_id_and_nonce(report.task_id, report.nonce)
-                        .await
-                    {
+                    match tx.get_client_report(report.task_id, report.nonce).await {
                         Ok(_) => {
                             warn!(?report.task_id, ?report.nonce, "report replayed");
                             // TODO (issue #34): change this error type.
@@ -409,16 +406,16 @@ where
                 let report_share_data = report_share_data.clone();
                 Box::pin(async move {
                     // Write aggregation job.
-                    let aggregation_job_id = tx.put_aggregation_job(&aggregation_job).await?;
+                    tx.put_aggregation_job(&aggregation_job).await?;
 
                     for (ord, share_data) in report_share_data.as_ref().iter().enumerate() {
                         // Write client report & report aggregation.
-                        let client_report_id = tx
-                            .put_report_share(task_id, &share_data.report_share)
+                        tx.put_report_share(task_id, &share_data.report_share)
                             .await?;
                         tx.put_report_aggregation(&ReportAggregation::<A> {
-                            aggregation_job_id,
-                            client_report_id,
+                            aggregation_job_id: job_id,
+                            task_id,
+                            nonce: share_data.report_share.nonce,
                             ord: ord as i64,
                             state: share_data.agg_state.clone(),
                         })
@@ -460,7 +457,7 @@ where
                 Box::pin(async move {
                     // Read existing state.
                     let (aggregation_job, report_aggregations) = future::try_join(
-                        tx.get_aggregation_job_by_aggregation_job_id::<A>(job_id),
+                        tx.get_aggregation_job::<A>(job_id),
                         tx.get_report_aggregations_by_aggregation_job_id::<A>(
                             &verify_param,
                             job_id,
@@ -471,7 +468,7 @@ where
                     // Handle each transition in the request.
                     let mut report_aggregations = report_aggregations.into_iter();
                     for transition in transitions.iter() {
-                        // Match current transition to existing report aggregation.
+                        // Match transition received from leader to stored report aggregation.
                         let report_aggregation = loop {
                             let report_agg = report_aggregations.next().ok_or_else(|| {
                                 datastore::Error::User(Box::new(Error::UnrecognizedMessage(
@@ -479,6 +476,30 @@ where
                                     task_id,
                                 )))
                             })?;
+                            if report_agg.nonce != transition.nonce {
+                                continue;
+                            }
+                            break report_agg;
+                        };
+                        let prep_step: A::PrepareStep;
+                        // XXX: retrieve prepare step from report aggregation (verifying it's in Waiting state)
+                        // XXX: how to deal with non-waiting state?
+
+                        // Parse transition.
+                        let prep_msg = if let TransitionTypeSpecificData::Continued { payload } =
+                            &transition.trans_data
+                        {
+                            A::PrepareMessage::decode_with_param(
+                                &prep_step,
+                                &mut Cursor::new(payload),
+                            )? // XXX: is returning appropriate here?
+                        } else {
+                            return Err(datastore::Error::User(Box::new(
+                                Error::UnrecognizedMessage(
+                                    "leader sent non-Continued transition in payload",
+                                    task_id,
+                                ),
+                            )));
                         };
                         // XXX
                     }
@@ -663,7 +684,7 @@ where
         return Err(Error::InvalidConfiguration("role is not an aggregator"));
     }
 
-    let hpke_config_encoded = hpke_recipient.config.get_encoded();
+    let hpke_config_encoded = hpke_recipient.config().get_encoded();
 
     let aggregator = Arc::new(Aggregator::new(
         vdaf,
@@ -963,7 +984,7 @@ mod tests {
         datastore::test_util::{ephemeral_datastore, DbHandle},
         hpke::{HpkeSender, Label},
         message::{AuthenticatedResponseDecoder, HpkeCiphertext, HpkeConfig, TaskId, Time},
-        task::TaskParameters,
+        task::{TaskParameters, Vdaf},
         time::tests::MockClock,
         trace::test_util::install_test_trace_subscriber,
     };
@@ -972,7 +993,7 @@ mod tests {
     use prio::{
         codec::Decode,
         vdaf::prio3::Prio3Aes128Count,
-        vdaf::{Vdaf, VdafError},
+        vdaf::{Vdaf as VdafTrait, VdafError},
     };
     use rand::{thread_rng, Rng};
     use ring::{hmac::HMAC_SHA256, rand::SystemRandom};
@@ -1084,7 +1105,7 @@ mod tests {
 
         let bytes = hyper::body::to_bytes(response.into_body()).await.unwrap();
         let hpke_config = HpkeConfig::decode(&mut Cursor::new(&bytes)).unwrap();
-        assert_eq!(hpke_config, hpke_recipient.config);
+        assert_eq!(&hpke_config, hpke_recipient.config());
         let sender = HpkeSender::from_recipient(&hpke_recipient);
 
         let message = b"this is a message";
@@ -1107,8 +1128,12 @@ mod tests {
             .run_tx(|tx| {
                 let fake_url = Url::parse("localhost:8080").unwrap();
 
-                let task_parameters =
-                    TaskParameters::new_dummy(task_id, vec![fake_url.clone(), fake_url]);
+                let task_parameters = TaskParameters::new_dummy(
+                    task_id,
+                    vec![fake_url.clone(), fake_url],
+                    Vdaf::Prio3Aes128Count,
+                    Role::Leader,
+                );
                 Box::pin(async move { tx.put_task(&task_parameters).await })
             })
             .await
@@ -1384,10 +1409,7 @@ mod tests {
 
         let got_report = datastore
             .run_tx(|tx| {
-                Box::pin(async move {
-                    tx.get_client_report_by_task_id_and_nonce(report.task_id, report.nonce)
-                        .await
-                })
+                Box::pin(async move { tx.get_client_report(report.task_id, report.nonce).await })
             })
             .await
             .unwrap();
@@ -1463,10 +1485,7 @@ mod tests {
 
         let got_report = datastore
             .run_tx(|tx| {
-                Box::pin(async move {
-                    tx.get_client_report_by_task_id_and_nonce(report.task_id, report.nonce)
-                        .await
-                })
+                Box::pin(async move { tx.get_client_report(report.task_id, report.nonce).await })
             })
             .await
             .unwrap();
@@ -1612,8 +1631,13 @@ mod tests {
         datastore
             .run_tx(|tx| {
                 Box::pin(async move {
-                    tx.put_task(&TaskParameters::new_dummy(task_id, Vec::new()))
-                        .await
+                    tx.put_task(&TaskParameters::new_dummy(
+                        task_id,
+                        Vec::new(),
+                        Vdaf::Prio3Aes128Count,
+                        Role::Helper,
+                    ))
+                    .await
                 })
             })
             .await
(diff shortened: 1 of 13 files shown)


janus_server/src/aggregator.rs
```diff
@@ -439,11 +442,163 @@ where
...
+                            if report_agg.nonce != transition.nonce {
+                                // TODO(brandon): assuming we aren't going to fail due to
+                                // out-of-order/unexpected transitions, any report aggregations that
+                                // get skipped here are skipped because the leader omitted them
+                                // because something failed on the leader's side. We move to INVALID
+                                // state to ensure we don't get later updates, but is that the
+                                // desired behavior?
+                                report_agg.state = ReportAggregationState::Invalid;
+                                tx.update_report_aggregation(&report_agg).await?;
+                                continue;
```
tgeoghegan, line 496: I'm not sure this is right. The [relevant protocol text](https://github.com/abetterinternet/ppm-specification/blob/b734ac2272e58b839b8c5d5511e518e42b324956/draft-gpew-priv-ppm.md?plain=1#L1211), from the interop branch:

```
Next, the helper processes the Transition messages from the leader. If any
message appears out of order or has an unrecognized nonce, or if any two
messages have the same nonce, then the helper MUST abort with error
"unrecognizedMessage".
```

My reading of this code is that we'll iterate over the received transitions, and for each one, we'll then iterate over the stored aggregations until we find one whose nonce matches the transition, skipping any stored reports that are unrecognized or in the wrong order. But my reading of the protocol text is that if the leader's i-th transition doesn't match our i-th `report_aggregation`, we should abort the entire `AggregateContinue`.

So I think all this can be simplified if you do something like:
```
for (transition, report_aggregation) in transitions.into_iter().zip(report_aggregations.into_iter()) {
    if transition.nonce != report_aggregation.nonce {
        return Err(/* Some error that aborts the whole request */
    }
    // handle the aggregate continue
}
```

tgeoghegan (changes requested):

janus_server/src/aggregator.rs
```diff
@@ -439,11 +442,163 @@ where
...
+                            job_id,
+                        ),
+                    )
+                    .await?;
+
+                    // Handle each transition in the request.
+                    let mut report_aggregations = report_aggregations.into_iter();
+                    let (mut saw_continue, mut saw_finish) = (false, false);
+                    let mut response_transitions = Vec::new();
+                    for transition in transitions.iter() {
```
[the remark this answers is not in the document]
divergentdave: Note that we should wrap our call to rayon in `tokio::task::spawn_blocking()`. (and, if there's enough CPU-bound work that we would benefit from a threadpool, maybe we should be using `spawn_blocking()` already)

janus_server/src/aggregator.rs
divergentdave: By my reading, I think the helper does need to handle cases where related aggregate continue requests have fewer transitions than preceding aggregate init/aggregate continue requests. We do need to be more liberal than zipping the iterators and comparing nonces, and the inner loop that advances `report_aggregations` looks right.

The end of the section "Aggregate Continuation Request" says "The leader processes this message as described in {{aggregate-init-request}}." That section says, at the end of request-response processing, "Finally, the leader removes from the report sequence any report for which the corresponding state is FAILED and proceeds as described in the next section, even if the sequence is empty." The prep-leader diagram says that if the helper failed, the leader's state for the report will be FAILED, but there is also an edge from the leader's invocation of prep_next to FAILED. (with the caveat that the edge label says `prep_next` `send Transition`) Thus, in between making requests to the helper's `/aggregate`, the leader can remove reports from the sequence for two reasons, either the helper's `prep_next()` failed, and it sent a transition of type `failed` to the leader, or the leader's `prep_next()` failed, and by my reading, the leader never sends a transition of type FAILED in a request.

As for the related question in the TODO comment of what we should do with the state enum: If we take the state machine literally, if a nonce stops showing up in requests, then it would stay at the WAITING state indefinitely. On the other hand, we can infer that either the leader's `prep_next()` failed, or the leader is not following the protocol correctly. As FAILED is a terminal state, I think it make sense to store that. If the leader operates correctly, then the helper will never look up this state again. If the leader errs by excluding a report from one continue request, then including it in another request, storing a FAILED state signifies that we don't expect to see the report again, so we would filter out the report when looking up report aggregations, and then correctly reply with `unrecognizedMessage` due to the mismatch.

BranLwyd: ^ Dave's response above is why I implemented it this way. A simple zip doesn't work because, in general, the leader will be sending transitions for a _subsequence_ of reports included in the aggregation job. I can't do something simple based on `Iterator::find`, either, because I want to update the report aggregations that end up being skipped.

The existing logic would also catch duplicate nonces in the leader's request; I updated the error message to note this.

I moved to a terminal state because I wanted to notice/fail out on the "leader omits a report from one continue request, then includes it on a later continue request" case. I think `FAILED` makes a lot of sense (over `INVALID`), but the one wrinkle is that I'm not sure what error code we would want to use -- none of them seem appropriate, and the leader doesn't tell us why it omitted the request (nor do I think we necessarily want to make it do so). `report-dropped` is the closest choice of the existing errors IMO, but currently it specifically indicates that a report was dropped from storage for being "too old" & I'm wary of overloading codes. I'm not sure if introducing a new code is appropriate without a spec change -- `TransitionError` is defined in the spec. Thoughts on the correct error code to use, if we want to switch from `INVALID` to `FAILED`?

BranLwyd: BTW: for future (probably post-interop-test, unless the CF folks think otherwise), I think we should probably drop the ordering requirement entirely.

The justification I've heard for the ordering requirement is that it saves us from having to do an `O(n lg n)` sort to match incoming transitions to stored report aggregation state. But even without the ordering requirement we could just put the report aggregations in e.g. a `HashMap<Nonce, ReportAggregation>` (in `O(n)` time), then look them up by nonce in `O(1)` time -- no `O(n lg n)`.

The downside of the ordering semantics is the additional implementation complexity, and additional failure modes.

Am I missing something? If not, I'll put this PR on my to-do list.

janus_server/src/aggregator.rs
```diff
@@ -1935,63 +2099,1055 @@ mod tests {
...
+            clock,
+            skew,
+            Role::Helper,
+            verify_params[1].clone(),
+            hpke_recipient,
+            hmac_key.clone(),
         )
+        .unwrap();
+
+        let mut response = warp::test::request()
```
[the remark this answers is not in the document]
BranLwyd: I disagree for a couple of reasons:
1. "Test the public interface." Arguably, the public interface is the `warp` filter. (or, going further, it might be the server itself, since it's `aggregator_server` that's marked `pub` -- but I'm OK viewing these as effectively identical.)

2. There actually is functionality under test that is part of the filter. Namely, authentication, request parsing, response encoding, and exposing the endpoint only if acting as a helper. I think it's useful/important to test these as well.


For the problem mentioned around #55, perhaps we should factor things so that "look up task parameters (including VDAF) from task ID" is a generic operation -- eventually we're going to want to be able to do this per-request, possibly falling back to a DB lookup for new tasks, so I think it makes sense to do so. Then tests could just use a lookup function that returns the desired parameters, including a `fake::Vdaf`.

janus_server/src/aggregator.rs
```diff
@@ -439,11 +442,163 @@ where
...
-        _job_id: AggregationJobId,
-        _transitions: Vec<Transition>,
+        task_id: TaskId,
+        job_id: AggregationJobId,
+        transitions: Vec<Transition>,
     ) -> Result<AggregateResp, Error> {
-        todo!() // TODO(brandon): implement
+        let vdaf = Arc::new(self.vdaf.clone());
+        let verify_param = Arc::new(self.verify_param.clone());
+        let transitions = Arc::new(transitions);
```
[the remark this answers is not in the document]
BranLwyd: Sadly, I don't think so; I tried a number of approaches and Rust complained about lifetimes in every case.

This makes some sense to me: the closure passed to `run_tx` is not a `FnOnce` but a `Fn` (and in the face of transaction retries, we actually will call the closure multiple times). If we moved `transitions` into the closure, we couldn't call the closure more than once, since the first call would consume `transitions`. So it seems that we do need _some_ way to keep `transitions` around for multiple calls, and cloning an `Arc` is a lot cheaper than cloning the whole structure. Keep in mind that `Arc::new` is moving just the constant-size "header" of the transitions vector, not the contents of the vector.

I'm more concerned about the other `Arcs` TBH -- we end up cloning the VDAF & the verify parameter, with no guarantee that either clone is "cheap" since the size of these structures is up to the VDAF in use. I think we may want to refactor the `Aggregator` struct to hold onto an `Arc<A>` and an `Arc<A::VerifyParam>` to avoid these clones, but I opted not to do this performance enhancement in this PR.

janus_server/src/aggregator.rs
```diff
@@ -439,11 +442,163 @@ where
...
+                        // Parse preparation message out of transition received from leader.
+                        let prep_msg = match &transition.trans_data {
+                            TransitionTypeSpecificData::Continued { payload } => {
+                                A::PrepareMessage::decode_with_param(
+                                    &prep_step,
+                                    &mut Cursor::new(payload),
+                                )?
+                            }
+                            _ => {
+                                // TODO(brandon): should this be reflected in the response?
```
divergentdave, line 527: Yes, the helper state machine says that if the leader did not continue, then the helper transitions to INVALID. Then, there's a sentence that says "If any state transition results in INVALID, the helper MUST abort with error 'unrecognizedMessage'".

divergentdave: Ah, this presents a conundrum. In this and other cases, we want to both record some state in the database, and return a "400 Bad Request" with problem details as a response. However, when we return `Err(Error::UnrecognizedMessage(...))`, `run_tx_once()` exits early, and the transaction rolls back. I think our transaction closures will either need to return a Result with a special error type, so that we can commit on certain errors and roll back on others, or a nested Result (which could be an ergonomic pain without `#![feature(try_blocks)]`).

janus_server/src/aggregator.rs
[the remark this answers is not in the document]
BranLwyd: Added a TODO. We currently believe that we're going to be IO-bound, so improving CPU efficiency will be of limited efficacy -- for that reason, I'm wary of adding too much complexity in the name of performance just yet, until we see how the system operates in practice.

A few potential performance issues I'm concerned about (but did not address yet) -- I added TODOs for these as well:
* We currently hold the database transaction open while we're doing the VDAF computations. This isn't necessary: a read-copy-update strategy could work (in a first transaction, read necessary data; outside transaction, compute updates; in a second transaction, reread data to verify it hasn't changed, then write updates -- starting over if data has changed). But it's unclear to me if "one transaction living longer" is worse than "multiple transactions".
* We currently do one network round-trip with the database for each report aggregation, waiting until the response is received before continuing with program logic. We could improve here by updating all of the report aggregations at once (https://stackoverflow.com/questions/18797608/update-multiple-rows-in-same-query-using-postgresql), or by pipelining writes the way we pipeline reads (https://docs.rs/tokio-postgres/0.7.5/tokio_postgres/#pipelining). Probably either approach would be fine, but both are somewhat technically challenging.

janus_server/src/aggregator.rs
tgeoghegan: I buy the argument of the leader filtering out `FAILED` reports. I tend to agree with you about removing the ordering requirement, or I had suggested always requiring ascending nonces within an aggregation job so that we get a canonical representation on the wire.

janus_server/src/aggregator.rs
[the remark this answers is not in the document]
tgeoghegan: Yes, I have to admit that part of my approach in #55 was to limit the size of the changes given that the PR is already large. In any case, since I'm on vacation from tomorrow, we should favour your change over mine in the name of unblocking yours and David's work.

tgeoghegan (approved):

divergentdave (changes requested):

janus_server/src/datastore.rs
```diff
@@ -561,6 +636,90 @@ fn single_row(rows: Vec<Row>) -> Result<Row, Error> {
     }
 }
 
+fn check_update(row_count: u64) -> Result<(), Error> {
+    match row_count {
+        0 => Err(Error::NotFound),
+        1 => Ok(()),
+        _ => panic!(
```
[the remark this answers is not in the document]
BranLwyd: The reason I panic is that this case is pure programmer error: this function is intended to be used with an `UPDATE` statement that can update at most one row (due to the way the UPDATE is written & the constraints in the DB schema). If we ever end up in the "impossible" situation that our at-most-one-row `UPDATE` actually updated more than one row, I panic to make the failure as visible as possible.

An error would work too, but might be more likely to be consumed/ignored (and accidentally update a whole bunch of rows). If folks feel strongly, I can change it -- my justification is basically just "impossible/programmer errors should crash (i.e. panic)".

janus_server/src/datastore.rs
```diff
@@ -423,85 +467,66 @@ impl Transaction<'_> {
...
+    {
+        let stmt = self
+            .tx
+            .prepare_cached(
+                "SELECT client_reports.nonce_time, client_reports.nonce_rand, report_aggregations.ord, report_aggregations.state, report_aggregations.vdaf_message, report_aggregations.error_code
+                FROM report_aggregations
+                JOIN client_reports ON client_reports.id = report_aggregations.client_report_id
+                WHERE report_aggregations.aggregation_job_id = (SELECT id FROM aggregation_jobs WHERE aggregation_job_id = $1)
+                  AND client_reports.task_id = (SELECT id FROM tasks WHERE task_id = $2)"
+            )
```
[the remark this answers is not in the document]
BranLwyd: Oof, nice catch -- idk how I didn't catch this with testing (but I suspect you're correct about the tests not being "big" enough to trigger whatever behavior internal to postgres that would cause the rows to be returned in non-insertion order).

janus_server/src/aggregator.rs
BranLwyd: I think `unrecognizedMessage` doesn't make sense here -- that's an `aggregator::Error` / what would be reflected in a problem document / an error at the level of an entire request.

Since we only want to fail this individual client report, not the entire request, we need to choose an error from [`TransitionError`](https://github.com/divviup/janus/blob/2c053feb6bfd01db8fd3c7c3c5024132262df549/janus_server/src/message.rs#L762-L773). That is, we want to choose from among the values: `BatchCollected`, `ReportReplayed`, `ReportDropped`, `HpkeUnknownConfigId`, `HpkeDecryptError`, `VdafPrepError`, `UnrecognizedNonce` (or standardize a new value).

janus_server/src/aggregator.rs
BranLwyd: I updated the semantics to abort as expected, thanks!

I think it's currently underspecified what to do if the entire request fails with `unrecognizedMessage` (or any other error): the leader has no way to know _which_ condition caused the failure, and therefore has no idea what state the helper might be in. In general, I'd be pretty surprised if a "bad request" (`unrecognizedMessage`) response caused state changes. I think the semantics we actually want are to not change state at all if the entire request is bad. TBF, that isn't what's written in the spec (i.e. in this case it will transition to INVALID and then abort), but I think we probably want to standardize on "aborting causes no state changes" -- which we'd then interpret as the helper realizing there's an INVALID transition, rolling back (undoing the invalid transition), and then responding with an error.

divergentdave (approved):

janus_server/src/aggregator.rs
BranLwyd: For now, I went with `ReportDropped`. I'm a little unhappy with the overloading of this error code, but I suppose it's the best one available (and I don't necessarily want to try to specify a new one).

BranLwyd: I also decided to only update the state if we aren't already in a non-WAITING state, to avoid potentially overwriting a previously-recorded error.

janus_server/src/aggregator.rs
```diff
@@ -439,11 +444,168 @@ where
...
+                                datastore::Error::User(Box::new(Error::UnrecognizedMessage(
+                                    "leader sent unexpected, duplicate, or out-of-order transitions",
+                                    task_id,
+                                )))
+                            })?;
+                            if report_agg.nonce != transition.nonce {
+                                // TODO(brandon): assuming we aren't going to fail due to
+                                // out-of-order/unexpected transitions, any report aggregations that
+                                // get skipped here are skipped because the leader omitted them
+                                // because something failed on the leader's side. We move to INVALID
```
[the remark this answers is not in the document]
BranLwyd: (the TODO change quite a bit, but I made sure the wording took this into account)

janus_server/src/aggregator.rs
```diff
@@ -439,11 +444,166 @@ where
...
+                                datastore::Error::User(Box::new(Error::UnrecognizedMessage(
+                                    "leader sent unexpected, duplicate, or out-of-order transitions",
+                                    task_id,
+                                )))
+                            })?;
+                            if report_agg.nonce != transition.nonce {
+                                // This report was omitted by the leader because of a prior failure.
+                                // Note that the report was dropped (if it's not already in an error
+                                // state) and continue.
+                                if !matches!(report_agg.state, ReportAggregationState::Waiting(_)) {
```
divergentdave, line 495: I think this conditional is backwards, presently it will only overwrite failed/invalid/finished states with failed.

PR #58, opened 2022-04-06
testcontainers: 0.12.0 -> 0.13.0.

BranLwyd:
A few notes:
 * Unfortunately, we can't use podman yet; attempting to start the
   Postgres container fails.
 * The testcontainers library apparently removed the ability to request
   a specific Postgres version, but fortunately the default version is
   high enough to have all of the features we care about.

Example error when using podman:
```
thread 'datastore::tests::roundtrip_task' panicked at 'called `Result::unwrap()` on an `Err` value: Error("invalid type: string \"docker-entrypoint.sh\", expected a sequence", line: 180, column: 48)', /home/bran/.cargo/registry/src/github.com-1ecc6299db9ec823/testcontainers-0.13.0/src/clients/cli.rs:323:88
```

### DIFF
--- a/janus_server/src/datastore.rs
+++ b/janus_server/src/datastore.rs
@@ -817,19 +817,19 @@ pub mod test_util {
     use deadpool_postgres::{Manager, Pool};
     use lazy_static::lazy_static;
     use std::str::{self, FromStr};
-    use testcontainers::{clients::Cli, images::postgres::Postgres, Container, Docker};
+    use testcontainers::{clients::Cli, images::postgres::Postgres, Container};
     use tokio_postgres::{Config, NoTls};
 
     const SCHEMA: &str = include_str!("../../db/schema.sql");
 
     // TODO(brandon): use podman instead of docker for container management once testcontainers supports it
     lazy_static! {
-        static ref DOCKER: Cli = Cli::default();
+        static ref CONTAINER_CLIENT: Cli = Cli::default();
     }
 
     /// DbHandle represents a handle to a running (ephemeral) database. Dropping this value causes
     /// the database to be shut down & cleaned up.
-    pub struct DbHandle(Container<'static, Cli, Postgres>);
+    pub struct DbHandle(Container<'static, Postgres>);
 
     /// ephemeral_datastore creates a new Datastore instance backed by an ephemeral database which
     /// has the Janus schema applied but is otherwise empty.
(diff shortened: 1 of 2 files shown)


divergentdave:
Hmm, that error message looks like it came from Serde. I presume there's a mismatch in JSON schemas between the podman binary and some crate, regarding the container's CMD.

BranLwyd:
> Hmm, that error message looks like it came from Serde. I presume there's a mismatch in JSON schemas between the podman binary and some crate, regarding the container's CMD.

My (very quick, possibly-incorrect) reading of the code in question is that it's trying to do some kind of parsing on the log lines generated by the container, and for whatever reason it fails when doing so via podman.

divergentdave:
This looks like a bug in podman, insofar as it's not matching the structure of `docker inspect`. The `testcontainers` crate tries to run `podman inspect {id}` before tearing down the container, and parses the output as `Vec<bollard_stubs::models::ContainerInspectResponse>`. (`bollard` is a crate for interfacing with the Docker API) The response structure contains a `Config` member, of type `bollard_stubs::models::ContainerConfig`, and this contains a member `entrypoint` of type `Option<Vec<String>>`. I ran `docker inspect` and `podman inspect` on postgres containers, and indeed Docker's output shows an array under `Entrypoint`, while Podman shows a bare string. Looking inside podman's models, there's a structure `InspectContainerConfig`, and here `Entrypoint` is a string, and not a slice of strings.

janus_server/src/datastore.rs
```diff
@@ -817,33 +817,33 @@ pub mod test_util {
...
+    pub struct DbHandle(Container<'static, Postgres>);
 
     /// ephemeral_datastore creates a new Datastore instance backed by an ephemeral database which
     /// has the Janus schema applied but is otherwise empty.
     ///
     /// Dropping the second return value causes the database to be shut down & cleaned up.
     pub async fn ephemeral_datastore() -> (Datastore, DbHandle) {
         // Start an instance of Postgres running in a container.
-        let db_container = DOCKER.run(Postgres::default().with_version(14));
+        let db_container = CONTAINER_CLIENT.run(Postgres::default());
```
divergentdave, line 840: We should add `.with_tag("14-alpine")` to pin our Postgres version here.
```suggestion
        let db_container = CONTAINER_CLIENT.run(RunnableImage::from(Postgres::default()).with_tag("14-alpine"));
```

divergentdave (approved):

PR #64, opened 2022-04-09
Implement datastore::Crypter, to be used to encrypt sensitive datastore values.

BranLwyd:
datastore::Crypter will encrypt/decrypt values to be stored in a
datastore. The ciphertexts are cryptographically bound to the specific
database cell (i.e. table, column, & row) they are to be stored in.

N.B. on use of LessSafeKey: the alternative that ring seems to want me
to use is a combination of OpeningKey/SealingKey, which are AEAD keys
associated with a "nonce sequence." While using a nonce sequence makes
sense for e.g. a wire protocol that can determine the nonce from a
packet number or similar value, for our use-case it would be challenging
to encrypt with a specific sequence of nonces & impossible to decrypt.
Instead, I opt to use randomly-generated nonces, including the nonce
values in the ciphertexts.

I think using random nonces is safe in this context: nonces are 12 bytes
in length, giving 2^(12*8) = 2^96 possible nonce values. The birthday
bound then tells us we would expect to see a nonce collision after
generating around sqrt(2^96) = 2^48 nonces. Given that we are planning
to use this to encrypt a number of different plaintexts that is bounded
by about 3 times the number of tasks, I suspect we won't be anywhere
near this limit for quite some time.

Part of https://github.com/divviup/janus/issues/48.


divergentdave (approved):
LGTM.

I notice the row passed to be used as AAD is a `&[u8]`, is the intent that we'll use task IDs and aggregation sub-job IDs as "row" identifiers? I figure that's better than using database-internal row numbers, as this way we'll get binding to the relevant PPM-level objects.

BranLwyd:
Yeah: the general intent is to use the "effective primary key" as the row identifier. For tasks, that would be the 32-byte PPM-specified task ID, as opposed to the artificial BIGINT primary key column used for joins.

I think tasks are the only table we plan to include encrypted values in currently, but e.g. using the 32-byte aggregation job ID for the aggregation jobs table would be my guess. The most important property is that the row identifier be actually unique to the row -- for some tables this would mean we'd need to concatenate or otherwise serialize multiple values together (e.g. the row might be a serialization of `(task_id, nonce_time, nonce_rand)` if we hypothetically wanted to encrypt a value into the `client_reports` table).

Other than binding to the PPM object, I do this because the relevant datastore model objects purposefully don't know the artificial primary key values; if we wanted to use the artificial IDs as the row identifiers, we'd have to read them from the DB before performing a write. (...and when writing new objects, we won't know the artificial primary key until the write occurs, so we'd need to write twice!)

PR #68, opened 2022-04-11
git: ignore /janus_server/target/ directory.

BranLwyd:


divergentdave (approved):
LGTM, but we should probably add a comment indicating this is an old binary directory from before the change to using a workspace.

PR #70, opened 2022-04-11
datastore: use Crypter to encrypt secret values.

BranLwyd:
Right now, that means tasks.vdaf_verify_param, tasks.agg_auth_key, and
tasks.hpke_private_key.

### DIFF
--- a/janus_server/src/bin/aggregator.rs
+++ b/janus_server/src/bin/aggregator.rs
@@ -4,7 +4,7 @@ use deadpool_postgres::{Manager, Pool};
 use janus_server::{
     aggregator::aggregator_server,
     config::AggregatorConfig,
-    datastore::Datastore,
+    datastore::{self, Datastore},
     hpke::{HpkeRecipient, Label},
     message::Role,
     message::TaskId,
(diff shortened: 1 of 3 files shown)


janus_server/src/bin/aggregator.rs
```diff
@@ -58,10 +60,21 @@ struct Options {
...
+
+    /// Datastore encryption keys.
+    #[structopt(
+        long,
+        env = "DATASTORE_KEYS",
+        takes_value = true,
+        required(true),
+        help = "datastore encryption keys, encoded in base64 then comma-separated"
+    )]
+    datastore_keys: String,
```
divergentdave, line 77: I think we can have clap/structopt take care of comma separation by making this a `Vec<String>`, and possibly adding `#[structopt(use_delimiter = true)]` as well. The docs for [`env()`](https://docs.rs/clap/2.34.0/clap/struct.Arg.html#method.env) say that environment variables can be used for multiple-valued arguments in such cases, and the default delimiter is the comma. (This will add an ellipsis to the argument's representation in `--help`)

divergentdave (approved):

janus_server/src/bin/aggregator.rs
BranLwyd: Testing quickly, your suggestion works out. 👍🏻

BranLwyd: Testing quickly, your suggestion works out. 👍🏻

PR #71, opened 2022-04-11
Get database password from env. var. if not in URL

divergentdave:
Currently, we read in a Postgres connection string from the configuration file, and pass that directly to the database library. This change will read a password from the environment variable `PGPASSWORD` if the connection string/URL does not include one, and use that. Splitting the password handling out in this way will make it easy to assemble everything but the password in Terraform, without having to worry about the secrecy of Terraform state files. The password can be plumbed via an environment variable from a Kubernetes secret, separate from the configuration.

### DIFF
--- a/janus_server/src/bin/aggregator.rs
+++ b/janus_server/src/bin/aggregator.rs
@@ -90,13 +90,18 @@ async fn main() -> Result<()> {
     let vdaf = Prio3Aes128Count::new(2).unwrap();
     let verify_param = vdaf.setup().unwrap().1.first().unwrap().clone();
 
-    let database_config = tokio_postgres::Config::from_str(config.database.url.as_str())
+    let mut database_config = tokio_postgres::Config::from_str(config.database.url.as_str())
         .with_context(|| {
             format!(
                 "failed to parse database connect string: {:?}",
                 config.database.url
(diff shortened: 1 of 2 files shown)


janus_server/src/bin/aggregator.rs
```diff
@@ -90,13 +90,18 @@ async fn main() -> Result<()> {
...
-    let database_config = tokio_postgres::Config::from_str(config.database.url.as_str())
+    let mut database_config = tokio_postgres::Config::from_str(config.database.url.as_str())
         .with_context(|| {
             format!(
                 "failed to parse database connect string: {:?}",
                 config.database.url
             )
         })?;
+    if database_config.get_password().is_none() {
+        if let Ok(password) = std::env::var("PGPASSWORD") {
```
BranLwyd, line 101: Specify this as part of the Options struct above? I know that we'll specify this via env var 100% of the time (well, 100% of the time we want our Postgres password to actually be secret), but I think it's worthwhile to keep a standard method of exposing configuration options & to get the documentation that comes along with using structopts.

BranLwyd (changes requested):

BranLwyd (approved):

PR #74, opened 2022-04-12
Add other HPKE algorithms, check vs. test vectors

divergentdave:
This fills in the remainder of the match statements for dispatching HPKE encryption and decryption. I added round-trip tests for all supported algorithms, and decryption of test vectors in the cases where they are available.


janus_server/src/hpke.rs
```diff
@@ -144,38 +144,166 @@ impl HpkeSender {
...
-                        self.label,
-                        self.sender_role,
-                        self.recipient_role,
-                    ),
+                    application_info,
                     plaintext,
                     associated_data,
                 )
             }
+            // This catch-all pattern is unreachable at time of writing, but it may be necessary
```
BranLwyd, line 303: WDYT about removing this case entirely (causing a compile failure if new enum values are ever added)?

This would force us to maintain compatibility with the version of the `hpke` crate we're using, since with this PR we're effectively opting into supporting everything `hpke` does.

janus_server/src/hpke.rs
```diff
@@ -144,38 +144,166 @@ impl HpkeSender {
...
+            self.sender_role,
+            self.recipient_role,
+        );
         // We must manually dispatch to each possible specialization of seal
         match (
             self.recipient_config.kem_id,
             self.recipient_config.kdf_id,
             self.recipient_config.aead_id,
         ) {
+            (HpkeKemId::P256HkdfSha256, HpkeKdfId::HkdfSha256, HpkeAeadId::Aes128Gcm) => {
```
BranLwyd, line 159: We could reduce repetition by having each `match` arm return the function to be called -- since every arm calls a function with the same signature we can put it in a variable, then call the variable.

That is, I replaced the body with the following in my pre-this-PR copy and it worked:
```
        // We must manually dispatch to each possible specialization of seal
        let seal = match (
            self.recipient_config.kem_id,
            self.recipient_config.kdf_id,
            self.recipient_config.aead_id,
        ) {
            (HpkeKemId::P256HkdfSha256, HpkeKdfId::HkdfSha256, HpkeAeadId::Aes256Gcm) => {
                seal::<AesGcm256, HkdfSha256, DhP256HkdfSha256>
            }
            (HpkeKemId::X25519HkdfSha256, HpkeKdfId::HkdfSha512, HpkeAeadId::ChaCha20Poly1305) => {
                seal::<ChaCha20Poly1305, HkdfSha512, X25519HkdfSha256>
            }
            (_, _, _) => {
                return Err(Error::InvalidConfiguration(
                    "unsupported set of HPKE algorithms",
                ))
            }
        };
        seal(
            &self.recipient_config,
            HpkeApplicationInfo::new(
                self.task_id,
                self.label,
                self.sender_role,
                self.recipient_role,
            ),
            plaintext,
            associated_data,
        )
```

BranLwyd, line 159: nit, not caused by this PR: it's a little confusing that the order of the items in the matched tuple is reversed from the order of the items in the generic type parameter list.

janus_server/src/hpke.rs
```diff
@@ -300,35 +428,177 @@ impl HpkeRecipient {
...
+    /// vectors.
+    fn open_internal(
+        &self,
+        ciphertext: &HpkeCiphertext,
+        application_info: HpkeApplicationInfo,
+        associated_data: &[u8],
     ) -> Result<Vec<u8>, Error> {
         // We must manually dispatch to each possible specialization of open
         match (self.config.kem_id, self.config.kdf_id, self.config.aead_id) {
+            (HpkeKemId::P256HkdfSha256, HpkeKdfId::HkdfSha256, HpkeAeadId::Aes128Gcm) => {
```
BranLwyd, line 454: Same comments as above re: reducing repetition by matching to a function, and matching the ordering of the tuple with the order of the generic type parameters.

janus_server/src/hpke.rs
```diff
@@ -469,4 +747,175 @@ mod tests {
             .open(&ciphertext, b"wrong associated data")
             .unwrap_err();
     }
+
+    fn round_trip_check<KEM: hpke::Kem, KDF: hpke::kdf::Kdf, AEAD: hpke::aead::Aead>(
+        task_id: TaskId,
```
BranLwyd, line 752: nit: why accept `task_id`? it looks like it's only used in one place, and it doesn't need to match up between calls to this function AFAICT. We could just call `TaskId::random()` (or a "zero" task ID) directly.

Similarly, instead of passing `&mut rng`, drop the parameter and call `thread_rng()` directly as needed?

janus_server/src/hpke.rs
```diff
@@ -469,4 +747,175 @@ mod tests {
...
+        serialized_public_key: Vec<u8>,
+        #[serde(with = "hex", rename = "skRm")]
+        serialized_private_key: Vec<u8>,
+        #[serde(with = "hex")]
+        base_nonce: Vec<u8>,
+        encryptions: Vec<EncryptionRecord>,
+    }
+
+    #[test]
+    fn decrypt_test_vectors() {
```
BranLwyd, line 841: This is an excellent test 👍🏻.

BranLwyd (changes requested):

BranLwyd (approved):

PR #77, opened 2022-04-13
Add caching to non-Docker CI build

divergentdave:
This invokes the cache action to cache cargo's repository index and our target directory. I'm going to force-push this in a bit and compare timing/outputs to see if it worked.


divergentdave:
Looks good, it saved more than a minute thanks to the cache hit.

PR #78, opened 2022-04-13
use hpke-dispatch

jbr:


janus_server/Cargo.toml
```diff
@@ -28,6 +28,7 @@ prio = { git = "https://github.com/divviup/libprio-rs", rev = "229cd9c45924c7dae
 rand = "0.8"
 reqwest = { version = "0.11.4", default-features = false, features = ["rustls-tls", "json"] }
 ring = "0.16.20"
+runtime-hpke = { git = "https://github.com/jbr/runtime-hpke/" }
```
BranLwyd, line 31: Could we cut a release of runtime-hpke? libprio-rs is currently at a specific commit rather than a released version, but that's because we're not the only team using libprio-rs -- I don't suppose that applies to runtime-hpke.

If we can't cut a release version, let's pin to a specific commit revision as libprio-rs does.

BranLwyd (changes requested):
I think you need to merge latest main into this PR, but otherwise it looks good modulo one nit.

janus_server/Cargo.toml
jbr: Released as runtime-hpke@0.1.0 under my own account

BranLwyd (approved):

tgeoghegan:
I think I understand the test failure here. The problem is converting between enum `janus_server::message::HpkeKemId` and `hpke_dispatch::Kem`. The `janus_server` enum uses [the constants exported from crate `hpke`](https://docs.rs/hpke/latest/src/hpke/kem/dhkem.rs.html#364-370), which are `0x10` and `0x20` for `DhP256HkdfSha256` and `X25519HkdfSha256`, respectively. The enum in `hpke_dispatch` uses [those same values, but in decimal](https://docs.rs/hpke-dispatch/latest/src/hpke_dispatch/lib.rs.html#92-100).

Technically there is the same problem for `hpke_dispatch`'s KDF and AEAD enums, but since all those values are smaller than 10, there's no problem.

jbr:
Oh interesting! Are the ids in hex in the paper? I missed that, will treat it as a bug

jbr:
released as https://github.com/jbr/hpke-dispatch/releases/tag/v0.1.1

jbr:
Anyone know why the ids are listed in hex in the paper? Seems like a strange choice for a human-readable number representation

cjpatton:
> Anyone know why the ids are listed in hex in the paper? Seems like a strange choice for a human-readable number representation

If by "paper" you mean RFC 9180: https://www.rfc-editor.org/rfc/rfc9180.html#name-key-encapsulation-mechanism

jbr:
Yes, by paper I meant draft spec / source document — I mistakenly copied the numbers as decimal from the tables in the draft spec. On reread, I'm guessing the integers are represented as hex in the draft spec because in that context it's documenting the bytewise wire protocol representation?

tgeoghegan:
@jbr Would you mind rebasing this change so we can merge it? Or I'm happy to do so if you are busy elsewhere.

jbr:
@tgeoghegan Rebased

PR #80, opened 2022-04-13
Datastore support for multiple HPKE configs.

BranLwyd:
Note that there is additional work to be done: the Aggregator type needs
to be taught how to deal with multiple HPKE configs. But this PR is
large enough already.

I also move a few test-only functions into a test_util crate.
HpkeRecipient::generate's removal might seem petty, but my next
refactoring step is going to turn it (and HpkeSender) into more
ephemeral types, intended to be created & used in a single line; places
that store HpkeRecipients would instead store (HpkeConfig,
HpkePrivateKey), places that store HpkeSenders would instead store
HpkeConfig.


janus_server/src/hpke.rs
```diff
@@ -362,9 +329,36 @@ fn open<Encrypt: Aead, Derive: Kdf, Encapsulate: Kem>(
...
+            HpkePrivateKey(private_key.to_bytes().as_slice().to_vec()),
+        )
+    }
+}
+
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use super::{test_util::generate_hpke_config_and_private_key, *};
     use crate::trace::test_util::install_test_trace_subscriber;
```
divergentdave, line 362: It looks like this is the sole merge conflict with main/

janus_server/src/datastore.rs
```diff
@@ -127,63 +129,180 @@ impl Transaction<'_> {
...
+            .prepare_cached(
+                "SELECT (SELECT tasks.task_id FROM tasks WHERE tasks.id = task_hpke_keys.task_id),
+            config_id, config, private_key FROM task_hpke_keys",
+            )
+            .await?;
+        let hpke_config_rows = self.tx.query(&stmt, &[]);
+
+        let (task_rows, hpke_config_rows) = try_join(task_rows, hpke_config_rows).await?;
+
+        let mut task_row_by_id = HashMap::new();
```
divergentdave, line 266: Nit, take it or leave it: I think this could be a `Vec<(TaskId, Row)>`, as we only iterate over it later.

divergentdave (changes requested):
Thus far, the tests only create tasks with a single HPKE keypair. We should kick the tires on round-tripping a task with two keypairs, and also check that our AAD scheme and conversions work for a config_id of at least 0x80, just in case.

Should we return an error at the datastore level if a task has zero HPKE keypairs associated with it? It may simplify the task lifecycle if we require that tasks have a keypair at creation.

janus_server/src/datastore.rs
BranLwyd: You're correct, and I suppose a `Vec` is slightly more efficient.

db/schema.sql
```diff
@@ -25,14 +25,24 @@ CREATE TABLE tasks(
...
 
-    -- TODO(timg): move vdaf_verify_param, agg_auth_key, hpke_config, hpke_private_key to new
+    -- TODO(timg): move vdaf_verify_param, agg_auth_key to new
     -- tables with many:1 relationships to tasks to allow for rotation of secrets
 );
 
+-- The HPKE public keys (aka configs) and private keys used by a given task.
+CREATE TABLE task_hpke_keys(
+    id BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,  -- artificial ID, internal-only
+    task_id BIGINT NOT NULL, -- XXX: document
```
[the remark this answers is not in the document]
BranLwyd: (one day I will learn to search the entire repo, and not just the `src/` directory, when looking for stray `XXX` markers...)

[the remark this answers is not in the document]
divergentdave: Here's a oneliner for a pre-commit hook: `bash -c '! git grep XXX'`. (The bash subprocess is presumably necessary because `set -e` only cares about subprocesses, and not expression evaluation)

BranLwyd:
Re testing: I modified `roundtrip_task` to create tasks with multiple HPKE configs, including a config with large `config_id`.

Re treatment of zero HPKE keypairs: instead of a datastore error, I implemented an error in `TaskParameters::new` if `hpke_keypairs` is empty. It'd be nice to make "there is at least one HPKE keypair" a data-type invariant, except that everything is `pub(crate)` at the moment so our invariants aren't so invariant.

divergentdave (approved):

PR #81, opened 2022-04-15
Aggregator support for multiple HPKE configs/keys.

BranLwyd:
The largest part of this PR is a refactoring of HpkeSender &
HpkeReceiver to be free functions seal/open instead. I think this is a
good refactoring because it binds the application info more directly to
the place it is used (i.e. we know from where we are in the code what
application info to use; passing around HpkeSender/Receiver would allow
potentially binding to _incorrect_ application data).

There are further refactorings possible--most importantly, I think
HpkePrivateKey should effectively always be paired with an HpkeConfig,
since knowing the config is required to use the private key. But this PR
is already big enough without adding even more refactorings on top.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -5,13 +5,13 @@ use crate::{
         models::{AggregationJob, AggregationJobState, ReportAggregation, ReportAggregationState},
         Datastore,
     },
-    hpke::HpkeRecipient,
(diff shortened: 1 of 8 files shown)


divergentdave (approved):

BranLwyd:
I also added a test for the new behavior in `aggregate_init` to bail if the HPKE config ID is unknown.

PR #87, opened 2022-04-18
Remove additional apk packages from Dockerfile

divergentdave:
These are no longer needed by console-subscriber. (which is no longer built by default, in the first place)


divergentdave:
Ah, I was too overeager with libc. I was fooled by local caching, which skipped rebuilding the affected crates, since they were previously built by a different container from when I still had the apk commands in.

BranLwyd (approved):

PR #88, opened 2022-04-19
Add support for OpenTelemetry traces, Jaeger

divergentdave:
This adds an option to export tracing spans and events via Jaeger. Up to now, our tracing subscribers have been primarily focused on events. Jaeger lets us visualize spans themselves, relations between spans, and the duration of each span. As with tokio-console, I have put it behind a feature and a configuration flag for now.


divergentdave:
Here's a screenshot of it in action (with #89 applied as well)
image

README.md
```diff
@@ -39,3 +39,18 @@ RUSTFLAGS="--cfg tokio_unstable" CARGO_TARGET_DIR=target/tokio_unstable cargo bu
 ```
 
 Install `tokio-console`, run the server, and run `tokio-console http://127.0.0.1:6669` to connect to it and monitor tasks.
+
+## OpenTelemetry with Jaeger
+
+There is experimental support for exporting application traces to Jaeger. When enabled, tracing spans will be mapped to OpenTelemetry spans, and trace data will be published to a local Jaeger agent via UDP. Enable this feature by compiling with the `jaeger` feature enabled, and add the following to the configuration file.
```
tgeoghegan, line 45: Maybe a link here to a website explaining what Jaeger is and how it relates to OpenTelemetry?

tgeoghegan (commented):
What happens if I run with `otel_jaeger: true` but no Jaeger agent is running?

Also I'd like to understand how this fits into our metrics strategy. I'd never heard of Jaeger before this PR. Are we now committing to using it in Divvi Up? Are we also exposing Prometheus metrics scrapes? I ask because Jaeger and Prometheus have two different strategies for exporting data (push to an agent vs. scrapes).

divergentdave:
With no agent running, I see `OpenTelemetry trace error occurred. Exporter jaeger encountered the following error(s): thrift agent failed with not open` printed to the terminal occasionally, lining up with ICMP "port unreachable" packets.

I'm not committed to using Jaeger. Between it, Grafana Tempo, and Honeycomb, it was the simplest to get started with.

One thing I've gathered recently is that OpenTelemetry only specifies an object model, and suggests some conventions for key names. Each product then handles the transport layer differently. I'll add a bit more exposition to the README.

BranLwyd:
I'm not pro- or anti-Jaeger; until this PR I had also never heard of it. I would have looked at Honeycomb first, but only because that's "the popular one."

Could we evaluate the different options before choosing one? It doesn't need to be formal, I just want to know that our choice makes sense vs the competition.

divergentdave:
After some more research, I found that Honeycomb and Jaeger both support the OTLP wire protocol, which the opentelemetry-rust SDK also includes an exporter crate for. I think I'll switch over to that and add instructions for using both.

It appears GCP Cloud Trace only supports uploading traces using their own proprietary API, and they don't provide a Rust-language SDK. However, it would be possible to export OTLP to an OpenTelemetry Collector process, which can in turn export to Cloud Trace, as it embeds GCP's Go SDK.

divergentdave:
I added the OLTP exporter, and instructions for connecting it to Honeycomb. I also tacked on a filter to weed out some `h2` spans that come from the OLTP exporter itself, which were drowning out our application's spans.

Jaeger is nice for local testing. It's easy to spin up without persistence, and if you only have a few traces, they're just a couple clicks away. Honeycomb's queries are definitely more powerful, and the query autocomplete makes it very easy to get started with.

janus_server/src/trace.rs
```diff
@@ -50,6 +67,19 @@ pub struct TokioConsoleConfiguration {
...
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct OtlpConfiguration {
+    /// Enable the OpenTelemetry tracing layer, with the OTLP exporter.
+    #[serde(default)]
+    pub enabled: bool,
+    /// gRPC endpoint for OTLP exporter.
+    pub endpoint: String,
+    /// Additional metadata/HTTP headers to be sent with OTLP requests.
+    #[serde(default)]
+    pub metadata: HashMap<String, String>,
```
tgeoghegan, line 80: IIUC the expectation is we'll put a Honeycomb API key into this metadata hashmap via the `aggregator` YAML config file. I don't think we should put a secret value like that in config. Can we plumb it through the aggregator binary's arguments, instead, so that it can be read from an environment variable populated from a Kubernetes secret?

janus_server/src/trace.rs
```diff
@@ -109,6 +139,82 @@ pub fn install_trace_subscriber(config: &TraceConfiguration) -> Result<(), Error
...
+                .with_trace_config(trace::config().with_resource(Resource::new(vec![
+                    KeyValue::new(SERVICE_NAME, "janus_server"),
+                ])))
+                .install_batch(opentelemetry::runtime::Tokio)?;
+
+        let filter = Targets::new()
+            .with_default(LevelFilter::TRACE)
+            .with_target("h2::proto::streams::prioritize", LevelFilter::OFF)
+            .with_target("h2::proto::streams::store", LevelFilter::OFF)
+            .with_target("h2::codec::framed_write", LevelFilter::OFF);
```
tgeoghegan, line 201: I'm sure these are sensible choices but a comment explaining why we are enabling (or disabling?) these features would be appreciated.

janus_server/src/trace.rs
```diff
@@ -13,6 +13,17 @@ pub enum Error {
     SetGlobalTracingSubscriber(#[from] tracing::subscriber::SetGlobalDefaultError),
     #[error("logging error: {0}")]
     SetGlobalLogger(#[from] tracing_log::log_tracer::SetLoggerError),
+    #[cfg(any(feature = "jaeger", feature = "otlp"))]
+    #[error(transparent)]
+    OTel(#[from] opentelemetry::trace::TraceError),
```
tgeoghegan, line 18:
```suggestion
    OpenTelemetry(#[from] opentelemetry::trace::TraceError),
```

janus_server/src/trace.rs
```diff
@@ -109,6 +139,82 @@ pub fn install_trace_subscriber(config: &TraceConfiguration) -> Result<(), Error
         );
     }
 
+    if config.otel_jaeger && config.otel_otlp.enabled {
+        return Err(Error::Other(anyhow::anyhow!(
+            "bad configuration, both Jaeger and OTLP OpenTelemetry layers cannot be enabled at the same time"
+        )));
```
tgeoghegan, line 145: I don't think we need to introduce `anyhow` here. IMO the point of `anyhow` is to wrap arbitrary error types from any number of dependencies, but here we're working with an error that we create. Could you just define `Error::Other(&'static str)` and do `return Err(Error::Other("bad configuration (...)"))`

tgeoghegan (changes requested):

janus_server/src/trace.rs
```diff
@@ -32,6 +43,12 @@ pub struct TraceConfiguration {
     /// (optional)
     #[serde(default)]
     pub tokio_console_config: TokioConsoleConfiguration,
+    /// Enable delivering OpenTelemetry traces to a Jaeger agent.
```
BranLwyd, line 46: I think we should document that `otel_jaeger` & `otel_otlp` are mutually-exclusive.

(Better yet would be making them mutually exclusive at the data structure level, e.g. similar to a Rust enum, but I'm not sure YAML encoding supports this kind of "oneof" functionality.)

janus_server/src/trace.rs
```diff
@@ -109,6 +139,84 @@ pub fn install_trace_subscriber(config: &TraceConfiguration) -> Result<(), Error
...
+        let tracer = opentelemetry_jaeger::new_pipeline()
+            .with_service_name("janus_server")
+            .install_batch(opentelemetry::runtime::Tokio)?;
+        let telemetry = tracing_opentelemetry::layer().with_tracer(tracer);
+        layers.push(telemetry.boxed());
+    }
+
+    #[cfg(not(feature = "jaeger"))]
+    if config.otel_jaeger {
+        eprintln!(
```
BranLwyd, line 159: IMO, trying to enable features not compiled in should result in a failure, rather than a warning. A warning may not be noticed & means we'll just keep rolling with a difference between what we think we're doing & what we're actually doing. A failure is a lot harder to ignore.

[same comment for the warning around L213]

BranLwyd (changes requested):

janus_server/src/trace.rs
divergentdave: We could put these two in their own enum, serde would turn that into another layer of nesting, like so:

```yaml
logging_config:
  otel:
    otlp:
      enabled: true
      endpoint: "https://www.example.com"
```

When parsing the enum, serde would complain loudly if it found a dictionary with both "jaeger" and "otlp" keys, rather than one or the other.

janus_server/src/trace.rs
```diff
@@ -50,6 +64,43 @@ pub struct TokioConsoleConfiguration {
...
+            OpenTelemetryTraceConfiguration::Jaeger(config) => config.enabled,
+            OpenTelemetryTraceConfiguration::Otlp(config) => config.enabled,
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq, Serialize, Deserialize)]
+pub struct JaegerConfiguration {
+    /// Enable the Jaeger exporter.
+    pub enabled: bool,
```
BranLwyd, line 88: Is there a good reason not to use the existence of the Jaeger/OTLP config as the enabled flag? That is, is there a use-case for setting this `enabled` flag to false?

If not, I suggest dropping this flag & `OtlpConfiguration.enabled`, and using the following semantics:
* If `open_telemetry_config` is omitted entirely, disabled Jaeger/OTLP.
* If `open_telemetry_config.jaeger` is specified, enable Jaeger.
* If `open_telemetry_config.otlp` is specified, enable OTLP.

BranLwyd (approved):
One comment (please address before merging), but LGTM otherwise.

tgeoghegan (approved):

PR #89, opened 2022-04-19
Instrument Datastore/Transaction methods w/ spans

divergentdave:
This adds tracing spans to `Datastore::run_tx_once()` and the various `Transaction` accessors. Database access is both the major system boundary in our deployment, and also the chief contributor to request latency, so this seems like a natural place to add spans.

### DIFF
--- a/janus_server/src/datastore.rs
+++ b/janus_server/src/datastore.rs
@@ -65,6 +65,7 @@ impl Datastore {
         }
     }
 
+    #[tracing::instrument(skip(self, f))]
     async fn run_tx_once<F, T>(&self, f: &F) -> Result<T, Error>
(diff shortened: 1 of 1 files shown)


janus_server/src/datastore.rs
```diff
@@ -65,6 +65,7 @@ impl Datastore {
         }
     }
 
+    #[tracing::instrument(skip(self, f))]
```
tgeoghegan, line 68: I think we should add `err` alongside `skip` so that we emit events when these functions return errors.

tgeoghegan (changes requested):

tgeoghegan (approved):

PR #93, opened 2022-04-20
Skeleton of the leader `/collect` endpoint

tgeoghegan:


janus_server/src/aggregator.rs
```diff
@@ -181,6 +188,23 @@ impl<C: Clock> Aggregator<C> {
...
     }
 
+    async fn handle_collect(&self, req_bytes: &[u8]) -> Result<Url, Error> {
+        let collect_req = CollectReq::get_decoded(req_bytes)?;
+
+        let task_aggregator = self.task_aggregator_for(collect_req.task_id).await?;
+
+        // Only the leader supports collect.
+        if task_aggregator.task.role != Role::Leader {
+            // TODO (timg): We should make sure that a helper returns HTTP 404 or 403 when this
```
BranLwyd, line 198: I think we (eventually) want a new error for this case; while UnrecognizedTask can map to a 404, semantically it's incorrect for the error here (which is more "wrong role for this task") which will lead to e.g. misleading problem document values.

BranLwyd (approved):

divergentdave (approved):

janus_server/src/aggregator.rs
```diff
@@ -181,6 +188,23 @@ impl<C: Clock> Aggregator<C> {
         Err(Error::InvalidHmac(task_id))
     }
 
+    async fn handle_collect(&self, req_bytes: &[u8]) -> Result<Url, Error> {
+        let collect_req = CollectReq::get_decoded(req_bytes)?;
```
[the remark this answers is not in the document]
tgeoghegan: I agree; this is tracked in https://github.com/ietf-wg-ppm/ppm-specification/issues/155

janus_server/src/aggregator.rs
tgeoghegan: I agree. I was reproducing what `Aggregator::handle_upload` does in the corresponding case, but I think both should probably just return HTTP 403, maybe without any body. I filed #96 for this.

PR #95, opened 2022-04-21
Make testcontainers a dev-dependency

divergentdave:
This is stacked on top of #94.

This moves the declaration of items necessary for an ephemeral datastore into a macro in a new crate, `test_util`. This macro is then invoked inside `janus_server` only in its unit tests, as well as separately in each integration test. As a result, `testcontainers` can be removed from `janus_server`'s runtime dependencies.

I'm using a macro tactic here to enable code reuse while keeping `ephemeral_datastore()` out of `janus_server`'s exposed API, and to be able to use it within `janus_server`'s unit tests without setting up a dev-dependency cycle.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1240,6 +1240,7 @@ dependencies = [
  "signal-hook-tokio",
  "structopt",
  "tempfile",
+ "test_util",
  "testcontainers",
  "thiserror",
  "tokio",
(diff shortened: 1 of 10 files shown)


tgeoghegan (approved):
I'm not crazy about the macro, but I don't have a better answer, short of factoring `janus_server:;datastore` out into a further crate that both `test_util` and `janus_server` could depend on.

BranLwyd (approved):
This is an extremely gross workaround/hack for Rust integration tests not running under `#[cfg(test)]` :-(. We're effectively duplicating code, though the macro at least lets us edit the duplicated code in one place.

But yeah, AFAIK there really is no better solution, so LGTM.

PR #97, opened 2022-04-22
Store collect jobs in the database

tgeoghegan:
Adds support for fetching collect job UUIDs from the database, and
creating collect jobs on demand.

### DIFF
--- a/janus_server/src/datastore.rs
+++ b/janus_server/src/datastore.rs
@@ -7,8 +7,8 @@ use self::models::{
 use crate::{
     hpke::HpkePrivateKey,
     message::{
-        AggregationJobId, Extension, HpkeCiphertext, HpkeConfig, Nonce, Report, ReportShare,
-        TaskId, Time,
+        AggregationJobId, Extension, HpkeCiphertext, HpkeConfig, Interval, Nonce, Report,
+        ReportShare, TaskId, Time,
     },
     task::{self, AggregatorAuthKey, Task},
 };
@@ -23,6 +23,7 @@ use ring::aead::{self, LessSafeKey, AES_128_GCM};
 use std::{convert::TryFrom, fmt::Display, future::Future, io::Cursor, mem::size_of, pin::Pin};
 use tokio_postgres::{error::SqlState, row::RowIndex, IsolationLevel, Row};
 use url::Url;
+use uuid::Uuid;
 
 // TODO(brandon): retry network-related & other transient failures once we know what they look like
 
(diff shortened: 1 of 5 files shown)


janus_server/src/datastore.rs
```diff
@@ -821,6 +822,74 @@ impl Transaction<'_> {
...
+        encoded_aggregation_parameter: &[u8],
+    ) -> Result<Uuid, Error> {
+        let batch_interval_start = batch_interval.start.as_naive_date_time();
+        let batch_interval_duration = i64::try_from(batch_interval.duration.0)?;
+
+        let stmt = self
+            .tx
+            .prepare_cached(
+                "SELECT collect_job_id FROM collect_jobs
+            WHERE task_id = (SELECT id FROM tasks WHERE task_id = $1)
```
BranLwyd, line 844: nit: indentation

janus_server/src/datastore.rs
```diff
@@ -821,6 +822,74 @@ impl Transaction<'_> {
                 .await?,
         )
     }
+
+    /// If a collect job corresponding to the provided values exists, its UUID
+    /// is returned, which may then be used to construct a collect job URI. If
+    /// that collect job does not exist, then a UUID is generated and the
+    /// collect job is stored, and the UUID is returned.
+    #[tracing::instrument(skip(self, encoded_aggregation_parameter), err)]
+    pub(crate) async fn lookup_collect_job_uri(
```
BranLwyd, line 831: IMO this shouldn't be named "lookup", since it creates collect jobs, too.

To match the API exposed by Transaction for other nounds, maybe create a pair of methods, "get_collect_job_for_parameters" (properly wordsmithed, returning NotFound if no collect job is returned), and "put_collect_job"? Ideally the getter would return a (new) CollectJob type that represents a collect job.

BranLwyd (changes requested):

divergentdave (approved):

janus_server/src/datastore.rs
```diff
@@ -823,27 +823,26 @@ impl Transaction<'_> {
...
         let stmt = self
             .tx
             .prepare_cached(
                 "SELECT collect_job_id FROM collect_jobs
-            WHERE task_id = (SELECT id FROM tasks WHERE task_id = $1)
-            AND batch_interval_start = $2 AND batch_interval_duration = $3
-            AND aggregation_param = $4",
+                WHERE task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                    AND batch_interval_start = $2 AND batch_interval_duration = $3
+                    AND aggregation_param = $4",
```
BranLwyd, line 845:
```suggestion
                AND batch_interval_start = $2 AND batch_interval_duration = $3
                AND aggregation_param = $4",
```

janus_server/src/datastore.rs
```diff
@@ -859,20 +858,29 @@ impl Transaction<'_> {
             )
             .await?;
 
-        // Found an existing collect job
-        if let Some(row) = row {
-            return Ok(row.get("collect_job_id"));
-        }
+        Ok(row.map(|row| row.get("collect_job_id")))
```
BranLwyd, line 861: Q: It's sorta strange that this returns an `Option` determined by whether the row exists -- all other methods return the found object directly, and if there is no such object, they return `Err(NotFound)`.

An `Option` is the more Rust-idiomatic way to return an optional value. OTOH, all other methods "expect" the object to exist, so it's an error-worthy surprising condition if they don't -- that doesn't apply here, I suppose.

I'm going to make this uniform across all methods soon; I was wondering if you had any thoughts on which way (`Option` vs `NotFound`) would be better.

BranLwyd (approved):

janus_server/src/datastore.rs
divergentdave: I'm inclined toward `Option` in cases like this where we may expect one or zero rows. I just wrote a match block that pulled out `Error::NotFound` on `get_task()`, and I can see us having to do this more as we implement all the anti-replay logic.

tgeoghegan: I prefer using an `Option` for the same reasons as David: if we use `Err()` only for errors, then we can always just propagate those with `?`.

janus_server/src/datastore.rs
tgeoghegan: Sounds like we need an inline SQL linter! Done.

PR #103, opened 2022-04-25
Helper `/aggregate_share` endpoint

tgeoghegan:
Implements the helper's `aggregate_share` endpoint. The assumption is
that the process of preparing input shares into output shares will
create rows in `batch_aggregations` and update the `checksum` and
`aggregate_share` columns as individual reports are prepared. Then, all
the `/aggregate_share` handler has to do is sum the aggregate shares it
finds.

Note that this does not include support for the protocol changes in [1].

[1]: https://github.com/ietf-wg-ppm/ppm-specification/pull/224

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -9,11 +9,12 @@ use crate::{
     message::{
         AggregateReq,
         AggregateReqBody::{AggregateContinueReq, AggregateInitReq},
-        AggregateResp, AggregationJobId, AuthenticatedDecodeError, AuthenticatedEncoder,
-        AuthenticatedRequestDecoder, CollectReq, HpkeConfig, HpkeConfigId, Interval, Nonce, Report,
-        ReportShare, Role, TaskId, Transition, TransitionError, TransitionTypeSpecificData,
+        AggregateResp, AggregateShareReq, AggregateShareResp, AggregationJobId,
+        AuthenticatedDecodeError, AuthenticatedEncoder, AuthenticatedRequestDecoder, CollectReq,
+        HpkeConfig, HpkeConfigId, Interval, Nonce, Report, ReportShare, Role, TaskId, Transition,
+        TransitionError, TransitionTypeSpecificData,
     },
-    task::{self, Task},
+    task::{self, AggregatorAuthKey, Task},
     time::Clock,
 };
 use bytes::Bytes;
@@ -24,10 +25,11 @@ use http::{
 };
 use prio::{
     codec::{Decode, Encode, ParameterizedDecode},
+    field::FieldError,
     vdaf::{
         self,
         prio3::{Prio3Aes128Count, Prio3Aes128Histogram, Prio3Aes128Sum},
-        PrepareTransition, Vdaf,
+        Aggregatable, PrepareTransition, Vdaf,
     },
 };
 use std::{
@@ -92,8 +94,30 @@ pub enum Error {
     /// A collect or aggregate share request was rejected because the interval is valid, per §4.6
     #[error("Invalid batch interval: {0} {1:?}")]
     InvalidBatchInterval(Interval, TaskId),
+    /// There are not enough reports in the batch interval to meet the task's minimum batch size.
+    #[error("Insufficient number of reports ({0}) for task {1:?}")]
+    InsufficientBatchSize(u64, TaskId),
     #[error("URL parse error: {0}")]
     Url(#[from] url::ParseError),
+    /// The checksum or report count in one aggregator's aggregate share does not match the other
+    /// aggregator's aggregate share, suggesting different sets of reports were aggregated.
+    #[error(
+        "Batch misalignment: own checksum: {own_checksum:?} own report count: {own_report_count} \
+peer checksum: {peer_checksum:?} peer report count: {peer_report_count}"
+    )]
+    BatchMisalignment {
+        task_id: TaskId,
+        own_checksum: [u8; 32],
+        own_report_count: u64,
+        peer_checksum: [u8; 32],
+        peer_report_count: u64,
+    },
+    /// Too many queries against a single batch.
+    #[error("Maxiumum batch lifetime for task {0:?} exceeded")]
+    BatchLifetimeExceeded(TaskId),
+    /// HPKE failure.
+    #[error("HPKE error: {0}")]
+    Hpke(#[from] crate::hpke::Error),
     /// An error representing a generic internal aggregation error; intended for "impossible"
     /// conditions.
     #[error("internal aggregator error: {0}")]
@@ -164,7 +188,7 @@ impl<C: Clock> Aggregator<C> {
         let resp = task_aggregator
             .handle_aggregate(&self.datastore, req)
             .await?;
-        let key = task_aggregator.task.agg_auth_keys.last().unwrap();
+        let key = task_aggregator.current_aggregator_auth_key();
         Ok(AuthenticatedEncoder::new(resp).encode(key.as_ref()))
     }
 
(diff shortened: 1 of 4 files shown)


janus_server/src/datastore.rs
```diff
@@ -898,6 +898,142 @@ impl Transaction<'_> {
...
+    /// Store a new `batch_aggregations` row in the datastore.
+    #[cfg(test)]
+    #[tracing::instrument(skip(self), err)]
+    pub(crate) async fn put_batch_aggregation<A>(
+        &self,
+        task_id: TaskId,
+        batch_interval_start: Time,
+        aggregate_share: &A,
+        report_count: u64,
+        checksum: &[u8],
```
tgeoghegan, line 911: Might be nice to have a `Checksum` newtype around `[u8; 32]` but this PR is already too big.

janus_server/src/aggregator.rs
```diff
@@ -933,6 +1010,171 @@ impl VdafOps {
...
+        // TODO(timg): What should we do if any of the aggregation jobs handling reports relevant to
+        // the aggregate share request aren't finished yet? Helper could return an error like
+        // "EAGAIN" to leader, or could trust leader not to issue an AggregateShareReq until all of
+        // leader's aggregate jobs are done (which should imply that helper's are done).
+
+        let total_aggregate_share = datastore
+            .run_tx(move |tx| {
+                let task = task.clone();
+                let aggregate_share_req = aggregate_share_req.clone();
+                Box::pin(async move {
```
tgeoghegan, line 1075: Note that I'm holding the DB transaction open from when we get the aggregations until we increment the consumed lifetime, which I think is necessary to account for multiple concurrent requests to `aggregate_share`.

janus_server/src/aggregator.rs
```diff
@@ -933,6 +1010,171 @@ impl VdafOps {
...
+                    let batch_aggregations = tx
+                        .get_batch_aggregations::<A, E>(task.id, aggregate_share_req.batch_interval)
+                        .await?;
+
+                    let mut total_report_count = 0;
+                    let mut total_checksum = [0u8; 32];
+                    let mut total_aggregate_share: Option<A> = None;
+
+                    for batch_aggregation in &batch_aggregations {
+                        // TODO(timg): This satisfies the max lifetime requirement in §4.6 but it
```
BranLwyd, line 1085: +1.

If we end up including the aggregation parameter in the `batch_aggregations` table, I think we could do this by counting the number of `batch_aggregations` there are for the `(batch window, aggregation parameter)`.

janus_server/src/aggregator.rs
```diff
@@ -3217,6 +3507,412 @@ mod tests {
...
+
+        // Interval is big enough, but report count doesn't match
+        let request = AggregateShareReq {
+            task_id,
+            batch_interval: Interval {
+                start: Time(0),
+                duration: message::Duration(2000),
+            },
+            report_count: 20,
+            checksum: [2; 32],
```
divergentdave, line 3809: This should be `3 ^ 2` as well, we may be taking the wrong branch to the right error.

janus_server/src/datastore.rs
```diff
@@ -2169,6 +2330,94 @@ mod tests {
         assert!(rows.len() == 2);
     }
 
+    #[tokio::test]
+    async fn roundtrip_batch_aggregation() {
+        install_test_trace_subscriber();
+
+        let task_id = TaskId::random();
+        let other_task_id = TaskId::random();
+        //let vdaf = Prio3Aes128Count::new(2).unwrap();
```
BranLwyd, line 2339: nit: drop this line

janus_server/src/datastore.rs
```diff
@@ -2169,6 +2330,94 @@ mod tests {
...
+                            duration: Duration(200),
+                        },
+                    )
+                    .await
+                })
+            })
+            .await
+            .unwrap();
+
+        let expected_batch_aggregations = [
```
BranLwyd, line 2399: pedantic nitpick: I think this is currently ordered, so if Postgres changes to return the rows in opposite order, this test will break. Drop ordering on the check in this test?

BranLwyd (changes requested):
almost all comments/questions, a couple of nitpicks

divergentdave (approved):

janus_server/src/aggregator.rs
```diff
@@ -933,6 +1010,171 @@ impl VdafOps {
...
+                            .iter_mut()
+                            .zip(batch_aggregation.checksum)
+                            .for_each(|(x, y)| *x ^= y);
+
+                        // §4.4.4.3: Sum all the report counts
+                        total_report_count += batch_aggregation.report_count;
+
+                        match &mut total_aggregate_share {
+                            Some(share) => {
+                                if let Err(err) = share.merge(&batch_aggregation.aggregate_share) {
```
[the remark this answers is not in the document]
divergentdave: VDAFs are defined with an "out_shares_to_agg_share" method, which takes an array of output shares, and produces an aggregate share. The end of that paragraph notes that "Some VDAFs may also support a 'streaming' form, where shares are processed one at a time." A non-streaming VDAF would require a different approach, as you note, but I think the benefits of accumulating aggregation jobs, then batches, then collect requests incrementally are worth it. (For example, a non-streaming VDAF's out_shares_to_agg_share may have a working set size that scales with the number of reports)

janus_server/src/datastore.rs
```diff
@@ -898,6 +898,142 @@ impl Transaction<'_> {
...
+    {
+        let batch_interval_start = interval.start.as_naive_date_time();
+        let batch_interval_end = interval.end().as_naive_date_time();
+
+        let stmt = self
+            .tx
+            .prepare_cached(
+                "SELECT batch_interval_start, aggregate_share, report_count, checksum, times_collected
+                FROM batch_aggregations
+                WHERE task_id = (SELECT id FROM tasks WHERE task_id = $1) AND batch_interval_start >= $2 AND batch_interval_start < $3",
```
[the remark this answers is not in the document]
tgeoghegan: Actually, this logic is plain wrong. What this function should do is return all the batch intervals that are completely inside of `interval`. As written, an interval is considered included if just its first instant is inside `interval`. I need to change the contract here.

janus_server/src/aggregator.rs
```diff
@@ -933,6 +1010,171 @@ impl VdafOps {
...
+    where
+        A: Aggregatable + Send + Sync,
+        E: std::fmt::Debug,
+        Vec<u8>: for<'a> From<&'a A>,
+        for<'a> A: TryFrom<&'a [u8], Error = E>,
+    {
+        // TODO(timg): What should we do if any of the aggregation jobs handling reports relevant to
+        // the aggregate share request aren't finished yet? Helper could return an error like
+        // "EAGAIN" to leader, or could trust leader not to issue an AggregateShareReq until all of
+        // leader's aggregate jobs are done (which should imply that helper's are done).
```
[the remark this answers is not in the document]
tgeoghegan: I wrote up #104 to pursue this discussion. I'll add a reference to the issue in the comment.

janus_server/src/aggregator.rs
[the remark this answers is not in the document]
tgeoghegan: Right, we can't use `prio::vdaf::Aggregator::aggregate` here because we are not working with output shares, but rather accumulations. `prio::vdaf::Collector::unshard` is what you're meant to use to unshard a sequence of `AggregateShare`s into an `AggregateResult`, so IMO this usage is precisely what `Aggregatable::merge` is for (I'm pretty sure I added `Aggregatable::merge` for exactly this purpose). However it sounds like we could use some better documentation in `libprio-rs` here, and maybe we want some corresponding definitions in the VDAF spec.

janus_server/src/aggregator.rs
tgeoghegan: Correctly enforcing max_batch_lifetime requires more persistence on the helper side, so I punted for now to avoid bloating this PR.

PR #107, opened 2022-04-27
Safer time calculations

tgeoghegan:
We had a number of risky casts between `chrono::Duration` and
`janus_server::message::Duration`, or between `i64` and `u64` in other
places. With this commit, we consistently use `message::{Duration, Time,
Interval}` everywhere. We also now have a safe constructor for
`Interval` that forbids the creation of an interval whose end cannot be
represented as a `Time`. We also have safe routines for doing math on
durations and times that return errors if overflows occur.


janus_server/src/message.rs
```diff
@@ -158,9 +166,20 @@ fn authenticated_decode<M: Decode>(
...
 
 impl Duration {
+    /// Create a duration representing the provided number of seconds.
     pub fn from_seconds(seconds: u64) -> Self {
         Self(seconds)
     }
+
+    /// Create a duration representing the provided number of minutes.
+    pub fn from_minutes(minutes: u64) -> Self {
+        Self::from_seconds(60 * minutes)
```
BranLwyd, line 176: Optional: since we're being pedantic -- do a checked multiply for `from_minutes` & `from_hours`?

BranLwyd (approved):

Issue #105, opened 2022-04-26
Collect protocol implementation


tgeoghegan:
To implement the collect protocol, we need:

- [ ] `[leader]/collect` endpoint (#93, #97)
- [ ] Collect job URI handler on leader
- [ ] `[helper]/aggregate_share` endpoint (#103)

tgeoghegan:
- [ ] `max_batch_lifetime` enforcement in helper

When servicing an `AggregateShareReq`, helper needs to check whether any of the minimum batch intervals that make up `AggregateShareReq.batch_interval` have already been included in some other `AggregateShareReq`. To that end, helper will need to persist the parameters of the `AggregateShareReq`s it services.

- [ ] `max_batch_lifetime` enforcement in leader

Leader needs similar enforcement. This should be done in the `/collect` handler.

- [ ] `AggregateShareReq` needs `aggregation_param` field (added in https://github.com/ietf-wg-ppm/ppm-specification/pull/224)

PR #111, opened 2022-04-29
Store aggregate share requests serviced by helper

tgeoghegan:
Adds a new database table `aggregate_share_jobs`, used by helper to
store the results of successfully serviced `AggregateShareReq`s. This
allows leaders to retry an `AggregateShareReq` indefinitely, provided
the parameters don't change. The leader's `collect_jobs` table now also
has some nullable columns where it can cache the leader and helper's
aggregate shares, to similarly allow the collector to retry requests to
the collect job URI.

Part of #105

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -2,8 +2,11 @@
 use crate::{
     datastore::{
         self,
-        models::{AggregationJob, AggregationJobState, ReportAggregation, ReportAggregationState},
-        Datastore,
+        models::{
+            AggregateShareJob, AggregationJob, AggregationJobState, ReportAggregation,
+            ReportAggregationState,
+        },
+        Datastore, Transaction,
     },
     hpke::{self, HpkeApplicationInfo, Label},
     message::{
@@ -46,7 +49,7 @@ use std::{
     time::Instant,
 };
 use tokio::sync::Mutex;
-use tracing::{error, warn};
+use tracing::{debug, error, warn};
 use url::Url;
 use warp::{
     filters::BoxedFilter,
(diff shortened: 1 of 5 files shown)


janus_server/src/aggregator.rs
```diff
@@ -1076,102 +1160,65 @@ impl VdafOps {
...
+                                ?aggregate_share_req,
+                                "cache miss, computing aggregate share job result"
+                            );
+                            // No cached aggregate job, so attempt to compute the aggregate share.
+                            Self::service_aggregate_share_request::<A, E>(
+                                tx,
+                                &task,
+                                &aggregate_share_req,
+                            )
+                            .await
```
divergentdave, line 1195: `service_aggregate_share_request()` makes one or two database queries, but if they error, the transaction closure will return `Ok(Err(Error::Datastore(datastore::Error::Db(...))))`, and `run_tx()` would fail to notice and retry serialization failures. I think we want `service_aggregate_share_request()` to return a nested `Result`, as is done with the calling closure. (datastore errors on the outside, aggregator errors for PPM problem documents on the inside)

divergentdave (changes requested):

janus_server/src/datastore.rs
```diff
@@ -1018,6 +1014,109 @@ impl Transaction<'_> {
 
         Ok(rows)
     }
+
+    /// Fetch an `aggregate_share_jobs` row from the datastore corresponding to the provided
+    /// [`AggregateShareRequest`], or `None` if no such job exists.
+    #[tracing::instrument(skip(self), err)]
+    pub(crate) async fn get_aggregate_share_job_by_request<P, A, E>(
```
BranLwyd, line 1021: Please make this take a single generic parameter, which is a `vdaf::Aggregator` (or maybe `vdaf::Vdaf`, if you don't need the additional functionality in the `Aggregator` trait), and state the constraints on the existing parameters in terms of the associated types of the `vdaf::Vdaf`.

`AggregateShareJob`'s generic parameters would be changed similarly, as would `put_aggregate_share_job`'s.

The intended meaning of these generic parameters is unnecessarily unclear, and I am not sure there is an upside to making this so flexible -- the intended use is always in the context of a specific `vdaf::Aggregator`. And we've already gone to the trouble of defining the associated types in libprio, let's make use of them. :-)

BranLwyd (changes requested):

janus_server/src/datastore.rs
BranLwyd: (I realized this was a preexisting issue, see https://github.com/divviup/janus/pull/130 for an example fix. I think the trickiest bit is that the auto-derived `PartialEq`/`Eq` implementations don't seem to play nicely. Maybe it's a GAT thing?)

divergentdave: I think the `#[derive(PartialEq)]` issue is referred to as "perfect derive" in [this roadmap blogpost](https://blog.rust-lang.org/inside-rust/2022/04/04/lang-roadmap-2024.html#looking-forward) FWIW, looks like it's orthogonal to GAT.

janus_server/src/aggregator.rs
tgeoghegan: I believe that this has been addressed by Brandon's suggestion of using `datastore::Error::User`, without requiring nested `Result`.

janus_server/src/aggregator.rs
```diff
@@ -1076,102 +1160,65 @@ impl VdafOps {
         // leader's aggregate jobs are done (which should imply that helper's are done).
         // See issue #104.
 
-        let total_aggregate_share = datastore
+        let aggregate_share_job = datastore
             .run_tx(move |tx| {
                 let task = task.clone();
                 let aggregate_share_req = aggregate_share_req.clone();
+                // This closure must return Result<T, datastore::Error>, so we smuggle
```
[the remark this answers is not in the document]
tgeoghegan: Much better, thank you for the suggestion.

divergentdave (approved):

BranLwyd (approved):

PR #113, opened 2022-04-29
OpenTelemetry Metrics

divergentdave:
This adds metrics support, using the OpenTelemetry SDK. I included both a Prometheus exporter and an OTLP exporter. (Honeycomb supports metrics via OTLP, though only as a premium feature) I have included a couple metrics around HTTP request handling, as a start.

FWIW, due to changes in the OpenTelemetry metrics specification, future crate versions will remove the "ValueRecorder" layer of indirection, deprecating it and replacing it with histograms.

Here's a sample scrape, using the Prometheus exporter.

```
# HELP aggregator_response Success and failure responses to incoming requests.
# TYPE aggregator_response counter
aggregator_response{endpoint="hpke_config",status="error"} 1
# HELP aggregator_response_time Elapsed time handling incoming requests.
# TYPE aggregator_response_time histogram
aggregator_response_time_bucket{endpoint="hpke_config",le="0.005"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="0.01"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="0.025"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="0.05"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="0.1"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="0.25"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="0.5"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="1"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="2.5"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="5"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="10"} 1
aggregator_response_time_bucket{endpoint="hpke_config",le="+Inf"} 1
aggregator_response_time_sum{endpoint="hpke_config"} 0.000439081
aggregator_response_time_count{endpoint="hpke_config"} 1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,4 +73,33 @@ logging_config:
         x-honeycomb-team: "YOUR_API_KEY"
 ```
 
-The gRPC metadata can also be specified on the command line, with `--otlp-metadata x-honeycomb-team=YOUR_API_KEY`.
+The gRPC metadata can also be specified on the command line, with `--otlp-metadata x-honeycomb-team=YOUR_API_KEY`, or through the environment variable `OTLP_METADATA`.
+
+## OpenTelemetry Metrics
+
+Application-level metrics from the server can be exported to one of the following services.
+
+### Prometheus
+
+When the Prometheus exporter is enabled, a server will listen on port 9464 for metrics scrape requests. Prometheus must be configured to scrape the server, either manually or via an auto-discovery mechanism. Compile `janus_server` with the `prometheus` feature enabled, and add the following to the configuration file.
+```yaml
+metrics_config:
+  exporter:
+    prometheus:
+```
+
+### Honeycomb
+
+Honeycomb also supports OpenTelemetry-formatted metrics, though only on the Enterprise and Pro plans. Compile `janus_server` with the `otlp` feature enabled, and add the following section to the configuration file. Note that the OTLP/gRPC exporter will push metrics at regular intervals.
+
+```yaml
+metrics_config:
+  exporter:
+    otlp:
+      endpoint: "https://api.honeycomb.io:443"
+      metadata:
+        x-honeycomb-team: "YOUR_API_KEY"
+        x-honeycomb-dataset: "YOUR_METRICS_DATASET"
+```
+
+The command line flag `--otlp-metadata` or environment variable `OTLP_METADATA` may alternately be used to supply gRPC metadata for the metrics exporter, as with the tracing exporter above.
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -23,7 +23,10 @@ use http::{
     header::{CACHE_CONTROL, LOCATION},
     StatusCode,
 };
-use opentelemetry::{metrics::Counter, KeyValue};
+use opentelemetry::{
+    metrics::{Counter, Unit, ValueRecorder},
+    KeyValue,
+};
 use prio::{
     codec::{Decode, Encode, ParameterizedDecode},
     field::FieldError,
@@ -47,7 +50,7 @@ use url::Url;
 use warp::{
     filters::BoxedFilter,
     reply::{self, Response},
-    trace, Filter, Reply,
+    trace, Filter, Rejection, Reply,
 };
 
 #[cfg(test)]
@@ -1370,6 +1373,33 @@ fn error_handler<R: Reply>(
     }
 }
 
+/// Factory that produces a closure that will wrap a `warp::Filter`, measuring the time that it
+/// takes to run, and recording it in metrics.
+fn timing_wrapper<F, T>(
+    value_recorder: &ValueRecorder<f64>,
+    name: &'static str,
+) -> impl Fn(F) -> BoxedFilter<(T,)>
+where
+    F: Filter<Extract = (T,), Error = Rejection> + Clone + Send + Sync + 'static,
+    T: Reply,
+{
+    let bound_value_recorder = value_recorder.bind(&[KeyValue::new("endpoint", name)]);
+    move |filter| {
+        warp::any()
+            .map(std::time::Instant::now)
+            .and(filter)
+            .map({
+                let bound_value_recorder = bound_value_recorder.clone();
+                move |start: std::time::Instant, reply| {
+                    let elapsed = start.elapsed().as_secs_f64();
+                    bound_value_recorder.record(elapsed);
+                    reply
+                }
+            })
+            .boxed()
+    }
+}
+
 /// Constructs a Warp filter with endpoints common to all aggregators.
 fn aggregator_filter<C>(
     datastore: Arc<Datastore>,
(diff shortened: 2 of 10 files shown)


janus_server/src/metrics.rs
```diff
@@ -0,0 +1,178 @@
...
+#[derive(Debug)]
+struct CustomAggregatorSelector;
+
+impl AggregatorSelector for CustomAggregatorSelector {
+    fn aggregator_for(&self, descriptor: &Descriptor) -> Option<Arc<dyn Aggregator + Send + Sync>> {
+        match descriptor.instrument_kind() {
+            opentelemetry::metrics::InstrumentKind::ValueRecorder => {
+                // The following boundaries are copied from the Ruby and Go Prometheus clients.
+                // Buckets could be specialized per-metric by matching on `descriptor.name()`.
+                let boundaries = &[
```
BranLwyd, line 75: If we create a metric with these boundaries, will we be able to switch later or are we "locked in" for that metric? I'm sure these are reasonable default buckets but the right buckets for a given metric are going to be heavily metric-dependent.

janus_server/src/metrics.rs
```diff
@@ -0,0 +1,178 @@
...
+                    match encoder.encode(&exporter.registry().gather(), &mut buffer) {
+                        Ok(()) => Response::builder()
+                            .header(hyper::header::CONTENT_TYPE, encoder.format_type())
+                            .body(buffer)
+                            // This unwrap is OK because the only possible source of errors is the
+                            // `header()` call, and its arguments are always valid.
+                            .unwrap()
+                            .into_response(),
+                        Err(err) => {
+                            tracing::error!(%err, "failed to encode Prometheus metrics");
```
BranLwyd, line 123:
```suggestion
                            tracing::error!(%err, "Failed to encode Prometheus metrics");
```

nit

janus_server/src/metrics.rs
```diff
@@ -0,0 +1,178 @@
...
+            use prometheus::{Encoder, TextEncoder};
+            use std::net::SocketAddr;
+            use warp::{Filter, Reply};
+
+            let exporter = opentelemetry_prometheus::exporter()
+                .with_aggregator_selector(CustomAggregatorSelector)
+                .try_init()?;
+
+            let filter = warp::path("metrics").and(warp::get()).map({
+                let exporter = exporter.clone();
```
BranLwyd, line 110: Wrap in an `Arc` or otherwise make sure we're cloning a reference, not the entire value.

janus_server/src/metrics.rs
```diff
@@ -0,0 +1,178 @@
...
+            use hyper::Response;
+            use prometheus::{Encoder, TextEncoder};
+            use std::net::SocketAddr;
+            use warp::{Filter, Reply};
+
+            let exporter = opentelemetry_prometheus::exporter()
+                .with_aggregator_selector(CustomAggregatorSelector)
+                .try_init()?;
+
+            let filter = warp::path("metrics").and(warp::get()).map({
```
BranLwyd, line 109: What is the security story here -- how do we keep arbitrary users of the Internet from hitting this endpoint and receiving metric data? Will we be configuring things such that this metrics port is only exposed internally for our prometheus instance to scrape from? And if so -- how do we specify the port? I don't see anything in this code or in the Prometheus config that would allow the port to be controlled.

README.md
```diff
@@ -73,4 +73,33 @@ logging_config:
...
+metrics_config:
+  exporter:
+    otlp:
+      endpoint: "https://api.honeycomb.io:443"
+      metadata:
+        x-honeycomb-team: "YOUR_API_KEY"
+        x-honeycomb-dataset: "YOUR_METRICS_DATASET"
+```
+
+The command line flag `--otlp-metadata` or environment variable `OTLP_METADATA` may alternately be used to supply gRPC metadata for the metrics exporter, as with the tracing exporter above.
```
BranLwyd, line 105: Q: is it possible we'll want to specify different OTLP metadata to the metrics vs the logging config? Or alternatively, a reason to believe we _won't_ want to specify different metadata?

BranLwyd (changes requested):

janus_server/src/metrics.rs
divergentdave: This was just 48 bytes (two Arcs, a String, and a u16) but I can actually avoid the clone altogether.

janus_server/src/metrics.rs
divergentdave: Yes, we can keep the exporter port internal only. The exporter configuration builder takes the port from an environment variable or a default, I'll document that and add an option in our configuration.

BranLwyd (approved):

BranLwyd:
(need to resolve merge conflicts, but LGTM)

janus_server/src/aggregator.rs
```diff
@@ -1362,6 +1382,33 @@ fn error_handler<R: Reply>() -> impl Fn(Result<R, Error>) -> warp::reply::Respon
...
+    name: &'static str,
+) -> impl Fn(F) -> BoxedFilter<(T,)>
+where
+    F: Filter<Extract = (T,), Error = Rejection> + Clone + Send + Sync + 'static,
+    T: Reply,
+{
+    let bound_value_recorder = value_recorder.bind(&[KeyValue::new("endpoint", name)]);
+    move |filter| {
+        warp::any()
+            .map(std::time::Instant::now)
```
tgeoghegan, line 1398: Any reason not to `use std::time::Instant` at the top of the file to declutter this?

janus_server/src/metrics.rs
```diff
@@ -0,0 +1,188 @@
...
+pub struct MetricsConfiguration {
+    /// Configuration for OpenTelemetry metrics, with a choice of exporters.
+    #[serde(default)]
+    pub exporter: Option<MetricsExporterConfiguration>,
+}
+
+/// Selection of an exporter for OpenTelemetry metrics.
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+pub enum MetricsExporterConfiguration {
+    #[serde(rename = "prometheus")]
```
tgeoghegan, line 38: nit: you could put `#[serde(rename_all = "lowercase"]` on the enum to avoid the repreated `#[serde(rename)]` blocks.

janus_server/src/metrics.rs
```diff
@@ -0,0 +1,188 @@
...
+        #[cfg(feature = "prometheus")]
+        Some(MetricsExporterConfiguration::Prometheus {
+            host: config_exporter_host,
+            port: config_exporter_port,
+        }) => {
+            use http::StatusCode;
+            use hyper::Response;
+            use prometheus::{Encoder, TextEncoder};
+            use std::net::SocketAddr;
+            use warp::{Filter, Reply};
```
tgeoghegan, line 106: I think moving these `use` statements (as well as the ones in other arms of this `match`) would declutter this method and make it easier to read. If you want to make the `use` conditional you can still do

```
#[cfg(feature="something")]
use {
    http::{...},
    hyper::{...},
    etc.
};
```

janus_server/src/metrics.rs
```diff
@@ -0,0 +1,188 @@
...
+    #[serde(default)]
+    pub metadata: HashMap<String, String>,
+}
+
+/// Choice of OpenTelemetry metrics exporter implementation.
+pub enum MetricsExporterHandle {
+    #[cfg(feature = "prometheus")]
+    Prometheus(tokio::task::JoinHandle<()>),
+    #[cfg(feature = "otlp")]
+    Otlp(opentelemetry::sdk::metrics::controllers::PushController),
```
tgeoghegan, line 62: I'd prefer to see these `use` statements at the top of the file, guarded by `#[cfg(...)]` as needed, to declutter these declarations.

tgeoghegan (approved):
I just have gripes about `use` statements, which suggests the change is good!

tgeoghegan (approved):

PR #115, opened 2022-04-29
Update base64 format used in problem details

divergentdave:
Earlier this month, the interop branch was updated to require that the task ID in problem details documents be encoded with base64url, with no padding. This updates our implementation to match.


BranLwyd (approved):
optional: make this the Display (or Debug, maybe?) implementation, and use that, rather than duplicating the encoding code everywhere it's needed.

divergentdave:
Good idea, I did so and got rid of the hex-encoding in Debug while I was at it.

BranLwyd (approved):

PR #130, opened 2022-05-02
Parameterize BatchUnitAggregation by vdaf::Aggregator.

BranLwyd:
Previously, it was parameterized by various messages, which was
unnecessarily flexible & somewhat confusing. I modified some tests to
work with Poplar1, to allow testing with a non-unit aggregation
parameter.

While working on this I noticed that
get_batch_unit_aggregations_for_task_in_interval assumes a deterministic
encoding for aggregation params, since it does an index lookup based on
the encoded aggregation parameter. I think this is true for all
aggregation parameters today (prio3 uses the unit type which certainly
qualifies; poplar1 uses BTreeSet which has a deterministic iteration
order). We'll need to ensure this remains true in the future to avoid
the potential for weird bugs -- I can't think of a (performant) way to
do better.

### DIFF
--- a/janus_server/src/datastore.rs
+++ b/janus_server/src/datastore.rs
@@ -903,14 +903,14 @@ impl Transaction<'_> {
     /// Store a new `batch_unit_aggregations` row in the datastore.
     #[cfg(test)]
     #[tracing::instrument(skip(self), err)]
-    pub(crate) async fn put_batch_unit_aggregation<P, A>(
+    pub(crate) async fn put_batch_unit_aggregation<A: vdaf::Aggregator>(
         &self,
-        batch_aggregation: &BatchUnitAggregation<P, A>,
+        batch_aggregation: &BatchUnitAggregation<A>,
     ) -> Result<(), Error>
     where
-        P: Encode + std::fmt::Debug,
-        A: std::fmt::Debug,
-        for<'a> &'a A: Into<Vec<u8>>,
+        A::AggregationParam: Encode + std::fmt::Debug,
+        A::AggregateShare: std::fmt::Debug,
+        for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
     {
         let unit_interval_start = batch_aggregation.unit_interval_start.as_naive_date_time();
         let encoded_aggregation_param = batch_aggregation.aggregation_param.get_encoded();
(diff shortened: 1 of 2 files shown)


janus_server/src/datastore.rs
```diff
@@ -1530,8 +1532,11 @@ pub mod models {
     /// This is the finest-grained possible aggregate share we can emit for this task, hence "batch
     /// unit". The aggregate share constructed to service a collect or aggregate share request
     /// consists of one or more `BatchUnitAggregation`s merged together.
-    #[derive(Clone, Debug, PartialEq, Eq)]
-    pub(crate) struct BatchUnitAggregation<P, A> {
+    #[derive(Clone, Debug)]
+    pub(crate) struct BatchUnitAggregation<A: vdaf::Aggregator>
+    where
+        for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
```
divergentdave, line 1538: One observation, take it or leave it: if we dropped constraints from the struct definition, it would let us simplify the Eq and PartialEq implementations.

divergentdave (approved):

janus_server/src/datastore.rs
BranLwyd: Unfortunately, I don't think we can: referring to `vdaf::Aggregator` anywhere without this bound leads to compilation errors like the following:

```
error[E0277]: the trait bound `for<'a> Vec<u8>: From<&'a <A as prio::vdaf::Vdaf>::AggregateShare>` is not satisfied
    --> janus_server/src/datastore.rs:1536:47
     |
1536 |     pub(crate) struct BatchUnitAggregation<A: vdaf::Aggregator> {
     |                                               ^^^^^^^^^^^^^^^^ the trait `for<'a> From<&'a <A as prio::vdaf::Vdaf>::AggregateShare>` is not implemented for `Vec<u8>`
     |
     = note: required because of the requirements on the impl of `for<'a> Into<Vec<u8>>` for `&'a <A as prio::vdaf::Vdaf>::AggregateShare`
note: required by a bound in `prio::vdaf::Aggregator`
    --> /home/bran/.cargo/git/checkouts/libprio-rs-adeec693b5802b42/229cd9c/src/vdaf.rs:162:39
     |
162  |     for<'a> &'a Self::AggregateShare: Into<Vec<u8>>,
     |                                       ^^^^^^^^^^^^^ required by this bound in `prio::vdaf::Aggregator`
help: consider introducing a `where` bound, but there might be an alternative better way to express this requirement
     |
1536 |     pub(crate) struct BatchUnitAggregation<A: vdaf::Aggregator> where Vec<u8>: for<'a> From<&'a <A as prio::vdaf::Vdaf>::AggregateShare> {
     |                                                                 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
```

I _think_ this can be resolved once GATs are stabilized, see https://github.com/divviup/libprio-rs/blob/662fc24a7df7c282fc2ad41c86442c4a8e7b7a80/src/vdaf.rs#L105-L107

PR #135, opened 2022-05-03
Adjust privacy of Time, Duration, and Interval

divergentdave:
Make fields private, and expose some additional methods as public.

This is part of #18. I also fixed one rustdoc warning I noticed.

### DIFF
--- a/janus_server/src/message.rs
+++ b/janus_server/src/message.rs
@@ -163,12 +163,17 @@ fn authenticated_decode<M: Decode>(
 
 /// PPM protocol message representing a duration with a resolution of seconds.
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
-pub struct Duration(pub(crate) u64);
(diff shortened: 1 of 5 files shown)


janus_server/src/message.rs
```diff
@@ -208,19 +213,31 @@ impl Display for Duration {
...
     }
 
+    /// Get the Unix timestamp represented by this [`Time`] (in seconds).
+    pub fn as_timestamp(&self) -> u64 {
+        self.0
+    }
+
+    /// Construct a [`Time`] from a Unix timestamp (in seconds).
+    pub fn from_timestamp(timestamp: u64) -> Self {
+        Self(timestamp)
```
tgeoghegan, line 236: I think `timestamp` is too ambiguous. At the risk of being verbose, how about `fn as_seconds_since_epoch()` and `from_seconds_since_epoch()`?

tgeoghegan (changes requested):

BranLwyd (approved):

tgeoghegan (approved):

PR #142, opened 2022-05-05
Max batch lifetime enforcement in the helper

tgeoghegan:


janus_server/src/datastore.rs
```diff
@@ -1059,7 +1059,7 @@ impl Transaction<'_> {
                 WHERE
                     task_id = (SELECT id FROM tasks)
                     AND unit_interval_start >= $2
-                    AND (unit_interval_start + (SELECT min_batch_duration FROM tasks) * interval '1 second') < $3
+                    AND (unit_interval_start + (SELECT min_batch_duration FROM tasks) * interval '1 second') <= $3
```
tgeoghegan, line 1062: I want to draw attention to this change: the intent here is that a batch unit with interval `[100, 200)` should be considered included in the batch interval `[0, 200)`.

janus_server/src/task.rs
```diff
@@ -136,6 +138,7 @@ pub struct Task {
     pub(crate) collector_hpke_config: HpkeConfig,
     /// Key used to authenticate messages sent to or received from the other
     /// aggregators.
+    #[derivative(Debug = "ignore")]
     pub(crate) agg_auth_keys: Vec<AggregatorAuthKey>,
     /// HPKE configurations & private keys used by this aggregator to decrypt client reports.
     pub(crate) hpke_keys: HashMap<HpkeConfigId, (HpkeConfig, HpkePrivateKey)>,
```
tgeoghegan, line 144: We should omit the `HpkePrivateKey` values from the debug output, too, but I'm not sure how to do that without replacing this `HashMap` with some custom type.

divergentdave: This is using our `HpkePrivateKey` type, which currently uses a derived `Debug` implementation. We could rewrite its implementation to be opaque, then we would get the behavior we want here. (The same tactic would work on `AggregatorAuthKey`) If we ever wanted to dump a key in a unit test, we'd have to take another step or two to get the key's inner bytes.

divergentdave (approved):

janus_server/src/task.rs
tgeoghegan: I wrote up #144 for this.

janus_server/src/datastore.rs
```diff
@@ -1163,6 +1163,44 @@ impl Transaction<'_> {
...
+                &stmt,
+                &[
+                    /* task_id */ &task_id.as_bytes(),
+                    &unit_interval_start,
+                    &unit_interval_end,
+                ],
+            )
+            .await?;
+
+        // PostgreSQL's `COUNT` aggregate function returns rows with a bigint-valued "count" column.
```
BranLwyd, line 1199: You can also name the returned column (as in `SELECT COUNT(*) AS count FROM...`) to remove the ambiguity.

janus_server/src/aggregator.rs
```diff
@@ -1091,11 +1086,68 @@ impl VdafOps {
...
+            if consumed_batch_lifetime == task.max_batch_lifetime {
+                warn!(
+                    ?task.id, ?aggregate_share_req, ?unit_interval,
+                    "refusing aggregate share request because lifetime for batch unit has been consumed"
+                );
+                return Err(datastore::Error::User(
+                    Error::BatchLifetimeExceeded(task.id).into(),
+                ));
+            }
+            if consumed_batch_lifetime > task.max_batch_lifetime {
```
BranLwyd, line 1143: Hypothetically, this branch could be taken if we modified a Task to lower its `max_batch_lifetime`. It may still be useful to have a special branch since this is extra-unexpected, but for that reason I'm somewhat wary of panicking.

janus_server/src/aggregator.rs
```diff
@@ -1091,11 +1086,68 @@ impl VdafOps {
...
+            //
+            // A further AggregateShareReq including either or both of the units is permissible, but
+            // if we queried how many rows overlap with that interval, we would get 2 and refuse the
+            // request. We must check the unit intervals individually to notice that each has enough
+            // remaining lifetime to permit the share request.
+            let unit_interval = Interval::new(
+                batch_unit_aggregation.unit_interval_start,
+                task.min_batch_duration,
+            )?;
+            let consumed_batch_lifetime = tx
```
BranLwyd, line 1131: These checks are independent, and can be done with concurrency: we could start each of the futures and join on them using e.g. `try_join_all`, then run the checks on each result.

An alternate strategy would be to update the datastore operation to return all of the necessary counts by batch unit interval rather than being a one-off, e.g. something like `SELECT batch_unit_start, COUNT(*) AS count FROM ... GROUP BY batch_unit_start`. I'm not sure which would be easier to implement.

janus_server/src/aggregator.rs
```diff
@@ -1091,11 +1086,68 @@ impl VdafOps {
...
+        // whatever batch unit aggregations are available now.
+        //
+        // See issue #104 for more discussion.
+
         let mut total_report_count = 0;
         let mut total_checksum = [0u8; 32];
         let mut total_aggregate_share: Option<A::AggregateShare> = None;
 
         for batch_unit_aggregation in &batch_unit_aggregations {
+            // Check how many rows in aggregate_share_jobs have a batch interval that includes this
```
BranLwyd, line 1113: I don't think these checks are sufficient to protect the privacy of client reports.

Say we have `match_batch_lifetime` = 2, and three consecutive batch units `A`, `B`, `C`. Each of these batch units does not have enough reports to meet the `min_batch_size`, but any two batch units do.

If a collector requested aggregation of `[A_start, C_end)`, followed by `[B_start, C_end)`, the current checks would allow both collections to occur. But then the collector could determine the aggregation of `[A_start, A_end)` by subtracting the second collection from the first, which violates the `min_batch_size` requirements.

One solution to this would be to require the intervals of different collection requests to either match exactly, or not overlap at all. But that may be too restrictive--perhaps there is a more flexible solution?

janus_server/src/datastore.rs
```diff
@@ -1163,6 +1163,44 @@ impl Transaction<'_> {
         }))
     }
 
+    /// Returns the number of rows in `aggregate_share_jobs` whose batch interval contains
+    /// `batch_unit_interval`.
+    #[tracing::instrument(skip(self), err)]
+    pub(crate) async fn get_aggregate_share_job_count_by_batch_unit(
```
BranLwyd, line 1169: nit: this interval isn't necessarily a batch unit, so the method is somewhat confusingly named. Maybe `..._by_interval` instead of `..._by_batch_unit`?

BranLwyd (changes requested):

janus_server/src/aggregator.rs
tgeoghegan: I'm going to try to write this as a single SQL query using [`WHERE ANY`](https://www.postgresql.org/docs/current/functions-comparisons.html#id-1.5.8.30.16).

janus_server/src/aggregator.rs
BranLwyd: I think that some task parameters are effectively immutable (e.g. anything distributed to clients), but other parameters might be mutable -- but I definitely agree that's something to be considered for DAP.

With that in mind, I filed https://github.com/ietf-wg-ppm/ppm-specification/issues/237. Moving forward with immutable tasks SGTM for now.

janus_server/src/aggregator.rs
BranLwyd: Fair enough; but from a practical standpoint, IMO this is an unresolved privacy vulnerability, so I think we need to resolve it _somehow_ before we can run "real" data through our system (though we might elect for something that we know we are going to throw away later, a la the recent aggregator<->aggregator auth changes).

That said, there's no reason to hold up this PR for it. Would you mind leaving a TODO or filing an issue for now?

tgeoghegan: Wrote this up in #149

BranLwyd (approved):

PR #143, opened 2022-05-05
Provide API base URL in aggregator config

tgeoghegan:
When servicing a collect request, the leader must generate a collect job
URI relative to the public base URL from which the API is served and
then stick that in a `Location` header. We now provide that base URL in
the aggregator's config file.

Part of #105


BranLwyd (approved):
We might want to eventually make this a per-task parameter, but since this is internal (non-standardized) config only, that migration would be easy enough.

tgeoghegan:
> We might want to eventually make this a per-task parameter, but since this is internal (non-standardized) config only, that migration would be easy enough.

The URL does go into the task anyway, because it has the [`aggregator_endpoints` field](https://github.com/divviup/janus/blob/684e15f536823a5f33dcf17d1d1a93f3cfb03a94/janus_server/src/task.rs#L120). Technically the URL in the task parameters doesn't have to match the one we use to construct collect job URIs (we are allowed to redirect to anything we want).

PR #145, opened 2022-05-05
max batch lifetime enforcement in the leader

tgeoghegan:
Refactors some existing code that supports the helper's
`max_batch_lifetime` enforcement so that it can be re-used in the
leader's `collect` endpoint. All the logic for that endpoint now moves
into methods on `VdafOps`.

part of #105

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -3,8 +3,8 @@ use crate::{
     datastore::{
         self,
         models::{
-            AggregateShareJob, AggregationJob, AggregationJobState, ReportAggregation,
-            ReportAggregationState,
+            AggregateShareJob, AggregationJob, AggregationJobState, BatchUnitAggregation,
+            ReportAggregation, ReportAggregationState,
         },
         Datastore, Transaction,
     },
@@ -53,6 +53,7 @@ use std::{
 use tokio::sync::Mutex;
 use tracing::{debug, error, warn};
 use url::Url;
+use uuid::Uuid;
 use warp::{
     filters::BoxedFilter,
     reply::{self, Response},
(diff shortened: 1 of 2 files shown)


janus_server/src/datastore.rs
```diff
@@ -1187,22 +1194,22 @@ impl Transaction<'_> {
...
             .tx
             .prepare_cached(
-                "WITH ranges AS (
+                &format!("WITH ranges AS (
                     SELECT tsrange(x.range_start, x.range_end) as interval
                     FROM unnest($1::TIMESTAMP[], $2::TIMESTAMP[]) AS x(range_start, range_end)
                 )
                 SELECT
-                    COUNT(aggregate_share_jobs.batch_interval_start) as overlap_count,
+                    COUNT({table}.batch_interval_start) as overlap_count,
```
tgeoghegan, line 1202: Is interpolating the table name into the query too clever? I think the Rigorous, Sensible thing to do might be to normalize the schema: replace `collect_jobs` and `aggregate_share_jobs` with a single table that has the common columns, and then introduce a new table that would have the `helper_aggregate_share`, `leader_aggregate_share`, `report_count` and `checksum` columns from `collect_jobs`, plus an FK to `collect_jobs`. The advantage is that the new table's columns could all be `NOT NULL`. But I'm not sure if this kind of normalization is a good idea yet.

BranLwyd: I like the idea of a single, shared table between leader & helper (as long as this doesn't introduce a bunch of complexity/confusion, but from your description it sounds like it would be fairly straightforward); we get away with something similar with the `aggregation_jobs` & `report_aggregations` tables.

That said, I'm OK with interpolating the table name if you want to keep it this way and/or address in a later PR, but doing so is definitely a sign that our schema could be improved IMO.

(if you do decide to modify the schema, please let me take another look for review)

janus_server/src/aggregator.rs
```diff
@@ -1022,6 +1001,108 @@ impl VdafOps {
...
 
+    async fn handle_collect(
+        &self,
+        datastore: &Datastore,
+        task: &Task,
+        collect_req: &CollectReq,
+    ) -> Result<Uuid, Error> {
+        match self {
+            VdafOps::Prio3Aes128Count(_, _) => {
+                Self::handle_collect_generic::<Prio3Aes128Count, FieldError>(
```
BranLwyd, line 1012: Q: is Rust smart enough to infer the second generic parameter if we specify the first? (e.g. does compilation succeed if we specify this as `<Prio3Aes128Count, _>`).

I'd like it even better if this took a single generic parameter that was the VDAF type since it's not immediately clear what the error type's relation/purpose is, but I think that might make the trait bounds on `handle_collect_generic` pretty confusing.

BranLwyd (approved):

janus_server/src/aggregator.rs
tgeoghegan: You were right, Rust can figure it out if we don't specify E! It'd be nice if there was a clippy warning or something for "unnecessary type parameter".

janus_server/src/datastore.rs
tgeoghegan: Looking at it again, we'd also have to do something about the `collect_job_id` column for the leader. I think I'm going to consider making that change in the future, because right now, I want to keep making progress on a collect implementation we can do interop testing with. #156

PR #147, opened 2022-05-06
Add a Dockerfile for aggregation_job_creator

divergentdave:
This adds a Dockerfile that builds the aggregation_job_creator binary. Thanks to caching, our two Dockerfiles can reuse the lion's share of the work they do.


BranLwyd (approved):

.github/workflows/ci-build.yml
```diff
@@ -43,3 +43,6 @@ jobs:
     - env:
         DOCKER_BUILDKIT: 1
       run: docker build .
```
tgeoghegan, line 45: We should rename the existing Dockerfile to `Dockerfile.aggregator` and use `-f` here, for consistency and to disambiguate.

Dockerfile.aggregation_job_creator
```diff
@@ -0,0 +1,14 @@
+FROM rust:1.60.0-alpine as builder
+RUN apk add libc-dev
+
+WORKDIR /src
+COPY Cargo.toml /src/Cargo.toml
+COPY Cargo.lock /src/Cargo.lock
+COPY janus_server /src/janus_server
+COPY monolithic_integration_test /src/monolithic_integration_test
+COPY test_util /src/test_util
+RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/src/target cargo build --release --bin aggregation_job_creator && cp /src/target/release/aggregation_job_creator /aggregation_job_creator
```
tgeoghegan, line 10: Is the only difference between the two dockerfiles `--bin aggregator` vs. `--bin aggregation_job_creator` and thenwhich path gets copied in from the builder image? If so, maybe we can use a single Dockerfile that takes a parameter for image name.

Dockerfile.aggregation_job_creator
```diff
@@ -0,0 +1,14 @@
...
+
+FROM alpine:3.15.4
```
tgeoghegan, line 12: The `aggregator` Dockerfile copes in `/src/db/schema.sql`. Do we not need that here?

tgeoghegan (changes requested):

Dockerfile.aggregation_job_creator
tgeoghegan: Obviously this suggestion contradicts my other comment on `.github/workflows/ci-build.yml`. I'm comfortable with either a single or dual Dockerfile solution, so please move ahead with whichever one you prefer.

Dockerfile.aggregation_job_creator
divergentdave: It doesn't get used in the container at runtime, as we only copy it into the container for use by our Terraform external data provider. This is now moot as I'm going with a single Dockerfile.

Dockerfile.aggregation_job_creator
divergentdave: I went with one parameterized file, nice idea! Build arguments can't be directly used in the ENTRYPOINT, unfortunately, so I decided to copy the argument into an environment variable (which becomes part of the image) and use that in a shell command. The alternate solution would be to rename the binary to something constant when copying it, and then put that common name in the ENTRYPOINT. I decided against that, as I think one PID plus a few megabytes of memory is worth the clarity of having argv[0] be the "real" binary name.

tgeoghegan (approved):

PR #148, opened 2022-05-06
Add public constructor for `MockClock`

tgeoghegan:
Also updates the `current_time` field to be a private `message::Time`.


divergentdave (approved):
LGTM. As we get further into integration, I figure we'll want to extend this with interior mutability.

BranLwyd:
> LGTM. As we get further into integration, I figure we'll want to extend this with interior mutability.

Indeed we will, this is coming with my next PR :)

PR #159, opened 2022-05-12
client: remove type param, add concrete upload fns

### LINKED ISSUE #152: Consider constructing client from `janus_server::task::Vdaf`
Otherwise this leads to a bit of awkwardness when having to import `prio::vdaf::Vdaf as LibPrioVdaf` and `janus::task::Vdaf as JanusVdaf`.

divergentdave:
This removes the type parameter on `Client` that specified a `vdaf::Client`, instead using our `crate::task::Vdaf` enum to specify a VDAF instantiation. Separate `upload_boolean()` and `upload_integer_u128()` methods are added, which take a measurement of a concrete type as an input, and perform dispatching based on the enum to the correct implementation. (Adding an `upload_byte_string()` method for Poplar1 is left as future work, for once the Poplar1 VDAF is better specified and implemented)

This fixes #152.

### DIFF
--- a/janus_server/src/client.rs
+++ b/janus_server/src/client.rs
@@ -1,6 +1,9 @@
 //! PPM protocol client
 
-use crate::hpke::{self, HpkeApplicationInfo, Label};
+use crate::{
+    hpke::{self, HpkeApplicationInfo, Label},
+    task::Vdaf,
+};
 use http::StatusCode;
 use janus::{
     hpke::associated_data_for_report_share,
@@ -32,6 +35,11 @@ pub enum Error {
     Hpke(#[from] crate::hpke::Error),
     #[error("invalid task parameters: {0}")]
     TaskParameters(#[from] crate::task::Error),
+    #[error("Wrong measurement type, expected {expected} and got {actual}")]
+    WrongMeasurementType {
+        expected: &'static str,
+        actual: &'static str,
+    },
 }
 
 static CLIENT_USER_AGENT: &str = concat!(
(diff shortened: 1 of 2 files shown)


BranLwyd (commented):
I think I disagree with this change; I think a Client really should be generic on the `vdaf::Client`. Doing a type-switch dispatch has the following downsides:

* The user of the `Client` has to Just Know which `upload_foo` method to call based on the VDAF type. There is no way to programmatically determine the correct type, and if the user gets it wrong they get a runtime error.
* We are now instantiating a `Vdaf` (cheap for existing VDAFs, but not zero-cost) & decoding a public parameter (which is technically fallible) with every measurement uploaded. [I think this could be technically worked around by having `Client` hold a VDAF-parameterized private inner type as a dynamic pointer, but of course that is a tradeoff since we'd then be doing dynamic runtime dispatch.]

Looking to #152 which led to this change: if we do want to do something, maybe we can rename our `task::Vdaf` to e.g. `VdafDefinition` or `VdafParameters` or the like? I like `VdafDefinition` since this is no longer just indicating a particular VDAF, but also the parameters that define how it operates.

tgeoghegan:
I'm not sure about this. I think we can provide a better client API that's less prone to failure because you called the wrong method. Compare what we have here to the API you usually get from Prometheus client libraries, like the [`prometheus` crate](https://docs.rs/prometheus/latest/prometheus/). There, you don't instantiate a `PrometheusClient`, you get distinct `Counter`, `Gauge` and `Histogram` types, each of which exposes a different set of methods specific to what kind of metric is being gathered, so that rules out the possibility of an error because you tried to register a gauge value into a histogram.

I think we should aim to expose something similar. Maybe we would have some kind of client factory that takes a `Task` and exposes methods to instantiate the different kinds of metrics registries. Then you have to deal with task misconfigurations only when setting up metrics clients, and they're pretty safe to use once you have them.

divergentdave:
Sounds good. I was considering `VdafInstance` as a new name for `task::Vdaf`, I'll open a new PR with that.

PR #160, opened 2022-05-12
Leader collect job URI handler

tgeoghegan:
Adds a warp filter for path `/collect_jobs/{collect_job_id}` to the
leader. Adds support for querying collect jobs from the datastore as
well as updating them with helper and leader aggregate shares. The
latter is currently only needed for tests, but will soon be used when
running collect jobs.

Part of #105

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -13,7 +13,8 @@ use crate::{
         AggregateReqBody::{AggregateContinueReq, AggregateInitReq},
         AggregateResp, AggregateShareReq, AggregateShareResp, AggregationJobId,
         AuthenticatedDecodeError, AuthenticatedEncoder, AuthenticatedRequestDecoder, CollectReq,
-        Interval, ReportShare, Transition, TransitionError, TransitionTypeSpecificData,
+        CollectResp, Interval, ReportShare, Transition, TransitionError,
+        TransitionTypeSpecificData,
     },
     task::{AggregatorAuthKey, Task, VdafInstance},
 };
@@ -89,6 +90,9 @@ pub enum Error {
     /// An attempt was made to act on an unknown aggregation job.
     #[error("unrecognized aggregation job: {0:?}")]
     UnrecognizedAggregationJob(AggregationJobId, TaskId),
+    /// An attempt was made to act on an unknown collect job.
+    #[error("unrecognized collect job: {0}")]
+    UnrecognizedCollectJob(Uuid),
     /// Corresponds to `outdatedHpkeConfig`, §3.1
     #[error("outdated HPKE config: {0} {1:?}")]
     OutdatedHpkeConfig(HpkeConfigId, TaskId),
@@ -233,6 +237,9 @@ impl<C: Clock> Aggregator<C> {
         Err(Error::InvalidHmac(task_id))
     }
 
+    /// Handle a collect request. Only supported by the leader. `req_bytes` is an encoded
+    /// [`CollectReq`]. Returns the URL at which a collector may poll for status of the collect job
+    /// corresponding to the `CollectReq`.
     async fn handle_collect(&self, req_bytes: &[u8]) -> Result<Url, Error> {
         let collect_req = CollectReq::get_decoded(req_bytes)?;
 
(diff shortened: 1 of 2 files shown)


janus_server/src/aggregator.rs
```diff
@@ -3915,10 +4084,115 @@ mod tests {
...
+                    let encrypted_helper_aggregate_share = hpke::seal(
+                        &collector_hpke_config,
+                        &HpkeApplicationInfo::new(
+                            task.id,
+                            Label::AggregateShare,
+                            Role::Helper,
+                            Role::Collector,
+                        ),
+                        &[],
+                        //&[0, 1, 2, 3, 4, 5],
```
BranLwyd, line 4112: nit: rm commented code

janus_server/src/aggregator.rs
```diff
@@ -3915,10 +4084,115 @@ mod tests {
...
+                task_id,
+                Label::AggregateShare,
+                Role::Leader,
+                Role::Collector,
+            ),
+            &collect_resp.encrypted_agg_shares[0],
+            &batch_interval.get_encoded(),
+        )
+        .unwrap();
+        fake::AggregateShare::try_from(leader_aggregate_share.as_ref()).unwrap();
```
BranLwyd, line 4161: Note that this isn't testing much right now, since `fake::AggregateShare` is just a newtype of the unit type.

We could use/adapt `run_vdaf` to produce values that would let us do an aggregation on a real VDAF, and check that the aggregation result is as expected. Up to you, I won't hold the PR up over this if you think it's not worthwhile.

janus_server/src/datastore.rs
```diff
@@ -1848,6 +2030,63 @@ pub mod models {
...
+    /// running collect jobs and store the results of completed ones.
+    #[derive(Clone, Derivative)]
+    #[derivative(Debug)]
+    pub(crate) struct CollectJob<A: vdaf::Aggregator>
+    where
+        for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
+    {
+        /// The unique identifier for the collect job.
+        pub(crate) collect_job_id: Uuid,
+        /// The task ID for this aggregate share .
```
BranLwyd, line 2043:
```suggestion
        /// The task ID for this aggregate share.
```

janus_server/src/datastore.rs
```diff
@@ -909,11 +909,104 @@ impl Transaction<'_> {
...
+        let collect_job_row = match self.tx.query_opt(&stmt, &[&collect_job_id]).await? {
+            Some(row) => row,
+            None => return Ok(None),
+        };
+        let task_id: i64 = collect_job_row.get("task_id");
+
+        let stmt = self
+            .tx
+            .prepare_cached("SELECT task_id FROM tasks WHERE id = $1")
+            .await?;
```
BranLwyd, line 964: To avoid the need for a second network round-trip, retrieve tasks.task_id with the original query via a join or subquery. Check out several of the other `get_foo` for example implementations.

janus_server/src/datastore.rs
```diff
@@ -2856,6 +3096,181 @@ mod tests {
         assert!(rows.len() == 2);
     }
 
+    #[tokio::test]
+    async fn lookup_collect_job_task_id() {
```
BranLwyd, line 3100:
```suggestion
    async fn get_collect_job_task_id() {
```

BranLwyd (approved):
Approved with a few nits.

janus_server/src/datastore.rs
```diff
@@ -1848,6 +2030,63 @@ pub mod models {
...
+                self.leader_aggregate_share.as_ref(),
+                self.report_count,
+                self.checksum,
+            ) {
+                (Some(_), Some(_), Some(_), Some(_)) => Ok(true),
+                (None, None, None, None) => Ok(false),
+                _ => Err(Error::DbState(format!(
+                    "collect job {} in inconsistent state",
+                    self.collect_job_id
+                ))),
```
divergentdave, line 2085: Depending on how we implement the rest of the leader worker processes, we may store the leader aggregate share, report count, and checksum immediately when they are computed, and then store the helper aggregate share once it has been requested and received. If so, we'd want to return `Ok(false)` for that combination. In the alternative, the leader worker would compute its fields, send a request, and only write to the database once upon receiving a successful response from the helper.

tgeoghegan: You're right: the leader must compute its aggregate share, report count and the checksum before it sends `AggregateShareReq` to helper, and there's no reason not to record all those values in the datastore while leader waits for helper's aggregate share. Neither would we want to hold a DB txn open while the leader awaits a helper response, so we should handle the case where we have leader share, report count and checksum.

janus_server/src/aggregator.rs
tgeoghegan: It was easy enough to adapt this test to use `Prio3Aes128Count` and then synthesize fake `AggregateShares` to verify that they get encoded and encrypted appropriately.

PR #161, opened 2022-05-12
improve generation of test HPKE configs

### LINKED ISSUE #158: New features for `hpke::test_util::generate_hpke_config_and_private_key()`
I'd like to request a couple features for this.
1. Currently this always choose config_id == 0, but consumers should tolerate any valid config id. It seems sensible to choose a random config ID so as to mitigate ossification.
2. The ciphersuite (X25519, HKDF-SHA512, ChaChaPoly) seems a bit unusual. It would be useful to either use somethhing more commen (see https://github.com/ietf-wg-ppm/ppm-specification/pull/247) or allow the caller to specify this.

tgeoghegan:
- randomize the generated config ID
  - use a less astonishing set of HPKE algos

This is stacked on #157 since that moved HPKE code around and I wanted to avoid merge conflicts.

Closes #158

### DIFF
--- a/janus/src/hpke.rs
+++ b/janus/src/hpke.rs
@@ -169,7 +169,7 @@ pub mod test_util {
         message::{HpkeAeadId, HpkeConfig, HpkeConfigId, HpkeKdfId, HpkeKemId, HpkePublicKey},
     };
     use hpke::{kem::X25519HkdfSha256, Kem, Serializable};
-    use rand::thread_rng;
+    use rand::{thread_rng, Rng};
 
     /// Generate a new HPKE keypair and return it as an HpkeConfig (public portion) and
     /// HpkePrivateKey (private portion).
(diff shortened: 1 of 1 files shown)


janus/src/hpke.rs
```diff
@@ -169,18 +169,18 @@ pub mod test_util {
...
         let (private_key, public_key) = X25519HkdfSha256::gen_keypair(&mut thread_rng());
         (
             HpkeConfig::new(
-                HpkeConfigId::from(0),
+                HpkeConfigId::from(thread_rng().gen::<u8>()),
                 HpkeKemId::X25519HkdfSha256,
-                HpkeKdfId::HkdfSha512,
-                HpkeAeadId::ChaCha20Poly1305,
+                HpkeKdfId::HkdfSha256,
+                HpkeAeadId::Aes128Gcm,
```
cjpatton, line 183: This solves my problem, but you may find it useful to support multiple ciphersuites for your own internal testing purposes.

divergentdave (approved):

janus/src/hpke.rs
tgeoghegan: True, but this change doesn't make us any worse off in terms of coverage across HPKE configs, so I would punt on testing the full space of HPKE configs to a future change.

PR #162, opened 2022-05-12
Move verify parameters to a separate table

divergentdave:
This creates a new `task_vdaf_verify_params` table, to allow storing more than one verification parameter per task.

(Note that rotating these parameters is not yet feasible in practice, as we need some mechanism at the specification level for aggregators to signal and synchronize which parameter they will use on any given report aggregation)

This implements the last part of #48.


tgeoghegan (approved):

db/schema.sql
```diff
@@ -28,7 +25,7 @@ CREATE TABLE task_aggregator_auth_keys(
     ord BIGINT NOT NULL,      -- a value used to specify the ordering of the authentication keys
     key BYTEA NOT NULL,       -- HMAC key used to authenticate messages to/from the other aggregator (encrypted)
 
-    CONSTRAINT unique_task_id_and_ord UNIQUE(task_id, ord),
+    CONSTRAINT auth_key_unique_task_id_and_ord UNIQUE(task_id, ord),
```
BranLwyd, line 28: Q: do these table-scoped constraints need to be uniquely named across the entire DB?

db/schema.sql
```diff
@@ -44,6 +41,17 @@ CREATE TABLE task_hpke_keys(
     CONSTRAINT fk_task_id FOREIGN KEY(task_id) REFERENCES tasks(id)
 );
 
+-- The VDAF verification parameters used by a given task.
+CREATE TABLE task_vdaf_verify_params(
+    id BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,  -- artificial ID, internal-only
+    task_id BIGINT NOT NULL,           -- task ID the verification parameter is associated with
+    ord BIGINT NOT NULL,               -- a value used to specify the ordering of verification parameters
```
BranLwyd, line 48: I'm a little nervous about implementing that the VDAF parameters are ordered before we specify/know how multiple verify parameters are used. I think we need to do _something_ specification-wise, and I strongly suspect whatever we do will end up communicating a "verify_parameter ID" of some kind (since the helper would otherwise have to guess, or try them all which is expensive), which may well not match up to the `ord` parameter we implement here.

WDYT about dropping the `ord` parameter, and making the `UNIQUE` constraint cover only `task_id`? This would enforce one-verify-param-per-task at a DB level, which I think is a good idea since that's what we want for now.

BranLwyd: Hmm. This does have the downside that we could no longer include `ord` in the crypter "row" parameter -- we'd need to do a data migration if/when we did introduce whatever verify parameter ID ends up being specified.

Still--if we wanted to switch from `ord` to a specified ID, we might need a data migration anyway, so maybe this point is moot?

BranLwyd (changes requested):

db/schema.sql
divergentdave: Yes, I got an error about a relation already existing before I made this change. The unique constraint creates an index with the given name, and both tables and indexes share the same namespace.

db/schema.sql
divergentdave: Guess-and-check on verification parameters wouldn't work, because we could only learn if we chose the "right" one when the leader combines its prepare message with the helper's prepare message, and applies the "decide" function to the result. The aggregators will either need to communicate some identifier with each aggregate request or select parameters a priori, for example by assigning a range of timestamps to each verification parameter.

Ideally, if there is any verification parameter identifier exposed in the protocol, we should use that in the AAD when encrypting the verification parameter. In that case, there's no way around a migration. I'll try out making the `UNIQUE` change.

Issue #164, opened 2022-05-12
Helper should write/update `batch_unit_aggregation` rows when it obtains an output share


tgeoghegan:
Fix [this TODO](https://github.com/divviup/janus/blob/5895b501f173514b448c33b74ba2f89e3ae9abca/janus_server/src/aggregator.rs#L952) in the helper's aggregate continue handler. Right now helper can successfully prepare input shares into output shares, but it doesn't write them anywhere, so the `aggregate_share` handler can't service requests.

Upon obtaining an output share, helper needs to check if a suitable `batch_unit_aggregation` row exists. If it does, accumulate the output share into the row. If not, make a new row an initialize it with the output share.

PR #165, opened 2022-05-13
Introduce `NonceChecksum`

tgeoghegan:
Adds a newtype around `[u8; 32]` to represent the checksum over report
nonces.


tgeoghegan:
I did this while solving #164 and broke it out into its own PR to keep the diff size under control.

divergentdave (approved):

BranLwyd (approved):

PR #166, opened 2022-05-13
Accumulate prepared output shares in helper

### LINKED ISSUE #164: Helper should write/update `batch_unit_aggregation` rows when it obtains an output share

tgeoghegan:
As the helper prepares inputs and obtains output shares, accumulate the
values into rows in `batch_unit_aggregations` (also updating checksum
and report count). These rows will later be used by the
`/aggregate_share` handler to service `AggregateShareReq` messages.

Resolves #164

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -1,5 +1,9 @@
 //! Common functionality for PPM aggregators
+
+mod accumulator;
+
 use crate::{
+    aggregator::accumulator::Accumulator,
     datastore::{
         self,
         models::{
@@ -562,7 +566,7 @@ impl VdafOps {
         }
     }
 
-    async fn handle_aggregate_generic<A: vdaf::Aggregator>(
+    async fn handle_aggregate_generic<A: vdaf::Aggregator, E>(
         datastore: &Datastore,
         vdaf: &A,
         task: &Task,
@@ -572,12 +576,14 @@ impl VdafOps {
     where
         A: 'static + Send + Sync,
         A::AggregationParam: Send + Sync,
+        A::AggregateShare: Send + Sync + for<'a> TryFrom<&'a [u8], Error = E>,
         for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
         A::PrepareStep: Send + Sync + Encode + ParameterizedDecode<A::VerifyParam>,
         A::PrepareMessage: Send + Sync,
         A::OutputShare: Send + Sync + for<'a> TryFrom<&'a [u8]>,
         for<'a> &'a A::OutputShare: Into<Vec<u8>>,
         A::VerifyParam: Send + Sync,
+        E: std::fmt::Display,
     {
         match req.body {
             AggregateInitReq { agg_param, seq } => {
@@ -608,7 +614,7 @@ impl VdafOps {
 
     /// Implements the aggregate initialization request portion of the `/aggregate` endpoint for the
     /// helper, described in §4.4.4.1 of draft-gpew-priv-ppm.
-    async fn handle_aggregate_init_generic<A: vdaf::Aggregator>(
+    async fn handle_aggregate_init_generic<A: vdaf::Aggregator, E>(
         datastore: &Datastore,
         vdaf: &A,
         task: &Task,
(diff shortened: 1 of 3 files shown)


janus_server/src/aggregator.rs
```diff
@@ -799,6 +804,20 @@ impl VdafOps {
...
+                        {
+                            Self::put_or_update_batch_unit_aggregation::<A, _>(
+                                tx,
+                                task_id,
+                                min_batch_duration,
+                                share_data.report_share.nonce,
+                                &aggregation_job.as_ref().aggregation_param,
+                                output_share,
+                            )
+                            .await?;
```
tgeoghegan, line 819: This case is a 0-round VDAF (or a DAF, I suppose), where an output share can be recovered without any peer prepare message. We don't currently exercise this because we don't have any 0-round VDAFs. I think we should consider adding a variant `task::Vdaf::FakeZeroRound` and using `aggregator::fake::Vdaf` to implement it, but I opted not to do that in this PR which is already a bit large, especially since the lack of test coverage for 0-round VDAFs was a problem before this change.

divergentdave (approved):
LGTM. I noticed that `put_or_update_batch_unit_aggregation()` is inefficient in that it requires performing one UPDATE per report, whereas we could accumulate changes for an entire aggregation job and only do one UPDATE per batch unit touched. I'd be okay with landing the logic as-is and filing an issue to batch the updates later.

janus_server/src/aggregator.rs
```diff
@@ -572,12 +572,14 @@ impl VdafOps {
...
         A: 'static + Send + Sync,
         A::AggregationParam: Send + Sync,
+        A::AggregateShare: Send + Sync + for<'a> TryFrom<&'a [u8], Error = E>,
         for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
         A::PrepareStep: Send + Sync + Encode + ParameterizedDecode<A::VerifyParam>,
         A::PrepareMessage: Send + Sync,
         A::OutputShare: Send + Sync + for<'a> TryFrom<&'a [u8]>,
         for<'a> &'a A::OutputShare: Into<Vec<u8>>,
         A::VerifyParam: Send + Sync,
+        E: std::fmt::Display,
```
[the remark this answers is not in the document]
tgeoghegan: This worked, but I had to update so many trait bounds on so many methods that it yields a pretty big diff. I put the needed changes on [this branch](https://github.com/divviup/janus/tree/timg/cleanup-lifetimes) and will make a PR with them in the future.

tgeoghegan:
> I noticed that `put_or_update_batch_unit_aggregation()` is inefficient in that it requires performing one UPDATE per report, whereas we could accumulate changes for an entire aggregation job and only do one UPDATE per batch unit touched.

This is a good idea. It shouldn't be too hard to maintain a `HashMap<(TaskId, Time), BatchUnitAggregation>` so we can accumulate output shares, checksum and report count in memory, then blast them all into the database in one go after all the transitions are handled. I'll do that first thing on Monday.

janus_server/src/aggregator/accumulator.rs
```diff
@@ -0,0 +1,165 @@
...
+            if let Some(batch_unit_aggregation) = batch_unit_aggregations.first_mut() {
+                debug!(
+                    unit_interval_start = ?unit_interval.start(),
+                    "accumulating into existing batch_unit_aggregation_row",
+                );
+                batch_unit_aggregation
+                    .aggregate_share
+                    .merge(&accumulate.aggregate_share)
+                    .map_err(|e| datastore::Error::User(e.into()))?;
+                batch_unit_aggregation.report_count += 1;
```
divergentdave, line 141: This update is incorrect
```suggestion
                batch_unit_aggregation.report_count += accumulate.report_count;
```

divergentdave (approved):
Looks good save for one straightforward error, thanks!

janus_server/src/aggregator/accumulator.rs
```diff
@@ -0,0 +1,165 @@
...
+
+            let mut batch_unit_aggregations = tx
+                .get_batch_unit_aggregations_for_task_in_interval::<A, _>(
+                    self.task_id,
+                    unit_interval,
+                    &self.aggregation_param,
+                )
+                .await?;
+
+            if batch_unit_aggregations.len() > 1 {
```
BranLwyd, line 123: nit, optional: we could make a separate tx getter that, instead of finding all batch unit aggregations over a given interval, instead takes a point in time and returns the batch unit interval (if any) including that point in time].

doing so would remove this error case, and result in slightly simpler queries.

BranLwyd (approved):

janus_server/src/aggregator/accumulator.rs
tgeoghegan: I see your point, but I prefer adding a ~6 line `if` statement here over adding a large method to `Transaction`.

PR #170, opened 2022-05-16
Accumulate prepared output shares in helper

### LINKED ISSUE #164: Helper should write/update `batch_unit_aggregation` rows when it obtains an output share

tgeoghegan:
As the helper prepares inputs and obtains output shares, accumulate the
values into rows in `batch_unit_aggregations` (also updating checksum
and report count). These rows will later be used by the
`/aggregate_share` handler to service `AggregateShareReq` messages.

Resolves #164

This is the same changes as #166, but I merged that PR into the wrong branch.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -1,5 +1,9 @@
 //! Common functionality for PPM aggregators
+
+mod accumulator;
+
 use crate::{
+    aggregator::accumulator::Accumulator,
     datastore::{
         self,
         models::{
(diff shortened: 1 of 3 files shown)


divergentdave (approved):
Carrying over R+ from the last one

PR #171, opened 2022-05-16
Cleanup: Error trait bounds and NonceChecksum

tgeoghegan:
A couple of follow up items from #166 / #170.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -39,7 +39,6 @@ use opentelemetry::{
 };
 use prio::{
     codec::{Decode, Encode, ParameterizedDecode},
-    field::FieldError,
     vdaf::{
         self,
         prio3::{Prio3Aes128Count, Prio3Aes128Histogram, Prio3Aes128Sum},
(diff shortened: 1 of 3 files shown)


janus_server/src/aggregator.rs
```diff
@@ -591,23 +590,44 @@ impl VdafOps {
     ) -> Result<AggregateResp, Error> {
         match self {
             VdafOps::Prio3Aes128Count(vdaf, verify_param) => {
-                Self::handle_aggregate_generic(datastore, vdaf, task, verify_param, req).await
+                Self::handle_aggregate_generic::<Prio3Aes128Count>(
```
BranLwyd, line 593: Q (genuine question): why do we have to specify the VDAF type, e.g. `Prio3Aes128Count`, now? I would've thought it would be inferred from the `vdaf` parameter, like before.

BranLwyd (approved):
Thanks for the generics change! I know it's pedantic but it's a nice cleanup IMO.

janus_server/src/aggregator.rs
tgeoghegan: If we don't specify the trait bound here, we get an error:

```
  593:17  error   `<_ as TryFrom<&'a [u8]>>::Error` doesn't implement `std::fmt::Display` ​rustc:
                  the trait `for<'a> std::fmt::Display` is not implemented for `<_ as TryFrom<&'a [u8]>>::Error`
                  in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
```

I don't really understand why that is. Clearly the compiler can figure out that `vdaf` is of type `Prio3Aes128Count`, and so it should be able to figure out that `Prio3Aes128Count::AggregateShare` satisfies the `TryFrom` bound we asked for, but the fact that the compiler says `<_ as TryFrom<&'a [u8]>>::Error` instead of `<AggregateShare as TryFrom<&'a [u8]>>::Error` suggests that maybe there's some limitation in how trait bounds are parsed?

PR #176, opened 2022-05-18
Introduce `janus_main`

tgeoghegan:
Factors out common binary startup work like parsing options and config,
setting up metrics and logging and setting up a datastore connection
into a function `binary_utils::janus_main`.

### DIFF
--- a/janus_server/src/bin/aggregator.rs
+++ b/janus_server/src/bin/aggregator.rs
@@ -1,25 +1,16 @@
-use anyhow::{anyhow, Context, Result};
+use anyhow::{Context, Result};
 use futures::StreamExt;
 use janus::time::RealClock;
 use janus_server::{
     aggregator::aggregator_server,
-    binary_utils::datastore,
+    binary_utils::{janus_main, BinaryOptions, CommonBinaryOptions},
     config::AggregatorConfig,
-    metrics::{install_metrics_exporter, MetricsExporterConfiguration},
-    trace::{cleanup_trace_subscriber, install_trace_subscriber, OpenTelemetryTraceConfiguration},
 };
-use std::{
-    fmt::{self, Debug, Formatter},
-    fs::File,
-    future::Future,
-    iter::Iterator,
-    path::PathBuf,
-    sync::Arc,
-};
+use std::{future::Future, iter::Iterator, sync::Arc};
 use structopt::StructOpt;
 use tracing::info;
 
-#[derive(StructOpt)]
+#[derive(Debug, StructOpt)]
 #[structopt(
     name = "janus-aggregator",
     about = "PPM aggregator server",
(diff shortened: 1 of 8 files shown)


tgeoghegan:
For #105, I am going to need to add `collect_job_creator` and `collect_job_driver` targets, so this will be useful for that. Next up, I'm going to factor out the task enumeration logic in `aggregation_job_creator` into something that can be re-used in `collect_job_creator`.

janus_server/src/bin/aggregator.rs
```diff
@@ -1,101 +1,30 @@
...
 
-#[derive(StructOpt)]
+#[derive(Debug, StructOpt)]
 #[structopt(
     name = "janus-aggregator",
     about = "PPM aggregator server",
     rename_all = "kebab-case",
     version = env!("CARGO_PKG_VERSION"),
 )]
 struct Options {
```
tgeoghegan, line 20: This struct (and the equivalent ones in `aggregation_job_driver.rs` and `aggregation_job_creator.rs` exist solely so we can provide a `#[structopt(...)]` block with the `name` and `about` values.

janus_server/src/binary_utils.rs
```diff
@@ -60,3 +70,149 @@ pub fn datastore<C: Clock>(
...
     Ok(Datastore::new(pool, Crypter::new(ds_keys), clock))
 }
+
+/// Options for Janus binaries.
+pub trait BinaryOptions: StructOpt + Debug {
+    /// Returns the common options.
+    fn common_options(&self) -> &CommonBinaryOptions;
+}
+
+// Common options that are used by all Janus binaries.
```
tgeoghegan, line 80: This is deliberately not a doccomment to work around [a known issue in structopt](https://github.com/TeXitoi/structopt/issues/333). It's not clear if/when that will be fixed, but in any case, I learned today that [structopt has been deprecated in favor of functionality built into `clap`](https://github.com/TeXitoi/structopt/issues/525).

janus_server/src/binary_utils.rs
```diff
@@ -60,3 +70,149 @@ pub fn datastore<C: Clock>(
...
+        }
+
+        config
+    };
+    install_trace_subscriber(config.logging_config())
+        .context("couldn't install tracing subscriber")?;
+    let _metrics_exporter = install_metrics_exporter(config.metrics_config())
+        .context("failed to install metrics exporter")?;
+
+    info!(?common_options, ?config, "Starting aggregation job creator");
```
BranLwyd, line 200: nit: this info line refers to the aggregation job creator specifically; I suspect we can get away without referencing which job is starting up, so the easiest fix might be to just change the string to something like `"Starting up"`.

BranLwyd (approved):

janus_server/src/binary_utils.rs
divergentdave: How about this?
```suggestion
#[cfg_attr(doc, doc = "Common options that are used by all Janus binaries.")]
```

divergentdave (approved):

janus_server/src/binary_utils.rs
tgeoghegan: Yeah, I think in practice it'll be obvious which binary logged this because the log event will be annotated with a Kubernetes pod or deployment.

janus_server/src/binary_utils.rs
tgeoghegan: That works, thanks!

PR #180, opened 2022-05-19
Introduce mod `job_creator`

tgeoghegan:
Factors logic for enumerating tasks and creating per-task jobs out of
`aggregation_job_creator` and into a new module. Also adds a skeleton of
`collect_job_creator` to show how this is used across multiple binary
targets.

Part of #105


tgeoghegan:
Now that I've posted this pull request, I think I've actually convinced myself I don't need it. Collect jobs get _created_ in the aggregator's `/collect` handler. What I wanted was to factor out the support to write a `collect_job_driver`. Downgrading to draft for now. Please don't spend any time reviewing until/unless I upgrade it back to a real PR!

PR #182, opened 2022-05-20
Introduce mod `job_driver`

tgeoghegan:
Factors logic for discovering incomplete jobs out of
`aggregation_job_driver` and into a new module.
Adds a skeleton of `collect_job_creator` to show how this is used across
multiple binary targets.

Part of #105

### DIFF
--- a/janus_server/src/bin/aggregation_job_driver.rs
+++ b/janus_server/src/bin/aggregation_job_driver.rs
@@ -6,15 +6,18 @@ use futures::{
 use http::header::CONTENT_TYPE;
 use janus::{
     hpke::{self, associated_data_for_report_share, HpkeApplicationInfo, Label},
-    message::{Duration, Report, Role, TaskId, Time},
+    message::{Duration, Report, Role, TaskId},
     time::{Clock, RealClock},
 };
 use janus_server::{
-    binary_utils::{janus_main, BinaryOptions, CommonBinaryOptions},
+    binary_utils::{janus_main, job_driver::JobDriver, BinaryOptions, CommonBinaryOptions},
     config::AggregationJobDriverConfig,
     datastore::{
         self,
-        models::{AggregationJob, AggregationJobState, ReportAggregation, ReportAggregationState},
+        models::{
+            AcquiredAggregationJob, AggregationJob, AggregationJobState, ReportAggregation,
+            ReportAggregationState,
+        },
         Datastore,
     },
     message::{
@@ -34,8 +37,7 @@ use prio::{
 };
 use std::{fmt::Debug, sync::Arc};
 use structopt::StructOpt;
-use tokio::{sync::Semaphore, task, time};
-use tracing::{debug, error, info, info_span, Instrument};
+use tracing::error;
 
 #[derive(Debug, StructOpt)]
 #[structopt(
@@ -69,31 +71,48 @@ async fn main() -> anyhow::Result<()> {
     janus_main::<Options, _, _, _, _>(
         RealClock::default(),
         |clock, config: AggregationJobDriverConfig, datastore| async move {
-            let http_client = reqwest::Client::builder()
-                .user_agent(CLIENT_USER_AGENT)
-                .build()
-                .context("couldn't create HTTP client")?;
-
+            let aggregation_job_driver = Arc::new(AggregationJobDriver {
+                http_client: reqwest::Client::builder()
+                    .user_agent(CLIENT_USER_AGENT)
+                    .build()
+                    .context("couldn't create HTTP client")?,
+            });
             // Start running.
-            Arc::new(AggregationJobDriver {
-                datastore,
+            Arc::new(JobDriver::new(
+                Arc::new(datastore),
                 clock,
-                http_client,
-                min_aggregation_job_discovery_delay: Duration::from_seconds(
-                    config.min_aggregation_job_discovery_delay_secs,
+                Duration::from_seconds(config.job_driver_config.min_job_discovery_delay_secs),
+                Duration::from_seconds(config.job_driver_config.max_job_discovery_delay_secs),
+                config.job_driver_config.max_concurrent_job_workers,
+                Duration::from_seconds(config.job_driver_config.worker_lease_duration_secs),
+                Duration::from_seconds(
+                    config
+                        .job_driver_config
+                        .worker_lease_clock_skew_allowance_secs,
                 ),
-                max_aggregation_job_discovery_delay: Duration::from_seconds(
-                    config.max_aggregation_job_discovery_delay_secs,
-                ),
-                max_concurrent_aggregation_job_workers: config
-                    .max_concurrent_aggregation_job_workers,
-                aggregation_worker_lease_duration: Duration::from_seconds(
-                    config.aggregation_worker_lease_duration_secs,
-                ),
-                aggregation_worker_lease_clock_skew_allowance: Duration::from_seconds(
-                    config.aggregation_worker_lease_clock_skew_allowance_secs,
-                ),
-            })
+                |datastore, lease_duration, max_acquire_count| async move {
+                    datastore
+                        .run_tx(|tx| {
+                            Box::pin(async move {
+                                // TODO(brandon): only acquire jobs whose batch units have not
+                                // already been collected (probably by modifying
+                                // acquire_incomplete_aggregation_jobs)
+                                tx.acquire_incomplete_aggregation_jobs(
+                                    lease_duration,
+                                    max_acquire_count,
+                                )
+                                .await
+                            })
+                        })
+                        .await
+                },
+                |datastore, acquired_aggregation_job, aggregation_job_driver| async move {
+                    aggregation_job_driver
+                        .step_aggregation_job(datastore, &acquired_aggregation_job)
+                        .await
+                },
+                aggregation_job_driver,
+            ))
             .run()
             .await;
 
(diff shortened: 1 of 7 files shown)


tgeoghegan:
This needs more test coverage (note TODOs in test function in `janus_server/src/binary_utils/job_driver.rs`), which I want to do in this PR, but this is far enough along that the approach can be critiqued, plus it's about EOD on a Friday so I felt I should push.

janus_server/src/binary_utils/job_driver.rs
```diff
@@ -0,0 +1,284 @@
...
+    job_stepper: Stepper,
+}
+
+impl<C, E, Acquirer, AcquirerFut, Stepper, StepperFut, JobId> JobDriver<C, Acquirer, Stepper>
+where
+    C: Clock,
+    E: Debug + Send + Sync + 'static,
+    Acquirer: Fn(Arc<Datastore<C>>, Duration, usize) -> AcquirerFut + Send + Sync + 'static,
+    AcquirerFut:
+        Future<Output = Result<Vec<(TaskId, VdafInstance, JobId, Time)>, datastore::Error>> + Send,
```
BranLwyd, line 48: I'm not sure this is the right abstraction for acquired jobs: I think we want something like `Vec<(UserDefinedStructure, Time)>`, i.e. replace all of `TaskId, VdafInstance, JobId` with a user-defined structure that ends up being passed to `Stepper`. I say this for a couple of reasons:

* I'm not sure `VdafInstance` is generally-applicable. The collect job driver is going to be making `AggregateShareReqs` as it can & storing the results in the datastore to eventually be included in a `CollectResp`, right? If so, I don't think there's anything that would require knowing the VDAF type. (the existing agg job driver retrieves the VDAF type here so it can do the generic-function match early, to drop one tx and/or some very awkward code in the stepper logic)
* I am not sure that every kind of job will be identified by a `TaskId` and one other kind of `JobId` -- eg we might have `TaskId, JobIdA, JobIdB`. While this could be worked around in the existing framework by making `JobId = (JobIdA, JobIdB)`, this is more awkward than just having the caller use a structure of their own. In the other direction, if we have a task-level driver, we wouldn't need `JobId` at all; using a user-defined structure in this case avoids the awkwardness of `JobId = ()`.

janus_server/src/config.rs
```diff
@@ -119,81 +108,60 @@ pub struct AggregationJobCreatorConfig {
...
-    /// Applies only when there are no aggregation jobs to be stepped.
-    pub max_aggregation_job_discovery_delay_secs: u64,
-    /// The maximum number of aggregation jobs being stepped at once. This parameter determines the
-    /// amount of per-process concurrency.
-    pub max_concurrent_aggregation_job_workers: usize,
-    /// The length of time, in seconds, workers will acquire a lease for the aggregation jobs they
-    /// are stepping. Along with aggregation_worker_lease_clock_skew_allowance, determines the
-    /// effective timeout of stepping a single aggregation job.
-    pub aggregation_worker_lease_duration_secs: u64,
+pub struct JobDriverConfig {
```
BranLwyd, line 139: I'm not sure we want to make `aggregation_job_driver` & `collect_job_driver` share the same config files, in case they need different options in the future.

OTOH, the "driver" options could be usefully factored out to their own sub-structure, included (perhaps with `#[serde(flatten)]`) in both job-specific configurations.

janus_server/src/binary_utils/job_driver.rs
```diff
@@ -0,0 +1,284 @@
...
+                    //
+                    // Unwrap safety: we have seen that at least `acquired_jobs.len()` permits are
+                    // available, and this task is the only task that acquires permits.
+                    let permit = Arc::clone(&sem).try_acquire_owned().unwrap();
+                    let this = Arc::clone(&self);
+                    async move {
+                        info!(?lease_expiry, "Stepping job");
+                        match time::timeout(
+                            this.effective_lease_duration(lease_expiry),
+                            (this.job_stepper)(this.datastore.clone(), vdaf, task_id, job_id),
```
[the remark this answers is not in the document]
tgeoghegan: I think what you propose is _possible_, but I'm not sure how to achieve it. The difficulty is convincing the compiler that the nested async blocks and function calls will complete before the `JobDriver` value gets dropped. I think that's doable if we put the right lifetimes in the trait bounds on the closures and the futures they return, but then it's tricky to convince `tokio::task::spawn` that the closure we pass to it has lifetime `'static` (cloning `Arc` is a very easy way to meet that last requirement). I gave myself about an hour to try to get this to work, then I actually tried for two and didn't get there, and now I think I really need to move on to more important problems.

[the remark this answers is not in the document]
BranLwyd: Ah, I didn't intend to nerd-snipe you :-) it's not a big deal if it's hard to make Rust do the right thing, here -- I suppose we're at most saving an `Arc` clone.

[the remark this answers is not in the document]
tgeoghegan: I spent more time on this and I think I got most of the way to a solution. Basically, we want something like:

```
struct JobDriver {
    datastore: Datastore<C>, // owned value
}
```

janus_server/src/bin/aggregation_job_driver.rs
```diff
@@ -66,32 +65,42 @@ const DAP_AUTH_HEADER: &str = "DAP-Auth-Token";
...
+                                    lease_duration,
+                                    max_acquire_count,
+                                )
+                                .await
+                            })
+                        })
+                        .await
+                },
+                |datastore, vdaf, task_id, aggregation_job_id| async move {
+                    let aggregation_job_driver = AggregationJobDriver {
```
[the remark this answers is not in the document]
divergentdave: Likewise, we should make one `reqwest::Client` and clone it for use in each task, so we can get the benefits of connection pooling.

janus_server/src/binary_utils/job_driver.rs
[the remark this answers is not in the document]
BranLwyd: Right. But in practice you have `this = Arc::clone(&self)` being moved into the spawned task already; can you use `&this.datastore` to avoid doubling up on `Arc`s? That is effectively what the previous implementation was doing (well, it passed an entire `&AggregationJobDriver`, and could freely access `AggregationJobDriver::datastore` from there).

This appears to work based on modifying the existing aggregation job driver code, and I don't see a reason why it wouldn't work here too.

janus_server/src/binary_utils/job_driver.rs
```diff
@@ -0,0 +1,388 @@
...
+                    // job-stepper task to ensure that acquiring a permit does not race with
+                    // checking how many permits we have available in the next iteration of this
+                    // loop, to maintain the invariant that this task is the only place we acquire
+                    // permits.
+                    //
+                    // Unwrap safety: we have seen that at least `acquired_jobs.len()` permits are
+                    // available, and this task is the only task that acquires permits.
+                    let permit = Arc::clone(&sem).try_acquire_owned().unwrap();
+                    let this = Arc::clone(&self);
+                    let acquired_job_clone = acquired_job.clone();
```
BranLwyd, line 141: Stick this in an `Arc` or otherwise avoid cloning a structure of unknown size, IMO.

tgeoghegan: Done, at the cost of some slightly ugly `Arc<AcquiredJob>` in various function signatures.

janus_server/src/binary_utils/job_driver.rs
```diff
@@ -0,0 +1,414 @@
...
+
+                            Ok(Err(err)) => {
+                                error!(?err, "Couldn't step job")
+                            }
+
+                            Err(err) => error!(?err, "Stepping job timed out"),
+                        }
+                        drop(permit);
+                    }
+                    .instrument(info_span!("Job stepper", acquired_job = ?acquired_job_clone))
```
divergentdave, line 186: This is minor, but we could avoid putting `acquired_job` in an Arc and cloning it if we constructed the span first, moved the acquired job into the future, then annotated the future with the span, and adjusted signatures to match.

divergentdave (approved):

janus_server/src/binary_utils/job_driver.rs
```diff
@@ -0,0 +1,414 @@
...
+            // TODO(brandon): only acquire jobs whose batch units have not already been collected (probably by modifying acquire_incomplete_aggregation_jobs)
+            let max_acquire_count = sem.available_permits();
+            info!(max_acquire_count, "Acquiring jobs");
+            let acquired_jobs = (self.incomplete_job_acquirer)(
+                self.datastore.clone(),
+                self.worker_lease_duration,
+                max_acquire_count,
+            )
+            .await;
+            let acquired_jobs = match acquired_jobs {
```
BranLwyd, line 128: I probably should have done this myself in the agg job driver, but especially now that the "acquire" process has been made generic, IMO we should have a check that the number of acquired jobs is <= `max_acquire_count`. This is required for some of the reasoning in the comments around semaphore acquiry to work out. An assert would be fine IMO since this can be viewed as programmer error.

BranLwyd (approved):

janus_server/src/binary_utils/job_driver.rs
tgeoghegan: Huh, I had thought that `info_span!` would take ownership of fields passed into it, but I guess it must be invoking `std::fmt::Debug::fmt` on the field at the time the span gets created instead of holding on to the value. I implemented your suggestion in #188. Thanks!

PR #192, opened 2022-05-25
acquire and release collect jobs in datastore

tgeoghegan:
Adds support for acquiring and releasing leases on collect jobs to the
datastore module, which will soon be used by the collect job driver to
drive jobs.

Part of #105

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -142,6 +142,7 @@ CREATE TABLE collect_jobs(
     batch_interval_start    TIMESTAMP NOT NULL, -- the start of the batch interval
     batch_interval_duration BIGINT NOT NULL,    -- the length of the batch interval in seconds
     aggregation_param       BYTEA NOT NULL,     -- the aggregation parameter (opaque VDAF message)
+    lease_expiry            TIMESTAMP NOT NULL DEFAULT TIMESTAMP '-infinity',  -- when lease on this collect job expires; -infinity implies no current lease
     helper_aggregate_share  BYTEA,              -- the helper's encrypted aggregate share; null until the helper has serviced the AggregateShareReq
     leader_aggregate_share  BYTEA,              -- the leader's unencrypted aggregate share; null until the leader has performed its aggregation for a CollectReq
     report_count            BIGINT,             -- the count of reports included in the leader's aggregate share; null until the leader has performed its aggregation for a CollectReq
@@ -150,6 +151,8 @@ CREATE TABLE collect_jobs(
     CONSTRAINT unique_collect_job_task_id_interval_aggregation_param UNIQUE(task_id, batch_interval_start, batch_interval_duration, aggregation_param),
     CONSTRAINT fk_task_id FOREIGN KEY(task_id) REFERENCES tasks(id)
 );
+-- TODO: what's the right index here?
+CREATE INDEX collect_jobs_lease_expiry ON collect_jobs(lease_expiry);
 
 -- The helper's view of aggregate share jobs.
 CREATE TABLE aggregate_share_jobs(
--- a/janus_server/src/datastore.rs
+++ b/janus_server/src/datastore.rs
@@ -1252,50 +1252,47 @@ impl<C: Clock> Transaction<'_, C> {
         let lease_expiry_time = now.add(lease_duration)?;
         let maximum_acquire_count: i64 = maximum_acquire_count.try_into()?;
 
-        let stmt = self.tx.prepare_cached(
-            "WITH updated as (
-            UPDATE collect_jobs SET lease_expiry = $1
-            -- Join on tasks table
-            FROM tasks
-            WHERE
-                -- Constraint for tasks table in FROM position
-                tasks.id = collect_jobs.task_id
-                -- Do not acquire collect jobs that have been completed
-                AND collect_jobs.helper_aggregate_share IS NULL
-                -- Do not acquire collect jobs with an unexpired lease
-                AND collect_jobs.lease_expiry <= $2
-                AND collect_jobs.id IN (
-                    SELECT collect_jobs.id
-                    FROM collect_jobs
-                    -- Join on aggregation jobs with matching task ID and aggregation parameter
-                    INNER JOIN aggregation_jobs
-                        ON collect_jobs.aggregation_param = aggregation_jobs.aggregation_param
-                        AND collect_jobs.task_id = aggregation_jobs.task_id
-                    -- Join on report aggregations with matching aggregation job ID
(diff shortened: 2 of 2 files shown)


janus_server/src/datastore.rs
```diff
@@ -1239,6 +1239,118 @@ impl<C: Clock> Transaction<'_, C> {
...
+                        AND client_reports.id = report_aggregations.client_report_id
+                    WHERE
+                        -- Do not acquire collect jobs that have been completed
+                        collect_jobs_alias.helper_aggregate_share IS NULL
+                        -- Do not acquire collect jobs with an unexpired lease
+                        AND collect_jobs_alias.lease_expiry <= $2
+                        -- Filter out collect jobs where any associated aggregation jobs are not
+                        -- finished
+                        -- XXX How to achieve this?
+                        -- AND collect_jobs_alias.id NOT IN (SELECT collect_jobs_alias.id FROM updated WHERE aggregation_jobs.state = 'IN_PROGRESS')
```
tgeoghegan, line 1285: I need some help with this. This big series of joins figures out which aggregation jobs are related to report shares whose nonce is within the collect job's batch interval. There may be multiple aggregation jobs for a single collect job. I can't quite figure out how to handle the case where some but not all of the aggregation jobs are in the `FINISHED` state. Basically, if there are any rows in the giant joined table where `aggregation_jobs.state` is `IN_PROGRESS`, then I want to remove all other rows that have the same `collect_jobs_alias.id`. It feels like I want to intersect the table with itself, and I'm not sure how to do that or if it's a good idea.

janus_server/src/datastore.rs
```diff
@@ -3812,6 +3932,475 @@ mod tests {
...
+                aggregation_jobs,
+                report_aggregations,
+                collect_job_test_cases,
+            },
+        )
+        .await;
+    }
+
+    #[tokio::test]
+    async fn collect_job_acquire_release_aggregation_job_in_progress() {
```
tgeoghegan, line 4333: This test fails because as I described elsewhere, I can't quite figure out the SQL for it.

janus_server/src/datastore.rs
divergentdave: I need to chew on this some more, but my idea is to put a SELECT subquery in the UPDATE's FROM position, use a GROUP BY in the SELECT, and impose our for-all conditions with the bool_and aggregate function. I think we definitely need to make use of subqueries somehow here, because our UPDATE query may otherwise return multiple rows per collect_job row. Per the [Postgres docs](https://www.postgresql.org/docs/14/sql-update.html#id-1.9.3.182.8): "When using FROM you should ensure that the join produces at most one output row for each row to be modified."

Other notes: I think we don't want `FROM collect_jobs as collect_jobs_alias`, as that gives us a self-join on that table, once from the UPDATE table, and once from the FROM clause. The `LIMIT $3` needs to be in the UPDATE subquery, rather than the wrapping `SELECT`, to keep us from taking out extra leases and dropping them on the ground.

divergentdave: Okay, I got the test passing at least: 5a2c0d07714c48ee8d46fdbb45c42631296b57ad

tgeoghegan: THANK YOU! I was going nuts trying to figure this out. Not to mention the other two or three crippling bugs you spotted!

I did have to make one tweak, to move the `WHERE` clauses into the inner `SELECT` where the `LIMIT` is applied. Otherwise, the `LIMIT` gets applied to query results before we check for collect job lease expiry or completion, and so we might end up discarding runnable jobs.

db/schema.sql
```diff
@@ -142,6 +142,7 @@ CREATE TABLE collect_jobs(
     batch_interval_start    TIMESTAMP NOT NULL, -- the start of the batch interval
     batch_interval_duration BIGINT NOT NULL,    -- the length of the batch interval in seconds
     aggregation_param       BYTEA NOT NULL,     -- the aggregation parameter (opaque VDAF message)
+    lease_expiry            TIMESTAMP NOT NULL DEFAULT TIMESTAMP '-infinity',  -- when lease on this aggregation job expires; -infinity implies no current lease
```
divergentdave, line 145: nit: copy-paste error
```suggestion
    lease_expiry            TIMESTAMP NOT NULL DEFAULT TIMESTAMP '-infinity',  -- when lease on this collect job expires; -infinity implies no current lease
```

janus_server/src/datastore.rs
```diff
@@ -1239,6 +1239,119 @@ impl<C: Clock> Transaction<'_, C> {
...
+                        collect_job_id,
+                    },
+                    lease_expiry_time,
+                ))
+            })
+            .collect()
+    }
+
+    /// release_collect_job releases an acquired (via e.g. acquire_incomplete_collect_jobs) collect
+    /// job. It returns an error if the collect job has no current lease.
```
divergentdave, line 1326: This method doesn't currently check for an existing lease. Should this be an IOU for when we add ticket IDs to both of our kinds of leases?

divergentdave (approved):

janus_server/src/datastore.rs
tgeoghegan: I spotted Brandon's comment in the corresponding `release_aggregation_job`. I don't think it makes sense to check for an active lease, because then the caller is racing with the lease expiring. Whether the lease is released due to explicit caller action or due to timeout, the desired postcondition (no lease is held) is achieved, so I don't think it makes sense to hand an error to the caller (what would they do with it?).

tgeoghegan: Oh, wait, I realize now you must both mean checking that the lease is not `-infinity`, rather than checking if there is an unexpired lease. That seems like a sensible guard against programmer error. I think this fits with a few other hardenings and improvements of our queue implementation we'll need that I think should be addressed beyond this PR. I have filed #193 as an umbrella issue to track that work.

BranLwyd: Yep, avoiding the ABA problem in the presence of significant clock skew (or implementation bugs in lease timeout logic) is the goal of that TODO. I think it's worthwhile to address because, while clock skew may not be very likely, the consequences of the problem if it does occur are effectively "data corruption" or "lost writes".

BranLwyd (approved):

janus_server/src/datastore.rs
tgeoghegan: Got it. I'm not arguing that this isn't a problem, but rather that we should address it in a later PR, fixing this for both aggregate jobs and collect jobs at once.

PR #197, opened 2022-05-28
Remove AggregateContinueRequest::task_id.

BranLwyd:
This brings Janus' messages fully in-line with the current DAP
specification.

#187 was merged with `task_id` in place, in the optimistic hope that https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/263 would be merged fairly quickly. I'm no longer as optimistic, and if it comes to pass that the DAP change is not merged by the time we are ready to begin end-to-end testing with CF's implementation, I don't want this difference to hold up testing.

I don't plan to merge this until https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/263 is resolved one way or another, unless/until this would become a blocker for interop testing.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -89,7 +89,7 @@ pub enum Error {
     UnrecognizedTask(TaskId),
     /// An attempt was made to act on an unknown aggregation job.
     #[error("unrecognized aggregation job: {0:?}")]
-    UnrecognizedAggregationJob(AggregationJobId, TaskId),
+    UnrecognizedAggregationJob(AggregationJobId, Option<TaskId>),
     /// An attempt was made to act on an unknown collect job.
     #[error("unrecognized collect job: {0}")]
     UnrecognizedCollectJob(Uuid),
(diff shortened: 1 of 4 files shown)


janus_server/src/aggregator.rs
```diff
@@ -241,9 +241,18 @@ impl<C: Clock> Aggregator<C> {
...
-        let task_id = TaskId::decode(&mut Cursor::new(req_bytes))?;
+        // Parse aggregation job ID early to avoid parsing the entire message before performing
+        // authentication. We need to know the task to perform authentication (since the auth token
+        // is a per-task configuration), and we need to know the aggregation job ID to look up the
+        // task.
+        let job_id = AggregationJobId::decode(&mut Cursor::new(req_bytes))?;
+        let task_id = self
+            .datastore
+            .run_tx(|tx| {
+                Box::pin(async move { tx.lookup_task_id_for_aggregation_job_id(job_id).await })
```
BranLwyd, line 252: Note: I considered e.g. `self.task_aggregator_for_aggregation_job_id(job_id)`. However:

1. If this ends up being a shim to support initial interop testing but we eventually include the `task_id` field, I don't want to spend too much time thinking about the caching logic.

2. If this ends up living forever, i.e. the WG doesn't include the `task_id` field or another indication of the task ID in aggregate continue requests, we might want to be careful about caching since aggregation jobs are pretty short-lived. E.g. for `prio3`, we have one AggregateInitializeReq & one AggregateContinueReq for each aggregation job, so we have one chance to get a cache hit -- if the AggregateContinueReq happens to go to a different process than the AggregateInitializeReq (pretty likely once we start scaling), simple in-memory caching will be ineffectual.

For that reason I decided just to look up the task ID for now.

tgeoghegan (approved):

BranLwyd:
Closing this pull request as https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/263 has been merged.

PR #208, opened 2022-05-31
remove direct dependency on `hpke`

tgeoghegan:
`hpke-dispatch` re-exports numerous items from `hpke`, so we run into
trouble when `hpke-dispatch` uses a different version of `hpke` than
`janus` (see for instance PR #201). We resolve this by removing our
direct dependency on `hpke` and only using it via `hpke-dispatch`.


janus/src/message.rs
```diff
@@ -557,11 +553,11 @@ impl Decode for HpkeKdfId {
 #[repr(u16)]
 pub enum HpkeAeadId {
```
jbr, line 554: Would it break a lot of code if this was just a newtype `HpkeAeadId(Aead)`?

janus/src/hpke.rs
```diff
@@ -330,24 +341,17 @@ mod tests {
...
-        round_trip_check::<X25519HkdfSha256, HkdfSha256, AesGcm128>();
-        round_trip_check::<X25519HkdfSha256, HkdfSha256, AesGcm256>();
-        round_trip_check::<X25519HkdfSha256, HkdfSha256, ChaCha20Poly1305>();
-        round_trip_check::<X25519HkdfSha256, HkdfSha384, AesGcm128>();
-        round_trip_check::<X25519HkdfSha256, HkdfSha384, AesGcm256>();
-        round_trip_check::<X25519HkdfSha256, HkdfSha384, ChaCha20Poly1305>();
-        round_trip_check::<X25519HkdfSha256, HkdfSha512, AesGcm128>();
-        round_trip_check::<X25519HkdfSha256, HkdfSha512, AesGcm256>();
-        round_trip_check::<X25519HkdfSha256, HkdfSha512, ChaCha20Poly1305>();
+        for kem_id in [HpkeKemId::P256HkdfSha256, HpkeKemId::X25519HkdfSha256] {
```
jbr, line 344: This could potentially be `hpke_dispatch::KEM_ALL` (and likewise `KDF_ALL`, and `AEAD_ALL`) but I don't have a good sense of whether it's actually an advantage to have the explicit reenumeration here

janus/src/message.rs
tgeoghegan: That might be better. TBH the only reason we need to define this or the `HpkeKemId` or `HpkeKdfId` enums in crate `janus` is so we can put `prio::codec::{Encode, Decode}` implementations on them and them send them over the wire. Otherwise your label on a `u16` is just as good as ours. I think maybe we could have `prio::codec` provide something like `impl<T> Encode for T where T: Into<u16>`, and then any enum with `repr(u16)` that we derive `num_enum::IntoPrimitive` on can be used. On the other hand, maybe `Into<u16>` is too widespread for that to be a good idea.

janus/src/hpke.rs
tgeoghegan: The idea here is to enumerate over `janus` representations of the KDF, KEM and AEADs. It's possible this test is entirely redundant if `hpke-dispatch` already exercises all these cases.

divergentdave (approved):

janus/src/message.rs
tgeoghegan: I gave this a try and it quickly become more complicated than I think it's worth. The short version is that it makes encoding fallible (since `TryFrom` is fallible) and then you have to plumb errors everywhere and I'm not convinced that's better than these newtypes.

janus/src/hpke.rs
tgeoghegan: This test is kinda redundant with `tests/round_trip.rs` in `hpke-dispatch`, except that this does exercise the Janus representation of stuff like algorithm IDs and HPKE configurations, so let's keep it.

PR #210, opened 2022-06-02
Add CORS filters to client-facing endpoints

### LINKED ISSUE #184: CORS headers on /upload and /hpke_config
We will need to set CORS headers to allow direct access from browser-based DAP clients.

divergentdave:
This fixes #184, using Warp's built-in CORS filter. I confirmed that these settings are sufficient for sending requests and reading responses with `fetch()` in Firefox and Chromium.

One limitation to be aware of: HTTP error responses that arise from Warp rejections do not carry an Access-Control-Allow-Origin header, so browser clients will see a CORS error rather than the underlying 404, 400, etc. error. (This affects rejections from unrecognized paths, mismatched HTTP methods, and mismatched Content-Type headers.) HTTP error responses that originate in our `error_handler()` filter do get decorated with CORS headers, so browser clients can read them properly. This is because the CORS wrapping filter passes rejections on unmodified, whereas it wraps reply futures to add headers before responses are sent. (Presumably it does not wrap or otherwise modify rejections because that would break `Rejection::find()`.) If all filters reject a request, the rejection is finally turned into a response inside the library, using a non-public trait.

If we really wanted to, we could add a filter within the wrapping CORS filter that recovers rejections, tries to match them, and produces responses, so that they could get CORS headers. I don't think this is worth the complication, as the affected error cases all represent programmer error (and thus do not require clients to be able to see error messages from script), plus omitting CORS headers on certain error responses seems to be common practice around the web.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -1732,6 +1732,13 @@ where
     }
 }
 
+/// The number of seconds we send in the Access-Control-Max-Age header. This determines for how
+/// long clients will cache the results of CORS preflight requests. Of popular browsers, Mozilla
+/// Firefox has the highest Max-Age cap, at 24 hours, so we use that. Our CORS preflight handlers
+/// are tightly scoped to relevant endpoints, and our CORS settings are unlikely to change.
+/// See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
+const CORS_PREFLIGHT_CACHE_AGE: u32 = 24 * 60 * 60;
+
 /// Constructs a Warp filter with endpoints common to all aggregators.
 fn aggregator_filter<C: Clock>(
     datastore: Arc<Datastore<C>>,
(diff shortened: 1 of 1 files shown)


BranLwyd (approved):

janus_server/src/aggregator.rs
```diff
@@ -1751,42 +1758,59 @@ fn aggregator_filter<C: Clock>(
...
+                            .map_err(|err| Error::Internal(format!("couldn't produce response: {}", err)))
+                    },
+                )
+                .map(error_handler(&response_counter, "hpke_config"))
+                .with(
+                    warp::cors()
+                        .allow_any_origin()
+                        .allow_method("GET")
+                        .max_age(CORS_PREFLIGHT_CACHE_AGE),
+                ),
```
tgeoghegan, line 1784: nit: I wonder if common stuff like these, setting up the timing wrapper and the named `tracing` spans could be factored out into a common function so that we don't have to repeat it across every HTTP endpoint.

tgeoghegan (approved):

janus_server/src/aggregator.rs
divergentdave: Yeah, that could be a good improvement. Maybe we could have a utility function take a routing filter, a response-making filter, and a name/label, and return a composite filter with appropriate wrappers. I'll try this out as part of #211.

PR #212, opened 2022-06-02
Leases: add tokens, on-release sanity checking, acquire counts.

### LINKED ISSUE #193: Improvements to job queues
We use a handrolled job queue to manage aggregation and collect jobs in the Janus leader. It works, but needs improvements:

- [x] check if a job lease has ever been acquired before releasing it ([discussion](https://github.com/divviup/janus/pull/192#discussion_r883012373), [TODO in datastore.rs](https://github.com/divviup/janus/blob/6654d7f1c39ff448ebdf997251fd120cc64f7f6a/janus_server/src/datastore.rs#L836)
- [x] add ticket IDs to job leases
- [x] exponential backoff on job message redelivery
- [x] track job retries and eventually give up / move job to a dead letter queue

BranLwyd:
* The tokens are a (probably-)unique opaque value associated with each
    lease, intended to ensure that a release operation affects the same
    lease that was acquired.
  * On-release sanity checking is implemented by checking that the lease
    timeout & token match what is recorded in the in-memory lease. If
    they don't, someone else has taken the lease out from under us & we
    must fail. Note that we might still be able to release if the lease
    expiry time is in the past--the lease expiry time is the first time
    someone else _can_ take the lease, if no one else has taken the
    lease we can still release it.
  * Tracking of acquire counts (since last successful release) will
    allow lease users to decide when to permanently fail. (to be
    implemented in a pair of upcoming PRs)

I also adapt the JobDriver abstraction to work with leases; I suppose
this isn't strictly necessary, and it enforces that all users of the
JobDriver need to work with leases in the future. I think that's OK.

Note that I don't implement exponential backoff. I think we need to tune
our batch jobs to be non-overloading even when running at maximum
throughput; we might choose to implement exponential backoff on the helper
request retries, once we have retries in place, since we don't have control
over how overloaded the helper is.

Closes #193.

### DIFF
--- a/janus_server/src/datastore.rs
+++ b/janus_server/src/datastore.rs
@@ -2,7 +2,7 @@
 
 use self::models::{
     AcquiredAggregationJob, AcquiredCollectJob, AggregateShareJob, AggregationJob, AggregatorRole,
-    BatchUnitAggregation, CollectJob, ReportAggregation, ReportAggregationState,
+    BatchUnitAggregation, CollectJob, Lease, LeaseToken, ReportAggregation, ReportAggregationState,
     ReportAggregationStateCode,
 };
 use crate::{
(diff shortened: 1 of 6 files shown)


janus_server/src/datastore.rs
```diff
@@ -830,27 +841,30 @@ impl<C: Clock> Transaction<'_, C> {
...
         check_update(
             self.tx
                 .execute(
                     &stmt,
                     &[
-                        /* task_id */ &task_id.as_bytes(),
-                        /* aggregation_job_id */ &aggregation_job_id.as_bytes(),
+                        /* task_id */ &lease.leased().task_id.as_bytes(),
+                        /* aggregation_job_id */
+                        &lease.leased().aggregation_job_id.as_bytes(),
```
tgeoghegan, line 865:
```suggestion
                        /* aggregation_job_id */ &lease.leased().aggregation_job_id.as_bytes(),
```

tgeoghegan (approved):

janus_server/src/datastore.rs
BranLwyd: I definitely agree, but the auto-formatter insists on this layout :-/

PR #213, opened 2022-06-02
JobDriver: remove JobStepperContext.

BranLwyd:
The future-generating function can own & clone any required contextual
state without assistance from the JobDriver.


tgeoghegan (approved):
Thank you, the exact right placement of `move` was eluding me when I wrote this.

PR #214, opened 2022-06-02
JobDriver: remove all non-required parameters from callbacks.

BranLwyd:
The callbacks (job-acquirer, job-stepper) now only receive information
that is managed/computed by the JobDriver itself; data that is
effectively "constant" that the JobDriver was holding onto is now
expected to be owned by the callback itself.

This change may be controversial because it pushes more logic into the
application rather than keeping it in the "helper" JobDriver. I think
it's worthwhile not just from a cleanliness perspective but also because
it opens the door to later cleanups/optimizations--as a minor example,
we could now place the datastore into the {Aggregation,Collect}JobDriver
and no longer need to manage an Arc\<Datastore\>.

### DIFF
--- a/janus_server/src/bin/aggregation_job_driver.rs
+++ b/janus_server/src/bin/aggregation_job_driver.rs
@@ -69,48 +69,59 @@ async fn main() -> anyhow::Result<()> {
     janus_main::<Options, _, _, _, _>(
         RealClock::default(),
         |clock, config: AggregationJobDriverConfig, datastore| async move {
+            let datastore = Arc::new(datastore);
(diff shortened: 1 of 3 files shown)


tgeoghegan (approved):
I thought there was some value to being able to ensure that the callbacks use the same `Datastore` as the `JobDriver`, but I suppose passing the `Arc<Datastore>` into the closures was never going to stop a client of `JobDriver` from using the wrong one in their closure. This LGTM and does indeed simplify things.

divergentdave (approved):

PR #215, opened 2022-06-02
Rearrange wrapping of Warp filters

### LINKED ISSUE #211: `timing_wrapper()` should only record after request routing is successful
I had an insight while working on the CORS headers that our `warp::wrap_fn(timing_wrapper(...))` filters are currently placed in the wrong spot. Currently, they each wrap a filter chain that begins with `warp::path(...)`. This means that, depending on the request and order of or'd filters, one request may enter timing filters multiple times, as it runs each path filter and rejects immediately before finding the correct one.

To solve this, we should rearrange all filters similar to how #210 did, by constructing a filter stack starting with the method filter, adding the wrapping timing filter to the stack at the end, and then placing that filter stack inside `warp::path("...").and(...)`. I'll take care of this as a follow-up.

divergentdave:
This moves metrics and tracing wrapping filters so that they don't fire until `warp::path` filters have succeeded. The relevant part of filter composition is now handled in a common helper method.

This fixes #211.

There are two other small behavioral changes: `collect_jobs_endpoint` now includes a `warp::get()` in its filter stack, and it was missing a trace filter previously. The new helper method is already paying off :smile:

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -1732,6 +1732,45 @@ where
     }
 }
 
+/// Convenience function to perform common composition of Warp filters for a single endpoint. A
+/// combined filter is returned, instrumented to measure request processing time for metrics, and
+/// with per-route named tracing spans.
+///
+/// `route_filter` should be a filter that determines whether the incoming request matches a
+/// given route or not. It should inspect the ambient request, and either extract the empty tuple
+/// or reject.
+///
+/// `response_filter` should be a filter that performs all response handling for this route, after
+/// the above `route_filter` has already determined the request is applicable to this route. It
+/// should only reject in response to malformed requests, not requests that may yet be served by a
+/// different route. This will ensure that a single request doesn't pass through multiple wrapping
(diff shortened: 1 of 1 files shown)


janus_server/src/aggregator.rs
```diff
@@ -1757,68 +1796,66 @@ fn aggregator_filter<C: Clock>(
...
-
-    let aggregate_endpoint = warp::path("aggregate")
-        .and(warp::post())
+        .map(error_handler(&response_counter, "hpke_config"))
+        .with(
+            warp::cors()
+                .allow_any_origin()
+                .allow_method("GET")
+                .max_age(CORS_PREFLIGHT_CACHE_AGE),
+        );
```
tgeoghegan, line 1822: Why not set up the error handler and the CORS headers in `compose_common_wrappers`?

tgeoghegan (changes requested):
Good catch on the collect job endpoint!

tgeoghegan (approved):

BranLwyd (approved):

PR #217, opened 2022-06-03
Aggregation job driver: give up on jobs that fail too many times.

BranLwyd:
Failure is implemented as moving the aggregation job to the FINISHED
state. We don't change the state of any of the included report
aggregations, on the theory that we may want to debug the issue &
keeping state around will help with that; this is the only way that an
aggregation job can reach the FINISHED state with report aggregations in
a non-terminal state.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -73,8 +73,9 @@ CREATE INDEX client_reports_task_and_time_index ON client_reports(task_id, nonce
 
 -- Specifies the possible state of an aggregation job.
 CREATE TYPE AGGREGATION_JOB_STATE AS ENUM(
-    'IN_PROGRESS', -- at least one included report is in a non-terminal (START, WAITING) state
-    'FINISHED'     -- all reports have reached a terminal state (FINISHED, FAILED, INVALID) or we have given up on the aggregation job entirely
+    'IN_PROGRESS', -- at least one included report is in a non-terminal (START, WAITING) state, processing can continue
+    'FINISHED',    -- all reports have reached a terminal state (FINISHED, FAILED, INVALID)
+    'ABANDONED'    -- we have given up on the aggregation job entirely
 );
 
 -- An aggregation job, representing the aggregation of a number of client reports.
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -74,7 +74,7 @@ CREATE INDEX client_reports_task_and_time_index ON client_reports(task_id, nonce
 -- Specifies the possible state of an aggregation job.
 CREATE TYPE AGGREGATION_JOB_STATE AS ENUM(
     'IN_PROGRESS', -- at least one included report is in a non-terminal (START, WAITING) state
-    'FINISHED'     -- all reports have reached a terminal state (FINISHED, FAILED, INVALID)
+    'FINISHED'     -- all reports have reached a terminal state (FINISHED, FAILED, INVALID) or we have given up on the aggregation job entirely
 );
 
 -- An aggregation job, representing the aggregation of a number of client reports.
--- a/janus_server/src/bin/aggregation_job_driver.rs
+++ b/janus_server/src/bin/aggregation_job_driver.rs
@@ -36,7 +36,7 @@ use prio::{
 };
 use std::{fmt::Debug, sync::Arc};
 use structopt::StructOpt;
-use tracing::error;
+use tracing::{error, info};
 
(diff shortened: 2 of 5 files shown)


db/schema.sql
```diff
@@ -74,7 +74,7 @@ CREATE INDEX client_reports_task_and_time_index ON client_reports(task_id, nonce
 -- Specifies the possible state of an aggregation job.
 CREATE TYPE AGGREGATION_JOB_STATE AS ENUM(
     'IN_PROGRESS', -- at least one included report is in a non-terminal (START, WAITING) state
-    'FINISHED'     -- all reports have reached a terminal state (FINISHED, FAILED, INVALID)
+    'FINISHED'     -- all reports have reached a terminal state (FINISHED, FAILED, INVALID) or we have given up on the aggregation job entirely
```
divergentdave, line 77: Would it be better to create a new state for aggregation jobs that exhaust retries? (maybe 'ABANDONED'?) This would make identifying affected jobs easier, and would simplify the correspondence between report aggregation states and aggregation job states.

janus_server/src/bin/aggregation_job_driver.rs
```diff
@@ -705,6 +715,76 @@ impl AggregationJobDriver {
             .await?;
         Ok(())
     }
+
+    // XXX: log that we are canceling?
```
divergentdave, line 719: This appears to be stale, we are logging at the callsite.

janus_server/src/bin/aggregation_job_driver.rs
```diff
@@ -1297,6 +1377,125 @@ mod tests {
...
+        // Run: create an aggregation job driver & cancel the aggregation job.
+        let aggregation_job_driver = AggregationJobDriver {
+            http_client: reqwest::Client::builder().build().unwrap(),
+        };
+        aggregation_job_driver
+            .cancel_aggregation_job(Arc::clone(&ds), lease)
+            .await
+            .unwrap();
+
+        // Verify: check that the datstore state is update as expected (the aggregation job is
```
divergentdave, line 1460: nit: typo
```suggestion
        // Verify: check that the datstore state is updated as expected (the aggregation job is
```

divergentdave (approved):

janus_server/src/bin/aggregation_job_driver.rs
```diff
@@ -117,6 +117,17 @@ async fn main() -> anyhow::Result<()> {
                         let (datastore, aggregation_job_driver) =
                             (Arc::clone(&datastore), Arc::clone(&aggregation_job_driver));
                         async move {
+                            if aggregation_job_lease.lease_attempts()
+                                >= config.job_driver_config.maximum_attempts_before_failure
+                            {
+                                info!(attempts = ?aggregation_job_lease.lease_attempts(),
```
tgeoghegan, line 123: Maybe this should be `warn!` so it stands out? But then we'll be able to search for this log statement easily anyway, and also we should be emitting metrics about this so we can alert on it. So probably a `TODO` comment so we don't forget to come back and wire up a metric suffices.

janus_server/src/bin/aggregation_job_driver.rs
BranLwyd: Gah, right you are, thanks.

janus_server/src/bin/aggregation_job_driver.rs
divergentdave: Whoops, one more
```suggestion
        // Verify: check that the datastore state is updated as expected (the aggregation job is
```

db/schema.sql
BranLwyd: I like this idea; I was trying to decide whether to do something like this myself. While I think it's not strictly necessary technically speaking, a separate state will certainly make analysis easier (as you note). And I suppose that a separate state may make later changes less error-prone -- while we don't currently assume that an aggregation job in state "FINISHED" has all-terminal report aggregations today, it'd be easy to assume so later since that would be _almost_ true.

I went ahead and made this change.

janus_server/src/bin/aggregation_job_driver.rs
BranLwyd: Yeah, that makes sense; we can search, but this is "exceptional" enough that it makes sense to make it stick out in the logs a bit.

PR #243, opened 2022-06-14
Update collect jobs to store an explicit status.

BranLwyd:
Most directly, I do this so I can introduce an "abandoned" status in an
upcoming PR, to allow us to give up on collect jobs if they fail
repeatedly. This refactoring also makes application code simpler/less
error-prone -- several error cases now no longer exist (because they
would be caught at time of DB read).

Also, drop the collect job checksum & report_count columns, since they
were not actually used for anything.


BranLwyd:
Part of #193.

divergentdave (approved):

PR #245, opened 2022-06-14
Clippy CI: also specify `--workspace` flag.

BranLwyd:
This checks all packages in the workspace.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -27,7 +27,7 @@ jobs:
     - name: lint
       run: cargo fmt --message-format human -- --check
     - name: clippy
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
This could be redundant, as Clippy seemed to be working for me before. Perhaps this was implicit when run in the context of a virtual manifest, but it certainly can't hurt.

divergentdave:
Here we go, this requires chasing it down in the `cargo check` docs.

> If the manifest is the root of a workspace then the workspaces default members are selected

The semantic difference of this change is that if we ever define a scoped-down `default-members` list in the manifest, then this `cargo clippy --workspace ...` invocation will continue to run on all workspace members, including non-default members. LGTM.

PR #246, opened 2022-06-14
Run rustdoc in CI

divergentdave:
We should check for documentation warnings in CI to catch them earlier.


BranLwyd (approved):
Thanks for this!

PR #247, opened 2022-06-15
Split VdafInstance out into the common crate

divergentdave:
This splits out `VdafInstance` into two enums, keeping the "fake" variants in `janus_server`. This is part of #18.

### DIFF
--- a/janus_server/src/task.rs
+++ b/janus_server/src/task.rs
@@ -25,13 +25,96 @@ pub enum Error {
 }
 
 /// Identifiers for VDAFs supported by this aggregator, corresponding to
-/// definitions in [draft-patton-cfrg-vdaf][1] and implementations in
+/// definitions in [draft-irtf-cfrg-vdaf-00][1] and implementations in
 /// [`prio::vdaf::prio3`].
 ///
-/// [1]: https://datatracker.ietf.org/doc/draft-patton-cfrg-vdaf/
+/// [1]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/00/
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum VdafInstance {
+    Real(janus::task::VdafInstance),
+
+    #[cfg(test)]
+    Fake,
+    #[cfg(test)]
+    FakeFailsPrepInit,
+    #[cfg(test)]
+    FakeFailsPrepStep,
+}
+
+impl From<janus::task::VdafInstance> for VdafInstance {
+    fn from(vdaf: janus::task::VdafInstance) -> Self {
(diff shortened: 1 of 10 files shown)


janus_server/src/task.rs
```diff
@@ -25,13 +25,96 @@ pub enum Error {
...
 /// Identifiers for VDAFs supported by this aggregator, corresponding to
-/// definitions in [draft-patton-cfrg-vdaf][1] and implementations in
+/// definitions in [draft-irtf-cfrg-vdaf-00][1] and implementations in
 /// [`prio::vdaf::prio3`].
 ///
-/// [1]: https://datatracker.ietf.org/doc/draft-patton-cfrg-vdaf/
+/// [1]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/00/
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum VdafInstance {
+    Real(janus::task::VdafInstance),
```
BranLwyd, line 34: I sorta dislike having an enum that (in non-test scenarios) will have exactly one value, containing an enum of a different type. This feels like unnecessary indirection.

What do you think about the following implementation strategy:
* Move `VdafInstance` to the common `janus` crate entirely, removing it from `janus_server`.
* Instead of marking the "fake" enum values `#[cfg(test)]`, create an e.g. `test-util` feature, and mark the "fake" enum values with this feature.
* Have `janus_server` include `janus` with the `test-util` feature as in `dev-dependencies`.

divergentdave: My immediate concern is the difficulty of synchronizing match blocks in `janus_server` with the enum definition in `janus`. The simplest solution of guarding match arms in `janus_server` with `#[cfg(test)]` would line up correctly when building unit tests, but when building integration tests, src/bin unit tests, examples, or benchmarks, then the `janus/test-util` feature would be set, and `janus_server`'s library crate would be built without `--cfg test`, leaving us with "missing match arm" errors.

One alternate solution would be mandating that a feature be set in `janus_server` when running tests, [like so](https://github.com/spacejam/sled/blob/60813522b579dbef4386c8fad8749fce4d64ad22/tests/common/mod.rs#L1-L7), and propagating to the `janus/test-util` feature. However, I'm hesitant to break the simple invocation of `cargo test`, and this would still leave open the possibility that a build with other crates could set the `janus/test-util` feature without the corresponding `janus_server` feature, which would give us the same problem.

If possible, I'd like to contain our coupled conditional compilation expressions to a single crate to avoid these issues.

BranLwyd: Fair enough, thanks for your thorough analysis. Let's keep it this way.

BranLwyd (approved):

Issue #234, opened 2022-06-13
Remove real (wallclock-consuming) sleeps from tests


branlwyd:
Some clock-based tests are currently racy (i.e. they use a real, wallclock-consuming sleep) for technical reasons. [typically, because `tokio::time::pause()` does not work as documented in all situations] These tests should be refactored to not require real sleeps.

inahga:
We're pretty good about not doing this, with some exceptions. We can address this as test flakes happen.

PR #249, opened 2022-06-15
Collect job driver: abandon collect jobs that fail repeatedly.

BranLwyd:
Part of #193.

### DIFF
--- a/janus_server/src/aggregator/aggregate_share.rs
+++ b/janus_server/src/aggregator/aggregate_share.rs
@@ -207,6 +207,19 @@ impl CollectJobDriver {
 
         Ok(())
     }
+
+    #[tracing::instrument(skip(self, datastore), err)]
+    pub async fn cancel_collect_job<C: Clock>(
+        &self,
+        datastore: Arc<Datastore<C>>,
+        lease: Lease<AcquiredCollectJob>,
+    ) -> Result<(), Error> {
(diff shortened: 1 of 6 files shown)


janus_server/src/aggregator/aggregate_share.rs
```diff
@@ -207,6 +207,19 @@ impl CollectJobDriver {
...
+
+    #[tracing::instrument(skip(self, datastore), err)]
+    pub async fn cancel_collect_job<C: Clock>(
+        &self,
+        datastore: Arc<Datastore<C>>,
+        lease: Lease<AcquiredCollectJob>,
+    ) -> Result<(), Error> {
+        let collect_job_id = lease.leased().collect_job_id;
+        datastore
+            .run_tx(|tx| Box::pin(async move { tx.cancel_collect_job(collect_job_id).await }))
```
BranLwyd, line 219: I need to also release the lease in this transaction (missed this on the initial implementation, whoops).

divergentdave (approved):
LGTM. As a follow-up, it might be good to create a test with helper aggregator mocks that always fail, and running the driver enough times that the job should be cancelled. That would require either sleeping or solving our detached task problem, see #234, so we may want to hold off for now.

BranLwyd:
I agree, I filed #253 to track this work.

PR #250, opened 2022-06-16
Create janus_client crate

divergentdave:
This pulls the client out into its own crate, and adds a new test-only tracing subscriber setup function in it. There's still room to improve the client API, but I think this could close out #18.


janus_client/src/lib.rs
```diff
@@ -202,10 +200,39 @@ where
...
 
+#[cfg(test)]
+mod test_util {
+    use std::sync::Once;
+    use tracing_log::LogTracer;
+    use tracing_subscriber::{prelude::*, EnvFilter, Registry};
+
+    /// Install a tracing subscriber for use in unit tests. This should be called at the beginning
+    /// of any test that requires a tracing subscriber.
+    pub fn install_test_trace_subscriber() {
```
tgeoghegan, line 211: This feels very, very similar to `janus_server::trace::test_util::install_test_trace_subscriber`. Maybe this should be in the `test_util` crate in the `janus` workspace so that both `janus_server` and `janus_client` can use it?

janus_client/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+reqwest = { version = "0.11.4", default-features = false, features = ["rustls-tls"] }
+thiserror = "1.0"
+tokio = { version = "^1.19", features = ["full"] }
+url = "2.2.2"
+
+[dev-dependencies]
+assert_matches = "1"
+janus_test_util = { path = "../test_util" }
+mockito = "0.31.0"
+tracing = "0.1.34"
```
tgeoghegan, line 21: I'm surprised that we only use `tracing` in test configs. Naturally we shouldn't pull in `tracing-log` and `tracing-subscriber`, but the production version of `janus_client` ought to be emitting trace events that clients can choose to subscribe to. IMO we should at least put a `#[tracing::instrument(..., err)]` on `Client::upload`, but I suppose that's beyond the scope of what this PR wants to accomplish.

tgeoghegan (changes requested):

janus_client/src/lib.rs
divergentdave: It is similar in effect to `janus_server::trace::test_util::install_test_trace_subscriber`, but pared down, without the configurability that `install_trace_subscriber()` has via `TraceConfiguration`. I'll give that move a try -- this could remove one more `#[doc(hidden)]` item in `janus_server`, which is nice.

tgeoghegan: It may not be worth the complexity of further parameterizing `janus_server::trace::test_util::install_test_trace_subscriber`, since the composed types get really gnarly, and since we probably won't have to change these tracing subscribers much, such that the benefit of de-duplicating code may be marginal. Your call.

test_util/src/lib.rs
```diff
@@ -288,3 +290,25 @@ where
         }
     }
 }
+
+/// Install a tracing subscriber for use in unit tests. This should be called at the beginning
```
tgeoghegan, line 294: super-nit: this is not exclusively for _unit_ tests
```suggestion
/// Install a tracing subscriber for use in tests. This should be called at the beginning
```

tgeoghegan (approved):

Issue #150, opened 2022-05-10
Consider using PostgreSQL's `tsrange`


tgeoghegan:
PostgreSQL has [range types](https://www.postgresql.org/docs/current/rangetypes.html), which among other things allow columns to represent a range of times. Ranges come with handy operators like `<@` that let you test whether a range contains another range or if a range contains a value. We could use `tsrange` ("range of timestamp without time zone") in tables like `aggregate_share_jobs` in lieu of storing `batch_interval_start` and `batch_interval_duration` as `TIMESTAMP` and `BIGINT`, respectively, which would make it easier to write some queries and (I think/hope) let us create efficient indices on the ranges.

divergentdave:
This was completed in #255.

PR #255, opened 2022-06-17
Use `tsrange` to represent batch intervals

divergentdave:
This implements #150. Note that I had to implement `FromSql`/`ToSql` in the common crate, due to the orphan rule. I put all of that behind a feature flag, so other downstream uses don't have to pull in PostgreSQL-related dependencies if they don't need it. I set up one GiST index over both the task ID and batch interval for each table as a starting point, but I haven't analyzed query performance of course. I do like how this change simplified SQL queries, and the relevant predicates should be much more index-friendly than constructing `tsrange`s on the fly from table columns.


janus/src/message.rs
```diff
@@ -226,6 +232,118 @@ impl Display for Interval {
...
+                Err("Interval can only represent timestamp ranges that are closed at the start and open at the end".into())
+            }
+            Nonempty(Inclusive(Some(start_raw)), Exclusive(Some(end_raw))) => {
+                // These timestamps represent the number of microseconds before (if negative) or
+                // after (if positive) midnight, 1/1/2000.
+                let start_timestamp = timestamp_from_sql(start_raw)?;
+                let end_timestamp = timestamp_from_sql(end_raw)?;
+
+                // Convert to Unix timestamp, in seconds since midnight 1/1/1970.
+                let negative = start_timestamp < 0;
```
BranLwyd, line 271: This code is a little complex.

What about defining a const `Time` that is "the zero SQL timestamp" (i.e. midnight, Jan 1, 2000). Then convert the SQL timestamps into a `Duration` based on a new `from_microseconds` constructor. Then use `add` or `sub` to apply the duration to the base timestamp -- both of these already use a `checked_` function to avoid overflow.

(Unfortunately, since `Duration` is unsigned, we do have to have the additional complexity of `add` or `sub` vs just creating a negative `Duration` and using `add`.)

janus/src/message.rs
```diff
@@ -226,6 +232,118 @@ impl Display for Interval {
...
+        _: &postgres_types::Type,
+        out: &mut bytes::BytesMut,
+    ) -> Result<postgres_types::IsNull, Box<dyn std::error::Error + Sync + Send>> {
+        let start_unix_timestamp_secs = i64::try_from(self.start().as_seconds_since_epoch())
+            .map_err(|_| "Interval start is out of range")?;
+        let duration_secs = i64::try_from(self.duration().as_seconds())
+            .map_err(|_| "Interval duration is out of range")?;
+
+        // Convert from the 2000 epoch to the 1970 epoch, and from seconds to microseconds.
+        let start_sql_usec = start_unix_timestamp_secs
```
BranLwyd, line 310: Similarly to above, if we had a const "zero SQL timestamp" `Time` value, we could just use `Time::difference` here to simplify the convert-epoch logic. A new `Duration::as_microseconds` would simplify the seconds-to-usec logic.

BranLwyd (changes requested):

BranLwyd (approved):

Issue #253, opened 2022-06-16
Job driver tests: test that repeated failure leads to cancellation


BranLwyd:
In #249, @divergentdave writes:
> As a follow-up, it might be good to create a test with helper aggregator mocks that always fail, and running the driver enough times that the job should be cancelled. That would require either sleeping or solving our detached task problem, see https://github.com/divviup/janus/issues/234, so we may want to hold off for now.

I agree, this would be a good test case to write.

PR #260, opened 2022-06-22
Monitor completion of some async tasks in tests

divergentdave:
This adds a new trait that abstracts over `tokio::task::spawn()`, and adds a test-only implementor that monitors when spawned async tasks complete. Tests can create multiple such `TestRuntime`s with different labels, and wait for some number of tasks to be completed before continuing. This addresses one of the two reasons we needed sleeps in tests for. Those tests are rewritten with this new functionality.

`TestRuntime` provides one additional feature: if a task spawned through it panics, then this will be detected, and the test's main task will panic when a related object drops.

This unblocks #253 and partially addresses #234.

### DIFF
--- a/test_util/src/runtime.rs
+++ b/test_util/src/runtime.rs
@@ -1,0 +1,204 @@
+use futures::FutureExt;
+use janus::Runtime;
+use std::{
+    collections::HashMap,
+    future::Future,
+    hash::Hash,
+    panic::{panic_any, AssertUnwindSafe},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+};
+use tokio::{
+    sync::watch::{self, Sender},
+    task::JoinHandle,
+};
+
+/// Tracks multiple instrumented [`Runtime`] objects, for use in tests. Each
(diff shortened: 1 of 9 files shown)


test_util/src/runtime.rs
```diff
@@ -0,0 +1,209 @@
...
+        self.map.entry(label).or_default().clone()
+    }
+
+    /// Wait for the runtime with the given label to reach some number of
+    /// completed tasks. Note that tasks that have already completed before
+    /// this method is called are included in the count.
+    pub async fn wait_for_completed_tasks(&mut self, label: L, target_count: usize) {
+        let labeled_runtime = self.with_label(label);
+        let mut receiver = labeled_runtime.inner.sender.subscribe();
+        if *receiver.borrow_and_update() >= target_count {
```
BranLwyd, line 50: nit: If I'm not missing something, this [L50-61] could be slightly simplified:

```rust
loop {
  if *receiver.borrow_and_update() >= target_count { break; }
  receiver.changed.await.expect("...");
}
```

(this may mark values as seen twice, but I think the code simplification is worth it)

PR #262, opened 2022-06-23
Move aggregation job creator/driver logic to the library target

divergentdave:
This moves the bulk of two `src/bin` files into the library target. Moving these items makes them available to `monolithic_integration_test` and opens the possibility of a single binary all-in-one aggregator for testing purposes.


BranLwyd (approved):
I was investigating launching the binaries to avoid this refactor, but since you've done it, I'll switch over to this. Thanks! :)

Issue #225, opened 2022-06-09
Support VDAFs with an aggregation parameter that isn't `()`


tgeoghegan:
Once we have a Poplar1 specification (https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/84) and implementation (https://github.com/divviup/libprio-rs/issues/141), we need `janus` to support it. Of course, a DAP implementation should be generic over VDAFs, so we shouldn't need any Poplar1 specific code, but what we will need to do is teach Janus to deal with VDAFs that actually have an aggregation parameter, where aggregation jobs cannot be created or driven until the collector hits `[leader]/collect`.

cjpatton:
You could probably already do Poplar with `ToyIdpf`. Should be enough to exercise this, though you'd have to keep the input length to a few bits.

BranLwyd:
As part of this, we will probably want to change the collect acquire operation since it races with the _creation_ of aggregation jobs--it should also check that there are no reports not yet included in an aggregation job over the relevant time period. This is included in this issue since current VDAFs (which use `()` as an aggregation parameter) support incremental aggregation, and thus are much less affected.

tgeoghegan:
#1508 deleted some test-only code that supported the non-trivial-aggregation-parameter case, which we might want to look at when we get around to reintroducing that support,

tgeoghegan:
Note from review on #3003 ([comment](https://github.com/divviup/janus/pull/3003/files#r1563158872)): in the case of the trivial aggregation parameter (Prio3), we eagerly scrub reports (i.e., discard input shares and public share) once they are prepared. We can't do that quite as easily when running Poplar1 or Mastic because we might need to aggregate the report again. However, if the aggregators notice they're at the last level of the tree, then they can scrub. We might need some libprio-rs plumbing so that aggregators can determine whether an aggregation parameter is for the leaves.

branlwyd:
Revisiting this: we need to implement, at least, creation of aggregation jobs w/ an aggregation parameter. There is currently test-only code which handles this for time-interval only.

Now that multi-collection of batches is gone, this may be all that is needed. But we should test to make sure that this is the _only_ place that needs to be updated.

Issue #224, opened 2022-06-09
Performance and load testing


tgeoghegan:
Devise and execute a plan for measuring and analyzing performance of `janus_server`. This will require a lot of instrumentation so we can get detailed breakdowns of where we spend our time while preparing or aggregating reports. We will also have to devise means of generating adequate volumes of representative data.

BranLwyd:
Note: I'm going through and converting TODOs to issues. I'm leaving the TODOs, but now with references to the underlying issues. Many performance-related TODOs now point at this issue.

branlwyd:
One tactical performance-improvement idea from @tgeoghegan: introduce bulk database operations which read/write multiple rows in a single query, to replace locations where we currently pipeline multiple individual requests.

We chose to use pipelined-individual-requests because we believe the performance/resource difference is marginal, and writing single-row operations is simpler than writing both single-row & bulk operations. I think this is still true: we should collect some evidence that pipelined operations are either unperformant or overly-expensive in resources before we spend engineering time making a change here.

tgeoghegan:
I don't think this issue is tracking any specific work anymore. There's different experiments and load tests we might consider but I think they are better tracked in specific issues.

PR #267, opened 2022-06-27
Add logic/DB query/tests for future Poplar support

divergentdave:
This adds datastore and aggregation job creator methods to support VDAFs with non-unit type aggregation parameters. For now, these are test-only methods. When full Poplar support is added in the future, this can be wired into existing `AggregationJobCreator` code that dispatches upon `VdafInstance`.

Aggregation job creation is likely the place where Janus's behavior will differ the most between Prio and Poplar, because the aggregation job creator must react to new collect jobs, rather than new client reports. Additionally, a report may be placed in more than one aggregate job, so that preparation can be run with multiple aggregation parameters.

This implements part of #225. My intent is to de-risk Poplar in Janus by getting a working query down for this part.

### DIFF
--- a/janus_server/src/aggregator/aggregation_job_creator.rs
+++ b/janus_server/src/aggregator/aggregation_job_creator.rs
@@ -288,11 +288,12 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
     // This is only used in tests thus far.
     #[cfg(test)]
     #[tracing::instrument(skip(self), err)]
-    async fn create_aggregation_jobs_for_task_with_param<const L: usize, A: vdaf::Aggregator<L>>(
+    async fn create_aggregation_jobs_for_task_with_param<const L: usize, A>(
         &self,
         task: &Task,
     ) -> anyhow::Result<()>
     where
+        A: vdaf::Aggregator<L>,
         for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
         A::PrepareMessage: Send + Sync,
         A::PrepareState: Send + Sync + Encode,
--- a/janus_server/src/aggregator/aggregation_job_creator.rs
+++ b/janus_server/src/aggregator/aggregation_job_creator.rs
@@ -17,6 +17,8 @@ use prio::vdaf;
 use prio::vdaf::prio3::{Prio3Aes128Count, Prio3Aes128Histogram, Prio3Aes128Sum};
 use rand::{thread_rng, Rng};
 use std::collections::HashMap;
+#[cfg(test)]
+use std::hash::Hash;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::select;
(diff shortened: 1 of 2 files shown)


janus_server/src/datastore.rs
```diff
@@ -2914,7 +2972,7 @@ mod tests {
                     Prio3Aes128Count,
                 > {
                     aggregation_job_id,
-                    task_id: unrelated_task_id,
+                    task_id,
```
tgeoghegan, line 2975: Is this change deliberate? I'm surprised that it doesn't cause the test to start failing since `get_unaggregated_client_report_nonces_for_task` is unchanged.

janus_server/src/datastore.rs
```diff
@@ -615,6 +615,63 @@ impl<C: Clock> Transaction<'_, C> {
...
+                    ON aggregation_jobs.id = report_aggregations.aggregation_job_id
+                    WHERE aggregation_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                ) AS report_aggs
+                ON report_aggs.client_report_id = client_reports.id
+                AND report_aggs.aggregation_param = collect_jobs.aggregation_param
+                WHERE client_reports.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                AND collect_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                AND collect_jobs.state = 'START'
+                AND report_aggs.id IS NULL
+                ORDER BY nonce_time DESC LIMIT 5000",
```
tgeoghegan, line 656: `LIMIT 5000` isn't going to work forever. We have a similar clause on the query `get_unaggregated_client_report_nonces_for_task` that references #224 though I admit I don't quite see how this relates to performance and load testing (I guess it's under load that looking at only the last 5,000 results would fail). Anyway let's make sure we've made a note somewhere that this needs to be revisited.

janus_server/src/aggregator/aggregation_job_creator.rs
```diff
@@ -279,6 +281,101 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
...
+    /// be used with VDAFs that have non-unit type aggregation parameters.
+    // This is only used in tests thus far.
+    #[cfg(test)]
+    #[tracing::instrument(skip(self), err)]
+    async fn create_aggregation_jobs_for_task_with_param<const L: usize, A: vdaf::Aggregator<L>>(
+        &self,
+        task: &Task,
+    ) -> anyhow::Result<()>
+    where
+        for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
```
tgeoghegan, line 296: nit:
```suggestion
    async fn create_aggregation_jobs_for_task_with_param<const L: usize, A>(
        &self,
        task: &Task,
    ) -> anyhow::Result<()>
    where
        A: vdaf::Aggregator<L>,
        for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
```

Having said that, what you have is consistent with some other methods in the file, so clearly we don't have an established or enforced style here. Meaning you should do whatever you like best.

janus_server/src/aggregator/aggregation_job_creator.rs
```diff
@@ -279,6 +281,101 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
...
+        A::PrepareMessage: Send + Sync,
+        A::PrepareState: Send + Sync + Encode,
+        A::OutputShare: Send + Sync,
+        for<'a> &'a A::OutputShare: Into<Vec<u8>>,
+        A::AggregationParam: Send + Sync + Eq + Hash,
```
tgeoghegan, line 301: Dubious idea: `create_aggregation_jobs_for_task_no_param` has a bound `A: vdaf::Aggregator<L, AggregationParam = ()>` ensuring it can only be used where the aggregation parameter is `()`. It'd be cool if we could ensure that `create_aggregation_jobs_for_task_with_param` is only used where `A::AggregationParam != ()`. Now, that's not possible out of the box in Rust, but what if we defined some new marker trait like `VdafWithAggregationParameter` and then "implemented" that on `prio::vdaf::Poplar1` and our test VDAF types? Then I think the compiler could ensure we don't accidentally use this path.

janus_server/src/datastore.rs
```diff
@@ -615,6 +615,63 @@ impl<C: Clock> Transaction<'_, C> {
             .collect::<Result<Vec<Nonce>, Error>>()
     }
 
+    /// get_unaggregated_client_report_nonces_by_collect_for_task returns pairs of nonces and
+    /// aggregation parameters, corresponding to client reports that have not yet been aggregated,
+    /// or not aggregated with a certain aggregation parameter, and for which there are collect
+    /// jobs, for a given task.
+    #[cfg(test)]
+    #[tracing::instrument(skip(self), err)]
+    pub async fn get_unaggregated_client_report_nonces_by_collect_for_task<
```
tgeoghegan, line 624: I think the doccomment here and on `get_unaggregated_client_report_nonces_for_task` should be clarified to explain that one should only be used for VDAFs with no aggregation parameter.

tgeoghegan, line 624: I don't love this method name. My attempt at doing better is `get_unaggregated_client_report_nonces_with_pending_collect_jobs_for_task`. However I think this points at a bigger nomenclature problem that we have, which is that we haven't spelled out the whole state machine for reports, aggregation jobs, collect jobs, and aggregate shares, so "pending" is kind of ambiguous there. We should probably stick with what you've got until we spell out a better state machine taxonomy and then revisit this.

tgeoghegan (changes requested):

janus_server/src/aggregator/aggregation_job_creator.rs
```diff
@@ -279,6 +289,103 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
...
+    // This is only used in tests thus far.
+    #[cfg(test)]
+    #[tracing::instrument(skip(self), err)]
+    async fn create_aggregation_jobs_for_task_with_param<const L: usize, A>(
+        &self,
+        task: &Task,
+    ) -> anyhow::Result<()>
+    where
+        A: vdaf::Aggregator<L>,
+        A: VdafHasAggregationParameter,
```
tgeoghegan, line 305: nit:
```suggestion
        A: vdaf::Aggregator<L> + VdafHasAggregationParameter,
```

janus_server/src/datastore.rs
```diff
@@ -615,6 +620,67 @@ impl<C: Clock> Transaction<'_, C> {
...
+    /// `get_unaggregated_client_report_nonces_for_task` should be used instead. In such cases, it
+    /// is not necessary to wait for a collect job to arrive before preparing reports.
+    #[cfg(test)]
+    #[tracing::instrument(skip(self), err)]
+    pub async fn get_unaggregated_client_report_nonces_by_collect_for_task<const L: usize, A>(
+        &self,
+        task_id: TaskId,
+    ) -> Result<Vec<(Nonce, A::AggregationParam)>, Error>
+    where
+        A: vdaf::Aggregator<L>,
```
tgeoghegan, line 639:
```suggestion
        A: vdaf::Aggregator<L> + VdafHasAggregationParameter,
```

Now that we have the marker trait, could we use it here (and `()` in `get_unaggregated_client_report_nonces_for_task`) for a bit more safety?

tgeoghegan (approved):

janus_server/src/datastore.rs
divergentdave: `get_unaggregated_client_report_nonces_for_task()` does not have any type parameters currently, so I'd be a bit loath to introduce one just for the constraint.

PR #268, opened 2022-06-27
test_util: spawn an interactive DB terminal

divergentdave:
This adds a new method to `janus_test_util::DbHandle` that spawns a new `gnome-terminal` window, to allow interacting with the database in the middle of a test. This method is intended as a debugging convenience only.

### DIFF
--- a/test_util/src/lib.rs
+++ b/test_util/src/lib.rs
@@ -35,16 +35,37 @@ macro_rules! define_ephemeral_datastore {
         pub struct DbHandle {
             _db_container: ::testcontainers::Container<'static, ::testcontainers::images::postgres::Postgres>,
             connection_string: String,
+            port_number: u16,
             datastore_key_bytes: Vec<u8>,
         }
 
         impl DbHandle {
+            /// Get a PostgreSQL connection string to connect to the temporary database.
             pub fn connection_string(&self) -> &str {
                 &self.connection_string
             }
 
             pub fn datastore_key_bytes(&self) -> &[u8] {
                 &self.datastore_key_bytes
+            }
+
+            /// Get the port number that the temporary database is exposed on, via the 127.0.0.1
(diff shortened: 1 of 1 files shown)


test_util/src/lib.rs
```diff
@@ -35,17 +35,38 @@ macro_rules! define_ephemeral_datastore {
...
+            pub fn port_number(&self) -> u16 {
+                self.port_number
+            }
+
+            /// Open an interactive terminal to the database in a new gnome-terminal window, and
+            /// block until the user exits from the terminal. This is intended to be used while
+            /// debugging tests.
+            pub fn interactive_db_terminal(&self) {
+                std::process::Command::new("gnome-terminal")
+                    .args(["--wait", "--", "psql", "--host=127.0.0.1", "--user=postgres", "-p"])
```
tgeoghegan, line 63: This only works for people using Gnome and its Terminal app. How about doing this by default, but consulting an environment variable like `JANUS_SHELL_CMD` so that users on macOS can do something else?

tgeoghegan (changes requested):

test_util/src/lib.rs
divergentdave: Done. From a brief look into Terminal.app, I couldn't find a way to launch it without the launching process kicking off the terminal in another process and exiting immediately. Hopefully one of the third-party terminals can slot in here. If worse comes to worse, xterm from XQuartz should work.

tgeoghegan: Yeah, all I could find on StackOverflow and the like was using `osascript(1)` to run an AppleScript that tells Terminal to do something, but if we make this an environment variable, then at least others can figure out how to do this. Plus it's good for Linuxes and Unixes that don't happen to run Gnome.

tgeoghegan (approved):

PR #270, opened 2022-06-29
Leader aggregation: update batch_unit_aggregations.

BranLwyd:
This will allow the collect protocol to complete successfully, since it
relies on the batch_unit_aggregations table being populated.

### DIFF
--- a/janus_server/src/aggregator/accumulator.rs
+++ b/janus_server/src/aggregator/accumulator.rs
@@ -80,8 +80,8 @@ where
             .time()
             .to_batch_unit_interval_start(self.min_batch_duration)
             .map_err(|e| datastore::Error::User(e.into()))?;
-        if let Some(accumulate) = self.accumulations.get_mut(&key) {
-            accumulate
+        if let Some(accumulation) = self.accumulations.get_mut(&key) {
+            accumulation
                 .update(output_share, nonce)
                 .map_err(|e| datastore::Error::User(e.into()))?;
         } else {
@@ -103,11 +103,11 @@ where
     /// aggregation exists, one is created and initialized with the accumulated values.
     #[tracing::instrument(skip(self, tx), err)]
     pub(super) async fn flush_to_datastore<C: Clock>(
-        self,
+        &self,
         tx: &Transaction<'_, C>,
     ) -> Result<(), datastore::Error> {
-        for (unit_interval_start, accumulate) in self.accumulations {
-            let unit_interval = Interval::new(unit_interval_start, self.min_batch_duration)?;
+        for (unit_interval_start, accumulation) in &self.accumulations {
+            let unit_interval = Interval::new(*unit_interval_start, self.min_batch_duration)?;
 
             let mut batch_unit_aggregations = tx
                 .get_batch_unit_aggregations_for_task_in_interval::<L, A>(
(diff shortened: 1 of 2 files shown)


janus_server/src/aggregator/aggregation_job_driver.rs
```diff
@@ -1324,14 +1376,19 @@ mod tests {
                         )
                         .await?
                         .unwrap();
-                    Ok((aggregation_job, report_aggregation))
+                    let batch_unit_aggregations = tx
+                        .get_batch_unit_aggregations_for_task_in_interval::<PRIO3_AES128_VERIFY_KEY_LENGTH, Prio3Aes128Count>(task_id, Interval::new(nonce.time().to_batch_unit_interval_start(task.min_batch_duration).unwrap(), task.min_batch_duration).unwrap(), &())
```
tgeoghegan, line 1380: I'm surprised `rustfmt` doesn't wrap this line. Could you please try to confine it to 100 columns or so?

janus_server/src/aggregator/accumulator.rs
```diff
@@ -103,11 +103,11 @@ where
     /// aggregation exists, one is created and initialized with the accumulated values.
     #[tracing::instrument(skip(self, tx), err)]
     pub(super) async fn flush_to_datastore<C: Clock>(
-        self,
+        &self,
```
tgeoghegan, line 106: I think I deliberately made this method consume `self` so that you can't continue accumulating into a flushed accumulator. Does making it take a reference make something else more ergonomic?

tgeoghegan (approved):

janus_server/src/aggregator/aggregation_job_driver.rs
BranLwyd: Yep, fixed. I line-broke some macro invocations too. (I'm not sure why it didn't apply TBH, it skips macro invocations sometimes but I've never seen it skip a non-macro function call...)

divergentdave: rustfmt is significantly more lax inside macro invocations, as it is not aware of that macro's DSL. For example, if it wraps an `assert_eq!()` onto multiple lines, the whitespace difference is safe, as it's no different at the token level, but rustfmt doesn't know whether the macro would be tolerant of adding a trailing comma to the last argument, now that it's on its own line.

I have seen rustfmt fail to act on normal code in certain places (somewhere in datastore.rs, I think) but I don't have any explanation for why. It could be we've reached some intentional nesting/complexity limit, or it could be a bug possibly. If so, minimizing a test case would be a bear.

janus_server/src/aggregator/accumulator.rs
BranLwyd: It's required for use in the leader aggregation implementation: since the computations around what to write (i.e. `update` calls) happen outside a transaction, but the writes (i.e. `flush_to_datastore`) happen inside a transaction, and transactions can be forced to retry due to serialization issues, the `flush_to_datastore` operation can't consume the accumulator since we might need to try writing again.

(We might need this in the helper aggregation implementation eventually too, if/when we break all of the aggregation computations out of the datastore transaction.)

We could replace this with a runtime check if we wanted to be extra-careful, but I am OK with code inspection + tests here to make sure we're doing the right thing.

tgeoghegan: Sounds reasonable.

Issue #222, opened 2022-06-09
Build and publish artifacts from `janus`


tgeoghegan:
Janus' automated release action should build artifacts and publish them to durable artifact stores. These include:

Ship to Google Artifact Registry (GAR) container registry:
- `aggregator` container
- `aggregate_job_creator` container
- `aggregate_job_driver` container
- `collect_job_driver` container

GAR blob store:
- `janus_server` database SQL schema

Ship to crates.io:
- Divvi Up Rust client crate

tgeoghegan:
Nice to have: compile docs over `janus_server` and put them somewhere (GitHub Pages?)

divergentdave:
FYI, it appears that GAR only supports a limited number of formats (Docker, Helm, Maven, NPM, Python packages, apt, and .rpms), so we'll need to do something else with our SQL schema to keep versioned copies.

tgeoghegan:
That's disappointing. I had hoped it would support opaque files. Well, maybe we can just upload the .sql files to a GCS bucket and interpolate a version number into the object key?

tgeoghegan:
Lots of good progress here last week. We have container images going out to GAR from `divviup/janus` now. Next step is to publish some Rust crates from that same project. That's nearly done in #280, but I'm going to have to pick that up after the long weekend. The outstanding things are to make sure the `janus -> janus_core` rename worked, that the `janus_client -> janus_core` dependency is properly handled both in local development and when publishing to crates.io and that crates get published as we expect.

tgeoghegan:
[`janus_core`](https://crates.io/crates/janus_core) and [`janus_client`](https://crates.io/crates/janus_client) are now published to crates.io. Last thing to solve here is publishing the SQL schema files.

tgeoghegan:
We now have containers going to our private GAR repository, crates going to crates.io, and finally SQL schema files going to a GCS bucket. This is done!

PR #280, opened 2022-07-01
Publish `janus_client` to crates.io

tgeoghegan:
Adds an action that runs on release that publishes the `janus_client`
crate.

Part of #222


divergentdave (commented):
We will need to publish both `janus` and `janus_client`, in that order, to satisfy dependency requirements. Also, a dry run reports that we need to update Cargo.toml path dependencies to include version numbers as well, so that those dependencies can be resolved into crates.io versions.

janus_client/Cargo.toml
```diff
@@ -7,7 +7,7 @@ rust-version = "1.58"
 
 [dependencies]
 http = "0.2.8"
-janus = { path = "../janus" }
+janus = "0.1.0"
```
divergentdave, line 10: The error message I got says the following:
```
$ cargo publish -p janus_client --dry-run
    Updating crates.io index
error: all dependencies must have a version specified when publishing.
dependency `janus` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```

We can thus use something like the following:

```suggestion
janus = { version = "0.1.0", path = "../janus" }
```

During normal development, Cargo will use the path dependency (or error if the version doesn't match, I think) and during publishing, it will rewrite the Cargo.toml file to refer to `crates.io` alone. (It looks like it also does a check build with `crates.io` dependencies, as I get errors from it trying to use the unrelated existing janus crate)

divergentdave (approved):

Issue #221, opened 2022-06-09
Verify that janus implements draft-ietf-ppm-dap-01


tgeoghegan:
Once draft-ietf-ppm-dap-01 is published, we should make sure that `janus_server` implements it. I believe this mostly consists of making sure we have the right HTTP problem document type strings and other naming things. We also should fix references in comments to draft sections, since those mostly refer to the sections from the interop branch of the DAP draft.

BranLwyd:
- [ ] We don't currently reject reports that fall into batch intervals that have already been collected.

tgeoghegan:
We've been doing interop testing with other draft-01 implementations for a little bit and we're not aware of other specific issues related to draft-01, so I'm going to close this.

PR #281, opened 2022-07-01
Return batch-collected & report-replayed from helper aggregation.

BranLwyd:
The `batch-collected` error is returned when aggregation is attempted
for a report that falls into a batch unit that has already been
collected. The `report-replayed` error occurs when aggregation is
attempted for a report that has already been aggregated.

Also, a few small finished-TODO & naming cleanups.

A few spec-related notes, I will drive the spec changes if these make
sense to reviewers:
  * The DAP specification language says to return `report-replayed` "if
    the report has already been aggregated." However, some VDAFs (e.g.
    poplar1) require aggregating the same report multiple times with
    varying aggregation parameters. I think the DAP specification should
    be updated to return `report-replayed` if the report has already
    been aggregated _with the same aggregation parameter_.
  * The DAP spec says that `report-dropped` may be returned under
    conditions prescribed in Section 4.4.6; however, the language of
    that section does not make it clear how this error code is distinct
    from `report-replayed`. I think the idea is that `report-dropped`
    would be returned if the aggregator has forgotten about reports in
    the appropriate batch window, but the spec language doesn't make
    this clear.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -89,6 +89,9 @@ pub enum Error {
     /// Corresponds to `reportTooLate`, §3.1
     #[error("stale report: {0} {1:?}")]
     ReportTooLate(Nonce, TaskId),
+    /// Corresponds to `reportTooEarly`, §3.1. A report was rejected becuase the timestamp is too far in the future, §4.3.4.
+    #[error("report from the future: {0} {1:?}")]
+    ReportTooEarly(Nonce, TaskId),
     /// Corresponds to `unrecognizedMessage`, §3.1
     #[error("unrecognized message: {0} {1:?}")]
     UnrecognizedMessage(&'static str, Option<TaskId>),
@@ -104,9 +107,6 @@ pub enum Error {
     /// Corresponds to `outdatedHpkeConfig`, §3.1
     #[error("outdated HPKE config: {0} {1:?}")]
     OutdatedHpkeConfig(HpkeConfigId, TaskId),
-    /// A report was rejected becuase the timestamp is too far in the future, §4.3.4.
-    #[error("report from the future: {0} {1:?}")]
-    ReportFromTheFuture(Nonce, TaskId),
     /// Corresponds to `unauthorizedRequest`, §3.1
     #[error("unauthorized request: {0:?}")]
     UnauthorizedRequest(TaskId),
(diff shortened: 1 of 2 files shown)


tgeoghegan (approved):
>I think the DAP specification should
>be updated to return report-replayed if the report has already
>been aggregated with the same aggregation parameter.

I think this has been discussed before, but I can't find an issue on DAP or anything to cite. Anyway, I think the concern is: suppose I send a collect request over some interval containing reports 1..100 with aggregation param `a_1`. Then, I wait for my victim to send report 101, falling into that same interval, and then send a collect request again with aggregation param `a_2`. Do VDAFs guarantee that there is no choice of `a_1, `a_2` such that the corresponding aggregation results reveal report 101?

This PR looks good given current protocol text though.

BranLwyd:
> > I think the DAP specification should
> > be updated to return report-replayed if the report has already
> > been aggregated with the same aggregation parameter.
> 
> I think this has been discussed before, but I can't find an issue on DAP or anything to cite. Anyway, I think the concern is: suppose I send a collect request over some interval containing reports 1..100 with aggregation param `a_1`. Then, I wait for my victim to send report 101, falling into that same interval, and then send a collect request again with aggregation param `a_2`. Do VDAFs guarantee that there is no choice of `a_1, a_2` such that the corresponding aggregation results reveal report 101?
> 
> This PR looks good given current protocol text though.

Hmm -- I think the existing language around `batch-collected` accounts for this case by not accepting any new reports in a batch unit that has started collection (and the argument you make shows why this is necessary to protect report privacy). The `batch-collected` wording also wouldn't block re-aggregation of an existing batch with a new aggregation parameter.

OTOH, I think `report-replayed` really does block re-aggregation of an existing batch with a new aggregation parameter; since we think such re-aggregation is necessary for `poplar1`, I think a change is necessary. I'll file an issue for these against the spec (making sure to call out the difference between this & `batch-collected`).

BranLwyd:
(FTR, I reported the first DAP spec issue as https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/283; the second DAP spec issue is close enough to the already-reported https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/180 to not be worth a second issue)

BranLwyd:
Belatedly, related to #221.

PR #288, opened 2022-07-05
Publish SQL schema to GCS bucket

tgeoghegan:
On release, we use `gsutil` to upload the SQL schema file to a GCS
bucket. We version the artifacts by uploading them under a "directory"
named after the git tag. This allows us to upload multiple SQL files in
the future and doesn't require the consumer of the SQL files to parse
out or otherwise ignore the version component.

Part of #222

### DIFF
--- a/.github/workflows/publish-sql-schema.yml
+++ b/.github/workflows/publish-sql-schema.yml
@@ -1,0 +1,34 @@
+name: publish-sql-schema
+
+on:
+  release:
+    types: [published]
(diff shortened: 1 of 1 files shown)


tgeoghegan:
We need https://github.com/divviup/janus-ops/pull/136 merged and deployed before this will work.

BranLwyd (approved):

divergentdave (approved):

PR #292, opened 2022-07-06
Bump Rust crates to 0.1.2

tgeoghegan:
Bump the `janus_core` and `janus_client` crate versions to 0.1.2 so that
we can cut a `janus` release and exercise the artifact publishing
workflows e2e.

Part of #222

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1276,7 +1276,7 @@ checksum = "1aab8fc367588b89dcee83ab0fd66b72b50b72fa1904d7095045ace2b0c81c35"
 
 [[package]]
 name = "janus_client"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
  "assert_matches",
  "http",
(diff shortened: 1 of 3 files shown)


divergentdave (changes requested):
The dependency in `janus_client/Cargo.toml` upon `janus_core` should have its version bumped as well. `cargo check` on the codebase as-is passes, and uses `janus_core v0.1.2` via the path dependency. `cargo publish -p janus_client --dry-run` uses `janus_core v0.1.1` from crates.io, and it runs successfully. Presumably, if there were a breaking change or new item in `janus_core` we needed, we would see an error in the compilation that `cargo publish` does.

tgeoghegan:
Maybe we should run `cargo publish --dry-run` in CI, though I'm not sure how to make that work if a PR makes `janus_client` depend on an unreleased `janus_core`. Also I don't understand how `Cargo.lock` affects publishing a library crate: the lockfile shouldn't be included in the version published to crates.io, right?

divergentdave:
Correct, on my machine, the unpacked crate at `~/.cargo/registry/src/github.com-1ecc6299db9ec823/janus_core-0.1.1` has no lockfile, just `Cargo.toml` and `Cargo.toml.orig`.

PR #295, opened 2022-07-06
Implement provision_task command.

### LINKED ISSUE #289: Tool to provision tasks for deployment testing
We should write a tool that can provision tasks for deployment testing scenarios (i.e. local-cluster tests, interop tests). This tool should probably be a Rust binary, as provisioning a task requires writing encrypted values to the datastore.

It would be very nice if this tool had at least rudimentary support for deploying the database schema. Properly handling versions is out-of-scope for this tool.

BranLwyd:
This command is intended to be run from a privileged operator
workstation. It will write some tasks (as defined by the binary config),
and optinally also attempt to write a schema (with no support for
versioning, etc).

Closes #289.


janus_server/src/bin/provision_task.rs
```diff
@@ -0,0 +1,203 @@
...
+    // Write all tasks requested.
+    if !ctx.config.tasks.is_empty() {
+        let tasks = Arc::new(ctx.config.tasks);
+        info!(task_count = tasks.len(), "Writing tasks");
+        ctx.datastore
+            .run_tx(|tx| {
+                let tasks = Arc::clone(&tasks);
+                Box::pin(async move {
+                    for task in tasks.iter() {
+                        tx.delete_task(task.id).await?;
```
divergentdave, line 38: I think this delete-then-put behavior could lead to confusing emergent behavior. For example, if this command is run with multiple tasks in its configuration, but one task already exists in the database and has a client report as well, then this whole transaction will roll back, and none of the other tasks will be created.

My preference would be that `delete_task()` should use `check_update()` on the number of rows deleted from the tasks table, and provision_task should just create tasks without deleting them first, perhaps skipping any task IDs that are already present.

test_util/src/lib.rs
```diff
@@ -42,6 +42,7 @@ macro_rules! define_ephemeral_datastore {
         }
 
         impl DbHandle {
+            // Retrieve a datastore attached to the ephemeral database.
```
divergentdave, line 45: nit: this and the comment before `pool()` should probably be doc comments.

divergentdave (approved):

PR #298, opened 2022-07-07
Round down timestamps in report nonces

divergentdave:
This implements the change in ietf-wg-ppm/draft-ietf-ppm-dap#281, rounding down timestamps of reports to the next multiple of `min_batch_duration`.

### DIFF
--- a/janus_core/src/message.rs
+++ b/janus_core/src/message.rs
@@ -372,12 +372,14 @@ impl Nonce {
         Nonce { time, rand }
     }
 
-    /// Generate a fresh nonce with the current time.
-    pub fn generate<C: Clock>(clock: &C) -> Nonce {
-        Nonce {
-            time: clock.now(),
+    /// Generate a fresh nonce based on the current time.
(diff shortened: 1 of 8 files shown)


janus_core/src/message.rs
```diff
@@ -372,12 +372,14 @@ impl Nonce {
         Nonce { time, rand }
     }
 
-    /// Generate a fresh nonce with the current time.
-    pub fn generate<C: Clock>(clock: &C) -> Nonce {
-        Nonce {
-            time: clock.now(),
+    /// Generate a fresh nonce based on the current time.
+    pub fn generate<C: Clock>(clock: &C, min_batch_duration: Duration) -> Result<Nonce, Error> {
```
BranLwyd, line 376: Could you also change the Nonce randomness to be 16 bytes (from 8, currently)? I think this would be as easy as changing L366 & updating the message-roundtrip tests, though I wouldn't be surprised if an assumption of Nonce size snuck into some higher-level tests as well.

BranLwyd (changes requested):

BranLwyd (approved):

tgeoghegan (approved):

PR #299, opened 2022-07-07
Remove define_ephemeral_datastore macro.

BranLwyd:
Instead, move the code inside the macro directly into the
datastore::test_util module, and update other uses of the macro to
refer to this version.

Enabled by #294, related to #141.


tgeoghegan (approved):
Nice! I was grappling with this when I did the PR to move `test_util` into `janus_core::test_util` but the trick I missed was not defining any of this as a macro anymore.

PR #306, opened 2022-07-11
Add metrics to job drivers & for partial failures in HTTP handlers.

### LINKED ISSUE #290: Augment metrics collection for interop testing
We should augment existing metrics -- most error or "non-happy-path" cases should likely increment a metric. Every current `info!`, `warn!`, `error!`, etc should very likely increment a metric.

BranLwyd:
Also, clean up some double-logging in the aggregator HTTP handlers, and tweak some log lines & error messages.

Closes #290.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -184,7 +184,7 @@ pub struct Aggregator<C: Clock> {
     /// Counter tracking the number of failed decryptions while handling the /upload endpoint.
     upload_decrypt_failure_counter: Counter<u64>,
     /// Counter tracking the number of failures to step client reports through the aggregation process.
-    aggregate_step_failure_counter: Arc<Counter<u64>>,
+    aggregate_step_failure_counter: Counter<u64>,
 }
 
 impl<C: Clock> Aggregator<C> {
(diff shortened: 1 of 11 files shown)


divergentdave (approved):

janus_server/src/aggregator.rs
```diff
@@ -404,7 +433,7 @@ impl TaskAggregator {
                     .clone()
                     .try_into()
                     .map_err(|_| Error::TaskParameters(task::Error::AggregatorAuthKeySize))?;
-                VdafOps::Prio3Aes128Count(vdaf, verify_key)
+                VdafOps::Prio3Aes128Count(Arc::new(vdaf), verify_key)
```
tgeoghegan, line 436: Is the `Arc` really necessary here? A VDAF implementation is basically just a bunch of function pointers with virtually no data, so just `.clone()`ing them (without an `Arc`) should be cheap enough. Plus [trait `Vdaf` already requires `Clone`](https://docs.rs/prio/latest/prio/vdaf/trait.Vdaf.html).

janus_server/src/aggregator.rs
```diff
@@ -835,10 +937,12 @@ impl VdafOps {
                 .hpke_keys
                 .get(&report_share.encrypted_input_share.config_id())
                 .ok_or_else(|| {
-                    warn!(
+                    error!(
```
tgeoghegan, line 940: Since this would result from bad input from a client rather than a mistake on our end, I think it ought to be no higher than `warn!`.

tgeoghegan:
Oh, whoops, looks like I'm late to the party. Fortunately I only had minor nits that don't need to block this change.

janus_server/src/aggregator.rs
BranLwyd: The justification is that since this VDAF will eventually be passed to code that is generic over the specific VDAF type, we don't (in general) know what a VDAF is. While that's true today, we don't know that some future VDAF doesn't contain a nontrivial amount of state. (precomputed data, perhaps?) It may not ever matter, but it's minor defensive programming IMO.

(Given the above argument, I'd suggest that `Vdaf` shouldn't require `Clone` since we shouldn't be cloning them since we can't guarantee they'll be small; but I don't care enough to try to make that change.)

PR #314, opened 2022-07-15
Provision task datastore key

BranLwyd:
kube/rustls issue: https://github.com/kube-rs/kube-rs/issues/153


janus_core/src/test_util/kubernetes.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// Creates & starts a new ephemeral Kubernetes cluster.
+    pub fn create() -> Self {
+        // Choose a cluster name.
+        let mut randomness = [0u8; 4];
+        thread_rng().fill(&mut randomness);
+        let cluster_name = format!("janus-ephemeral-{}", hex::encode(&randomness));
+
+        // Use kind to start the cluster.
+        assert!(Command::new("kind")
+            .args(["create", "cluster", "--name", &cluster_name])
```
divergentdave, line 23: I'd recommend that we use a separate temporary kubecfg file for this, rather than having test commands touch the default one. One big benefit is that this would avoid different commands trying to concurrently modify the file when running multiple tests. `kind` takes a `--kubeconfig` argument, as does `kubectl`.

Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM rust:1.62.0-alpine as builder
 ARG BINARY=aggregator
-RUN apk add libc-dev
+RUN apk add libc-dev openssl-dev
```
divergentdave, line 3: FYI, you'll need to set `RUSTFLAGS=-C target-feature=-crt-static` or flip on an `openssl-sys` feature for statically linking OpenSSL, otherwise you'll get issues at runtime. https://users.rust-lang.org/t/sigsegv-with-program-linked-against-openssl-in-an-alpine-container/52172/6

BranLwyd: Hmm, I didn't see any problems with `cargo test` or manually-running dev builds on my workstation (and release builds won't include OpenSSL at all). Maybe things have changed since that discussion?

divergentdave (approved):

janus_server/src/bin/janus_cli.rs
```diff
@@ -0,0 +1,307 @@
+use anyhow::{Context, Result};
```
tgeoghegan, line 1: I suggest using [`color-eyre`](https://docs.rs/color-eyre/latest/color_eyre/) so we get super pretty errors printed to stderr.

janus_server/src/bin/janus_cli.rs
```diff
@@ -0,0 +1,307 @@
...
+};
+use structopt::StructOpt;
+use tokio::fs;
+use tracing::info;
+
+static SCHEMA: &str = include_str!("../../../db/schema.sql");
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    janus_main::<_, Options, Config, _, _>(RealClock::default(), |ctx| async move {
```
tgeoghegan, line 29: We could tweak the trace subscriber config so that it does things better suited to a utility that will primarily run in operator shells instead of a Kubernetes cluster. For instance, most of its trace output should probably go to stderr, or maybe there's other nice formatters we could turn on. I'm happy to defer that to a future PR, so logging an issue is sufficient.

tgeoghegan (approved):
Looks fine, though I have a couple suggestions that can be punted to future enhancements. Last comment is that we should document the dependency on having `kind` in `$PATH` in a README.

janus_server/src/bin/janus_cli.rs
BranLwyd: Filed https://github.com/divviup/janus/issues/315.

janus_server/src/bin/janus_cli.rs
BranLwyd: Unfortunately, it looks like `color-eyre` requires use of its `color_eyre::Result` type to get the nicely-colored output. This is incompatible with `janus_main` right now since `janus_main` requires use of an `anyhow::Result`. It wouldn't be too hard of a refactor, but I've already decided to leave a larger `janus_main` refactoring for later -- I'll let this get picked up in #315.

BranLwyd:
Updated the README to note the new `kind` dependency for tests.

Issue #315, opened 2022-07-15
Refactor binary utilities to be more flexible for non-server use-cases


BranLwyd:
The new `janus_cli` tool is intended to be used as a command on an operator workstation, rather than as a server/service running deployed in a cloud somewhere.

Our current `binary_utils` module could be refactored to better-support this use case. For example:
* Log-line formatting could be changed to be displayed by default, and formatted in a way that a human operator will find easy to read.
* Color could be added, e.g. with the `color_eyre` crate. (right now the `janus_main` function requires returning an `anyhow::Result`, which isn't compatible with use of `color_eyre` since `color_eyre` has its own `Result` type)
* Even binaries that don't need a datastore get one (and, more importantly, must receive valid configuration for one). For `janus_cli`, some commands will need a datastore, and some won't.

I recommend factoring the `binary_utils` crate for increased flexibility:
* Move useful chunks of functionality (option/config parsing, logging configuration, datastore creation) out to their own separate functions. Generalize the ones that need different behavior for a CLI tool (or provide two versions).
* Drop `janus_main` entirely in favor of having binaries call the setup functions they need, when they need them. (alternatively, maybe call it `janus_server_main` since `janus_main`'s behavior is reasonable for all of our server components so far?)

BranLwyd:
I took a look at getting `color_eyre` working, and AFAICT there is not good support for converting from `anyhow::Error` to `color_eyre::eyre::Report` (`eyre`'s equivalent type). The root cause seems to be that `anyhow::Error` does not implement the stdlib's `std::error::Error` trait:

```
the trait bound `anyhow::Error: StdError` is not satisfied
the trait `StdError` is not implemented for `anyhow::Error`rustc[E0277](https://doc.rust-lang.org/error-index.html#E0277)
error.rs(22, 12): required by a bound in `eyre::error::<impl ErrReport>::new`
```

Since colored output is a nice-to-have, I'm not going to spend more effort on this for now.

Issue #36, opened 2022-03-24
Kubernetes compatible liveness/startup/readiness/health endpoint


tgeoghegan:
Aggregators should perform health checks at startup and report on an HTTP endpoint that can be used by (among other things) Kubernetes to determine if a running container is healthy.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

tgeoghegan:
#35 is one such check.

divergentdave:
We could run `SELECT COUNT(*) FROM tasks`, which would serve dual purposes of testing the database connection and confirming the database is provisioned with tasks.

divergentdave:
Should we check if `SELECT now()` is close to the local clock's time at startup as part of readiness checks? Clock skew could cause issues with leases, rejection of uploaded reports, and eager scheduling of aggregation jobs. If each worker checked its time against the database, we could hopefully catch glaring issues earlier.

If we don't want to gate on clock skew, perhaps processes should check their skew versus the database server periodically, and expose it as a gauge metric.

tgeoghegan:
#323 added endpoints to the `aggregator` component for startup and health probes. Some more improvements we could make:

- expose health checks on all the Janus components
- expose health checks on a different port than the rest of the API so that we can avoid exposing healthz endpoints to the Kubernetes service that is wired up to external ingress

BranLwyd:
> expose health checks on a different port than the rest of the API so that we can avoid exposing healthz endpoints to the Kubernetes service that is wired up to external ingress

Thinking about this a little more, we might be able to avoid exposing `/healthz` endpoints to the Internet if some higher-level routing component allows us to not route certain paths -- that would be another way to achieve the same goal.

divergentdave:
I looked into this a bit, and the preferred method appears to involve running a deployment of `k8s.gcr.io/defaultbackend-amd64`, and adding rules to the ingress that route `/healthz` etc. to it. The container combines a static file server with an empty directory, so that it returns 404s for every request.

I think I'd prefer moving `/healthz` to another port. It appears that our load balancer health checks also support selecting different ports, so this would be the cleanest solution.

tgeoghegan:
FWIW we already serve metrics scrapes from a separate port so one more port for healthz doesn't seem too bad to me.

tgeoghegan:
I think we have accomplished everything listed in this ticket, so I'm going to close it. Please re-open if there's still something valuable tracked here.

PR #322, opened 2022-07-19
Set timeouts in deadpool_postgres

tgeoghegan:
When deployed to GCP, Janus connects to its Postgres database through
the `cloud-sql-proxy`, which is deployed as a sidecar container to Janus
itself. Kubernetes provides no guarantees about the order in which it
will start the containers in a pod, which means that it's possible for
Janus components to attempt to connect to the proxy before it has
started and established a connection to the real database, which will
fail with "connection refused". This is a problem for the
`aggregation_job_creator`: it is configured to query the database once
an hour for defined tasks, so if it fails to do so at startup, it won't
retry for an hour. Other Janus components either query the DB in
response to API traffic or poll the database more often.

With this change, we now wait up to 60 seconds while trying to connect
to the database, which should help `aggregation_job_creator` get started
more reliably.

### DIFF
--- a/janus_server/src/binary_utils.rs
+++ b/janus_server/src/binary_utils.rs
@@ -9,7 +9,7 @@ use crate::{
     trace::{cleanup_trace_subscriber, install_trace_subscriber, OpenTelemetryTraceConfiguration},
 };
 use anyhow::{anyhow, Context, Result};
-use deadpool_postgres::{Manager, Pool};
+use deadpool_postgres::{Manager, Pool, Runtime, Timeouts};
 use janus_core::time::Clock;
 use ring::aead::{LessSafeKey, UnboundKey, AES_128_GCM};
 use std::{
@@ -18,6 +18,7 @@ use std::{
     future::Future,
     path::PathBuf,
     str::FromStr,
+    time::Duration,
 };
 use structopt::StructOpt;
 use tokio_postgres::NoTls;
@@ -52,8 +53,16 @@ pub fn datastore<C: Clock>(
         database_config.password(pass);
     }
 
+    let connection_pool_timeout = Duration::from_secs(db_config.connection_pool_timeouts_secs);
+
     let conn_mgr = Manager::new(database_config, NoTls);
     let pool = Pool::builder(conn_mgr)
+        .runtime(Runtime::Tokio1)
+        .timeouts(Timeouts {
+            wait: Some(connection_pool_timeout),
+            create: Some(connection_pool_timeout),
+            recycle: Some(connection_pool_timeout),
+        })
         .build()
         .context("failed to create database connection pool")?;
     let ds_keys = ds_keys
@@ -73,8 +82,14 @@ pub fn datastore<C: Clock>(
     if ds_keys.is_empty() {
         return Err(anyhow!("ds_keys is empty"));
     }
+
     Ok((
-        Datastore::new(pool.clone(), Crypter::new(ds_keys), clock),
+        Datastore::new(
+            pool.clone(),
+            connection_pool_timeout,
+            Crypter::new(ds_keys),
+            clock,
+        ),
         pool,
     ))
 }
@@ -223,7 +238,11 @@ where
         &common_options.database_password,
         &common_options.datastore_keys,
     )
-    .context("couldn't connect to database")?;
+    .context("couldn't create database connection pool")?;
+    datastore
+        .check_connection_pool()
+        .await
+        .context("couldn't connect to database")?;
 
     let logging_config = config.common_config().logging_config.clone();
 
--- a/janus_server/src/config.rs
+++ b/janus_server/src/config.rs
@@ -29,7 +29,18 @@ pub trait BinaryConfig: Debug + DeserializeOwned {
 pub struct DbConfig {
     /// URL at which to connect to the database.
     pub url: Url,
+    /// Timeout in seconds to apply when creating, waiting for, or recycling
+    /// connection pool objects. This value will be used to construct a
+    /// `deadpool_postgres::Timeouts` value.
+    #[serde(default = "DbConfig::default_connection_pool_timeout")]
+    pub connection_pool_timeouts_secs: u64,
     // TODO(#231): add option for connecting to database over TLS, if necessary
+}
+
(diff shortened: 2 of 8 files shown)


janus_server/src/binary_utils.rs
```diff
@@ -53,7 +54,13 @@ pub fn datastore<C: Clock>(
     }
 
     let conn_mgr = Manager::new(database_config, NoTls);
+    let timeout = Some(Duration::from_secs(db_config.connection_pool_timeouts_secs));
     let pool = Pool::builder(conn_mgr)
+        .timeouts(Timeouts {
```
divergentdave, line 59: FYI we will also need to call `.runtime(Runtime::Tokio1)` on the builder. Currently this fails with the following error.

```
Error: couldn't connect to database

Caused by:
    0: failed to create database connection pool
    1: Error occurred while building the pool: NoRuntimeSpecified: Timeouts require a runtime
```

tgeoghegan:
In any case, it turns out that neither `deadpool_postgres` or `tokio-postgres` will retry connecting to the DB on `ECONNREFUSED`, so the timeout doesn't help. I suspect it's probably a good idea to set a non-zero timeout, but we'll also want to wrap our call to `pool.get()` in `Datastore::run_tx_once()` in a retry loop.

janus_server/src/datastore.rs
```diff
@@ -109,8 +111,26 @@ impl<C: Clock> Datastore<C> {
...
+            multiplier: 2.0,
+            max_elapsed_time: Some(StdDuration::from_secs(600)),
+            ..Default::default()
+        };
+
+        // Open transaction, retrying if we encounter timeouts or connection refused errors (which
+        // occur if the Cloud SQL Proxy hasn't started yet and manifest as `PoolError::Backend`).
+        let mut client = retry(backoff, || async {
+            self.pool.get().await.map_err(|error| match error {
+                PoolError::Timeout(_) | PoolError::Backend(_) => {
```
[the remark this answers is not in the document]
tgeoghegan: Unfortunately, the error we get for `ECONNREFUSED` doesn't materialize as a `DbError`. I believe it'll be an [`ErrorInner`](https://docs.rs/tokio-postgres/0.7.6/src/tokio_postgres/error/mod.rs.html#366) with `kind: Kind::Connect`, but neither the `kind` field nor the `Kind` enum are public, so we can't match on the error type. I think I'll do as you suggest, and do a startup check instead of putting retries on every single call to `self.pool.get()`.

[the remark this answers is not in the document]
tgeoghegan: This is tricky. The error we get if the connection is refused because the sidecar isn't up yet looks like:
`Backend(Error { kind: Connect, cause: Some(Os { code: 111, kind: ConnectionRefused, message: \"Connection refused\" }) })`

The error we get if we provide an unroutable DB address (as the test `aggregator::health_check` does) is:
`Backend(Error { kind: Connect, cause: Some(Custom { kind: Uncategorized, error: "failed to lookup address information: Name or service not known" }) })`

We can't distinguish them programmatically, because the `kind` and `cause` fields are not public. The best I could come up with was to have test configurations give up retrying after 1 second, but to retry for up to a minute otherwise.

janus_server/src/binary_utils.rs
```diff
@@ -223,7 +238,11 @@ where
         &common_options.database_password,
         &common_options.datastore_keys,
     )
-    .context("couldn't connect to database")?;
+    .context("couldn't create database connection pool")?;
+    datastore
+        .check_connection_pool()
+        .await
+        .context("couldn't connect to database")?;
```
tgeoghegan, line 245: Now, all binaries will try to connect to the database at startup. This arguably makes the call to `Datastore::check_connection_pool` in the aggregator's `healthz/startup` endpoint redundant (it couldn't get so far as to serve that request if this call to `Datastore::check_connection_pool` hadn't already succeeded), so perhaps we can delete the `startup` endpoint.

BranLwyd: `janus_cli` doesn't want this since not all commands want a datastore connection at all, but we already plan to change `janus_cli` to not use the "standard server binary startup" codepath in #315, so I think it's OK to add this check for now.

I do think we should expose health/startup endpoints from all server components. Maybe these "z-page" (internal status) pages could be factored out to a common module, and maybe run on a different port (to ensure that they are never accessible from outside the cluster)? I think with the way that `/healthz` is currently implemented on the aggregator, once we're launched, anyone on the Internet could hit `foo.api.divviup.org/healthz` and see our internal healthz response. While that may not indicate much of a practical problem, I suspect we're going to add more status pages as we go, and some of those might contain information we would prefer not to be public.

janus_server/src/config.rs
```diff
@@ -29,6 +29,10 @@ pub trait BinaryConfig: Debug + DeserializeOwned {
 pub struct DbConfig {
     /// URL at which to connect to the database.
     pub url: Url,
+    /// Timeout in seconds to apply when creating, waiting for, or recycling
+    /// connection pool objects. This value will be used to construct a
+    /// `deadpool_postgres::Timeouts` value.
+    pub connection_pool_timeouts_secs: u64,
```
BranLwyd, line 35: Let's make this optional, defaulting to some reasonable value (60s SGTM) -- we should have a config knob for this, but I suspect we aren't going to want to turn it very often/ever.

janus_server/src/binary_utils.rs
BranLwyd: (all of the above, of course, doesn't require a change in this PR)

BranLwyd (approved):

janus_server/src/binary_utils.rs
tgeoghegan: I agree with both your points and I made a note in #36 about it.

janus_server/src/config.rs
tgeoghegan: Good point. Annoyingly we can't just implement `std::default::Default` on `DbConfig` (what would we use for `url`?) and the `#[serde(default)]` annotation doesn't let us provide a literal, just the name of a function to call, so this takes a bit more code than I'd like but it's tolerable.

PR #323, opened 2022-07-19
Add a simple health check endpoint

divergentdave:
This is a first effort towards #36, adding a health check endpoint that tests if our database connection pool is able to connect. We can use this in Kubernetes readiness checks and in GCP load balancer health checks for our Ingress.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -2132,7 +2132,7 @@ fn aggregator_filter<C: Clock>(
             CONTENT_TYPE.as_str(),
             AggregateShareReq::MEDIA_TYPE,
         ))
-        .and(with_cloned_value(aggregator))
+        .and(with_cloned_value(aggregator.clone()))
         .and(warp::body::bytes())
         .and(warp::header::optional::<String>(DAP_AUTH_HEADER))
         .then(
@@ -2149,8 +2149,26 @@ fn aggregator_filter<C: Clock>(
         aggregate_share_routing,
         aggregate_share_responding,
         warp::cors().build(),
-        response_time_recorder,
+        response_time_recorder.clone(),
         "aggregate_share",
+    );
+
+    let health_check_routing = warp::path("_health_check");
+    let health_check_responding = warp::get().and(with_cloned_value(aggregator)).then(
+        |aggregator: Arc<Aggregator<C>>| async move {
+            aggregator.datastore.check_connection_pool().await?;
+            http::Response::builder()
+                .status(StatusCode::OK)
+                .body(Vec::new())
+                .map_err(|err| Error::Internal(format!("couldn't produce response: {}", err)))
+        },
+    );
+    let health_check_endpoint = compose_common_wrappers(
+        health_check_routing,
+        health_check_responding,
+        warp::cors().build(),
+        response_time_recorder,
+        "health_check",
     );
 
     Ok(hpke_config_endpoint
@@ -2159,6 +2177,7 @@ fn aggregator_filter<C: Clock>(
         .or(collect_endpoint)
         .or(collect_jobs_endpoint)
         .or(aggregate_share_endpoint)
+        .or(health_check_endpoint)
         .boxed())
 }
 
@@ -2182,7 +2201,8 @@ mod tests {
     use crate::{
         datastore::{
             models::BatchUnitAggregation,
-            test_util::{ephemeral_datastore, DbHandle},
+            test_util::{ephemeral_datastore, generate_aead_key, DbHandle},
+            Crypter,
         },
         task::{
             test_util::{generate_aggregator_auth_token, new_dummy_task},
@@ -2190,6 +2210,7 @@ mod tests {
         },
     };
     use assert_matches::assert_matches;
+    use deadpool_postgres::{Manager, Pool};
     use http::Method;
     use hyper::body;
     use janus_core::{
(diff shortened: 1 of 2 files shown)


janus_server/src/aggregator.rs
```diff
@@ -2132,7 +2132,7 @@ fn aggregator_filter<C: Clock>(
             CONTENT_TYPE.as_str(),
             AggregateShareReq::MEDIA_TYPE,
         ))
-        .and(with_cloned_value(aggregator))
+        .and(with_cloned_value(aggregator.clone()))
```
BranLwyd, line 2135: nit: `Arc::clone(&aggregator)` (& please fixup the other instances in this function)

janus_server/src/aggregator.rs
```diff
@@ -2149,16 +2149,35 @@ fn aggregator_filter<C: Clock>(
         aggregate_share_routing,
         aggregate_share_responding,
         warp::cors().build(),
-        response_time_recorder,
+        response_time_recorder.clone(),
         "aggregate_share",
     );
 
+    let health_check_routing = warp::path("_health_check");
```
tgeoghegan, line 2156: Maybe instead of prefixing routes like this with `_`, we should try to namespace our routes to avoid future collisions. So all the HTTP routes specified by DAP could be under `/dap/` (`/dap/hpke_config`, `/dap/upload`, etc.), and then we could call our health check route whatever we want and never have to worry about colliding with new routes defined by DAP.

janus_server/src/datastore.rs
```diff
@@ -128,6 +128,15 @@ impl<C: Clock> Datastore<C> {
         tx.tx.commit().await?;
         Ok(rslt)
     }
+
+    /// Attempt to fetch a connection from the connection pool, to check if
+    /// the database is accessible. This will either create a new database
+    /// connection or recycle an existing one, and then return the connection
+    /// back to the pool.
+    pub async fn check_connection_pool(&self) -> Result<(), Error> {
+        let _client = self.pool.get().await?;
```
BranLwyd, line 137: I'm not sure I like the idea that a health check will (potentially) create new connections to the database. This will introduce additional connections, which can be a valuable resource. More importantly, if we ever introduce a maximum number of connections per process (which seems likely), a process that's simply busy might appear unhealthy. It would be taken down, which would lead to other processes taking on this load, and themselves becoming busier -- at worst, this could lead to cascading failure due to (self-inflicted) overload.

If all we want is a health check, IMO the health-check endpoint should just return 200.

If we want an effective "initial liveness" check as well, perhaps the datastore could track whether it has ever succeeded at getting a client, and return true from here if so? (and if it hasn't ever gotten a client, then it would do so here, but only at most once up to concurrency of health-check requests)

But an even better to the initial liveness check might be to retrieve a client once at process startup (ie in the main function somewhere), before we start listening on a port for requests. Then we could just do the just-return-200 in the health check.

janus_server/src/datastore.rs
```diff
@@ -128,6 +128,15 @@ impl<C: Clock> Datastore<C> {
         tx.tx.commit().await?;
         Ok(rslt)
     }
+
+    /// Attempt to fetch a connection from the connection pool, to check if
+    /// the database is accessible. This will either create a new database
+    /// connection or recycle an existing one, and then return the connection
+    /// back to the pool.
+    pub async fn check_connection_pool(&self) -> Result<(), Error> {
```
tgeoghegan, line 136: #322 adds retries to getting connections from the pool, so I think we'll want to refactor this so that both this and `run_tx_once` use the same code to get a connection... but I can take care of that by rebasing #322 after we merge this change.

janus_server/src/aggregator.rs
```diff
@@ -2149,16 +2149,35 @@ fn aggregator_filter<C: Clock>(
...
+    let health_check_responding = warp::get().and(with_cloned_value(aggregator)).then(
+        |aggregator: Arc<Aggregator<C>>| async move {
+            aggregator.datastore.check_connection_pool().await?;
+            http::Response::builder()
+                .status(StatusCode::OK)
+                .body(Vec::new())
```
tgeoghegan, line 2162: nit: [`http::Response::Builder::body`](https://docs.rs/http/latest/http/response/struct.Builder.html#method.body) lets you pass `()` so I think we can save an allocation:
```suggestion
                .body(())
```

BranLwyd (changes requested):

janus_server/src/aggregator.rs
BranLwyd: There's a convention in Kubernetes/Docker for using `/healthz` for this. (https://stackoverflow.com/questions/43380939/where-does-the-convention-of-using-healthz-for-application-health-checks-come-f)

janus_server/src/datastore.rs
tgeoghegan: It's possible that what we want here is a _startup probe_ as well as a health check. IIUC, Kubernetes won't consider a container to be started up (and hence won't add it to the backends of a service) until the startup probe succeeds, but once it does succeed, it won't hit the startup probe path again. So that would be a good place to check that we can connect to the database, that the necessary cryptography keys are available, etc..

janus_server/src/aggregator.rs
divergentdave: There's no `impl From<()> for hyper::body::Body`, but I can use a static slice for the same win.

BranLwyd: Note that `Vec::new()` doesn't allocate an array on the heap since it has zero length; unless the `Vec` value itself is moved to the heap, I think there is no allocation, and if that's the case then I think a static slice would also end up moved to the heap.

BranLwyd (approved):

PR #324, opened 2022-07-19
Pin Kind node image used in EphemeralCluster

tgeoghegan:
By default, `kind` would install a node image for Kubernetes 1.24, which
is a couple minor versions past what we deploy to cloud environments and
which runs afoul of some Terraform bugs[1]. Pin the Kind node image we
use to a Kubernetes 1.22 image, to match the Kind we use in other CI
contexts.

[1]: https://github.com/hashicorp/terraform-provider-kubernetes/issues/1724

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -24,6 +24,11 @@ jobs:
           ~/.cargo/git/db/
           target/
         key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+    - name: Install latest version of Kind
+      # We should always make sure that the `kind` CLI we install is from the
+      # same release as the node image version used by
+      # `janus_core::test_util::kubernetes::EphemeralCluster`
+      run: go install sigs.k8s.io/kind@v0.14.0
     - name: lint
       run: cargo fmt --message-format human -- --check
     - name: clippy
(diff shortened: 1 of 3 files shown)


tgeoghegan:
Besides being consistent with the Kind image we use over in `janus-ops`, I found that running tests on my workstation failed if I let `kind` choose for itself what node image to boot.

BranLwyd (approved):
(I didn't see errors due to `kind`, but no objections to fixing the node version.)

divergentdave:
Hmm, we could have skew between the versions of kind we have installed locally. Could you add what version of kind this node image is from, both as a comment here and under "Running tests" in the readme?

BranLwyd:
FWIW, at latest `main` I still don't see test failures. This points to either (a) `kind` (or Docker) is internally caching whatever version it first grabbed or (b) this is a `kind` version issue rather than a node version issue.

On my workstation:
```
$ kind --version
kind version 0.14.0
```

tgeoghegan:
OK, I should dig a bit more to understand why cluster creation is failing on my workstation. I still think pinning the image makes sense in order to keep things in sync with what janus-ops will later do (though I acknowledge David's point about version drift). Another thing I thought of last night is that in janus-ops, we explicitly control what version of kind is installed, but here in janus we rely on the one that's installed on the [Ubuntu GitHub actions runner](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2204-Readme.md#tools). We (I) should do something about that in this PR, too.

tgeoghegan:
I figured out my problem: I had an existing `kind` cluster running, and I didn't have enough `inofity` instances left to start two more `kind` clusters (the `systemd` that runs in each `kind` node uses `inotify` to watch paths, and that consumes the host's inotify resources). I was able to get past this by bumping `fs.inotify.max_user_instances` to 512 from 128.

PR #325, opened 2022-07-20
Increment versions to 0.1.3

divergentdave:
In the interest of getting a new aggregator image out that includes the health check endpoints, I suggest we cut a release for version 0.1.3. I chose to increment the version of the `janus_server` crate as well, so that container tags and `--version` output match. I fixed up the CLI reference tests with wildcards as well.


tgeoghegan (approved):
LGTM, though I may well cut 0.1.4 once #322 is ready (don't wait on me though).

PR #326, opened 2022-07-21
Document inotify issue and fix

divergentdave:
This adds a section to the README with advice on raising the inotify instance limit.


tgeoghegan (approved):
This is fine but we could also link to [`kind`'s own documentation on this known issue](https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files) to save ourselves some README weight.

PR #337, opened 2022-07-25
New configuration option, extra response headers

divergentdave:
This adds a new optional field to the configuration file, with extra response headers to be added to outgoing responses.

Note that, as this uses Warp filters, it will not add the headers to requests that are rejected by our filter stack. If we wanted to do so, we would have to write some boilerplate to turn the Warp filter into a `hyper::service::Service`, add a header-setting middleware to it, and then use Hyper's TCP server directly. I opted not to go this route until we have a need for headers on error responses. (for example, only eTLD+1s are eligible for automatic enrollment in HSTS preloading, so it's okay if Janus can't serve a preload enrollment header on 404 responses for `/`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1764,6 +1764,7 @@ dependencies = [
  "chrono",
  "deadpool-postgres",
  "futures",
+ "http",
  "janus_client",
  "janus_core",
  "janus_server",
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ It is currently in active development.
 
 ## Running janus\_server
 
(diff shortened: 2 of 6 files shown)


BranLwyd (approved):
I didn't consider that the HSTS preload list is eTLD+1 -- I suppose that means we'd need to preload all of divviup.org. When we're ready to work on this, in addition to the technical work, it'd be good to poke the comms team to make sure they're OK committing to serving only HTTPS. (I strongly suspect they will be given the ISRG's mission but better to overcommunicate than undercommunicate.)

PR #346, opened 2022-07-31
Daphne integration test.

### LINKED ISSUE #341: Integration test against an independent DAP implementation
We should create an integration test that spins up an independent (i.e. non-Janus) DAP implementation, and tests that Janus & the other implementation can successfully receive & aggregate reports.

This should test both the Janus-leader & Janus-helper scenarios.

The most mature non-Janus DAP implementation I am aware of is [Daphne](https://github.com/cloudflare/daphne). This is likely our best bet for the non-Janus aggregator.

BranLwyd:
Also, refactor the existing Janus-only integration test to use the same
code as the new Daphne-Janus integration tests.

The hermetic Daphne instances are implemented by including a compiled
version of Daphne as an artifact, and running it inside miniflare, based
heavily on the manual instructions in the Daphne repository. Miniflare
is now a required dependency for running tests; instructions for running
tests & updating the version of Daphne used for tests is included in the
README.md. Ideally, we would start a container, but Daphne does not yet
build out a suitable container.

My next steps will be chatting with the Daphne folks to see if I can
contribute most of this logic to Daphne itself, which would be a more
maintainable arrangement.

The `daphne_janus` test is currently flaky; I believe that this reveals
a real functional issue, but I'm going to dig in separately from this
PR.

Fixes #341.


monolithic_integration_test/src/janus.rs
```diff
@@ -0,0 +1,182 @@
...
+        let collect_job_driver_handle = task::spawn(async move {
+            select! {
+                _ = collect_job_driver.run() => unreachable!(),
+                _ = &mut collect_job_driver_shutdown_receiver => (),
+            }
+        });
+
+        // Start the "shutdown" task, allowing us to do asynchronous shutdown work in the
+        // synchronous drop implementation. (We use an async oneshot channel for "start shutdown"
+        // and a sync mpsc channel for "shutdown complete" to us to do the required operations from
```
divergentdave, line 143: nit: missing a word
```suggestion
        // and a sync mpsc channel for "shutdown complete" to allow us to do the required operations from
```

divergentdave (approved):

divergentdave:
I think we could use a tracing subscriber in each of the new test functions as well.

BranLwyd:
N.B. we are currently working around the Daphne issue that made the test flaky. We work around the issue instead of disabling the test in order to maintain that most of the functionality (i.e. the protocol back-and-forth) still works.

PR #356, opened 2022-08-01
Serve health check requests on a separate port

divergentdave:
This moves the `/healthz` endpoint to a separate HTTP server, on its own port, and adds health checks to the three remaining server components. This change adds a new required parameter to the configuration files. Part of #36.

### DIFF
--- a/janus_server/src/bin/aggregation_job_creator.rs
+++ b/janus_server/src/bin/aggregation_job_creator.rs
@@ -100,7 +100,7 @@ mod tests {
         },
         CommonConfig,
     };
-    use std::net::{Ipv4Addr, SocketAddr, SocketAddrV4};
+    use std::net::{Ipv4Addr, SocketAddr};
 
     #[test]
     fn roundtrip_config() {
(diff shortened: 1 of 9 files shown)


janus_server/src/bin/aggregation_job_creator.rs
```diff
@@ -108,6 +109,10 @@ mod tests {
                 database: generate_db_config(),
                 logging_config: generate_trace_config(),
                 metrics_config: generate_metrics_config(),
+                health_check_listen_address: SocketAddr::V4(SocketAddrV4::new(
```
BranLwyd, line 112: nit: a slightly more concise way to generate this value would be `SocketAddr::from((Ipv4Addr::LOCALHOST, 8080))`.

Separately, I think it would be slightly more realistic to specify `Ipv4Addr::UNSPECIFIED` (i.e. `0.0.0.0`) rather than `Ipv4Addr::LOCALHOST` (i.e. `127.0.0.1`), since IIUC that's what we'd want to specify in "real" configurations.

BranLwyd (approved):

PR #357, opened 2022-08-01
Authenticate collector requests.

### LINKED ISSUE #343: DAP compliance: collect requests should be authenticated
According to [DAP-01 section 4.4.1](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-01.html#name-collection-initialization), the `/collect` endpoint `MUST` be authenticated. (The intended authentication scheme, like for other endpoints, is the [bearer-token scheme described in section 3.2](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-01.html#name-https-sender-authentication).)

Currently, Janus [does not authenticate the `/collect` endpoint](https://github.com/divviup/janus/blob/4716fe32d77b9f4dc55503df9e49534e214cf5e8/janus_server/src/aggregator.rs#L5001-L5009). We should begin doing so.

A few related items:
  * We do (correctly, per-spec) authenticate the `/aggregate_share` endpoint.
  * We don't authenticate requests to the collect job URI (returned from the `/collect` endpoint), either. The spec doesn't call this out as required. Should Janus require authentication, and should we try to specify authentication for these requests in DAP? (IMO, yes to both questions)

BranLwyd:
Authentication of the /collect endpoint is mandated by the DAP
specification. Authentication of the collect job endpoints is not, but
IMO it's a good idea.

Closes #343.

### DIFF
--- a/janus_server/src/task.rs
+++ b/janus_server/src/task.rs
@@ -137,25 +137,25 @@ enum VdafSerialization {
     FakeFailsPrepStep,
 }
 
-/// An authentication (bearer) token used by aggregators for aggregator-to-aggregator
-/// authentication.
+/// An authentication (bearer) token used by aggregators for aggregator-to-aggregator &
+/// collector-to-aggregator authentication.
 #[derive(Clone)]
-pub struct AggregatorAuthenticationToken(Vec<u8>);
+pub struct AuthenticationToken(Vec<u8>);
 
-impl From<Vec<u8>> for AggregatorAuthenticationToken {
+impl From<Vec<u8>> for AuthenticationToken {
     fn from(token: Vec<u8>) -> Self {
         Self(token)
     }
 }
 
-impl AggregatorAuthenticationToken {
+impl AuthenticationToken {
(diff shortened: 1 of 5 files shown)


janus_server/src/task.rs
```diff
@@ -303,7 +326,8 @@ struct SerializedTask {
     min_batch_duration: Duration,
     tolerable_clock_skew: Duration,
     collector_hpke_config: SerializedHpkeConfig,
-    agg_auth_tokens: Vec<String>,          // in unpadded base64url
+    aggregator_auth_tokens: Vec<String>, // in unpadded base64url
```
divergentdave, line 329: Note that this rename will change our task definition file format. (in addition to the new member below) I think we have exactly one task YAML file, so backwards compatibility isn't a concern.

divergentdave (approved):

janus_server/src/task.rs
BranLwyd: Agreed -- for now, backwards-compatibility of the Task serialization format is not a requirement.

PR #361, opened 2022-08-03
Dockerfile: remove openssl-dev dependency.

BranLwyd:
This was added in the Daphne integration test PR, but in retrospect does
not appear to be required.


BranLwyd:
Looks like it is required after all...

PR #362, opened 2022-08-04
Add features to monolithic_integration_test

divergentdave:
This moves most of monolithic_integration_test and its novel dependencies behind a new "test-util" feature flag, similar to the other
crates. With this, our release builds will no longer have the feature flags janus_core/test-util, janus_server/test-util, nor reqwest/default-tls set, due to crate feature flags being unified across the workspace. As a result, we can once again eliminate OpenSSL as a build-time dependency.

I also fixed a tangentially related inaccuracy in some comments that Tim pointed out.

### DIFF
--- a/janus_core/Cargo.toml
+++ b/janus_core/Cargo.toml
@@ -58,6 +58,6 @@ hex = { version = "*", features = ["serde"] }
 janus_core = { path = ".", features = ["test-util"] }
 # Enable `kube`'s `openssl-tls` feature (which takes precedence over the
 # `rustls-tls` feature when creating a default client) to work around rustls's
-# lack of support for IP addresses in certificates when connecting to most
+# lack of support for connecting to servers by IP addresses, which affects many
 # Kubernetes clusters.
 kube = { version = "*", features = ["openssl-tls"] }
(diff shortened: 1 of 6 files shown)


BranLwyd:
While I don't mind doing this as a workaround, we really shouldn't have to introduce a `test-util` feature on a crate that is already test-only just to keep Cargo from bleeding its dependencies into other, unrelated crates. Every user of `monolithic_integration_test` is going to want to specify the `test-util` feature!

Could we instead factor this as `janus` & `daphne` features? (The former would enable use of the `janus` module, the latter the `daphne` module.) This would make usage of a feature more reasonable. And I suppose I exposed this functionality in the first place with the thought that others may want to build integration tests of their own against Janus; such a user would want to enable the `janus` feature but not the `daphne` feature, I suppose.

janus_core/Cargo.toml
```diff
@@ -58,6 +58,6 @@ hex = { version = "*", features = ["serde"] }
 janus_core = { path = ".", features = ["test-util"] }
 # Enable `kube`'s `openssl-tls` feature (which takes precedence over the
 # `rustls-tls` feature when creating a default client) to work around rustls's
-# lack of support for IP addresses in certificates when connecting to a Kind
-# cluster.
+# lack of support for IP addresses in certificates when connecting to most
```
BranLwyd, line 61: I think the specific condition is if the cluser is identified by IP, rather than DNS name -- IIUC rustls does not yet support IP-address-based certs.

Maybe "Kubernetes clusters identified by IP" or "local Kubernetes clusters" instead of "most Kubernetes clusters"? Both are more specific to the conditions that cause breakage.

BranLwyd (approved):

PR #375, opened 2022-08-09
Implement interoperation test API

divergentdave:
This adds three new binaries (and container images) to support draft-dcook-ppm-dap-interop-test-design. It includes an end-to-end test that exercises the binaries using the test APIs.

The following items are made public or moved out from behind "test-util" features to support this:

- `janus_server::aggregator::aggregator_filter()`
- `janus_core::hpke::generate_hpke_config_and_private_key()`
- `janus_core::time::MockClock`

### DIFF
--- a/interop_binaries/src/bin/janus_interop_aggregator.rs
+++ b/interop_binaries/src/bin/janus_interop_aggregator.rs
@@ -114,15 +114,7 @@ async fn handle_add_task(
                 collector_authentication_token.into_bytes(),
             )],
         ),
-        (1, _) => (
-            Role::Helper,
-            // TODO(issue #370): Task::new() requires that we have a collector authentication
-            // token, but draft-dcook-ppm-dap-interop-test-design-00 only provides such a token
-            // to the leader.
-            vec![AuthenticationToken::from(
-                b"This is a fake collector authentication token for the helper".to_vec(),
-            )],
-        ),
+        (1, _) => (Role::Helper, Vec::new()),
         _ => return Err(anyhow::anyhow!("Invalid \"aggregator_id\" value")),
     };
 
--- a/interop_binaries/src/bin/janus_interop_aggregator.rs
+++ b/interop_binaries/src/bin/janus_interop_aggregator.rs
@@ -35,14 +35,6 @@ use std::{
 use url::Url;
 use warp::{hyper::StatusCode, reply::Response, Filter, Reply};
 
-#[derive(Debug, Deserialize)]
-#[serde(rename_all = "camelCase")]
-struct EndpointRequest {
-    _task_id: String,
-    _aggregator_id: u8,
-    _hostname_and_port: String,
-}
-
 #[derive(Debug, Serialize)]
 struct EndpointResponse {
     status: &'static str,
@@ -58,8 +50,8 @@ static ENDPOINT_RESPONSE: EndpointResponse = EndpointResponse {
 #[serde(rename_all = "camelCase")]
 struct AddTaskRequest {
     task_id: String,
-    leader: String,
-    helper: String,
+    leader: Url,
+    helper: Url,
     vdaf: VdafObject,
     leader_authentication_token: String,
     #[serde(default)]
@@ -86,8 +78,6 @@ async fn handle_add_task(
     let task_id_bytes = base64::decode_config(request.task_id, base64::URL_SAFE_NO_PAD)
         .context("invalid base64url content in \"taskId\"")?;
     let task_id = TaskId::get_decoded(&task_id_bytes).context("invalid length of TaskId")?;
-    let leader_url = Url::parse(&request.leader).context("bad leader URL")?;
-    let helper_url = Url::parse(&request.helper).context("bad helper URL")?;
     let vdaf: janus_core::task::VdafInstance = request.vdaf.into();
     let vdaf: janus_server::task::VdafInstance = vdaf.into();
     let leader_authentication_token =
@@ -129,7 +119,7 @@ async fn handle_add_task(
 
     let task = Task::new(
         task_id,
-        vec![leader_url, helper_url],
+        vec![request.leader, request.helper],
         vdaf,
         role,
         vec![verify_key],
@@ -161,12 +151,10 @@ fn make_filter(
     let clock = janus_core::time::RealClock::default();
     let dap_filter = janus_server::aggregator::aggregator_filter(Arc::clone(&datastore), clock)?;
 
-    let endpoint_filter = warp::path!("endpoint_for_task")
-        .and(warp::body::json())
-        .map(|_request: EndpointRequest| {
-            warp::reply::with_status(warp::reply::json(&ENDPOINT_RESPONSE), StatusCode::OK)
-                .into_response()
-        });
+    let endpoint_filter = warp::path!("endpoint_for_task").map(|| {
+        warp::reply::with_status(warp::reply::json(&ENDPOINT_RESPONSE), StatusCode::OK)
+            .into_response()
+    });
     let add_task_filter =
         warp::path!("add_task")
             .and(warp::body::json())
@@ -236,6 +224,8 @@ async fn main() -> anyhow::Result<()> {
     client
         .batch_execute(include_str!("../../../db/schema.sql"))
         .await?;
+    // Return the database connection we used to deploy the schema back to the pool, so it can be
+    // reused.
     drop(client);
     let datastore = Arc::new(Datastore::new(pool, crypter, clock));
 
--- a/interop_binaries/src/bin/janus_interop_aggregator.rs
+++ b/interop_binaries/src/bin/janus_interop_aggregator.rs
@@ -198,10 +198,10 @@ fn app() -> clap::Command<'static> {
                 .help("Port number to listen on."),
         )
         .arg(
-            Arg::new("postgres-port")
-                .long("postgres-port")
-                .default_value("5432")
-                .help("Port number of PostgreSQL database on localhost."),
+            Arg::new("postgres-url")
+                .long("postgres-url")
+                .default_value("postgres://postgres@127.0.0.1:5432/postgres")
+                .help("PostgreSQL database connection URL."),
         )
 }
 
@@ -210,7 +210,7 @@ async fn main() -> anyhow::Result<()> {
     install_tracing_subscriber()?;
     let matches = app().get_matches();
     let http_port = matches.value_of_t::<u16>("port")?;
-    let postgres_port = matches.value_of_t::<u16>("postgres-port")?;
+    let postgres_url = matches.value_of_t::<Url>("postgres-url")?;
 
     // Make an ephemeral datastore key.
     let mut key_bytes = [0u8; 16];
@@ -220,11 +220,7 @@ async fn main() -> anyhow::Result<()> {
 
     // Connect to database, apply schema, and set up datastore.
     let db_config = DbConfig {
-        url: Url::parse(&format!(
-            "postgres://postgres@127.0.0.1:{}/postgres",
-            postgres_port
-        ))
-        .unwrap(),
+        url: postgres_url,
         connection_pool_timeouts_secs: 30,
     };
     let pool = database_pool(&db_config, &None).await?;
--- a/interop_binaries/src/bin/janus_interop_aggregator.rs
+++ b/interop_binaries/src/bin/janus_interop_aggregator.rs
@@ -36,12 +36,10 @@ use url::Url;
 use warp::{hyper::StatusCode, reply::Response, Filter, Reply};
 
 #[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
 struct EndpointRequest {
-    #[serde(rename = "taskId")]
     _task_id: String,
-    #[serde(rename = "aggregatorId")]
     _aggregator_id: u8,
-    #[serde(rename = "hostnameAndPort")]
     _hostname_and_port: String,
 }
 
@@ -57,27 +55,20 @@ static ENDPOINT_RESPONSE: EndpointResponse = EndpointResponse {
 };
 
 #[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
 struct AddTaskRequest {
-    #[serde(rename = "taskId")]
     task_id: String,
     leader: String,
     helper: String,
     vdaf: VdafObject,
-    #[serde(rename = "leaderAuthenticationToken")]
     leader_authentication_token: String,
-    #[serde(default, rename = "collectorAuthenticationToken")]
+    #[serde(default)]
     collector_authentication_token: Option<String>,
-    #[serde(rename = "aggregatorId")]
     aggregator_id: u8,
-    #[serde(rename = "verifyKey")]
     verify_key: String,
-    #[serde(rename = "maxBatchLifetime")]
     max_batch_lifetime: u64,
-    #[serde(rename = "minBatchSize")]
     min_batch_size: u64,
-    #[serde(rename = "minBatchDuration")]
     min_batch_duration: u64,
-    #[serde(rename = "collectorHpkeConfig")]
     collector_hpke_config: String,
 }
 
--- a/interop_binaries/src/bin/janus_interop_aggregator.rs
+++ b/interop_binaries/src/bin/janus_interop_aggregator.rs
@@ -2,11 +2,12 @@ use anyhow::Context;
 use base64::URL_SAFE_NO_PAD;
 use clap::{Arg, Command};
 use interop_binaries::{
-    generate_hpke_keypair, install_tracing_subscriber,
+    install_tracing_subscriber,
     status::{ERROR, SUCCESS},
     VdafObject,
 };
 use janus_core::{
+    hpke::generate_hpke_config_and_private_key,
     message::{Duration, HpkeConfig, Role, TaskId},
     time::RealClock,
     TokioRuntime,
@@ -134,7 +135,7 @@ async fn handle_add_task(
         _ => return Err(anyhow::anyhow!("Invalid \"aggregator_id\" value")),
     };
 
-    let (hpke_config, private_key) = generate_hpke_keypair();
+    let (hpke_config, private_key) = generate_hpke_config_and_private_key();
 
     let task = Task::new(
         task_id,
--- a/interop_binaries/src/bin/janus_interop_aggregator.rs
+++ b/interop_binaries/src/bin/janus_interop_aggregator.rs
@@ -84,29 +84,29 @@ async fn handle_add_task(
     request: AddTaskRequest,
 ) -> anyhow::Result<()> {
     let task_id_bytes = base64::decode_config(request.task_id, base64::URL_SAFE_NO_PAD)
-        .context("Invalid base64url content in \"taskId\"")?;
-    let task_id = TaskId::get_decoded(&task_id_bytes).context("Invalid length of TaskId")?;
-    let leader_url = Url::parse(&request.leader).context("Bad leader URL")?;
-    let helper_url = Url::parse(&request.helper).context("Bad helper URL")?;
+        .context("invalid base64url content in \"taskId\"")?;
+    let task_id = TaskId::get_decoded(&task_id_bytes).context("invalid length of TaskId")?;
+    let leader_url = Url::parse(&request.leader).context("bad leader URL")?;
+    let helper_url = Url::parse(&request.helper).context("bad helper URL")?;
     let vdaf: janus_core::task::VdafInstance = request.vdaf.into();
     let vdaf: janus_server::task::VdafInstance = vdaf.into();
     let leader_authentication_token =
         AuthenticationToken::from(request.leader_authentication_token.into_bytes());
     let verify_key = base64::decode_config(request.verify_key, URL_SAFE_NO_PAD)
-        .context("Invalid base64url content in \"verifyKey\"")?;
+        .context("invalid base64url content in \"verifyKey\"")?;
     let min_batch_duration = Duration::from_seconds(request.min_batch_duration);
     let collector_hpke_config_bytes =
         base64::decode_config(request.collector_hpke_config, URL_SAFE_NO_PAD)
-            .context("Invalid base64url content in \"collectorHpkeConfig\"")?;
+            .context("invalid base64url content in \"collectorHpkeConfig\"")?;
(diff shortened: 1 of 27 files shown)


interop_binaries/tests/end_to_end.rs
```diff
@@ -0,0 +1,611 @@
...
+        BTreeSet::from([client_port, leader_port, helper_port, collector_port]).len(),
+        4,
+        "Ports selected for HTTP servers were not unique",
+    );
+
+    // Step 1: Create and start containers. (here, we just run the binaries instead)
+    // We use std::process instead of tokio::process so that we can kill the child processes from
+    // a Drop implementation. tokio::process::Child::kill() is async, and could not be called from
+    // there.
+    let mut client_command = Command::new(env!("CARGO_BIN_EXE_janus_interop_client"));
```
BranLwyd, line 140: Why do these need to be written as multiple statements? I think e.g. `arg` is meant to be called in a chained fashion, so this could be written as:

```
let client_command = Command::new(...)
    .arg("--port")
    .arg(format!("{}", client_port))
    .stout(Stdio::piped())
    .stderr(Stdio::piped())
    .spawn();
```

interop_binaries/tests/end_to_end.rs
```diff
@@ -0,0 +1,611 @@
...
+        sleep(StdDuration::from_millis(200)).await;
+    }
+    Err(anyhow::anyhow!(
+        "Timed out waiting for a server to accept on port {}",
+        port,
+    ))
+}
+
+/// RAII guard to ensure that child processes are cleaned up during test failures.
+struct DropGuard(Vec<Child>);
```
BranLwyd, line 54: I think the `DropGuard` name is a little non-obvious (the same name might apply to many different types). Maybe something like `KillOnDropChild` (...a rather gruesome name)?

interop_binaries/tests/end_to_end.rs
```diff
@@ -0,0 +1,611 @@
...
+
+impl Drop for DropGuard {
+    fn drop(&mut self) {
+        let mut any_error = false;
+        for mut process in self.0.drain(..) {
```
BranLwyd, line 59: This type is complicated by having to deal with multiple `Child`ren IMO. It would be more easily stated if each `DropGuard` owned a single `Child` and implemented the desired drop logic for it.

interop_binaries/tests/end_to_end.rs
BranLwyd, line 54: Optional: creating a `DropGuard` could also set up forwarding of its stdout/stderr. This would simplify setup later, but also merges mostly-unrelated concerns into a single type. But you could write the process creation as a one-liner e.g.:

```
let my_child = DropGuard::new(Command::new(...)
   .arg(...)
   .stdin(...)
   .etc()
   .spawn());
```

If you take this suggestion, maybe name the type something like `TestProcessChild`, reflecting the more-general purpose of this type?

interop_binaries/tests/end_to_end.rs
```diff
@@ -0,0 +1,611 @@
...
+    let leader_port = pick_unused_port().context("Couldn't pick a port for the leader")?;
+    let helper_port = pick_unused_port().context("Couldn't pick a port for the helper")?;
+    let collector_port = pick_unused_port().context("Couldn't pick a port for the collector")?;
+    assert_eq!(
+        BTreeSet::from([client_port, leader_port, helper_port, collector_port]).len(),
+        4,
+        "Ports selected for HTTP servers were not unique",
+    );
+
+    // Step 1: Create and start containers. (here, we just run the binaries instead)
```
BranLwyd, line 136: Q: is the reason we start binaries instead of containers "container build is not automated"?

janus_core/src/hpke.rs
```diff
@@ -170,38 +174,28 @@ pub fn open(
...
-                HpkeKdfId::HkdfSha256,
-                HpkeAeadId::Aes128Gcm,
-                HpkePublicKey::new(public_key),
-            ),
-            HpkePrivateKey::new(private_key),
-        )
-    }
+/// Generate a new HPKE keypair and return it as an HpkeConfig (public portion) and
+/// HpkePrivateKey (private portion).
+pub fn generate_hpke_config_and_private_key() -> (HpkeConfig, HpkePrivateKey) {
```
BranLwyd, line 179: I think this function should either remain gated by the `test-util` feature, or be generalized to allow different `Kem`, `Kdf`, `Aead` selections & allow the config ID to be specified.

janus_core/src/time.rs
```diff
@@ -32,49 +35,40 @@ impl Debug for RealClock {
...
-    #[non_exhaustive]
-    pub struct MockClock {
-        /// The time that this clock will return from [`Self::now`].
-        current_time: Arc<Mutex<Time>>,
-    }
+/// A mock clock for use in testing. Clones are identical: all clones of a given MockClock will
+/// be controlled by a controller retrieved from any of the clones.
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub struct MockClock {
```
BranLwyd, line 42: I think `MockClock` should remain guarded by the `test-util` feature, since it is intended for use only in tests. (What's the technical reason you'd like to drop this feature guard?)

BranLwyd (commented):
This is really nice; I like the container definitions. Have you done any (possibly manual/one-off) testing to make sure the networking etc works out for these containers to allow a realistic test to be done?

I was working on containerizing the functionality in `monolithic_integration_test`; I'm probably going to refactor that work to use these binaries instead, and it'd be fantastic if I could run them in a container. (Unfortunately, I think until Daphne provides an implementation of the test API too, we'll need a separate interface for the `monolithic_integration_test` tests, even if Daphne itself is containerized. I guess the cleanest way to implement that is to phrase the `monolithic_integration_test` functionality in terms of these official-test-API containers, so at least the bulk of the functionality is only written once.)

divergentdave:
Yes, I rigged up a quick and dirty test of the containers themselves with bash, curl, and jq.

interop_binaries/src/bin/janus_interop_aggregator.rs
```diff
@@ -0,0 +1,337 @@
...
+fn app() -> clap::Command<'static> {
+    Command::new("Janus interoperation test aggregator")
+        .arg(
+            Arg::new("port")
+                .long("port")
+                .short('p')
+                .default_value("8080")
+                .help("Port number to listen on."),
+        )
+        .arg(
```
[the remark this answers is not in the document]
divergentdave: Yes, this is for the `end_to_end` test. I'll make that change.

interop_binaries/tests/end_to_end.rs
divergentdave: The builder methods each return `&mut Command`, so at minimum I need the let binding, since I'm not spawning from a temporary value. Beyond that, I pair arguments here for aesthetics, and then chain the rest of the builder methods at once within the for loop.

interop_binaries/tests/end_to_end.rs
divergentdave: Yes, plus Cargo ensures that binary targets are up-to-date before running this integration test.

janus_core/src/time.rs
divergentdave: I need this functionality to allow submitting reports with crafted timestamps. If I were to turn on the `janus_core/test-util` feature from the `interop_binaries` dependencies, then that would result in it being enabled for all builds throughout the workspace, and we'd be building `kube` from our `janus_server` release builds, etc.

interop_binaries/tests/end_to_end.rs
```diff
@@ -0,0 +1,601 @@
...
+        tokio::spawn(forward_stderr(ChildStderr::from_std(
+            drop_guard.0.stderr.take().unwrap(),
+        )?));
+        drop_guards.push(drop_guard);
+    }
+
+    // Step 3: Try opening a TCP connection to each container's port, and
+    // retry until it succeeds.
+    for port in [client_port, leader_port, helper_port, collector_port] {
+        wait_for_tcp_server(port).await?;
```
tgeoghegan, line 170: I don't think we can assume for all DAP implementations that the TCP port being bound necessarily means the server is ready for requests. I think you'll need to do some retries in the HTTP client accessing `/internal/test/endpoint_for_task` anyway, at which point you might as well not bother polling the TCP port.

interop_binaries/tests/end_to_end.rs
```diff
@@ -0,0 +1,601 @@
...
+    let leader_port = pick_unused_port().context("couldn't pick a port for the leader")?;
+    let helper_port = pick_unused_port().context("couldn't pick a port for the helper")?;
+    let collector_port = pick_unused_port().context("couldn't pick a port for the collector")?;
+    assert_eq!(
+        BTreeSet::from([client_port, leader_port, helper_port, collector_port]).len(),
+        4,
+        "Ports selected for HTTP servers were not unique",
+    );
+
+    // Step 1: Create and start containers. (here, we just run the binaries instead)
```
tgeoghegan, line 122: Editorial nit: numbered sequences in comments like this have a bad habit of getting out of sequence as new steps are added or steps are deleted. Might be best to just remove the numeration.

interop_binaries/src/bin/janus_interop_collector.rs
```diff
@@ -0,0 +1,503 @@
...
+use serde::{Deserialize, Serialize};
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    net::{Ipv4Addr, SocketAddr},
+    sync::Arc,
+};
+use tokio::sync::Mutex;
+use warp::{hyper::StatusCode, reply::Response, Filter, Reply};
+
+static DAP_AUTH_TOKEN: &str = "DAP-Auth-Token";
```
tgeoghegan, line 41: There's `janus_server::task::DAP_AUTH_HEADER` for this

interop_binaries/src/bin/janus_interop_collector.rs
```diff
@@ -0,0 +1,503 @@
...
+    let task_id = TaskId::get_decoded(&task_id_bytes).context("invalid length of TaskId")?;
+
+    let mut tasks_guard = tasks.lock().await;
+    let entry = tasks_guard.entry(task_id);
+    if let Entry::Occupied(_) = &entry {
+        return Err(anyhow::anyhow!("cannot add a task with a duplicate ID"));
+    }
+
+    let (hpke_config, private_key) = generate_hpke_config_and_private_key(
+        HpkeConfigId::from(0u8),
```
tgeoghegan, line 141: If `handle_add_task` is called multiple times, won't the aggregator end up with multiple HPKE configs with the same ID? Is the notion that HPKE configs only need to be unique in the scope of a task?

interop_binaries/src/bin/janus_interop_collector.rs
```diff
@@ -0,0 +1,503 @@
...
+#[derive(Debug, Deserialize)]
+struct CollectPollRequest {
+    handle: String,
+}
+
+#[derive(Debug, Serialize)]
+#[serde(untagged)]
+enum AggregationResult {
+    Number(u64),
+    NumberArray(Vec<u64>),
```
tgeoghegan, line 88: Is `u64` big enough? `Prio3Aes128Sum` uses `Field128`, so in the end, the result type is `u128`.  Making this generic over `prio::vdaf::Vdaf` would be kind of a pain but since this struct is just used for serializing to JSON, maybe using `u128` instead of `u64` suffices.

interop_binaries/src/bin/janus_interop_collector.rs
```diff
@@ -0,0 +1,503 @@
...
+            let vdaf = Prio3::new_aes128_histogram(2, buckets)
+                .context("failed to construct Prio3Aes128Histogram VDAF")?;
+            let aggregate_result = vdaf
+                .unshard(&(), [leader_aggregate_share, helper_aggregate_share])
+                .context("could not unshard aggregate result")?;
+            let converted = aggregate_result
+                .into_iter()
+                .map(|counter| {
+                    u64::try_from(counter).context(
+                        "entry in aggregate result was too large to represent natively in JSON",
```
tgeoghegan, line 361: JSON numbers are floating point, which should be big enough to represent `u128`. I'm not sure if `u128` is big enough that loss of precision can occur, though.

interop_binaries/src/bin/janus_interop_collector.rs
```diff
@@ -0,0 +1,503 @@
+use anyhow::Context;
```
tgeoghegan, line 1: We should consider factoring some of this logic out into a standalone collector library that handles making collect requests, polling collect URIs and combining aggregate shares into aggregate results. We could distribute it in `janus_client`. Not needed in this PR, but a future enhancement.

tgeoghegan (changes requested):

janus_core/src/hpke.rs
divergentdave: I'd like to avoid it, because turning on `test-util` there would have the effect of turning it on for every build in the workspace. This was half of the cause of our last run-in with openssl-dev build-time dependencies.

interop_binaries/src/bin/janus_interop_aggregator.rs
```diff
@@ -0,0 +1,337 @@
...
+    _task_id: String,
+    #[serde(rename = "aggregatorId")]
+    _aggregator_id: u8,
+    #[serde(rename = "hostnameAndPort")]
+    _hostname_and_port: String,
+}
+
+#[derive(Debug, Serialize)]
+struct EndpointResponse {
+    status: &'static str,
```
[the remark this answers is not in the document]
divergentdave: I wanted to reserve the HTTP status for issues between the interop test runner and the interop container, such as unsupported endpoints, invalid requests. Plus, I anticipate that some tests will expect particular DAP operations to fail, so this way the test runner doesn't have to worry about teasing apart 400s or 500s that are due to the expected failure versus a bug in the `/internal/test/...` code, and can easily ignore those errors when they are expected.

interop_binaries/src/bin/janus_interop_aggregator.rs
```diff
@@ -0,0 +1,337 @@
...
+        .unwrap(),
+        connection_pool_timeouts_secs: 30,
+    };
+    let pool = database_pool(&db_config, &None).await?;
+    let clock = janus_core::time::RealClock::default();
+    let client = pool.get().await?;
+    client
+        .batch_execute(include_str!("../../../db/schema.sql"))
+        .await?;
+    drop(client);
```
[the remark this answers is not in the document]
divergentdave: Not hugely, it just lets that one connection be recycled by the pool. As it's just one, we're not in danger of exhausting the pool.

interop_binaries/tests/end_to_end.rs
divergentdave: This test is already making assumptions that it's just talking to Janus's interop test servers (i.e., it hardcodes the aggregator URLs assuming that `endpoint_for_task` returns "/") but I'll open an issue to track this more generally on the design document repository.

interop_binaries/src/bin/janus_interop_collector.rs
divergentdave: I'm not aware of any semantics for this ID, beyond that it gets copied into ciphertext structs, but intuitively I figure that would have to be only unique per task. Aggregators can definitely use different HPKE configs for different tasks, so it stands to reason collectors could too. I imagine a big enough aggregator or collector could run out of configuration identifier space with many tasks if they tried to not reuse any HpkeConfigId across tasks.

interop_binaries/src/bin/janus_interop_collector.rs
BranLwyd: JSON numbers can only "safely" represent integers up to about 53 bits[1]. If we care to fix this (maybe not for now, since this is test code & can simply arrange to not use such large numbers), the standard method I've seen is to represent the integer as a JSON string like `"12345"`.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

interop_binaries/src/bin/janus_interop_collector.rs
divergentdave: I specified that the result should be encoded as a JSON number, and `serde_json`'s support for u128 is off by default, and has some warts (due to the limitations of JSON). I'll open an issue on the document to consider changing result types.

interop_binaries/src/bin/janus_interop_collector.rs
tgeoghegan: Oh, right, this is only the collector. This should be OK for now, but might become a problem in the future when DAP revisits how the collector's HPKE config is distributed (baking it into the task makes key rotation hard).

interop_binaries/src/bin/janus_interop_aggregator.rs
```diff
@@ -0,0 +1,323 @@
...
+            vec![AuthenticationToken::from(
+                collector_authentication_token.into_bytes(),
+            )],
+        ),
+        (1, _) => (Role::Helper, Vec::new()),
+        _ => return Err(anyhow::anyhow!("invalid \"aggregator_id\" value")),
+    };
+
+    let (hpke_config, private_key) = generate_hpke_config_and_private_key(
+        HpkeConfigId::from(0u8),
```
tgeoghegan, line 122: Teleporting over from a comment I made on the collector code... This works fine if we assume that HPKE config IDs are unique per-task, but I think that won't be true anymore once https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/289 lands in DAP.

interop_binaries/src/bin/janus_interop_collector.rs
divergentdave: Agreed, but maybe it should go in a separate crate, to keep janus_client slim for WASM users.

interop_binaries/tests/end_to_end.rs
```diff
@@ -0,0 +1,598 @@
...
+}
+
+/// RAII guard to ensure that child processes are cleaned up during test failures.
+struct ChildProcessCleanupDropGuard(Child);
+
+impl Drop for ChildProcessCleanupDropGuard {
+    fn drop(&mut self) {
+        if self.0.try_wait().unwrap().is_none() {
+            self.0.kill().unwrap();
+            self.0.wait().unwrap();
```
tgeoghegan, line 60: I think the call to `Child::wait` is unnecessary after `Child::kill`. The [docs](https://doc.rust-lang.org/std/process/struct.Child.html#method.kill) aren't really clear, but empirically I've found that no zombie is left behind after calling `Child::kill`, and the sentence "This is equivalent to sending a SIGKILL on Unix platforms." implies you don't need to wait.

interop_binaries/tests/end_to_end.rs
```diff
@@ -0,0 +1,598 @@
...
+    );
+
+    // Send a /internal/test/add_task request to the collector.
+    let collector_add_task_response = http_client
+        .post(format!(
+            "http://127.0.0.1:{}/internal/test/add_task",
+            collector_port,
+        ))
+        .header(CONTENT_TYPE, JSON_MEDIA_TYPE)
+        .json(&json!({
```
tgeoghegan, line 274:
```suggestion
        .json(&json!({
```
[`RequestBuilder::json` automatically sets `Content-Type: application/json`](https://docs.rs/reqwest/latest/src/reqwest/async_impl/request.rs.html#441-457) so I think quite a few calls to `RequestBuilder::header` can be deleted.

tgeoghegan (approved):
A few more nits but I think this is a sound implementation of https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html

interop_binaries/tests/end_to_end.rs
divergentdave: Makes sense, that way we rely on init to clean up the process once the test exits.

interop_binaries/tests/end_to_end.rs
```diff
@@ -0,0 +1,589 @@
...
+        port,
+    ))
+}
+
+/// RAII guard to ensure that child processes are cleaned up during test failures.
+struct ChildProcessCleanupDropGuard(Child);
+
+impl Drop for ChildProcessCleanupDropGuard {
+    fn drop(&mut self) {
+        if self.0.try_wait().unwrap().is_none() {
```
branlwyd, line 58: This code is racy: if the child process has not exited when the `try_wait()` call occurs, but exits before the call to `kill()`, the `unwrap()` on the return value to `kill()` will panic since `kill()` will return `Err(InvalidInput)`.

I think the intent is to kill the process if it's still running. If that's accurate, I recommend calling `kill()` unconditionally and ignoring the error if it is `InvalidInput`.

interop_binaries/tests/end_to_end.rs
```diff
@@ -0,0 +1,589 @@
...
+        if count == 0 {
+            return Ok(());
+        }
+        print!("{}", line);
+    }
+}
+
+/// Pass output from a child process's stderr pipe to eprint!(), so that it can be captured and
+/// stored by the test harness.
+async fn forward_stderr(stdout: ChildStderr) -> io::Result<()> {
```
branlwyd, line 81:
```suggestion
async fn forward_stderr(stderr: ChildStderr) -> io::Result<()> {
```

(mega-nit)

janus_core/src/time.rs
branlwyd: That's unfortunate. I suppose this is another artifact of the interplay between workspaces & features/how feature unification is implemented.

I think one of the solutions at https://nickb.dev/blog/cargo-workspace-and-the-feature-unification-pitfall might help us. I think our release builds also effectively use solution #2 as well since we specify the `--bin` flag to `cargo build` (but maybe specifying `-p` is also required?). Are you sure our release builds would encounter this pitfall?

In any case, I don't think we need to hold this PR up over this issue, but I'd definitely like to resolve it somehow. Would you mind filing an issue? I'd like to be able to use features w/o dependency resolution changing what features are enabled on a per-package basis; hopefully one of the blog solutions will work, though I'd be happier if we didn't have to go as far as introducing sub-workspaces.

branlwyd (changes requested):

janus_core/src/time.rs
divergentdave: I tested `cargo --bin aggregator` with and without `-p janus_server`, and the package specification is necessary to keep the feature from leaking over.

branlwyd (approved):

Issue #196, opened 2022-05-27
Retries in HTTP clients


tgeoghegan:
We currently use `reqwest::HttpClient` in the Janus client (to upload reports to aggregators), the aggregate job driver (for leaders to drive the aggregation protocol with helpers) and the collect job driver (for leaders to obtain helper aggregate shares). In all of these places, we should wire up retries on request failures with exponential backoffs.

tgeoghegan:
We can crib from `prio-server/facilitator` here.

BranLwyd:
(We should think through the semantics of retries if e.g. the server actually processed the request successfully, but that's somewhat beyond the scope of both this bug & implementations.)

BranLwyd:
We should also consider retries for database operations. Our database library (`tokio_postgres`) is not documented as supporting automatic retries. We should determine which errors from the DB server are "retryable" and automatically retry them, likely using logic similar to that used to retry "serialization" failures.

divergentdave:
On a related note, in the three places where our leader makes an HTTP request to the helper, we do not currently inspect the HTTP status code. The spec says the helper should respond with 200 OK in each case. We should check this and error out with relevant information, before attempting to decode the response body.

PR #379, opened 2022-08-10
Run e2e integration tests against Kubernetes cluster

tgeoghegan:


monolithic_integration_test/tests/janus.rs
```diff
@@ -1,39 +1,163 @@
...
+        let (leader_port, helper_port) = pick_two_unused_ports();
+        let (collector_hpke_config, collector_private_key) =
+            generate_test_hpke_config_and_private_key();
+        let (mut leader_task, mut helper_task) =
+            create_test_tasks(leader_port, helper_port, &collector_hpke_config);
+
+        let kubeconfig_path = env::var("JANUS_E2E_KUBE_CONFIG_PATH");
+        let kubectl_context_name = env::var("JANUS_E2E_KUBECTL_CONTEXT_NAME");
+        let leader_namespace = env::var("JANUS_E2E_LEADER_NAMESPACE");
+        let helper_namespace = env::var("JANUS_E2E_HELPER_NAMESPACE");
```
tgeoghegan, line 66: My thinking here is that developers at their workstations will do `cargo test` and run the tests against the in-process Janus, as will `janus` CI. But in `janus-ops`, we'll check out `janus`, set all these environment variables based on the Kind cluster that CI sets up, and run `cargo --package monolithic_integration_test --features kube-openssl`. However I'm open to other ways to solve this. We could also expose a feature in `monolithic_integration_test` that governs which way the test is run.

monolithic_integration_test/tests/janus.rs
```diff
@@ -1,39 +1,163 @@
...
+impl JanusPair {
+    fn in_cluster_aggregator_url(namespace: &str) -> Url {
+        Url::parse(&format!(
+            "http://aggregator.{namespace}.svc.cluster.local:80"
+        ))
+        .unwrap()
+    }
+
+    /// Set up a new Janus test instance with the provided task. If the environment variables listed
+    /// below are set, then the test instance is abcked by an already-running Kubernetes cluster.
```
divergentdave, line 36: typo:
```suggestion
    /// below are set, then the test instance is backed by an already-running Kubernetes cluster.
```

monolithic_integration_test/tests/janus.rs
```diff
@@ -1,39 +1,163 @@
...
+    /// Otherwise, an ephemeral, in-process instance of Janus is spawned. In either case, the Janus
+    /// aggregators' API endpoints will be available on the local loopback interface, at the URLs in
+    /// `Self.leader_task.aggregator_endpoints`.
+    ///
+    /// If connecting to the Kubernetes cluster API (k8s API, not DAP API) at an IP address over
+    /// HTTPS (e.g., "https://127.0.0.1:42356"), then the `monolithic_integration_test` package must
+    /// be built with the `kube-openssl` feature, as the default rustls can't validate IP addresses
+    /// in certificates.
+    /// e.g., `cargo test --features kube-openssl --packagep monolithic_integration_test`
```
divergentdave, line 45: typo fix:
```suggestion
    /// e.g., `cargo test --features kube-openssl --package monolithic_integration_test`
```

monolithic_integration_test/tests/janus.rs
divergentdave: I prefer this environment variable solution over a feature.

janus_core/src/test_util/kubernetes.rs
```diff
@@ -2,21 +2,92 @@
...
+            "--context",
+            &self.context_name,
+            "--namespace",
+            namespace,
+            "port-forward",
+            &format!("service/{service}"),
+            &format!("{local_port}:{service_port}"),
+        ];
+        debug!(?args, "Invoking kubectl");
+        let child = Command::new("kubectl")
```
branlwyd, line 68: Using this is inherently racy: the child process takes some time to set up port forwarding.

Maybe tail stdout for a while until you see the expected logging, or alternatively poke the port for a while until something is accepting connections on the port? (the former depends on `kubectl`'s output being stable, I believe the latter assumes that something will already be listening behind the forwarded port; maybe a cleverer solution is possible?)

janus_core/src/test_util/kubernetes.rs
```diff
@@ -77,9 +132,9 @@ impl Drop for EphemeralCluster {
                 "delete",
                 "cluster",
                 "--kubeconfig",
-                &self.kubeconfig_path.to_string_lossy(),
+                &self.cluster.kubeconfig_path.to_string_lossy(),
                 "--name",
-                &self.name,
+                &self.cluster.context_name,
```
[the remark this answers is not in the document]
branlwyd: I think the idea is idempotency: deleting a cluster twice should succeed both times. But from the `kind` tool's POV, the second delete just looks like the cluster doesn't exist at all. (https://kind.sigs.k8s.io/docs/user/quick-start/#deleting-a-cluster)

janus_core/src/test_util/kubernetes.rs
```diff
@@ -2,21 +2,92 @@
...
+        PortForwardGuard { child }
+    }
+}
 
 /// EphemeralCluster represents a running ephemeral Kubernetes cluster for testing. Dropping an
 /// EphemeralCluster will cause the associated Kubernetes cluster to be stopped & cleaned up.
 pub struct EphemeralCluster {
-    name: String,
-    kubeconfig_path: TempPath,
+    pub cluster: Cluster,
```
branlwyd, line 83: rather than making this `pub`, I suggest adding a getter that returns a reference to the `&Cluster`.

janus_core/src/test_util/kubernetes.rs
```diff
@@ -90,6 +145,22 @@ impl Drop for EphemeralCluster {
...
+pub struct PortForwardGuard {
+    child: Child,
+}
+
+impl Drop for PortForwardGuard {
+    fn drop(&mut self) {
+        debug!(?self.child, "dropping port forward");
+        // We kill and reap child `kubectl` processes here to ensure that we don't create zombies on
+        // test failures. This is normally a questionable practice, but acceptable since this is
+        // only used in test code.
```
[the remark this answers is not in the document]
branlwyd: I think waiting on the process post-`kill()` is reasonable: other than being more accurate to the comment, I think it would also be slightly cleaner if dropping the value didn't complete until the port forward (i.e. process) is actually gone, rather than just "signalled to die soon in a way that it can't(?) override".

janus_server/src/datastore.rs
```diff
@@ -138,6 +138,15 @@ impl<C: Clock> Datastore<C> {
         tx.tx.commit().await?;
         Ok(rslt)
     }
+
+    /// Write a task into the datastore.
+    pub async fn put_task(&self, task: &Task) -> Result<(), Error> {
```
branlwyd, line 143: I think this should be `test-util` since it always clones the given `Task` at least once. (non-test code _should_, and I believe _does_, wrap the task in an `Arc` & clone that)

janus_core/src/test_util/kubernetes.rs
```diff
@@ -90,6 +145,22 @@ impl Drop for EphemeralCluster {
     }
 }
 
+/// A guard on a running `kubectl port-forward` process. The process will be killed and reaped when
+/// the guard is dropped.
+pub struct PortForwardGuard {
```
branlwyd, line 152: Name this a `PortForward` since that's what it is -- the fact that it implements a "drop-guard" pattern is somewhat an implementation detail. (e.g., we don't call the type `ArcGuard`, we call it `Arc`.)

janus_core/src/test_util/kubernetes.rs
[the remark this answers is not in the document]
tgeoghegan: Good catch! I now separately track the context name and the Kind cluster name so we can reliably tear it down.

branlwyd (changes requested):

divergentdave (approved):

janus_core/src/test_util/kubernetes.rs
tgeoghegan: My plan for dealing with this race is to resolve https://github.com/divviup/janus/issues/196. IMO polling on a port or watching for logs doesn't reliably solve the problem--the process might still not be ready to handle messages. We could block until the Kubernetes health or liveness probe on the pod or service succeeds, but that doesn't actually help with the race with `kubectl port-forward`. I think the only way to do this reliably is for the HTTP clients to do reasonable retries if they encounter connection refused errors. Hence I argue for resolving this by fixing #196 instead of doing something in the test harness.

branlwyd: I think looking to the logging does reliably solve the problem of detecting when the port forward is "ready" as far as `kubectl port-forward` is concerned, since it will only log once the port is "ready". Attempting to connect to the port would work to detect readiness too, as long as (a) there is some process in K8S to listening on the forwarded port (false negative) and (b) nothing else is listening on that port yet (false positive).

If the underlying process running in K8S accepts connections but isn't ready to handle them, that's a bug in the underlying process -- we should just fix the bug in the process, which we can do since we are writing the software under test. (That is, a correct server won't start listening to accept requests until it can handle them.) More to the point, I'd like to not have to add retries to all of our test-setup requests.

tgeoghegan: What I'm worried about is that no matter what kind of preflight checks we do on the availability and functionality of the `port-forward`, the component that ends up making HTTP requests to the forwarded port must be robust against transient failures. Otherwise you end up with TOCTOU races. So if we have to deal with that robustness problem anyway, why not spare ourselves the hassle of writing the preflight checks (which would be brittle anyway -- the string that `kubectl` logs is not an API contract)?

branlwyd: > the component that ends up making HTTP requests to the forwarded port must be robust against transient failures

My argument is that:
(a) this is true for non-test request code running in production (though even then a direct retry may or may not be the right answer, depending on the request)
(b) ...but it's definitely not true for test-only code (such as code adding the task under test & doing other test setup), which are generally going to be talking over IPC (for in-process or Kind-based tests), which we can assume to be reliable against transient communication failures. An eventual test against an actually-deployed-to-a-real-GCP-K8S-cluster would be talking over the Internet, but IMO even then I'd like to accept an occasional flake vs trying to add retries to every bit of test-setup code we end up writing -- integration tests against "real" deployed environments tend to be inherently flaky, anyway.

This code also doesn't check that the port-forwarding actually succeeded. If it fails for any reason, this code won't notice. In the worst case, if the forwarding failed because the port to forward was already bound, later code might end up sending requests to the wrong service! I think noticing failure would require writing something like the log-check I suggest--which would give you the wait-until-port-is-forwarded check, too.

tgeoghegan: We're not going to fail to notice that the port-forwarding failed, because the subsequent test will fail because it can't connect to anything to upload reports. I implemented a polling loop that waits until the forwarded port can be connected to and is writable, since that's the observable effect of `kubectl port-forward` we care about. I believe parsing the output of `kubectl` is too brittle to be of much value beyond checking that we can connect to the port.

branlwyd: Right: the test failing in this case is bad, because it can happen even if everything is working correctly & the test code happened to run fast enough to get to the first request before the port-forward actually took effect. That's why I want to check that the port-forward completes before continuing.

monolithic_integration_test/tests/janus.rs
branlwyd: Same: I think we'll need to specify the data in the env vars somehow, so even if we went with a feature, we'd still need an env var (or moral equivalent) to pass along the connection data.

branlwyd (approved):

janus_core/src/test_util/kubernetes.rs
tgeoghegan: OK, I added a check to the TCP port polling loop that verifies that the child process is still running. Now, if we fail to connect with `ConnectionRefused`, and the child process has exited, we'll immediately bail out of the test. I believe this achieves what you want.

PR #380, opened 2022-08-11
Fix usage of Url::join.

BranLwyd:
There are fixes to two separate unexpected behaviors of the Url::join method:
  * If a URL which does not end with a slash is joined with something, the final portion of the path is removed from the URL before the joined portion is added.
  * If a URL is joined with something which starts with a slash (i.e. an absolute path), the original path is entirely replaced by the joined absolute path. (This one isn't documented!)

I also update the Janus-Janus monolithic integration test to exercise the previously-untested case of an aggregator endpoint which has a nontrivial path.

### DIFF
--- a/janus_client/src/lib.rs
+++ b/janus_client/src/lib.rs
@@ -57,9 +57,18 @@ impl ClientParameters {
     /// Creates a new set of client task parameters.
     pub fn new(
         task_id: TaskId,
-        aggregator_endpoints: Vec<Url>,
+        mut aggregator_endpoints: Vec<Url>,
         min_batch_duration: Duration,
     ) -> Self {
+        // Ensure provided aggregator endpoints end with a slash, as we will be joining additional
+        // path segments into these endpoints & the Url::join implementation is persnickety about
+        // the slash at the end of the path.
+        for url in &mut aggregator_endpoints {
+            if !url.as_str().ends_with('/') {
+                url.set_path(&format!("{}/", url.path()));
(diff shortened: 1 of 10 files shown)


divergentdave (approved):

tgeoghegan (changes requested):
We should put a similar check in [`ClientParameters::new`](https://github.com/divviup/janus/blob/9d33c1f58b14c6fce535cd96b9720bb1dc8b3838/janus_client/src/lib.rs#L58), since `Url::join` is also used to construct the URLs that the client will access for HPKE configs and uploads.

janus_client/src/lib.rs
```diff
@@ -57,9 +57,18 @@ impl ClientParameters {
...
         min_batch_duration: Duration,
     ) -> Self {
+        // Ensure provided aggregator endpoints end with a slash, as we will be joining additional
+        // path segments into these endpoints & the Url::join implementation is persnickety about
+        // the slash at the end of the path.
+        for url in &mut aggregator_endpoints {
+            if !url.as_str().ends_with('/') {
+                url.set_path(&format!("{}/", url.path()));
+            }
+        }
```
tgeoghegan, line 70: Meganit: we could put an extension trait on `Url` in `janus_core` that defines a method `Url::normalize` that does this. Then again this code probably won't change so we can live with a little copy-pasta.

tgeoghegan (approved):

Issue #235, opened 2022-06-13
Consider aborting aggregation jobs entirely on certain errors from helper


branlwyd:
When in the leader position, the aggregation job driver will currently retry (until normal too-many-retries logic causes us to stop) on _any_ HTTP failure. Certain HTTP error codes and/or DAP error codes should probably cause us to abandon the aggregation job immediately. We may also want to give up if the helper responds with an unknown/unexpected content type.

tgeoghegan:
See also #1180 which has a lot of discussion around avoiding scheduling aggregation jobs that can't succeed. We need to articulate state machines for reports, aggregation jobs, collections and aggregate shares, see where they intersect and then implement that coherently, which might also feed back into DAP itself.

PR #382, opened 2022-08-11
Check HTTP status codes from other aggregator

divergentdave:
This adds a new error variant to the aggregator's `Error` type, and checks status codes received from `/aggregate` and `/aggregate_share` endpoints.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -144,12 +144,15 @@ peer checksum: {peer_checksum:?} peer report count: {peer_report_count}"
     /// HPKE failure.
     #[error("HPKE error: {0}")]
     Hpke(#[from] janus_core::hpke::Error),
-    /// Error handling task parameters
+    /// Error handling task parameters.
     #[error("invalid task parameters: {0}")]
     TaskParameters(#[from] crate::task::Error),
-    /// Error making an HTTP request
+    /// Error making an HTTP request.
     #[error("HTTP client error: {0}")]
     HttpClient(#[from] reqwest::Error),
+    /// HTTP server returned an error status code.
+    #[error("HTTP response status {0}")]
+    Http(StatusCode),
     /// An error representing a generic internal aggregation error; intended for "impossible"
     /// conditions.
     #[error("internal aggregator error: {0}")]
@@ -181,6 +184,7 @@ impl Error {
             Error::Hpke(_) => "hpke",
             Error::TaskParameters(_) => "task_parameters",
             Error::HttpClient(_) => "http_client",
+            Error::Http(_) => "http",
             Error::Internal(_) => "internal",
         }
     }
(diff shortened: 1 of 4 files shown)


janus_server/src/aggregator/aggregate_share.rs
```diff
@@ -193,6 +193,11 @@ impl CollectJobDriver {
             .body(req.get_encoded())
             .send()
             .await?;
+        let status = response.status();
+        if !status.is_success() {
+            // TODO(#381): Attempt to decode a problem details response.
+            return Err(Error::Http(status));
+        }
```
tgeoghegan, line 200: In cases like this where we treat anything in the 4xx-5xx range uniformly as an error, I think we could use [`reqwest::Response::error_for_status`](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.error_for_status), which would save us a little code and should be handled by the existing `aggregator::Error::HttpClient` case which wraps `request::Error`.
```suggestion
            .await?
            .error_for_status()?;
```
Though maybe we shouldn't do this, because it's not apparent that `request::Error` lets you get at the response body, which we will eventually want to decode. Really what I'd like is for this to be harder to get wrong. Maybe we should write a wrapper around `reqwest::RequestBuilder` so we can do something like `RequestBuilder::send_with_expected_status(request, 201)`, and then you get back a `Result<Response, Error>`, which will be an error if the response status was anything but the expected HTTP status.

But we should just go ahead with your change for now.

tgeoghegan (approved):
This will solve our present problem, but I left a couple suggestions for you to consider (or perhaps file issues for).

janus_server/src/aggregator.rs
```diff
@@ -1996,6 +2000,7 @@ where
                     Err(Error::Url(_)) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
                     Err(Error::Message(_)) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
                     Err(Error::HttpClient(_)) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
+                    Err(Error::Http(_)) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
```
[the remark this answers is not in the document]
divergentdave: Thinking about this some more, this line should be unreachable, as this variant is only constructed in the aggregate job driver and the collect job driver. With that said, I'll leave it as mapping to 500, as that's good enough.

janus_server/src/aggregator/aggregate_share.rs
BranLwyd: I think we are going to want to eventually attach different behaviors to different error types from the peer aggregator; at least, some errors will be immediately terminal, some will allow a retry, based on the content of the problem document. (This is tracked in #235.)

That said, I think any abstraction we build should capture that ability (or at least not need to be unwound once we decide to build out #235).

PR #385, opened 2022-08-11
Increment versions to 0.1.5.

BranLwyd:
Based on #339.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1592,7 +1592,7 @@ checksum = "1aab8fc367588b89dcee83ab0fd66b72b50b72fa1904d7095045ace2b0c81c35"
 
 [[package]]
 name = "janus_client"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "assert_matches",
  "http",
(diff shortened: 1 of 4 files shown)


divergentdave (commented):
The version specified in `janus_client`'s dependency on `janus_core` should be bumped as well, otherwise LGTM.

BranLwyd:
I removed the version specification of `janus_core` in `janus_client`'s Cargo.toml entirely -- AFAICT it's not needed, since we specify a direct path to the crate's source code, which we can do since it's in the same repo. (This also matches how other `janus_foo` crates specify their dependencies on one another.)

divergentdave:
Removing the version won't work, as I now get the following error. The other crates can get away with no version because they all have `publish = false`.

```
$ cargo publish --dry-run -p janus_client
    Updating crates.io index
error: all dependencies must have a version specified when publishing.
dependency `janus_core` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```

tgeoghegan (approved):

BranLwyd:
OK. I specified version `0.1` (rather than `0.1.5`) in an attempt to reduce how often we need to bump this version number.

divergentdave (approved):

PR #408, opened 2022-08-18
Container-based interop testing

### LINKED ISSUE #394: Revisit workspace feature coalescing
Following up from https://github.com/divviup/janus/pull/375#discussion_r944826764:

We would like to be able to re-enable "test-util" feature flags in "above the fold" dependencies in crates that are only used for testing purposes, without affecting builds used in production. There are two ways we could accomplish this: either specifying `--package` for every production build (i.e., in the Dockerfile) or creating separate workspaces, so that we have isolated target directories and independent dependency resolution.

branlwyd:
* We build the interop test Janus container during build time, so that
   we can embed it in the interop test binary. This allows the interop
   testcontainer library to load the container at runtime as needed.
 * If we published a container (or with some magic to check if we want
   to try pulling from a private repository?), we could test with a
   published container version by default and trigger this behavior only
   with a feature/environment variable. This would eliminate the
   buildtime hit. (Or, if the buildtime hit is acceptable, we might even
   want to run different versions against the code under test as a test
   against regressions against previous versions of our code.)
 * I build a shim container for Daphne; this is a temporary solution to
   allow easy networking setup & task configuration. The next step is to
   contribute to Daphne a Dockerfile to build something like this
   container, then set up our repository to build (or, eventually, load)
   that container instead.
 * My next steps are the contribution to Daphne I mention above, and
   then using this so that the end-to-end test uses the real interop
   containers (extending it to exercise the Dockerfiles as well).
 * I moved the non-`/` serving path regression test to the new interop
   API end-to-end test, since it's now easier to express there. But we
   may want to move it back once the end-to-end test uses the
   containers too, to keep this test API-centric. If it's controversial,
   I'll figure out a way to move the regression test back in this PR.

Also sets `-p` on builds to avoid workspace feature coalescence enabling unexpected features in our binary builds (see #394 for more details).

Closes #394.


interop_binaries/build.rs
```diff
@@ -0,0 +1,62 @@
...
+        // Make a best-effort attempt to clean up after ourselves.
+        let rmi_output = Command::new("docker")
+            .args(["rmi", image_id])
+            .output()
+            .expect("Failed to execute `docker rmi` for interop aggregator");
+        assert!(
+            rmi_output.status.success(),
+            "Docker image removal of interop aggregator failed:\n{}",
+            String::from_utf8_lossy(&build_output.stderr)
+        );
```
divergentdave, line 60: This is causing problems for me in rust-analyzer. I see the following error.

```
error: failed to run custom build command for `interop_binaries v0.1.0 (/home/david/Code/ppm/janus/interop_binaries)`

Caused by:
  process didn't exit successfully: `/home/david/Code/ppm/janus/target/debug/build/interop_binaries-b4d5fa1e6926891a/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at 'Docker image removal of interop aggregator failed:
  ', interop_binaries/build.rs:56:9
  stack backtrace:
     0: rust_begin_unwind
               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/std/src/panicking.rs:584:5
     1: core::panicking::panic_fmt
               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/panicking.rs:142:14
     2: build_script_build::main
               at ./build.rs:56:9
     3: core::ops::function::FnOnce::call_once
               at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/ops/function.rs:248:5
  note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

I wonder if rust-analyzer is running the build script multiple times concurrently, and the later one stomps on the same image ID (since all the layers could be reused) and fails to delete it?  Perhaps we should give it a name and a randomly generated tag, and then delete by that tag.

monolithic_integration_test/build.rs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        use std::{env, process::Command};
+
+        // Build & save off a container image for the test Daphne instance.
+        // Note: `docker build` has an `--output` flag which writes the output to somewhere, which
+        // may be a tarfile. But `docker build --output` only exports the image filesystem, and not
+        // any other image metadata (such as exposed ports, the entrypoint, etc), so we can't easily
+        // use it.
+        let build_output = Command::new("docker")
+            .args(["build", "--file=Dockerfile.test_daphne", "--quiet", "."])
```
divergentdave, line 12: Similar to the other build script, we should tell Cargo to rerun this when the Dockerfile or any of the artifacts directory changes.

divergentdave (changes requested):

Dockerfile.interop_aggregator
```diff
@@ -10,13 +10,13 @@ COPY janus_server /src/janus_server
...
-RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/src/target cargo build --release --bin janus_interop_aggregator && cp /src/target/release/janus_interop_aggregator /janus_interop_aggregator
+RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/src/target cargo build --bin janus_interop_aggregator && cp /src/target/release/janus_interop_aggregator /janus_interop_aggregator
 
 FROM postgres:14-bullseye
 RUN mkdir /logs
 RUN apt-get update && apt-get install -y supervisor
 COPY interop_binaries/supervisord.conf /supervisord.conf
 COPY --from=builder /src/db/schema.sql /db/schema.sql
 COPY --from=builder /janus_interop_aggregator /janus_interop_aggregator
-EXPOSE 8080
```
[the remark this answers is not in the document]
branlwyd: I'm fine using the same port for integration & other uses -- I switched back to 8080.

I think we could have gotten away with using 80, BTW, even with rootless containers, with some capability twiddling. This seems to be borne out by https://stackoverflow.com/questions/42830495/side-effects-of-allowing-a-container-to-listen-on-port-80.

Dockerfile.interop
```diff
@@ -11,7 +11,7 @@ COPY janus_server /src/janus_server
 COPY monolithic_integration_test /src/monolithic_integration_test
 COPY interop_binaries /src/interop_binaries
 COPY db/schema.sql /src/db/schema.sql
-RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/src/target cargo build --release --bin $BINARY && cp /src/target/release/$BINARY /$BINARY
+RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/src/target cargo build --bin $BINARY && cp /src/target/release/$BINARY /$BINARY
```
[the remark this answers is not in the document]
branlwyd: Ugh, I knew I shouldn't have modified this as one of the last things I did.

Still, this worked on my own workstation yesterday; I guess I was pulling from my local build cache somehow? i.e. further updates to the interop binaries would not have been reflected in the containers which were built.

interop_binaries/build.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        // Build & save off a container image for the interop_aggregator.
+        // Note: `docker build` has an `--output` flag which writes the output to somewhere, which
+        // may be a tarfile. But `docker build --output` only exports the image filesystem, and not
+        // any other image metadata (such as exposed ports, the entrypoint, etc), so we can't easily
+        // use it.
+        let build_output = Command::new("docker")
+            .args([
+                "build",
+                "--file=Dockerfile.interop_aggregator",
```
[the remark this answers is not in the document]
branlwyd: I pulled in the Dockerfile itself & the dependencies it copies into these directives. (I had naively thought that not specifying directives would cause the script to be rerun if anything in the workspace changed, but re-reading the docs, I think the default is to re-run the script if anything in the package changes -- which isn't what we want.)

[While doing this, I noticed we were copying `monolithic_integration_test` into the build layer for the interop binaries; I removed this as it doesn't appear to be necessary, but calling it out in case this is undesirable.]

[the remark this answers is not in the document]
divergentdave: That was generally required because cargo would read `Cargo.toml`, see the list of workspace members, and read `Cargo.toml` from each subdirectory, plus check for `src/lib.rs` or `src/main.rs`. I'm not sure if we can get away without copying it in or rewriting files, but if so, sounds good.

interop_binaries/build.rs
branlwyd: Interesting -- I didn't see anything like that. FWIW, if I change the build script to attempt deletion _twice_, I do see a similar crash on the second attempt, so I also suspect that a concurrency issue is at play.

Since image-removal is best-effort, I think the best solution is just to ignore failures. If the image has already been deleted, great! (Indeed, I think the common case of running `cargo test` will build the image/save it off/delete it during build, then load it again during tests, so the image still ends up in the local docker.)

However, I do wonder if this could run into trouble with concurrent builds in the same system: the `docker rmi` in one build could happen before the `docker save` in another build, causing the latter build to fail. But I think "parallel builds of the same code on the same machine" isn't something we need to support well. I am curious if you're still running into trouble with this change in place, however.

interop_binaries/src/testcontainer.rs
```diff
@@ -0,0 +1,89 @@
...
+    io::{Read, Write},
+    process::{Command, Stdio},
+    sync::Mutex,
+    thread,
+};
+use testcontainers::{core::WaitFor, Image};
+
+// INTEROP_AGGREGATOR_IMAGE_BYTES / interop_aggregator.tar are created by this package's build.rs.
+const INTEROP_AGGREGATOR_IMAGE_BYTES: &[u8] =
+    include_bytes!(concat!(env!("OUT_DIR"), "/interop_aggregator.tar"));
```
tgeoghegan, line 15: How big is the interop aggregator image? If we pull it in with `include_bytes!`, does it all get loaded into memory at once, or will the `child_stdin.write_all(INTEROP_AGGREGATOR_IMAGE_BYTES)` call below gracefully read a chunk of it at a time?

interop_binaries/src/testcontainer.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        Self {}
+    }
+}
+
+impl Image for Aggregator {
+    type Args = ();
+
+    fn name(&self) -> String {
+        // This works around a quirk in testconatiners: it will always generated the image name
```
tgeoghegan, line 71:
```suggestion
        // This works around a quirk in testcontainers: it will always generate the image name
```

tgeoghegan (commented):
I am a little concerned about the implications here for build time. I wonder if there's some config file we can drop in the repo so that we could have rust-analyzer opt out of building the container images? Or maybe there's a Cargo.toml section that rust-analyzer honors? Additionally, maybe invoking `cargo test` on one's workstation shouldn't have to build all this stuff. It's kinda painful having to wait on container builds when I might just quickly want to see if a unit test passes.

interop_binaries/src/testcontainer.rs
divergentdave: Oh dear, the tar file from when I built it is 750MB! It would get mapped in as part of the `.rdata` section, I think. The two biggest contributors are the final layer, with `janus_interop_aggregator` only, and the layer that installed postgres, perl, etc.

tgeoghegan: Streaming the bytes from a `.tar` on disk and then into `child_stdin` seems like it ought to be about the same as streaming the bytes from the binary on disk, but yeah, maybe 750 MB in `.rdata` is a bit much.

branlwyd: (that said, no objection from me if we want to try to shrink these binaries/images somehow. ~330 MB for the amount of functionality in an aggregator, or ~750 MB for aggregator + DB, is wild to me.)

interop_binaries/build.rs
divergentdave: I commented out the assertion on the `docker rmi` status and that got things working again.

branlwyd:
> I am a little concerned about the implications here for build time. I wonder if there's some config file we can drop in the repo so that we could have rust-analyzer opt out of building the container images?

There is a rust-analyzer setting for this: `rust-analyzer.cargo.buildScripts.enable` (https://rust-analyzer.github.io/manual.html). I can't imagine setting this will substantively change the output of the (IIUC) `cargo check` that rust-analyzer does to get its symbols.

I'm not sure if we can configure our repo to have rust-analyzer do this by default (I'm going to do a little more digging here), but we can set it up in VS Code.


> Additionally, maybe invoking `cargo test` on one's workstation shouldn't have to build all this stuff. It's kinda painful having to wait on container builds when I might just quickly want to see if a unit test passes.

Today running only unit tests would be e.g. `cargo test -p janus_server`. An unmitigated `cargo test` will run both unit & integration tests. That said, maybe we don't want to run integration tests by default? We could hide it behind a feature. I don't object to doing this if desirable, as long as there is some way for contributors to run these tests on their workstation/pre-CI -- otherwise, if most of our behavioral testing is in integration tests, contributors will be likely to run into CI-only errors.

branlwyd:
(also it's pretty surprising to me that rust-analyzer runs build scripts by default -- if we were e.g. packaging some C library, every rust-analyzer run would try to build the library!)

branlwyd:
I just pushed a change to specify `-p janus_server` for our release builds, too. This is important for a couple of reasons:

1) It lets us use features like `test-util` in e.g. `monolithic_integration_test` without having the workspace convalescence feature also enable them while building binaries in totally different packages.

2) More importantly, I believe it skips trying to build packages in the workspace that aren't included in the binary, which is even more important now that building some packages will end up building a container.

I think our `cargo clippy` & `cargo check` are also running the build script; I guess that's a good idea, but we don't need an actual image to be built. I'm going to investigate having our CI invocations of `clippy`/`check` skip building the image.

tgeoghegan (approved):
I say: let's take the change, and see where the pain is, and we can iterate and improve on this stuff later.

monolithic_integration_test/src/lib.rs
```diff
@@ -2,5 +2,40 @@
 
 #[cfg(feature = "daphne")]
 pub mod daphne;
-#[cfg(feature = "janus")]
 pub mod janus;
+
+lazy_static::lazy_static! {
+    static ref CONTAINER_CLIENT: testcontainers::clients::Cli =
+        testcontainers::clients::Cli::default();
```
divergentdave, line 9: FYI, the Drop implementation of `Cli`'s internals is responsible for cleaning up Docker networks, so this will leak them. We could accept this, as they just represent some BoltDB rows and iptables rules. Alternately, we could create a new client object at the top of each test, or use a global `RwLock<Weak<Cli>>` with lazy initialization.

janus_core/src/test_util/kubernetes.rs
```diff
@@ -104,7 +104,7 @@ impl Cluster {
         .await
         .expect("timeout waiting for forwarded port to become writable");
 
-        PortForward { child }
+        PortForward { local_port, child }
```
divergentdave, line 107: I think this is a pre-existing rustdoc error, but the doc comment for this method needs to be updated to refer to `PortForward`.

interop_binaries/src/testcontainer.rs
divergentdave: I checked out 10c92e7, cleaned my target directory, and triggered the OOM killer a few times when trying to build, either with `cargo build` or `cargo test`. For example:

```
error: could not compile `interop_binaries`

Caused by:
  process didn't exit successfully: `rustc --crate-name interop_binaries --edition=2021 interop_binaries/src/lib.rs --error-format=json --json=diagnostic-rendered-ansi,artifacts,future-incompat --crate-type lib --emit=dep-info,metadata,link -C embed-bitcode=no -C debuginfo=2 --cfg 'feature="testcontainer"' -C metadata=fd8a365b13b98fba -C extra-filename=-fd8a365b13b98fba --out-dir /home/david/Code/ppm/janus/target/debug/deps -C incremental=/home/david/Code/ppm/janus/target/debug/incremental -L dependency=/home/david/Code/ppm/janus/target/debug/deps --extern ... -L native=/home/david/Code/ppm/janus/target/debug/build/ring-bce3495b5618ba54/out` (signal: 9, SIGKILL: kill)
```

```
kernel: Out of memory: Killed process 287248 (rustc) total-vm:9795372kB, anon-rss:3633724kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:13796kB oom_score_adj:0
```

```
kernel: Out of memory: Killed process 284952 (rustc) total-vm:25262432kB, anon-rss:9123824kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:27788kB oom_score_adj:0
```

Building succeeded eventually after some retries. I figure the SIGKILL in the CI logs is due to the same cause. The `interop_aggregator.tar` file is 750M, as before. I figure rustc and LLVM are both making several copies of the byte array as it passes through the pipeline.

How about we add `RUSTFLAGS="-C debuginfo=0"` in the build.rs invocation, via a new argument?

divergentdave (commented):
LGTM, my one remaining blocker is a solution for OOM-killer in CI

monolithic_integration_test/src/lib.rs
branlwyd: I think you're correct, but I'm going to leave this cleanup for now. `Container` instances include an internal reference to the client that created them, so I think I'd need to do some refactoring to make the lifetimes work out.

branlwyd:
I didn't see OOMs on my own workstation, but I'm pretty sure the CI failure is due to OOM.

I compressed the images with `zstd`, which is fast & has an excellent compression ratio for its speed -- at the cost of a few seconds of buildtime, it compresses the image to about ~25% of the original size. Anecdotally, I don't really notice a difference in buildtime or test-run time.

branlwyd:
OK, I think this is good to go, or at least small enough that it doesn't regularly crash our CI actions. 😆

I forgot to mention in the commit description that the approaches used to minimize the binary size came from https://github.com/johnthagen/min-sized-rust.

divergentdave (approved):

Issue #392, opened 2022-08-12
Audit `tracing` spans for large or secret objects


tgeoghegan:
There's some places where we include an entire `Task` structure in tracing spans, which can yield rather large messages. Additionally, some of those objects may contain secret values like HPKE private keys, which we never, ever want to log. We should audit our tracing spans to ensure that no values that are too large or sensitive get logged.

tgeoghegan:
It would be nice if there was some way to tag or taint structs that we know to contain secrets (like `HpkePrivateKey`) so that they are unloggable. Perhaps we can provide [`tracing::Value`](https://docs.rs/tracing/latest/tracing/trait.Value.html) implementations that panic?

divergentdave:
The instrument macro defaults to recording arguments with `tracing::field::Value`, with a fallback to `std::fmt::Debug`, and options to use `std::fmt::Display` instead. The simplest solution would be to customize the Debug and Display implementations for any secret struct, so that they are opaque when printed. Implementing `Value` on `HpkePrivateKey` would not have any effect on how `Task` gets recorded in a span, as `Task` is recorded as a string, using the Debug fallback, and will not touch `Value::record()`. Moreover, `Value` is a sealed trait, so we can't implement it, and I don't think tracing supports compound objects in fields.

divergentdave:
Should we exclude `CollectReq.agg_param` from logging, in anticipation of Poplar1? It could include an ASCII-encoded aggregate result, which may make it sensitive enough to exclude.

branlwyd:
I'm OK excluding it right now:
(a) prio3 doesn't use agg param at all, so we don't lose any logging usefulness currently.
(b) IMO an aggregate result is sensitive enough that we probably don't want to log it
(c) if later VDAFs have an aggregation parameter that is both non-trivial & non-sensitive & we decide we want to log them, we can always change our logging strategy at that point.

divergentdave:
Our log messages look better now, closing this out.

PR #410, opened 2022-08-18
Separate secret information from Debug implementations

divergentdave:
This makes `impl Debug for HpkePrivateKey` opaque, and introduces new wrapper types for verification keys. HPKE private keys will no longer be logged as part of tasks anymore. Verification keys were not being logged, but these new types will prevent accidentally logging them in the future. This is the bottom-up half of #392.


janus_core/src/hpke.rs
```diff
@@ -126,6 +129,12 @@ impl FromStr for HpkePrivateKey {
     }
 }
 
+impl Debug for HpkePrivateKey {
```
branlwyd, line 132: Any reason not to use `#[derivative(Debug)]` and mark the desired fields `#[derivative(Debug="ignore")]`?

janus_server/src/task.rs
```diff
@@ -137,6 +137,44 @@ enum VdafSerialization {
     FakeFailsPrepStep,
 }
 
+/// A verification key for a VDAF, with a variable length. It must be kept secret from clients to
+/// maintain robustness, and it must be shared between aggregators.
+#[derive(Clone, PartialEq, Eq)]
+pub struct VerifyKeyDynamicSize(Vec<u8>);
```
branlwyd, line 143: I don't like that we have two types for verification keys based on whether we happen to know the expected size of the verification key. The names are also quite a mouthful. Also, `VerifyKeyDynamicSize` will never actually be used as a verification key -- using it requires converting to a `VerifyKeyStaticSize` first, since all users of verification keys are generic over the length of the verify key.

How about we replace `VerifyKeyDynamicSize` by a purpose-generic type called e.g. `SecretBytes` whose purpose is to hold sensitive bytes of unknown purpose. Then rename `VerifyKeyStaticSize` to `VerifyKey`.

This IMO improves the naming a lot. A downside is that we no longer know the purpose of an arbitrary `SecretBytes` value. However, I think this downside is mitigated by the fact that we are storing these values in fields whose names describe their purpose, and always converting these values into a "purposeful" type (`VerifyKey`) before use -- once `Task` fields are private, I think the verify key getter should return a `VerifyKey` by performing the conversion as necessary, with the fact that the bytes are stored as `SecretBytes` inside the struct being an implementation detail. WDYT?

branlwyd (changes requested):

janus_server/src/task.rs
```diff
@@ -359,6 +345,25 @@ impl Task {
...
+    /// other aggregators.
+    ///
+    /// # Errors
+    ///
+    /// If the verify key is not the correct length as required by the VDAF, an error will be
+    /// returned.
+    pub fn primary_vdaf_verify_key<const L: usize>(&self) -> Result<VerifyKey<L>, Error> {
+        // We can safely unwrap this because we maintain an invariant that this vector is
+        // non-empty.
+        let secret_bytes = self.vdaf_verify_keys.first().unwrap();
```
branlwyd, line 359: This changes from `last()` to `first()` to determine the primary verification key. OTOH, I think that's fine--it's convention, and we currently don't know how to rotate these values anyway, so no one is relying on the convention not changing. :-)

branlwyd (approved):

janus_server/src/task.rs
divergentdave: Yeah, I was a bit unnerved by that, but our table constraints currently enforce only one verification key. Good time to get our story straight.

tgeoghegan (approved):

PR #411, opened 2022-08-19
Trim down overly verbose Debug output

divergentdave:
This makes the following changes to slim down Debug output:

- Skip logging OpenTelemetry SDK instruments. These are always going to be the same, and they are only included in structs or function arguments for performance.
- Skip logging a Tokio channel receiver in one location.
- Modify Debug implementations to print out `Url` objects with their Display formatting, for conciseness.

Examples of "before" output:

```
"job_creation_time_recorder":"ValueRecorder(SyncInstrument { instrument: SyncInstrument { instrument: Instrument { descriptor: Descriptor { name: \"janus_job_creation_time\", instrument_kind: ValueRecorder, number_kind: F64, config: InstrumentConfig { description: Some(\"Time spent creating aggregation jobs.\"), unit: Some(Unit(\"seconds\")), instrumentation_library: InstrumentationLibrary { name: \"aggregation_job_creator\", version: None } }, attribute_hash: 3686441284350234660 }, meter: \"Accumulator\" } }, _marker: PhantomData })"

"shutdown":"Receiver { inner: Some(Inner { state: State { is_complete: false, is_closed: false, is_rx_task_set: false, is_tx_task_set: false } }) }"

aggregator_endpoints: [Url { scheme: \"http\", cannot_be_a_base: false, username: \"\", password: None, host: Some(Domain(\"localhost\")), port: Some(22718), path: \"/\", query: None, fragment: None }, Url { scheme: \"http\", cannot_be_a_base: false, username: \"\", password: None, host: Some(Domain(\"aggregator.kind-ci-pollux.svc.cluster.local\")), port: None, path: \"/\", query: None, fragment: None }]
```

New URL output:

```
aggregator_endpoints: ["http://leader_endpoint/", "http://helper_endpoint/"]
```

This is part of #392.

### DIFF
--- a/janus_server/src/aggregator/aggregate_share.rs
+++ b/janus_server/src/aggregator/aggregate_share.rs
@@ -12,6 +12,7 @@ use crate::{
     task::{Task, PRIO3_AES128_VERIFY_KEY_LENGTH},
     task::{VdafInstance, DAP_AUTH_HEADER},
 };
+use derivative::Derivative;
 use futures::{future::BoxFuture, try_join};
 use http::header::CONTENT_TYPE;
 #[cfg(test)]
@@ -29,13 +30,15 @@ use prio::{
         Aggregatable,
     },
 };
-use std::{fmt::Debug, sync::Arc};
+use std::sync::Arc;
 use tracing::{debug, error, warn};
 
 /// Drives a collect job.
-#[derive(Debug)]
+#[derive(Derivative)]
+#[derivative(Debug)]
 pub struct CollectJobDriver {
     http_client: reqwest::Client,
+    #[derivative(Debug = "ignore")]
     job_cancel_counter: BoundCounter<u64>,
 }
 
--- a/janus_server/src/task.rs
+++ b/janus_server/src/task.rs
@@ -11,7 +11,10 @@ use janus_core::{
 };
 use ring::constant_time;
 use serde::{de::Error as _, Deserialize, Deserializer, Serialize, Serializer};
-use std::collections::HashMap;
+use std::{
+    collections::HashMap,
+    fmt::{self, Formatter},
+};
 use url::Url;
 
 /// HTTP header where auth tokens are provided in inter-aggregator messages.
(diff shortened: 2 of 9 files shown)


janus_server/src/aggregator/aggregate_share.rs
```diff
@@ -29,13 +30,15 @@ use prio::{
...
-use std::{fmt::Debug, sync::Arc};
+use std::sync::Arc;
 use tracing::{debug, error, warn};
 
 /// Drives a collect job.
-#[derive(Debug)]
+#[derive(Derivative)]
+#[derivative(Debug)]
 pub struct CollectJobDriver {
     http_client: reqwest::Client,
```
branlwyd, line 40: Q: do `reqwest::Client`s emit useful information in their `Debug` impls, to the point that we'd want to log them every time we render a `CollectJobDriver`?

janus_server/src/task.rs
```diff
@@ -323,6 +327,14 @@ impl Task {
     }
 }
 
+fn fmt_vector_of_urls(urls: &Vec<Url>, f: &mut Formatter<'_>) -> fmt::Result {
```
branlwyd, line 330: I sorta don't like that this duplicates code with the same function in `janus_client`'s lib.rs, but it's simple enough that I don't think this warrants what would be necessary to provide it to both w/o also exposing it publicly from our published crates (i.e. creating a new "utility" unpublished crate that both crates could rely on).

branlwyd (approved):
LGTM, one Q around whether we can cut down debug output further in one place.

janus_server/src/aggregator/aggregate_share.rs
divergentdave: The settings inside could be relevant, on the other hand, many of our clients are built with just the default settings. I noticed the custom Debug implementation collapses down a number of options if they are not set/enabled/present. A quick spot check shows we do not yet use CollectJobDriver's Debug implementation anywhere. Here's a sample output:

    Client {
        accepts: Accepts,
        proxies: [
            Proxy(
                System(
                    {},
                ),
                None,
            ),
        ],
        referer: true,
        default_headers: {
            "accept": "*/*",
        },
    }

Issue #393, opened 2022-08-12
Collector library


divergentdave:
Currently, writing a collector with Janus requires lots of plumbing code, between an HTTP client, Janus's message serialization and deserialization, HPKE decryption, VDAF aggregate share parsing, and unsharding. We should write a library that encapsulates all this, and encodes the DAP semantics regarding the HTTP responses.

tgeoghegan:
We should make sure this collector library uses `janus_core::retries` to retry requests appropriately.

branlwyd:
SGTM; part of this work should update our test container implementation to use this collector.

(There are a number of tests that effectively implement a collector, too -- we should consider migrating those tests to use the "official" collector library as well.)

PR #414, opened 2022-08-19
Retries in HTTP clients

### LINKED ISSUE #196: Retries in HTTP clients

tgeoghegan:
Adds a new module `janus_core::retries` which uses crate `backoff` to
allow retrying HTTP requests based on whether an error is transient. We
retry if we encounter:

  - failures to connect (i.e., connection refused or reset);
  - timeouts;
  - HTTP errors in the 500 range.

The error classification depends on methods on `reqwest::Error` and
walking the chain of `std::error::Error::source` errors to see if there
is an `std::io::Error` under the covers.

We also use `ClientBuilder::timeout` to set a 10 second timeout on
individual HTTP requests (note this is distinct from the overall 600s
timeout applied to a sequence of retries of HTTP requests).

Resolves #196

### DIFF
--- a/janus_core/src/retries.rs
+++ b/janus_core/src/retries.rs
@@ -1,0 +1,311 @@
+//! Provides a simple interface for retrying fallible HTTP requests.
+
+use backoff::{future::retry, ExponentialBackoff};
+use futures::Future;
+use reqwest::StatusCode;
+use std::{error::Error as StdError, time::Duration};
+use tracing::{debug, warn};
+
+/// Traverse chain of source errors looking for an `std::io::Error`.
+fn find_io_error(original_error: &reqwest::Error) -> Option<&std::io::Error> {
+    let mut cause = original_error.source();
+    while let Some(err) = cause {
+        if let Some(typed) = err.downcast_ref() {
+            return Some(typed);
+        }
+        cause = err.source();
+    }
+
+    None
+}
+
+/// An [`ExponentialBackoff`] with parameters suitable for most HTTP requests. The parameters are
+/// copied from the parameters used in the GCP Go SDK[1].
+///
+/// AWS doesn't give us specific guidance on what intervals to use, but the GCP implementation cites
+/// AWS blog posts so the same parameters are probably fine for both, and most HTTP APIs for that
+/// matter.
+///
+/// [1]: https://github.com/googleapis/gax-go/blob/fbaf9882acf3297573f3a7cb832e54c7d8f40635/v2/call_option.go#L120
+pub fn http_request_exponential_backoff() -> ExponentialBackoff {
+    ExponentialBackoff {
+        initial_interval: Duration::from_secs(1),
+        max_interval: Duration::from_secs(30),
+        multiplier: 2.0,
+        max_elapsed_time: Some(Duration::from_secs(600)),
+        ..Default::default()
+    }
+}
+/// An [`ExponentialBackoff`] with parameters tuned for tests where we don't want to be retrying
+/// for 10 minutes.
+#[cfg(feature = "test-util")]
+pub fn test_http_request_exponential_backoff() -> ExponentialBackoff {
+    ExponentialBackoff {
+        initial_interval: Duration::from_nanos(1),
+        max_interval: Duration::from_nanos(30),
+        multiplier: 2.0,
+        max_elapsed_time: Some(Duration::from_millis(10)),
+        ..Default::default()
+    }
+}
+
+/// Executes the provided request function and awaits the returned future, retrying using the
+/// parameters in the provided `ExponentialBackoff` if the [`reqwest::Error`] returned by
+/// `request_fn` is:
+///
+///   - a timeout
+///   - a problem establishing a connection
+///   - an HTTP status code indicating a server error
+///
+/// If the request eventually succeeds, the value returned by `request_fn` is returned. If an
+/// unretryable failure occurs or enough transient failures occur, then `Err(ret)` is returned,
+/// where `ret` is the `Result<reqwest::Response, reqwest::Error>` returned by the last call to
+/// `request_fn`. Retryable failures are logged.
+///
+/// # TODOs:
+///
+/// This function could take a list of HTTP status codes that should be considered retryable, so
+/// that a caller could opt to retry when it sees 408 Request Timeout or 429 Too Many Requests, but
+/// since none of the servers this is currently used to communicate with ever return those statuses,
+/// we don't yet need that feature.
+pub async fn retry_http_request<RequestFn, ResultFuture>(
+    backoff: ExponentialBackoff,
+    request_fn: RequestFn,
+) -> Result<reqwest::Response, Result<reqwest::Response, reqwest::Error>>
+where
+    RequestFn: Fn() -> ResultFuture,
+    ResultFuture: Future<Output = Result<reqwest::Response, reqwest::Error>>,
+{
+    retry(backoff, || async {
+        // In all branches in this match, we wrap the reqwest::Response or reqwest::Error up in a
+        // Result<reqwest::Response, backoff::Error<Result<reqwest::Response, reqwest::Error>>>>,
+        // which allows us to retry on certain HTTP status codes without discarding the
+        // reqwest::Response, which the caller may need in order to examine its body or headers.
+        match request_fn().await {
+            Ok(response) => {
+                if response.status().is_server_error()
+                    && response.status() != StatusCode::NOT_IMPLEMENTED
+                {
+                    warn!(?response, "encountered retryable server error");
+                    return Err(backoff::Error::transient(Ok(response)));
+                }
+
+                Ok(response)
+            }
+            Err(error) => {
+                if error.is_timeout() || error.is_connect() {
+                    warn!(?error, "encountered retryable error");
+                    return Err(backoff::Error::transient(Err(error)));
+                }
+
+                if let Some(io_error) = find_io_error(&error) {
+                    if let std::io::ErrorKind::ConnectionRefused
+                    | std::io::ErrorKind::ConnectionReset
+                    | std::io::ErrorKind::ConnectionAborted = io_error.kind()
+                    {
+                        warn!(?error, "encountered retryable error");
+                        return Err(backoff::Error::transient(Err(error)));
+                    }
+                }
+
+                debug!("encountered non-retryable error");
+                Err(backoff::Error::permanent(Err(error)))
+            }
+        }
+    })
+    .await
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::test_util::install_test_trace_subscriber;
+    use mockito::mock;
+    use tokio::net::TcpListener;
+    use url::Url;
+
+    #[tokio::test]
+    async fn http_retry_client_error() {
+        install_test_trace_subscriber();
+
+        let mock_404 = mock("GET", "/")
+            .with_status(StatusCode::NOT_FOUND.as_u16().into())
+            .with_header("some-header", "some-value")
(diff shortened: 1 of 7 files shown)


tgeoghegan:
Using `reqwest::Error::error_for_status` has mangled some existing test cases in ways that need some more thought. Downgrading to draft while I iterate on this.

branlwyd:
BTW, I think we should think about our retry strategy:
* Retries in the collector & client libraries probably make sense.
* Retries in things backed by job drivers, i.e. everything in `janus_server`, should IMO lean on the existing job-based retry mechanism rather than doubling up on retries. (of course, the existing retry mechanism might need tuning!)
* IMO, we should _not_ add a retry to the integration tests. The network communication there is really IPC, e.g. it's reliable, so "transient" failures are likely to indicate real problems in our code worthy of investigation.

tgeoghegan:
> BTW, I think we should think about our retry strategy:
> 
>     * Retries in the collector & client libraries probably make sense.

👍🏻 

>     * IMO, we should _not_ add a retry to the integration tests. The network communication there is really IPC, e.g. it's reliable, so "transient" failures are likely to indicate real problems in our code worthy of investigation.

👍🏻 

>     * Retries in things backed by job drivers, i.e. everything in `janus_server`, should IMO lean on the existing job-based retry mechanism rather than doubling up on retries. (of course, the existing retry mechanism might need tuning!)

In `prio-server`, we do retries both at the per-request level and via SQS/PubSub message redelivery. The reason is that if we're doing some work that involves making 1,000 API requests, and 999 of them succeed, but the 1,000th fails, then we'd like to immediately retry that last API request. If we wait until the next time the task is dequeued, then the 999 previous things have to be re-done.

Having said that, this should be less of a problem in the Janus components, because we do less work per task, so less ork has to be re-done when a task is redelivered. Plus I agree with your intuition that we don't want two mechanisms for doing retries, so I'll do as you suggest, and then we can see in the future whether we end up with a lot of wasted work due to task retries.

tgeoghegan:
In this PR, we only add retries to the DAP client implementation. We should, as Brandon argued, do so for requests made by the collector, but since we don't yet have a create that implements that role, the only place we could add retries would be integration tests where we don't want retries. I made a note in #393 that the eventual collector library should make retries.

Also while working on this, I noticed that we don't ever set client timeouts at the `reqwest` layer, which can cause us to hang forever in some circumstances. I didn't fix that here to keep the PR focused, but #426 tracks fixing that.

janus_core/src/retries.rs
```diff
@@ -0,0 +1,315 @@
...
+                debug!("encountered non-retryable error");
+                Err(backoff::Error::permanent(Err(error)))
+            }
+        }
+    })
+    .await;
+
+    match result {
+        Ok(response) => Ok(response),
+        Err(result) => result,
```
divergentdave, line 122: Coalescing the two different `Ok` variants here makes me worry about the retry function and its caller agreeing on the semantics of success/error. This concern is more applicable to future support for configuring which status codes result in retries.

One idea, for argument's sake: If the retry function returned a two-variant error enum inside `Err()`, with one variant for `reqwest::Error`, and the other for `reqwest::Response` with a failing status, then we could do all our error categorization "inside" the retry handler, and avoid redundant checks on the result of the retry method. (This would also bring a fringe benefit of making it impossible to forget to check the status.)

janus_core/src/retries.rs
```diff
@@ -0,0 +1,315 @@
...
+        assert_eq!(
+            response.headers().get("some-header").unwrap(),
+            &"some-value"
+        );
+        assert_eq!(response.text().await.unwrap(), "some-body".to_string());
+        mock_500.assert();
+    }
+
+    #[tokio::test]
+    async fn http_retry_server_err1or_unimplemented() {
```
divergentdave, line 197: nit: typo
```suggestion
    async fn http_retry_server_error_unimplemented() {
```

janus_core/src/retries.rs
```diff
@@ -0,0 +1,315 @@
...
+        install_test_trace_subscriber();
+
+        let mock_500 = mock("GET", "/")
```
divergentdave, line 200: This is probably where that 1 was headed :)

```suggestion
        let mock_501 = mock("GET", "/")
```

janus_core/src/retries.rs
```diff
@@ -0,0 +1,315 @@
...
+        })
+        .await
+        .unwrap();
+
+        assert_eq!(response.status(), StatusCode::NOT_IMPLEMENTED);
+        mock_500.assert();
+    }
+
+    #[tokio::test]
+    async fn http_retry_server_err1or_eventuall1y_succeeds() {
```
divergentdave, line 218:
```suggestion
    async fn http_retry_server_error_eventually_succeeds() {
```

divergentdave (approved):

janus_core/src/retries.rs
```diff
@@ -0,0 +1,315 @@
...
+        max_interval: Duration::from_secs(30),
+        multiplier: 2.0,
+        max_elapsed_time: Some(Duration::from_secs(600)),
+        ..Default::default()
+    }
+}
+/// An [`ExponentialBackoff`] with parameters tuned for tests where we don't want to be retrying
+/// for 10 minutes.
+#[cfg(feature = "test-util")]
+pub fn test_http_request_exponential_backoff() -> ExponentialBackoff {
```
branlwyd, line 42: This function is only used in this package's tests; move into the test module as a private function?

janus_core/src/retries.rs
```diff
@@ -0,0 +1,315 @@
...
+
+use backoff::{future::retry, ExponentialBackoff};
+use futures::Future;
+use reqwest::StatusCode;
+use std::time::Duration;
+use tracing::{debug, warn};
+
+/// Traverse chain of source errors looking for an `std::io::Error`.
+fn find_io_error(original_error: &dyn std::error::Error) -> Option<&std::io::Error> {
+    let mut cause = original_error.source();
```
branlwyd, line 11: nit: if `original_error` is already a `std::io::Error`, this function won't return it. This is only a nit because I think that doesn't matter for how the function is currently used. The easiest fix is probably to change this line to something like `let mut cause = Some(original_error)`, i.e. start walking the chain from `original_error` rather than `original_error`'s source.

janus_core/src/retries.rs
```diff
@@ -0,0 +1,315 @@
...
+/// This function could take a list of HTTP status codes that should be considered retryable, so
+/// that a caller could opt to retry when it sees 408 Request Timeout or 429 Too Many Requests, but
+/// since none of the servers this is currently used to communicate with ever return those statuses,
+/// we don't yet need that feature.
+pub async fn retry_http_request<Request, ResultFuture>(
+    backoff: ExponentialBackoff,
+    request_fn: Request,
+) -> Result<reqwest::Response, reqwest::Error>
+where
+    Request: Clone + FnMut() -> ResultFuture,
```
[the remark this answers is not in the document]
branlwyd: Also, consider naming this parameter something like `RequestFn` to indicate it's a function.

janus_core/src/retries.rs
tgeoghegan: It's worse than that: I was just trying to force `cargo test`'s test name matcher to ignore those tests.

janus_core/src/retries.rs
tgeoghegan: I had a similar thought, but a two-variant error enum where one variant is `reqwest::Error` and the other  is `reqwest::Response` is exactly `Result<reqwest::Response, reqwest::Error>`, which is what I'm sticking inside the error, so I don't think this would make a difference.

I don't see what error categorization is happening outside of the call to `backoff::retry`. All we do with the result from `retry` is flatten the nested `Err()` into the type expected by the caller of `retry_http_request`.

This can be simplified (arguably) by calling `Result::or_else` instead of using a match to flatten the error case.

branlwyd: +1, IMO we should allow the caller to determine whether the retry mechanism considered the request a failure or not.

The suggestion to introduce an error enum would work. We could also use `Result<reqwest::Response, reqwest::Error>` as this enum; this is very close to what is returned by the `retry` call & would make the final return value of this function `Result<reqwest::Response, Result<reqwest::Response, reqwest::Error>>`.

divergentdave: Sorry for the ambiguity -- yeah, the caller of `retry_http_request()` currently has to determine whether the `Response` represents a success case or failure case.

branlwyd: > other is reqwest::Response is exactly Result<reqwest::Response, reqwest::Error>, which is what I'm sticking inside the error, so I don't think this would make a difference.

The return type is currently `Result<reqwest::Response, reqwest::Error>`. A successful-from-reqwest's-perspective request that leads to too many transient retries will be returned as `Ok(response)`, giving the caller no indication that the success response they're getting is actually a failure as far as the retry machinery is concerned.

I think we should modify this function's return type to `Result<reqwest::Response, Result<reqwest::Response, reqwest::Error>>`. `Ok(response)` would be returned only if the response was not a retry; `Err(Ok(response))` would indicate that this response was considered a failure due to retry issues even though reqwest viewed it as a successful HTTP transaction; and `Err(Err(err))` would indicate some HTTP-or-lower-level error reported by reqwest.

janus_core/src/retries.rs
branlwyd: (I'd also be OK with inlining this at points of call, or defining it in a const/static way somewhere -- this function returns a contant value so it may be better expressed as something other than a function)

tgeoghegan: Good catch: the intention was to use this in `janus_client`'s tests so that they wouldn't possibly block for ten minutes.

janus_core/src/retries.rs
tgeoghegan: Generalizing as you suggest is surprisingly tricky because of the `'static` lifetime in the signature of `std::error::Error::source`. In this module, we know that `original_error` will only ever be `reqwest:Error`, so I think the easiest thing is to change the signature of `find_io_error` to reflect that.

janus_core/src/retries.rs
tgeoghegan: I gave this a try in 788e7ae957dc5fbdb9614cb24de8184d2d93fbde and I'm not convinced it's better. Nothing that currently uses `retry_http_request` cares about the distinction between `Ok(response)` and `Err(Ok(response))`, so the upshot is that we now force every caller of `retry_http_request` to deal with that ambiguity, whereas otherwise they can just use `?` and then examine the HTTP status code in `Ok(response)`.

If you both think this approach is better, I'm happy to go with the majority decision, though.

branlwyd: IMO, the new approach isn't ambiguous, the original approach is: it makes it ambiguous whether an `Ok(response)` return value indicates (a) success or (b) a transient failure that ran out of time to retry. As David pointed out, this forces every caller to implement (matching!) error-detection logic.

If nested `Result`s are too ugly / too hard to deal with, we could map transient failures to an arbitrary error enum indicating the reason for failure. This is easier to consume, but will lose some fidelity if there's ever a caller that wants to inspect e.g. the response even if it is a failure.

tgeoghegan: Well, the good news is, we can change `retry_http_request` in lockstep with all its users in the future, so we don't have to get it perfectly right today. I'll merge this with your preferred approach.

PR #415, opened 2022-08-19
CI: Mitigation for disk exhaustion

divergentdave:


divergentdave:
Freeing up 2G from Docker was not enough. Closing in favor of #416.

PR #416, opened 2022-08-19
CI buildtime improvements.

branlwyd:
* Split the main build steps from the lint-like steps to ensure the cache does not include the lint-like steps' outputs.
* Use a more precise caching strategy.
* Disable incremental compilation.
* Disable debuginfo.
* Allow `cargo clippy` to skip building the interop test container.

Before these changes, a CI run I kicked off this morning took 45 minutes: https://github.com/divviup/janus/actions/runs/2891132620
With these changes, CI runs are more like 21 minutes: https://github.com/divviup/janus/actions/runs/2892438113
... or 30 minutes on a cache miss: https://github.com/divviup/janus/actions/runs/2892294130

Most of the strategies in this PR were taken from [Fast Rust Builds](https://matklad.github.io/2021/09/04/fast-rust-builds.html).

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -15,6 +15,8 @@ jobs:
       RUSTFLAGS: "-D warnings"
     steps:
     - uses: actions/checkout@v3
+    - name: Setup Go toolchain
+      uses: actions/setup-go@v3
     - name: Install Kind
       # We should always make sure that the `kind` CLI we install is from the
       # same release as the node image version used by
@@ -31,8 +33,7 @@ jobs:
     - name: Build
       run: cargo test --no-run --locked
     - name: Test
-      # set PATH to ensure we use go binaries installed above
-      run: PATH=$(go env GOPATH)/bin:$PATH cargo test
+      run: cargo test
   
   janus_lints:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -29,13 +29,15 @@ jobs:
         components: clippy, rustfmt
     - name: Cache dependencies
       uses: Swatinem/rust-cache@v2
-    - name: lint
+    - name: Lint
       run: cargo fmt --message-format human -- --check
-    - name: clippy
+    - name: Clippy
+      env:
+        CARGO_TARGET_DIR: "/tmp/target"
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-build.yml
```diff
@@ -6,49 +6,59 @@ on:
...
-      run: PATH=$(go env GOPATH)/bin:$PATH cargo test --verbose
-    - name: document
-      env:
-        CARGO_TARGET_DIR: "target/doc-deny"
-        RUSTDOCFLAGS: "-Dwarnings"
-      run: cargo doc
+    - name: Clippy
+      run: cargo clippy --workspace --all-targets
+    - name: Document
+      run: cargo doc -- -D warnings
```
divergentdave, line 59: nit: I think the `-D warnings` here is redundant with the environment variable set above.

.github/workflows/ci-build.yml
```diff
@@ -6,49 +6,59 @@ on:
...
-      env:
-        RUST_LOG: info
-      # set PATH to ensure we use go binaries installed above
-      run: PATH=$(go env GOPATH)/bin:$PATH cargo test --verbose
-    - name: document
-      env:
-        CARGO_TARGET_DIR: "target/doc-deny"
-        RUSTDOCFLAGS: "-Dwarnings"
-      run: cargo doc
+    - name: Clippy
```
divergentdave, line 56: We may want to use the rust-cache action on the lint job as well (perhaps with a separate cache key) to keep Clippy fast.

branlwyd:
(N.B. this is a draft currently, just to allow me to run CI; review away if you wish but I'm still planning on a few changes)

.github/workflows/ci-build.yml
branlwyd: Yup, it was -- in fact, specifying it caused the action to fail entirely!

.github/workflows/ci-build.yml
branlwyd: Instead, I moved the `clippy` & other lint-like checks to a different build step to ensure they aren't poisoning the build cache (and maybe get a little perf boost from the improved parallelism). I also implemented a small change so that `clippy` can skip building the container -- it runs the build script (and I can't figure out a way to make it not do so), but it definitely doesn't need a correctly-built container image to do its job.

divergentdave (approved):

PR #417, opened 2022-08-19
Remove aggregation parameter from logs and Debug implementations

divergentdave:
This removes aggregation parameters from logs, as they may be sensitive with Poplar. Part of #392.


branlwyd (approved):
This is off-topic from this PR, but it's somehwat unfortunate that Poplar aggregation params are considered sensitive (IIUC, they imply the string being searched for?): in terms of VDAF evaluation, this is the only sensitive value that an aggregator is directly exposed to.

I wonder if some work on Poplar would allow implementations to have a non-sensitive aggregation parameter. I'm not sure how feasible an ask this would be -- I'm going to file an issue in the VDAF repo to ask.

PR #432, opened 2022-08-23
Build Daphne in use for tests.

branlwyd:
Also, update Daphne to the latest version.


branlwyd:
OK, this is ready for review.

branlwyd:
A few notes:
* My next step in this line of work is going to be doing the work to allow a prebuilt Daphne container to be used, and to use one in practice for CI & make it easy for us to use prebuilt Daphne on our local machines.
* Sadly, the aggregator-endpoint path issue we ran into in our deployment would have been caught by the integration test -- but no one ran them on a version that includes the issue. (This doesn't matter much since our current deployments are test-only, but IMO this highlights the importance of maintaining strong tests.)

divergentdave (approved):

PR #434, opened 2022-08-25
Use containers for interop API tests.

branlwyd:
With this commit, the `interop_binaries` E2E tests now test the interop
test containers, rather than the binaries that are added to the
containers. This is important since the container build definitions
include important wiring -- before this change, the container builds
could be broken (eg Janus can't talk to datastore), but the E2E tests
would still pass.


branlwyd:
Notes:
* This temporarily removes the can't-handle-non-`/`-endpoints issue regression test. This is reintroduced in https://github.com/divviup/janus/pull/432 (by the Daphne tests, since Daphne will always serve from `/v01/`). I tried wiring the interop aggregator to always serve from `/dap/`, but humorously, this broke the Janus-Daphne integration tests: it appears that the version of Daphne under test has the same URL-path handling bug that led to the issue in Janus. (I believe this is fixed in the latest versions of Daphne, but https://github.com/divviup/janus/pull/432 has to land before we're testing against those, and I decided not to introduce a dependency between these two PRs.)
* The new images build quickly since they all use the same cache of Rust build artifacts, and since they're basically an alpine baseimage plus a single binary, they're fairly small: just over 5.5 MiB each, after zstd compression.

interop_binaries/src/testcontainer.rs
```diff
@@ -1,24 +1,90 @@
...
+const INTEROP_COLLECTOR_IMAGE_BYTES: &[u8] =
+    include_bytes!(concat!(env!("OUT_DIR"), "/interop_collector.tar.zst"));
+static INTEROP_COLLECTOR_IMAGE_HASH: Mutex<Option<String>> = Mutex::new(None);
+
+/// Represents a Janus Client as a testcontainer image.
+#[non_exhaustive]
+pub struct Client {}
+
+impl Client {
+    // The internal port that the Client serves on.
```
divergentdave, line 25: This may as well be a doc comment.
```suggestion
    /// The internal port that the Client serves on.
```

interop_binaries/src/testcontainer.rs
```diff
@@ -54,3 +120,56 @@ impl Image for Aggregator {
         Vec::new()
     }
 }
+
+/// Represents a Janus Collector as a testcontainer image.
+#[non_exhaustive]
+pub struct Collector {}
+
+impl Collector {
+    // The internal port that the Collector serves on.
```
divergentdave, line 129:
```suggestion
    /// The internal port that the Collector serves on.
```

interop_binaries/tests/end_to_end.rs
```diff
@@ -8,173 +12,87 @@ use janus_server::task::PRIO3_AES128_VERIFY_KEY_LENGTH;
...
-        4,
-        "Ports selected for HTTP servers were not unique",
-    );
-
-    // Create and start containers. (here, we just run the binaries instead)
+) -> serde_json::Value {
+    // Create and start containers.
     // We use std::process instead of tokio::process so that we can kill the child processes from
     // a Drop implementation. tokio::process::Child::kill() is async, and could not be called from
     // there.
```
divergentdave, line 39: This part of the comment is obsolete now.
```suggestion
```

interop_binaries/tests/end_to_end.rs
```diff
@@ -8,173 +12,87 @@ use janus_server::task::PRIO3_AES128_VERIFY_KEY_LENGTH;
...
-    //
-    // We set up the helper task to serve out of the `/dap/` prefix. This tests for regression
-    // against a bug where Janus in the leader position would not work with a helper whose endpoint
-    // had a non-`/` path.
+    let network = generate_network_name();
+
+    let client_port = pick_unused_port().expect("couldn't pick a port for the client");
```
divergentdave, line 42: The containers each EXPOSE port 8080, so instead of picking ports here, we could let Docker handle port selection, and then look it up the usual way via `Container::get_host_port_ipv4()`.

divergentdave (approved):

interop_binaries/tests/end_to_end.rs
branlwyd: 👍🏻 (btw, this is entirely moot with this change & even before this change, using a stdlib non-async process was fine, but you can work around the technical issue of not being able to run async code in a Drop implementation by using a "shutdown" signal that the Drop implementation signals, along with a "shutdown complete" signal that the Drop implementation then waits on; see [Daphne's drop implementation](https://github.com/divviup/janus/blob/e842b61f0527843c4edba44881ade8196cd7594c/monolithic_integration_test/src/daphne.rs#L226-L233) for an example)

interop_binaries/tests/end_to_end.rs
branlwyd: 👍🏻 I also played a little code golf with the image/container instantiations while I was at it, and cleaned up a `Command` instantiation that I had accidentally left behind (!).

tgeoghegan (approved):

PR #443, opened 2022-08-29
Fix build with "otlp" feature

divergentdave:
A version bump last week broke builds with the "otlp" feature enabled. We construct some objects using the `tonic` crate, and pass them to `opentelemery-otlp`, which does not re-export any of the `tonic` types it expects. The difference in crate versions was causing errors due to mismatching types. This PR reverts that version bump, makes note of the constraint between the two packages, and adds an `--all-features` step to the CI to catch such issues in the future.

### DIFF
--- a/janus_server/Cargo.toml
+++ b/janus_server/Cargo.toml
@@ -59,7 +59,7 @@ testcontainers = { version = "0.14.0", optional = true }
 thiserror = "1.0"
 tokio = { version = "^1.20", features = ["full", "tracing"] }
 tokio-postgres = { version = "0.7.7", features = ["with-chrono-0_4", "with-serde_json-1", "with-uuid-1", "array-impls"] }
-tonic = { version = "0.8.0", optional = true, features = ["tls", "tls-webpki-roots"] }
+tonic = { version = "0.6.2", optional = true, features = ["tls", "tls-webpki-roots"] }  # keep this version in sync with what opentelemetry-otlp uses
 tracing = "0.1.36"
 tracing-log = "0.1.3"
 tracing-opentelemetry = { version = "0.17.4", optional = true }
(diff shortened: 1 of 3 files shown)


janus_server/Cargo.toml
```diff
@@ -59,7 +59,7 @@ testcontainers = { version = "0.14.0", optional = true }
 thiserror = "1.0"
 tokio = { version = "^1.20", features = ["full", "tracing"] }
 tokio-postgres = { version = "0.7.7", features = ["with-chrono-0_4", "with-serde_json-1", "with-uuid-1", "array-impls"] }
-tonic = { version = "0.8.0", optional = true, features = ["tls", "tls-webpki-roots"] }
+tonic = { version = "0.6.2", optional = true, features = ["tls", "tls-webpki-roots"] }  # keep this version in sync with what opentelemetry-otlp uses
```
branlwyd, line 62: Consider also adding a comment to the line for `opentelemetry-otlp`, as someone clicking through a dependabot PR to update that package may not notice this comment.

branlwyd (approved):

janus_server/Cargo.toml
tgeoghegan: I wonder if dependabot will preserve the trailing comment if/when it bumps either version. Maybe it's best to put the comment above the dependencies, just in case?

tgeoghegan (approved):

janus_server/Cargo.toml
divergentdave: I checked [the source code](https://github.com/dependabot/dependabot-core/blob/main/cargo/lib/dependabot/cargo/file_updater/manifest_updater.rb), and it does a series of regex replacements, so our comments should be fine.

PR #448, opened 2022-08-30
`janus_cli`: automatically discover datastore keys

tgeoghegan:
In order to provision tasks, `janus_cli` needs the datastore key, used
encrypt some database rows. This key can be provided with
`--datastore-keys`, but since `janus_cli` has to talk to a Kubernetes
cluster anyway, it can also fetch the secret value from Kubernetes
secrets, which is fewer steps for an operator (no need to fetch the
secret before running `janus_cli`) and less risk of exposing the
datastore key to an operator workstation (or worse, an operator's
clipboard).

### DIFF
--- a/janus_server/src/bin/janus_cli.rs
+++ b/janus_server/src/bin/janus_cli.rs
@@ -60,6 +60,9 @@ enum Command {
         #[structopt(flatten)]
         common_options: CommonBinaryOptions,
 
+        #[structopt(flatten)]
+        kubernetes_secret_options: KubernetesSecretOptions,
+
         /// A YAML file containing a list of tasks to be written. Existing tasks (matching by task
         /// ID) will be overwritten.
         tasks_file: PathBuf,
@@ -70,11 +73,8 @@ enum Command {
         #[structopt(flatten)]
         common_options: CommonBinaryOptions,
 
-        /// The Kubernetes namespace to create the datastore key secret in.
-        k8s_namespace: String,
-
-        /// The name of the Kubernetes secret to place the datastore key in.
-        k8s_secret_name: String,
+        #[structopt(flatten)]
+        kubernetes_secret_options: KubernetesSecretOptions,
     },
 
     /// Decode a single Distributed Aggregation Protocol message.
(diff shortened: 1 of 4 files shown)


divergentdave (approved):

janus_server/src/bin/janus_cli.rs
```diff
@@ -56,17 +56,15 @@ enum Command {
...
-        k8s_secret_name: String,
-    },
+    CreateDatastoreKey,
 }
 
 impl Command {
     async fn execute(&self, options: &Options, config: &Config) -> Result<()> {
+        let k8s_client = kube::Client::try_default()
+            .await
+            .context("couldn't connect to Kubernetes environment")?;
```
divergentdave, line 66: With `janus_cli decode-dap-message` on the way, we should avoid constructing this client unless it is required by the subcommand, so that `write-schema` and `decode-dap-message` can be used on systems without kubeconfig files. `once_cell::unsync::Lazy` would be a convenient way to do so without duplication.

branlwyd: +1, we should maintain that we don't access/connect to dependencies that are not used for the particular subcommand.

branlwyd (approved):

tgeoghegan:
Downgrading to draft while I re-work the argument handling here and in #450

tgeoghegan:
- [ ] This is stacked on #450, merge that and retarget this as needed!

janus_server/src/bin/janus_cli.rs
```diff
@@ -101,6 +101,10 @@ enum Command {
 
 impl Command {
     async fn execute(&self) -> Result<()> {
+        let kube_client = kube::Client::try_default()
+            .await
+            .context("couldn't connect to Kubernetes environment")?;
+
```
divergentdave, line 107: We need to defer this, and only run it inside the commands that require it. Otherwise, we may get the following error when it's not necessary.

```
Error: couldn't connect to Kubernetes environment

Caused by:
    0: Failed to infer configuration: failed to infer config: in-cluster: (failed to read the default namespace: No such file or directory (os error 2)), kubeconfig: (failed to read kubeconfig from '"/root/.kube/config"': No such file or directory (os error 2))
    1: failed to infer config: in-cluster: (failed to read the default namespace: No such file or directory (os error 2)), kubeconfig: (failed to read kubeconfig from '"/root/.kube/config"': No such file or directory (os error 2))
    2: failed to read kubeconfig from '"/root/.kube/config"': No such file or directory (os error 2)
    3: No such file or directory (os error 2)
```

divergentdave (changes requested):

janus_server/src/bin/janus_cli.rs
tgeoghegan: Whoops, I was sure I had caught this, but I think I got mixed up while rebasing.

divergentdave (approved):

tgeoghegan:
Another day, another PR merged to the wrong target.

PR #453, opened 2022-08-31
Rename monolithic_integration_test -> integration_tests.

branlwyd:
Trivial/pedantic cleanup; arguably, we already have more than one
integration test (in the sense that we test against ourselves & Daphne),
and we'll surely be adding more in the future.


divergentdave (approved):

branlwyd:
(btw, if merging this would cause another PR to have merge conflicts, I don't mind waiting to merge -- this is both low-priority & straightforward, so it's likely that resolving merge conflicts in this PR will be easier than in a more substantial PR. let me know if you'd like me to wait.)

Cargo.toml
```diff
@@ -1,5 +1,12 @@
 [workspace]
-members = ["build_script_utils", "interop_binaries", "janus_core", "janus_client", "janus_server", "monolithic_integration_test"]
+members = [
```
branlwyd, line 2: Off-topic from this PR: Tim, you've expressed concern about how the recent test changes will affect build times.

If the build times do turn out to be overly painful, we can specify [`default-members`](https://doc.rust-lang.org/cargo/reference/workspaces.html#package-selection) so that a `cargo test` invocation will e.g. run the `janus_*` tests but not `integration_tests` by default. (We'd just need to be careful that our CI specify `--workspace` in all relevant `cargo` invocations, so that the integration tests run in CI.)

tgeoghegan: I'll be working exclusively from my less powerful laptop over the next few months so I will get to find out first hand whether build and test times are especially long.

tgeoghegan (approved):
No worries about merge conflicts. I have more work to do on everything I have in-flight at the moment anyway.

Cargo.toml
branlwyd: SGTM. I'd also note that you can do e.g. `cargo test -p janus_server` to run only the `janus_server` tests; the build/tests are set up such that the containers will only be built if there is a test being built that actually needs to use them.

PR #454, opened 2022-08-31
Ensure we check container readiness on IPv4

tgeoghegan:
`interop_binaries::test_util::await_http_server` blocks until HTTP
requests are responded to on a specified port. The assumption is that
this port is bound on localhost with IPv4, but because we were
connecting to `localhost:{port}`, it could sometimes connect to the port
bound on IPv6, which might be bound by some totally different container,
meaning that tests prematurely will consider a container's port to be
available.

We now connect to `http://127.0.0.1:{port}`, which ought to ensure we
connect to IPv4.


divergentdave (commented):
LGTM. There are also some other combinations of `localhost` and `get_host_port_ipv4()` in `integration_tests/src/janus.rs`, `integration_tests/src/daphne.rs`, and `integration_tests/tests/common/mod.rs` we should probably clean up.

branlwyd (approved):
Nice catch!

divergentdave (approved):

PR #456, opened 2022-09-01
`janus_cli`: discover datastore keys in k8s secrets

### LINKED ISSUE #409: `janus_cli`: automatically discover database secrets in Kubernetes secrets
Currently, `janus_cli` needs to be provided the datastore encryption key with the `--datastore-keys` option, and also sometimes the Postgres password in `--database-password` or the `PGPASSWORD` environment variable. It inherits these flags from the other Janus binaries that run in-cluster and receive these values by having Kubernetes secrets plugged into their environment variables. However when running `janus_cli`, it's a hassle for operators to have to first fish secrets out of the Kubernetes cluster. `janus_cli` uses the Kubernetes API to work with secrets anyway (since we use it to create datastore keys), so when it is provisioning tasks, we should teach it to read the correct datastore key (and any other secrets it may need) itself.

tgeoghegan:
In order to provision tasks, `janus_cli` needs the datastore key, used
encrypt some database rows. This key can be provided with
`--datastore-keys`, but since `janus_cli` has to talk to a Kubernetes
cluster anyway, it can also fetch the secret value from Kubernetes
secrets, which is fewer steps for an operator (no need to fetch the
secret before running `janus_cli`) and less risk of exposing the
datastore key to an operator workstation (or worse, an operator's
clipboard).

Resolves #409

### DIFF
--- a/janus_server/src/bin/janus_cli.rs
+++ b/janus_server/src/bin/janus_cli.rs
@@ -60,6 +60,9 @@ enum Command {
         #[structopt(flatten)]
         common_options: CommonBinaryOptions,
 
+        #[structopt(flatten)]
+        kubernetes_secret_options: KubernetesSecretOptions,
+
         /// A YAML file containing a list of tasks to be written. Existing tasks (matching by task
         /// ID) will be overwritten.
         tasks_file: PathBuf,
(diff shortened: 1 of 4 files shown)


tgeoghegan:
Re-created because I merged #448 to the wrong place.

divergentdave (approved):

Issue #455, opened 2022-09-01
`integration_tests::daphne` is flaky (maybe)


tgeoghegan:
I think there is still some flakiness in integration tests. Here's two failures I saw in recent ci-build runs:

https://github.com/divviup/janus/runs/8122772496?check_suite_focus=true
https://github.com/divviup/janus/runs/8142726381?check_suite_focus=true

Problem is, the failure just looks like:
``` running 2 tests
test daphne_janus ... ok
test janus_daphne has been running for over 60 seconds
test janus_daphne ... FAILED

failures:

---- janus_daphne stdout ----
thread 'janus_daphne' panicked at 'assertion failed: Instant::now() < collect_job_poll_timeout', integration_tests/tests/common/mod.rs:209:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    janus_daphne
```

So not a ton of information there. I think we will want to augment the tests so that we scrape logs out of all the interop containers, and then attach them to the test run as an artifact, like we do over in `janus-ops` (https://github.com/divviup/janus-ops/blob/723f8a3be226e3e1bad4f969a1378bb3da9c17ba/.github/workflows/ci-build.yml#L147). Once we learn more about what is going wrong, we can dig in and debug.

Besides grabbing container logs, we should have the test harness log some useful information like the ports it's connecting to, which might give us a leg up detecting wacky things like IPv4/IPv6 port collisions on `localhost`.

branlwyd:
Can this be closed? I think it was fixed with https://github.com/cloudflare/daphne/commit/80b53c4b0f2c93d5f9df66dfce237b20756c9147 (& related work on Janus to use that commit in integration tests).

tgeoghegan:
Yeah, we can write new issues if we continue to see test failures.

PR #477, opened 2022-09-06
Scrape logs out of containers when integration tests fail

tgeoghegan:
When tests run against interop containers fail, we now copy logs out of them so that they can be rolled up by GitHub Actions and attached to failing test runs as an artifact.

Part of #455


tgeoghegan:
This seems to be working! I made a change to force test failures. Here's a failing run: https://github.com/divviup/janus/actions/runs/3004476197

We get a link to download the logs artifact. Its contents:
```
./janus_daphne
./janus_daphne/leader-d2d45a824ddd18673c402f61ed2d71fe44c04eed627d9a084720dfd58374d7bc
./janus_daphne/leader-d2d45a824ddd18673c402f61ed2d71fe44c04eed627d9a084720dfd58374d7bc/postgres_stdout.log
./janus_daphne/leader-d2d45a824ddd18673c402f61ed2d71fe44c04eed627d9a084720dfd58374d7bc/postgres_stderr.log
./janus_daphne/leader-d2d45a824ddd18673c402f61ed2d71fe44c04eed627d9a084720dfd58374d7bc/aggregator_stdout.log
./janus_daphne/leader-d2d45a824ddd18673c402f61ed2d71fe44c04eed627d9a084720dfd58374d7bc/aggregator_stderr.log
./janus_daphne/helper-9104509d90cc3fedea7dcbefb98c2ecbf348073163f15d24458673fe581d25f7
./janus_daphne/helper-9104509d90cc3fedea7dcbefb98c2ecbf348073163f15d24458673fe581d25f7/stdout.log
./janus_daphne/helper-9104509d90cc3fedea7dcbefb98c2ecbf348073163f15d24458673fe581d25f7/stderr.log
./daphne_janus
./daphne_janus/leader-ce359b3f0b57462955a524afe420d06ccaef55fe98f6cbf840b6a42b42dbf285
./daphne_janus/leader-ce359b3f0b57462955a524afe420d06ccaef55fe98f6cbf840b6a42b42dbf285/stdout.log
./daphne_janus/leader-ce359b3f0b57462955a524afe420d06ccaef55fe98f6cbf840b6a42b42dbf285/stderr.log
./daphne_janus/helper-e3dd356a2babf57492b8139670885a4201bca6ce030f2b8625a7dd432d82bef7
./daphne_janus/helper-e3dd356a2babf57492b8139670885a4201bca6ce030f2b8625a7dd432d82bef7/postgres_stdout.log
./daphne_janus/helper-e3dd356a2babf57492b8139670885a4201bca6ce030f2b8625a7dd432d82bef7/postgres_stderr.log
./daphne_janus/helper-e3dd356a2babf57492b8139670885a4201bca6ce030f2b8625a7dd432d82bef7/aggregator_stdout.log
./daphne_janus/helper-e3dd356a2babf57492b8139670885a4201bca6ce030f2b8625a7dd432d82bef7/aggregator_stderr.log
```
The top level path is the name of the failing test (or rather, a label chosen by the test and passed into the log fetcher function). Then we create a directory named after the role being played and the Docker container ID, inside of which are the stdout and stderrs of the container's processes.

integration_tests/tests/common/mod.rs
```diff
@@ -88,11 +105,33 @@ pub fn translate_url_for_external_access(url: &Url, external_port: u16) -> Url {
...
+    Leader: CopyLogs,
+    Fut: Future<Output = Result<()>>,
+    Test: FnMut() -> Fut,
+>(
+    test_name: &str,
+    helper: &Helper,
+    leader: &Leader,
+    mut test: Test,
+) {
+    let result = test().await;
```
divergentdave, line 120: We should probably use `std::panic::catch_unwind()` here, both to ensure we emit logs on any panics we miss, and to make writing tests in this context more ergonomic. (We can use `std::panic::panic_any()` to propagate resulting outer errors)

divergentdave (approved):

integration_tests/tests/common/mod.rs
tgeoghegan: I tried doing that, but `std::panic::catch_unwind` can't be passed an async closure. The async equivalent is [`futures::future::FutureExt::catch_unwind`](https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.catch_unwind), but it requires `UnwindSafe`, which the closure we want to test is decidedly not, and so I backed off from this approach and instead made `submit_measurements_and_verify_aggregate` fallible.

divergentdave: Ah right, of course. There are future wrappers that apply this to poll, FWIW: https://docs.rs/futures/latest/futures/future/struct.CatchUnwind.html

PR #479, opened 2022-09-07
Introduce new collector library

### LINKED ISSUE #393: Collector library

divergentdave:
This introduces a new `janus_collector` crate, and rewrites `janus_interop_collector` and `integration_tests` to use it. The two main methods are `Collector::start_collection()` and `Collector::poll_once()`, and there's additionally a convenience method `Collector::poll_until_complete()` that handles making requests until the collect job has completed. I moved `CollectReq` and `CollectResp` from `janus_server` to `janus_core` to enable this. This closes #393.


janus_collector/src/lib.rs
```diff
@@ -0,0 +1,866 @@
...
+    "collector"
+);
+
+/// Authentication configuration for communication with the leader aggregator.
+#[derive(Derivative)]
+#[derivative(Debug)]
+#[non_exhaustive]
+pub enum Authentication {
+    /// Beearer token authentication, via the `DAP-Auth-Token` header.
+    BearerToken(#[derivative(Debug = "ignore")] String),
```
tgeoghegan, line 66: I'm not sure it's correct to call this a bearer token since we're not doing `Authorization: Bearer <token>` but instead our bespoke `DAP-Auth-Token: <token>`. In any case, it seems like this should use `janus_server::task::AuthenticationToken`, moving that struct into `janus_core` as needed.

janus_collector/src/lib.rs
```diff
@@ -0,0 +1,866 @@
...
+        hpke_private_key: HpkePrivateKey,
+        http_request_retry_parameters: ExponentialBackoff,
+        collect_poll_wait_parameters: ExponentialBackoff,
+    ) -> CollectorParameters {
+        // Ensure the provided leader endpoint ends with a slash, as we will be joining additional
+        // path segments into this endpoint. The Url::join implementation will drop the last
+        // path component of the base URL if it does not end with a slash.
+        if !leader_endpoint.as_str().ends_with('/') {
+            leader_endpoint.set_path(&format!("{}/", leader_endpoint.path()));
+        }
```
tgeoghegan, line 134: This is our third copy of this URL sanitization code, after `janus_client/src/lib.rs` and `janus_server/src/task.rs`. I think it's time to put something like `fn sanitize_url()` in `janus_core`.

janus_collector/src/lib.rs
```diff
@@ -0,0 +1,866 @@
...
+        )
+        .await
+        .map_err(|res| match res {
+            Ok(response) => Error::Http(response.status()),
+            Err(error) => Error::HttpClient(error),
+        })?;
+
+        let status = response.status();
+        match status {
+            StatusCode::ACCEPTED => return Ok(None),
```
tgeoghegan, line 315: IIRC the protocol text urges aggregator implementations to set a [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) header, so it'd be nice to check that here and perhaps provide it to the caller so they can decide when to try again. i.e.,
```
enum PollResult {
    AggregateResult(V::AggregateResult),
    NextAttempt(StdDuration),
}
fn poll_once(...) -> Result<PollResult, Error> { ... }
```
That said, Janus doesn't yet ever set a `Retry-After` header, so I think we could just write up an issue about this and move on. On the other hand, if we wire this up in the collector library now, then we won't have to change the public API later.

tgeoghegan (changes requested):
I believe you'll need some changes to `.github/workflows/publish-crates.yml` to publish the `janus_collector` crate.

janus_collector/src/lib.rs
```diff
@@ -0,0 +1,866 @@
...
+}
+
+/// The PPM collector's view of task parameters.
+#[derive(Derivative)]
+#[derivative(Debug)]
+pub struct CollectorParameters {
+    /// Unique identifier for the task.
+    task_id: TaskId,
+    /// The base URL of the leader's aggregator API endpoints.
+    #[derivative(Debug(format_with = "std::fmt::Display::fmt"))]
```
branlwyd, line 76: neat! this nicely sidesteps the ugly default Debug implementation for `Url`

janus_collector/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+thiserror = "1.0"
+tracing = "0.1.36"
+url = "2.2.2"
+
+[dev-dependencies]
+assert_matches = "1"
+janus_core = { path = "../janus_core", features = ["test-util"]}
+mockito = "0.31.0"
+rand = "0.8"
+tokio = { version = "^1.20", features = ["full"] } # TODO: rebase and bump to 1.21
```
branlwyd, line 27: I think the `TODO` on this line can/should be done before merging this PR.

janus_collector/src/lib.rs
branlwyd: +1 to a different name, `DAP-Auth-Token` is (in the generic sense) a bearer token but the potential for confusion is high.

We could be potentially overly-literal and call this variant literally `DAPAuthToken`; I guess the upside is that it has a high chance of not conflicting with other auth mechanisms introduced later.

janus_collector/src/lib.rs
branlwyd: Alternatively, maybe we should just have our own `actually_join_url()` with correct semantics, and enforce that it's called instead of `Url::join` everywhere (via auditing/code review; static analysis would work too but is probably too much effort).

The upside (IMO) is that it's a lot easier to audit for a particular method (not) being used than that every path to storing a URL happens to call a `sanitize_url` method.

janus_collector/src/lib.rs
```diff
@@ -0,0 +1,866 @@
...
+                multiplier: 1.5,
+                max_elapsed_time: None,
+                ..Default::default()
+            },
+        )
+    }
+
+    /// Creates a new set of collector task parameters with non-default HTTP request retry
+    /// parameters.
+    #[allow(clippy::too_many_arguments)]
```
branlwyd, line 119: optional: instead of having two "constructor" functions, we could use a builder-like pattern. I suppose `new` would stay as it is (since every parameter to `new` is a required parameter w/ no useful default value); but then we could introduce `.with_http_request_backoff(self, ExponentialBackoff)` & `.with_collect_poll_backoff(self, ExponentialBackoff)` which consume an existing `CollectorParameters` and update the given field.

The upside is that we don't have to create `2^n` constructor variants for every set of fields we decide to independently support -- we have a single constructor variant & `O(n)` "builder" methods.

janus_collector/src/lib.rs
```diff
@@ -0,0 +1,866 @@
...
+    pub fn batch_interval(&self) -> Interval {
+        self.batch_interval
+    }
+
+    pub fn aggregation_parameter(&self) -> &V::AggregationParam {
+        &self.aggregation_parameter
+    }
+}
+
+/// A PPM collector.
```
branlwyd, line 206:
```suggestion
/// A DAP collector.
```

janus_collector/src/lib.rs
branlwyd: I suspect a common way to use this type will be to call `poll_until_complete` (which calls this `poll_once` method as a subroutine) -- I think the eventual change to support `Retry-After` headers should also make `poll_until_complete` respect the `Retry-After` header, if specified.

(this would be pretty hard to do while also respecting the backoff set by the user--which is one more good reason to figure out the API sooner rather than later, IMO)

branlwyd (changes requested):

janus_collector/src/lib.rs
```diff
@@ -0,0 +1,1046 @@
...
+use retry_after::{FromHeaderValueError, RetryAfter};
+use std::{
+    convert::TryFrom,
+    time::{Duration as StdDuration, SystemTime},
+};
+use tokio::time::{sleep, Instant};
+use url::Url;
+
+/// HTTP header where auth tokens are provided in messages to the leader.
+pub const DAP_AUTH_HEADER: &str = "DAP-Auth-Token";
```
tgeoghegan, line 28: nit: this constant could be exported from `janus_core` but this business should get replaced with a more robust authentication story before too long anyway, so it's not a big deal.

janus_collector/src/lib.rs
tgeoghegan: +1 to this, especially if we can find some nice macros that generate the builder functions based on a struct declaration.

tgeoghegan (approved):

janus_collector/src/lib.rs
```diff
@@ -0,0 +1,1032 @@
...
+        http_client: &reqwest::Client,
+    ) -> Collector<V> {
+        Collector {
+            parameters,
+            vdaf_collector,
+            http_client: http_client.clone(),
+        }
+    }
+
+    /// Send a collect request to the leader aggregator.
```
branlwyd, line 230: Q. Right now, I think the expected usage of this type is:

1) Call `start_collection` to receive a `CollectJob`.
2) Either:
 a) Repeatedly call `poll_once` until the collect job is completed, or
 b) Call `poll_until_complete` to have the library handle retrying for you.

WDYT about a simpler API for collection: `collect`, which receives the same arguments as `start_collection`, and returns a future that resolves to the collection result (or an error). That is, it would effectively fuse the existing `start_collection` & `poll_until_complete` methods.

The upside is mostly in simplification -- the API would be simpler, and we'd no longer need to expose several of the helper types like `CollectJob` or `PollResult`.

I can think of a couple potential downsides, but IMO they don't break this idea. I wonder if I'm missing something else:
1) Users of this library can no longer pass in an arbitrary `CollectJob`. I'd argue that's OK: the only thing they could add to the collect job over what they're passing to `start_collection` is the collect job URI, and that is an "implementation deatil" of DAP from the perpsective of a collect library user.

2) Users of this library can no longer call `poll_once` to effectively interleave polling of multiple futures. I think that's OK: standard utilities like `try_join!` or `select!` exist to allow multiple futures to be driven concurrently.

(Putting things another way: what is the use case for `poll_once` being public? If there isn't one, I think simplifying things in this way makes sense.)

janus_collector/src/lib.rs
```diff
@@ -0,0 +1,1032 @@
...
+        }
+    }
+
+    /// Replace the exponential backoff settings used for HTTP requests.
+    pub fn with_http_request_backoff(mut self, backoff: ExponentialBackoff) -> CollectorParameters {
+        self.http_request_retry_parameters = backoff;
+        self
+    }
+
+    /// Replace the exponential backoff settings used while polling for aggregate shares.
```
branlwyd, line 131: IIUC, this backoff would effectively be overridden by the `Retry-After` value, correct? If so, we should document this.

However, I wonder if these are the desired semantics: currently, if I specify a backoff strategy that says to retry after (say) 10 seconds, and the `Retry-After` header says to retry after 5 seconds, the library will attempt to retry after 5 seconds. Effectively, my `collect_poll_backoff` settings are not being respected in this scenario.

I think better semantics might be to sleep for an amount of time equal to `max(retry-after header, exponential backoff's next wait time)` -- this would respect both of the retry values. WDYT?

(another solution might be to remove backoff configuration entirely, and always rely on `Retry-After`, settling on a reasonable default if the server doesn't set the `Retry-After` header. this is even more opinionated, however.)

janus_collector/src/lib.rs
branlwyd: (the current API setup also creates some tension in configuration: if the user only ever calls `poll_once`, the `collect_poll_wait_parameters` are effectively ignored/unused state)

branlwyd (changes requested):

janus_collector/src/lib.rs
divergentdave: I think it might be worth having `start_collection` and `poll_once` exposed for the sake of users that want to persist the state of collect jobs that are in flight. (i.e. a long-lived service that needs to handle process restarts gracefully, or a distributed service that needs to pass that intermediate state between servers) I do like the idea of dropping `poll_until_complete` in favor of an end-to-end `collect`.

While Janus will hand out an already-created collect job URI if it receives a duplicate CollectReq, I don't think DAP requires that leaders do so. Thus I'd like to provide downstream users a way to resume polling an existing URI, rather than making them send a new CollectReq.

branlwyd: Hmm -- I think users that wish to persist state could get away with persisting only the batch interval/aggregation parameter, and not the collect job URI. But the reason I think this is up to DAP spec interpretation:

> While Janus will hand out an already-created collect job URI if it receives a duplicate CollectReq, I don't think DAP requires that leaders do so. Thus I'd like to provide downstream users a way to resume polling an existing URI, rather than making them send a new CollectReq.

I think you are correct that Leaders are not required to return identical collect job URIs if they receive identical `CollectReq`s. However, all of the text around how aggregators validate repeated collections is phrased around the batch & aggregation parameter being collected, rather than the collect job URI. That is, if a collector sends two identical `CollectReq`s and receives two different collect job URIs, the correct interpretation of those URIs is as two "aliases" to the same collect job (i.e. batch & aggregation parameter), rather than as IDs to two separate collect jobs. IMO the latter interpretation does not fit with the DAP specification, as if these two `CollectReq`s are treated as separate collect jobs, I think we'd fall afoul of the following text:

> Check if the report has already been aggregated with this aggregation parameter. If this check fails, the input share MUST be marked as invalid with the error report-replayed.

... since treating the two `CollectReq`s as separate aggregations would effectively cause losing the first collect job to block reattempting collection. (and since the Collector updating its durable state is not transactional with the aggregators' state updates in handling `CollectReq`, treating two identical `CollectReq`s as resulting in separate aggregations would be another "data loss by design" scenario.)

All that said, the exact relation between "aggregations", "batches", "collect jobs", & "collect job URIs" might just be underspecified. Please correct me if your interpretation is different or I'm missing something.

Given the above, I don't see much value in exposing the collect job URI to the user -- a Collector that persists the state of collected data would do so indexed on `(batch_interval, aggregation_parameter)`. If the Collector crashed while waiting on a collection, it would figure out its state from what was last written to durable storage, then resend an identical `CollectReq` to resume collection (albeit with a potentially-different collect job URI). Indeed, even if the Collector wanted to store the collect job URI to avoid the need to send an identical `CollectReq`, if it crashed after receiving the URI but before storing it, it would need to send an identical `CollectReq` anyway. Since the gains to exposing the collect job URI are marginal, I suggest we simply don't expose it, and we remove `poll_once` in favor of `collect`.

(also, the details of the collect job URI might change eventually, e.g. with https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/278; not exposing the collect job URI to users of this library would likely let us do the necesssary updates without requiring a breaking change)

branlwyd (changes requested):

janus_core/src/message.rs
```diff
@@ -1038,11 +1038,78 @@ impl Decode for Report {
     }
 }
 
+/// PPM protocol message representing a request from the collector to the leader to provide
```
branlwyd, line 1041:
```suggestion
/// DAP protocol message representing a request from the collector to the leader to provide
```

janus_core/src/message.rs
```diff
@@ -1038,11 +1038,78 @@ impl Decode for Report {
...
+
+        Ok(Self {
+            task_id,
+            batch_interval,
+            agg_param,
+        })
+    }
+}
+
+/// PPM protocol message representing a leader's response to the collector's request to provide
```
branlwyd, line 1079:
```suggestion
/// DAP protocol message representing a leader's response to the collector's request to provide
```

branlwyd (approved):

PR #482, opened 2022-09-09
Update tokio requirement to 1.21

divergentdave:
This updates our version requirement on Tokio to 1.21. Note that there is no change to Cargo.lock; our lock file's version of Tokio is already at 1.21.0 due to some other recent upgrades. Since we upgraded `console-subscriber`, this is the only possible version we can select now (as it requires `tokio = ^1.21`), and later changes that regenerated the lockfile with Cargo reified this.


tgeoghegan (approved):
According to the [docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements), `^1.21` is equivalent to `1.21`, so I think we could delete the caret just to eliminate the suggestion that something more complicated is happening here.

PR #486, opened 2022-09-09
Migrate from `structopt` to `clap::Parser`

### LINKED ISSUE #177: Replace `structopt` with `clap`
Apparently [`structopt` is in maintenance mode](https://github.com/TeXitoi/structopt/issues/525), and we should now be using [`clap`](https://docs.rs/clap).

tgeoghegan:
`structopt` has been migrated into `clap`, so we now use `clap::Parser`, which is virtually a drop-in replacement for `structopt::Structopt`, using the instructions in `clap`'s handy migration guide[1]. The name `Structopt` is still exported from `clap`, but it's easy enough to adopt `clap::Parser` instead. There's no changes except some minor diff in the help output of binaries. We also add simple tests based on `clap::App::debug_assert` as suggested in the migration guide.

[1]: https://github.com/clap-rs/clap/blob/v3-master/CHANGELOG.md?plain=1#L584

Resolves #177

### DIFF
--- a/janus_server/tests/cmd/aggregation_job_creator.trycmd
+++ b/janus_server/tests/cmd/aggregation_job_creator.trycmd
@@ -4,23 +4,30 @@ janus-aggregation-job-creator [..]
 Janus aggregation job creator
 
 USAGE:
-    aggregation_job_creator [OPTIONS] --config-file <config-file>
-
-FLAGS:
-    -h, --help       Prints help information
-    -V, --version    Prints version information
+    aggregation_job_creator [OPTIONS] --config-file <CONFIG_FILE>
 
 OPTIONS:
-        --config-file <config-file>                path to configuration file [env: CONFIG_FILE=]
-        --database-password <database-password>    PostgreSQL password [env: PGPASSWORD=]
-        --datastore-keys <datastore-keys>...
-            datastore encryption keys, encoded in base64 then comma-separated [env: DATASTORE_KEYS=]
+        --config-file <CONFIG_FILE>
+            path to configuration file [env: CONFIG_FILE=]
 
-        --otlp-metrics-metadata <KEY=value>...
-            additional OTLP/gRPC metadata key/value pairs for the metrics exporter [env: OTLP_METRICS_METADATA=]
+        --database-password <DATABASE_PASSWORD>
+            PostgreSQL password [env: PGPASSWORD]
 
-        --otlp-tracing-metadata <KEY=value>...
-            additional OTLP/gRPC metadata key/value pairs for the tracing exporter [env: OTLP_TRACING_METADATA=]
+        --datastore-keys <DATASTORE_KEYS>
+            datastore encryption keys, encoded in base64 then comma-separated [env: DATASTORE_KEYS]
 
+    -h, --help
+            Print help information
+
+        --otlp-metrics-metadata <KEY=value>
+            additional OTLP/gRPC metadata key/value pairs for the metrics exporter [env:
+            OTLP_METRICS_METADATA=]
+
+        --otlp-tracing-metadata <KEY=value>
+            additional OTLP/gRPC metadata key/value pairs for the tracing exporter [env:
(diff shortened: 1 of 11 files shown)


janus_server/tests/cmd/aggregation_job_creator.trycmd
```diff
@@ -4,23 +4,30 @@ janus-aggregation-job-creator [..]
...
+    aggregation_job_creator [OPTIONS] --config-file <CONFIG_FILE>
 
 OPTIONS:
-        --config-file <config-file>                path to configuration file [env: CONFIG_FILE=]
-        --database-password <database-password>    PostgreSQL password [env: PGPASSWORD=]
-        --datastore-keys <datastore-keys>...
+        --config-file <CONFIG_FILE>
+            path to configuration file [env: CONFIG_FILE=]
+
+        --database-password <DATABASE_PASSWORD>
```
branlwyd, line 13: not caused by this PR: does `PGPASSWORD=` mean that, if we ever have `PGPASSWORD` set but then run the executable with a bad set of flags, the program will "helpfully" print our Postgres password on stdout?

(same for `DATASTORE_KEYS`, and presumably many more sensitive values we attempt to pass via environment variables)

branlwyd: It appears that `clap`'s help does indeed reflect the values of any environment variables it is configured to care about. From my workstation at current `main`:

```
$ PGPASSWORD=foo DATASTORE_KEYS=bar aggregation_job_creator --help
janus-aggregation-job-creator 0.1.11
Janus aggregation job creator

USAGE:
    aggregation_job_creator [OPTIONS] --config-file <config-file>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
        --config-file <config-file>                path to configuration file [env: CONFIG_FILE=]
        --database-password <database-password>    PostgreSQL password [env: PGPASSWORD=foo]
        --datastore-keys <datastore-keys>...
            datastore encryption keys, encoded in base64 then comma-separated [env: DATASTORE_KEYS=bar]

        --otlp-metrics-metadata <KEY=value>...
            additional OTLP/gRPC metadata key/value pairs for the metrics exporter [env: OTLP_METRICS_METADATA=]

        --otlp-tracing-metadata <KEY=value>...
            additional OTLP/gRPC metadata key/value pairs for the tracing exporter [env: OTLP_TRACING_METADATA=]
```

IMO this is a security vulnerability -- anyone using `clap` for argument handling with sensitive environment variables is one misconfigured push away from logging all secrets being passed to the binary via environment variables. Currently that includes us.

It looks like the expected fix is to call [`hide_env_values`](https://docs.rs/clap/3.2.20/clap/builder/struct.Arg.html#method.hide_env_values) on necessary `Args`. This method can be called when using derive-based flags -- see the "Raw attributes" subsection of [Arg Attributes](https://docs.rs/clap/3.2.20/clap/_derive/index.html#arg-attributes) for how this is done. [I'd really prefer this to be the default, given how commonly environment variables are used as a way to specify secret/sensitive values... but me editorializing on API design is less than helpful here. :-)]

Would you mind fixing this in this PR? (another PR is OK too, I just figure it's a small high-impact change & it touches a lot of code that you are already touching)

branlwyd: Hmm, an unknown/misspelled flag gets a shorter error message which does not emit environment variable values, so we're less likely to hit the "one bad push" scenario (the push would have to be so bad that our production binaries specified the `--help` flag, I think).

Still--we should fix this, IMO.

divergentdave (approved):
nit: We still have some `#[structopt]` attributes in a few places, I think we should swap those out for `#[clap]`. The derive macro for Parser will treat both the same, but the macros for Args and Subcommand only take `#[clap]`. Using `#[clap]` throughout would be good for consistency and prevent future surprises.

PR #491, opened 2022-09-12
Disable Daphne integration tests.

branlwyd:
These will be re-enabled once our DAP-02 implementation is complete.

Also clean up the integration_tests' Cargo.toml. Mark a few dependencies non-optional (previously they were gated on the daphne feature, but were actually required whether or not that feature was enabled), and mark a few others optional or remove them entirely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1567,7 +1567,6 @@ dependencies = [
  "futures",
  "hex",
  "http",
- "integration_tests",
  "interop_binaries",
  "itertools",
  "janus_client",
--- a/integration_tests/Cargo.toml
+++ b/integration_tests/Cargo.toml
@@ -10,11 +10,9 @@ publish = false
 daphne = [
     "dep:build_script_utils",
     "dep:daphne",
-    "dep:hex",
-    "dep:rand",
     "dep:serde",
     "dep:serde_json",
-    "dep:tokio",
+    "dep:tempfile",
     "interop_binaries/test-util",
 ]
 kube-openssl = ["kube/openssl-tls"]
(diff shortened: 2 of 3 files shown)


divergentdave (approved):
LGTM, but my personal preference would be to use `#[ignore]` over conditional compilation.

branlwyd:
Mind if I ask why the preference for `#[ignore]`? I don't mind changing if that approach is better.

I used conditional compilation because that lets me drop unnecessary dependencies -- AFAIK `#[ignore]` still requires the test to build, which requires building all of the relevant dependencies. Since the dependencies are in this case significant (building/pulling a Daphne image) I decided to go this route.

branlwyd:
(BTW, filed https://github.com/divviup/janus/issues/493 to track the need to re-enable this test.)

divergentdave:
Two thoughts in this direction:

- If we turn it off with conditional compilation, then there's a greater chance of masking compilation issues over time that we'd have to deal with when re-enabling it. (This is mitigated by `cargo clippy --workspace --all-targets --all-features` in our CI)
- Turning off a known-failing test with a feature means that we can't do `cargo test --all-features` anymore. (Though that's of low salience here, since I don't think we have any features that affect tests, like `multithreaded` over on libprio)

branlwyd:
Hmm, I hadn't considered that `cargo test --all-features` issue before. For now I'm going to go ahead with the conditional-compilation approach, but I'll switch to `#[ignore]` if conditional compilation ends up causing pain.

(FWIW, I'm less concerned about the first point because compilation failures can occur due to removing features just as easily as adding features. That is, you point out a real problem, but it's no more a problem with this approach than without it -- e.g. I already had to clean up a few compilation failures before I could successfully disable the `daphne` feature. I guess a "real" solution to this problem would be to configure our CI to build everything with every possible set of features, but I'm not sure we want to take on the combinatorial explosion of feature sets that would imply.)

tgeoghegan (approved):

PR #505, opened 2022-09-14
Introduce `ReportMetadata` and adopt it

tgeoghegan:


janus_server/src/aggregator.rs
```diff
@@ -4882,7 +5041,7 @@ mod tests {
             job_id: aggregation_job_id,
             prepare_steps: vec![PrepareStep {
                 nonce: Nonce::new(
-                    Time::from_seconds_since_epoch(54321),
+                    //Time::from_seconds_since_epoch(54321),
```
branlwyd, line 5044: nit: drop commented code

branlwyd (approved):

Issue #508, opened 2022-09-14
`run_test_capturing_logs` does not capture logs if the test panics


branlwyd:
The intent of `run_test_capturing_logs` is to run a test (represented by a function which is executed to run the test logic), capturing logs from relevant containers & writing them to the filesystem if the test fails.

`run_test_capturing_logs` expects failure to be indicated by the test function returning an "error" result value. However, common convention is for test code to fail by panicking -- it is common for test code to use macros such as `assert!`, `assert_eq!`, etc which result in a panic if the assertion fails. More concretely, some of the functions we test against (such as `submit_measurements_and_verify_aggregate`) will currently panic for many failure modes. We should support test failure via panic.

For example, if I set `JANUS_E2E_LOGS_PATH`, then induce a panic by adding a line to the start of `submit_measurements_and_verify_aggregate` to `panic!("uh oh")`, then run `cargo test -p integration_tests --test janus`, the test fails but nothing ends up written to the desired path. If instead I modify `submit_measurements_and_verify_aggregate` to `return Err(anyhow!("uh oh"))`, the test fails & logs are written to the desired path. My expectations would be that logs are written to the desired path in both cases.

Suggestions (each bullet point an alternative):
* `std::panic::catch_unwind` would let us catch the panic, and write logs at that point (before presumably re-panicking). I suppose the risk is that if the logs-writing process itself panics, we lose the original panic.
* `Drop` implementations are still run on panic. We could arrange that relevant types have a Drop implementation that writes logs on being dropped. We could likely use the `std::thread::panicking` function to determine if we are panicking during the drop, to allow us to maintain that we only write logs on test failure.

divergentdave:
I think we should use `std::panic::catch_unwind()`. Since a double panic results in an abort, the alternative solution would require that all log handling code be written panic-free. Moreover, `catch_unwind` directly encodes the behavior we want.

branlwyd:
Hmm, I'm not sure I agree:

First, since this code is for tests only, aborting the process on test failure is ultimately OK as long as we don't lose any important information about why the test failed.

* In the `catch_unwind` approach, if the log-writing code panics, the original panic might be lost entirely (we could hypothetically log it before we start writing logs, but I'd prefer to `resume_unwind` to ensure the original panic is rendered as closely as possible to if we didn't have this "log-on-failure" code at all -- the idea being that the fact this logs-writing functionality exists should be as transparent as possible). Logs will be lost based on where in the log-handling code the panic occurs.

[...]

PR #509, opened 2022-09-14
Fix log-export functionality to export logs on panic, too.

branlwyd:
This is required since the standard way for a test to fail is to panic.


branlwyd:
Closes #508.

tgeoghegan (approved):
I had explored using `catch_unwind`, but found it very awkward to do so with an async function. This solution neatly sidesteps all of that, though.

PR #511, opened 2022-09-15
Add example binary to perform collection from CLI

divergentdave:
This takes the new collector library and exposes it via a Clap CLI. Unit tests cover command-line argument handling, and I confirmed that it works end-to-end against interop-00. Here's the help text, for reference.

```
collect 0.1.11
Command-line DAP-PPM collector from ISRG's Divvi Up

USAGE:
    collect [OPTIONS] --task-id <TASK_ID> --leader <LEADER> --auth-token <AUTH_TOKEN> --hpke-config <HPKE_CONFIG> --hpke-private-key <HPKE_PRIVATE_KEY> --vdaf <VDAF> --batch-interval-start <BATCH_INTERVAL_START> --batch-interval-duration <BATCH_INTERVAL_DURATION>

OPTIONS:
    -h, --help       Print help information
    -V, --version    Print version information

DAP TASK PARAMETERS:
        --task-id <TASK_ID>
            DAP task identifier, encoded with base64url

        --leader <LEADER>
            The leader aggregator's endpoint URL

        --auth-token <AUTH_TOKEN>
            Authentication token for the DAP-Auth-Token HTTP header [env: AUTH_TOKEN=]

        --hpke-config <HPKE_CONFIG>
            DAP message for the collector's HPKE configuration, encoded with base64url

        --hpke-private-key <HPKE_PRIVATE_KEY>
            The collector's HPKE private key, encoded with base64url [env: HPKE_PRIVATE_KEY=]

VDAF ALGORITHM AND PARAMETERS:
        --vdaf <VDAF>          VDAF algorithm [possible values: count, countvec, sum, histogram]
        --bits <BITS>          Bit length of measurements, for use with --vdaf=sum
        --buckets <BUCKETS>    Comma-separated list of bucket boundaries, for use with
                               --vdaf=histogram
        --length <LENGTH>      Number of vector elements, for use with --vdaf=countvec

COLLECT REQUEST PARAMETERS:
        --batch-interval-duration <BATCH_INTERVAL_DURATION>
            Duration of the collection batch interval, in seconds

        --batch-interval-start <BATCH_INTERVAL_START>
            Start of the collection batch interval, as the number of seconds since the Unix epoch
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1653,9 +1653,12 @@ dependencies = [
 name = "janus_collector"
 version = "0.1.11"
 dependencies = [
+ "anyhow",
  "assert_matches",
  "backoff",
+ "base64",
  "chrono",
+ "clap",
  "derivative",
  "janus_core",
  "mockito",
@@ -1666,6 +1669,8 @@ dependencies = [
  "thiserror",
  "tokio",
  "tracing",
+ "tracing-log",
+ "tracing-subscriber",
  "url",
 ]
 
--- a/janus_collector/Cargo.toml
+++ b/janus_collector/Cargo.toml
@@ -25,8 +25,13 @@ tracing = "0.1.36"
 url = "2.2.2"
 
 [dev-dependencies]
+anyhow = "1"
 assert_matches = "1"
+base64 = "0.13.0"
 chrono = "0.4"
-janus_core = { path = "../janus_core", features = ["test-util"]}
+clap = { version = "3.2.20", features = ["derive", "env"] }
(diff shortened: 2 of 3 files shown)


tgeoghegan (approved):
LGTM! Any reason not to promote this to a binary built out of the `janus_collector` crate, so that users can get it just by doing `cargo build` on that crate?

divergentdave:
I think my main hesitation is regarding the extra dependencies that library users don't need. We'd either have to always depend on them, or make a feature depend on them, which then sets you back in ease of use.

Issue #445, opened 2022-08-30
`janus_core::message::{Time,Duration}` should be used only for message serialization/deserialization


branlwyd:
The `janus_core::message::{Time,Duration}` types are currently used for two purposes:
* DAP message serialization/deserialization.
* Representation of times/durations in various places throughout the codebase, such as configurations for various jobs.

Using our `Time`/`Duration` types for DAP message serialization/deserialization makes sense -- that is the purpose of those types.

OTOH, we should not use our `Time`/`Duration` for e.g. internally representing a duration of time in configurations or when performing time-based calculations:
* Doing so leads to us writing our own time math, which can be surprisingly error-prone.
* In the long run, DAP may move away from the need to represent timestamps or durations of time at all. It would be unfortunate if we were still using `Time`/`Duration` as a generic time/duration type at that point -- I suppose we'd end up having our own custom types for times & durations.

My suggestion would be that our internal `Time`/`Duration` types be used only "at the edge" -- converted from a standard representation of times/durations into the message types, just before serialization. For other purposes, we should use the appropriate types from the `chrono` crate, as we already do in some instances throughout our codebase.

(I think this same argument would apply to the `janus_core::message::Interval` type, except I'm not sure we have any "bad" usages in our codebase currently.)

inahga:
We have a few examples of this still outstanding, but don't really need to address this proactively. We can pull this out into the style guide and address it over time.

PR #515, opened 2022-09-16
Extend collect timeout in end_to_end test

divergentdave:
This should fix some CI flakiness I've been seeing today. The end_to_end integration test was taking 12-14s on my machine, and the code gave it ~15s (plus request time) to complete, which is uncomfortably close. I have extended that limit to 60s flat, and rewritten the loop using the backoff crate, to be less boutique.

I think the recent rewrite of `janus_interop_collector` likely caused this, because it made `/internal/test/collect_poll` requests much faster, as they no longer make subsidiary requests, instead only checking sync primitives in memory. This effectively decreased the amount of time we allowed the test to run before timing out. Combined with the constrained resources of CI builders, this pushed it past the deadline too often.

I separately plan to tune the various timers in each interop test container, so that aggregations finish quicker. (I have anecdotally noticed many database transaction conflicts)

I also fixed a compiler warning that only appears when building janus_server without the test-util feature.

### DIFF
--- a/interop_binaries/src/bin/janus_interop_aggregator.rs
+++ b/interop_binaries/src/bin/janus_interop_aggregator.rs
@@ -243,7 +243,7 @@ async fn main() -> anyhow::Result<()> {
     let aggregation_job_creator = Arc::new(AggregationJobCreator::new(
         Datastore::new(pool, crypter, clock),
         clock,
-        StdDuration::from_secs(5),
+        StdDuration::from_secs(2),
         StdDuration::from_secs(1),
         1,
         100,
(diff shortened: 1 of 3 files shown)


divergentdave:
I ran an aggregation with these binaries while capturing traces, and the main thing consuming time was the delay before jobs were picked up by the various leader components. I lowered those intervals, and it cut the total test time by an average of four seconds.

branlwyd (approved):
> I ran an aggregation with these binaries while capturing traces, and the main thing consuming time was the delay before jobs were picked up by the various leader components. I lowered those intervals, and it cut the total test time by an average of four seconds.

In the longer-term, I'd very much like to change job-configurations from using seconds-resolution Duration type to a type with a smaller resolution to allow our tests to wait much less time. (and that said -- the timeouts you selected are fine, but as the aggregation job creator is already using `std::time::Duration`, we could go even lower. I'd be OK with almost any nonzero delay, maybe down to a few milliseconds?)

divergentdave:
Agreed on swapping out the duration type, that sounds good. (this is tracked in #445) A few milliseconds might be too fast for now, as I worry that running the job discovery queries that frequently may worsen rates of transaction conflicts.

PR #517, opened 2022-09-19
Change field name for errors, use Display to format them

divergentdave:
tracing::instrument [uses the field name "error" when emitting an event upon returning an error](https://docs.rs/tracing-attributes/latest/src/tracing_attributes/expand.rs.html#207). We should use the same field name in our events for consistency. Additionally, errors are now formatted with Display rather than Debug where possible, again for consistency.

### DIFF
--- a/janus_server/src/binary_utils/job_driver.rs
+++ b/janus_server/src/binary_utils/job_driver.rs
@@ -10,7 +10,11 @@ use opentelemetry::{
     metrics::{Meter, Unit},
     KeyValue,
 };
-use std::{fmt::Debug, future::Future, sync::Arc};
+use std::{
+    fmt::{Debug, Display},
+    future::Future,
+    sync::Arc,
+};
 use tokio::{
     sync::Semaphore,
     time::{self, Instant},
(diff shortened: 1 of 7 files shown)


janus_server/src/binary_utils/job_driver.rs
```diff
@@ -195,12 +199,12 @@ where
...
-                            Ok(Err(err)) => {
-                                error!(?err, "Couldn't step job");
+                            Ok(Err(error)) => {
+                                error!(%error, "Couldn't step job");
                                 status = "error"
                             }
-                            Err(err) => {
-                                error!(?err, "Stepping job timed out");
+                            Err(_err) => {
+                                error!("Stepping job timed out");
```
divergentdave, line 207: Note that the error here is of type `tokio::time::error::Elapsed`. This doesn't provide any information that isn't conveyed in the message, so I dropped it. Only Debug is implemented on Elapsed, so this would have just printed `Elapsed(())` previously.

tgeoghegan (approved):

Issue #518, opened 2022-09-19
`janus_messages`: Export all DAP message structs from a new crate with minimal dependencies


tgeoghegan:
Janus currently splits its message definitions across `janus_core` and `janus_server`, both of which have `messages` modules. We'd like other DAP implementations to be able to use Janus' representations of protocol messages without having to pull in Janus' considerable dependency graph. The strategy we agreed on is to move all message definitions and support for encoding and decoding them into a single, new crate called `janus_messages`, which will otherwise have minimal dependencies (i.e., no cryptography).

tgeoghegan:
It would make sense to take the message decoder feature from `janus_cli` and put it in `janus_messages`, too.

tgeoghegan:
This has landed on `release/0.1` and `main`, but I want to keep this issue open until the crate gets published and @simon-friedberger confirms that it meets his needs, so that any further changes can be tracked under this issue.

PR #532, opened 2022-09-20
`release/0.1`: Introduce `janus_messages` crate

tgeoghegan:
Adds a new crate that contains Janus' Distributed Aggregation Protocol
message definitions and support for encoding and decoding them to/from
the wire format. By default, `janus_messages` does not pull in any
cryptography dependencies in order to make it easy to use.

Message definitions are moved from `janus_core::message` and
`janus_server::message` and are now all exported from `janus_messages`.
`janus_core::time` is also moved into `janus_messages` since its `Clock`
implementations are useful for generating messages.

The DAP message decoding feature is moved from a subcommand on
`janus_server/src/bin/janus_cli.rs` into a new tool, `dap-decode`, built
from the `janus_messages` crate.

The `janus_messages` crate is published to crates.io on Janus releases,
and the CI build verifies that `janus_messages` builds without any
features enabled.

Part of #518


tgeoghegan:
@simon-friedberger: this PR adds the `janus_messages` crate. It'd be helpful if you could let me know whether it meets your needs. In particular, I'm trying to figure out which of its dependencies I should make optional. I've gated the `ring` dependency on a feature,  but what about [`rand`](https://crates.io/crates/rand)?

janus_client/Cargo.toml
```diff
@@ -13,7 +13,8 @@ rust-version = "1.63"
 backoff = { version = "0.4.0", features = ["tokio"] }
 derivative = "2.2.0"
 http = "0.2.8"
-janus_core = { version = "0.1", path = "../janus_core" }
+janus_core = { path = "../janus_core" }
+janus_messages = { path = "../janus_messages" }
```
divergentdave, line 17: We will need versions for both of these dependencies before we next publish janus_client

janus_collector/Cargo.toml
```diff
@@ -15,7 +15,8 @@ test-util = []
 [dependencies]
 backoff = { version = "0.4.0", features = ["tokio"] }
 derivative = "2.2.0"
-janus_core = { version = "0.1", path = "../janus_core" }
+janus_core = { path = "../janus_core" }
+janus_messages = { path = "../janus_messages" }
```
divergentdave, line 19: Ditto, this needs version specifications

janus_collector/src/lib.rs
```diff
@@ -19,10 +20,10 @@
 //! // Supply DAP task paramenters.
 //! # let task_id = TaskId::random();
 //! # let (hpke_config, private_key) = janus_core::hpke::generate_hpke_config_and_private_key(
-//! #     janus_core::message::HpkeConfigId::from(0),
```
divergentdave, line 22: Given the changes here and in the use directive above, we should probably drop these `#` comments to show the reader what we're doing with generate_hpke_config_and_private_key and the rest.

divergentdave (changes requested):

janus_messages/src/time.rs
```diff
@@ -1,6 +1,6 @@
 //! Utilities for timestamps and durations.
```
tgeoghegan, line 1: FWIW: Brandon's message refactor in #534 means that we don't have to move this module from `janus_core` to `janus_messages` on `main`.

janus_client/Cargo.toml
tgeoghegan: Oops, I knew there must have been a good reason that was there! I also added this to `janus_core`'s dep on `janus_messages`.

divergentdave (approved):

janus_messages/src/lib.rs
```diff
@@ -1104,13 +1109,397 @@ impl Decode for CollectResp {
...
+        })
+    }
+}
+
+#[cfg(feature = "test-util")]
+pub mod test_util {
+    use super::{Nonce, Report, TaskId, Time};
+    use rand::{thread_rng, Rng};
+
+    pub fn new_dummy_report(task_id: TaskId, when: Time) -> Report {
```
branlwyd, line 1490: I think this method (and in general, all test-util functionality that is not used in `janus_messages` tests?) should live in Janus.

janus_messages/src/time.rs
branlwyd, line 1: I think this module should continue to live in `janus_core`, for a couple of reasons:

* The intent of `janus_messages` is to be a message serialization/deserialization crate. Including a clock abstraction is outside this crate's wheelhouse.
* I am unsure _we_ want to use this `Clock` abstraction in the long run (#445), and therefore even less sure we want to export this as part of our public interface. (it already is as part of `janus_core`, natch, but anything in `janus_messages` is not just technically public but explictly moved to allow others to depend on it)

My suggestion would be to:
* Move this module back to `janus_core`.
* Remove any functionality in `janus_messages` that relies on a `Clock`.
* Replace `Nonce::generate` with `Nonce::new` that receives takes a `Time` instead of generating one from a `Clock`.
* Re-implement the removed `Nonce::generate` functionality in Janus (`janus_core`?), likely via an extension trait on `Nonce`.

This implementation strategy would be minimally impactful on existing callers to e.g. `Nonce::generate()`; any files that called `Nonce::generate` would need to import the new extension trait, but would otherwise be unchanged.

branlwyd (changes requested):

janus_messages/src/lib.rs
```diff
@@ -419,15 +424,17 @@ impl Decode for Nonce {
 
 /// Checksum over DAP report nonces, defined in §4.4.4.3.
 #[derive(Copy, Clone, Debug, Default, Hash, PartialEq, Eq)]
-pub struct NonceChecksum([u8; SHA256_OUTPUT_LEN]);
+pub struct NonceChecksum([u8; 32]);
```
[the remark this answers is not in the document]
tgeoghegan: `NonceChecksum` does appear in the protocol message `AggregateShareReq`. However we could make `AggregateShareReq.checksum` an opaque `[u8; 32]` to keep the `NonceChecksum` type and its logic out of `janus_messages`.

janus_messages/src/lib.rs
```diff
@@ -193,7 +198,7 @@ impl Decode for Time {
     }
 }
 
-/// PPM protocol message representing a half-open interval of time with a resolution of seconds;
+/// DAP protocol message representing a half-open interval of time with a resolution of seconds;
 /// the start of the interval is included while the end of the interval is excluded.
 #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
 pub struct Interval {
```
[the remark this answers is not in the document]
tgeoghegan: A wrapper type for `janus_messages::Interval` struck me as the easiest way to integrate gracefully with `Row::get` and the stuff for constructing queries, so that's what I did here.

janus_messages/src/lib.rs
```diff
@@ -100,7 +105,7 @@ impl Display for Duration {
     }
 }
 
-/// PPM protocol message representing an instant in time with a resolution of seconds.
+/// DAP protocol message representing an instant in time with a resolution of seconds.
```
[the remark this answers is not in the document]
tgeoghegan: OK, I will move all the time math stuff into extension traits in either janus_core or janus_server, depending on where it's all used.

janus_messages/src/lib.rs
```diff
@@ -204,11 +108,8 @@ pub struct Interval {
 }
 
 impl Interval {
-    /// Create a new [`Interval`] from the provided start and duration. Returns an error if the end
-    /// of the interval cannot be represented as a [`Time`].
+    /// Create a new [`Interval`] from the provided start and duration.
     pub fn new(start: Time, duration: Duration) -> Result<Self, Error> {
-        start.add(duration)?;
-
```
tgeoghegan, line 211: I deleted this check because the `Time::add` method it uses is now up in `janus_server`. I'm not sure that it's good that we don't verify that start + duration doesn't overflow `Time`, but then technically such a message could be encoded on the wire, and it'd be the responsibility of some other representation of time (such as one introduced by #445) to detect that.

tgeoghegan: I decided to put an overflow check that doesn't depend on `Time::add` back in. It's gross to duplicate the code that checks for overflow, but this way we aren't changing the semantics of anything, and this can be cleaned up properly on `main`.

janus_messages/src/lib.rs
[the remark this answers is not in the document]
branlwyd: So, the DAP spec doesn't define a type called `NonceChecksum` at all -- `AggregateShareReq.checksum` is defined as `opaque checksum[32]`. We made the implementation choice to define our own type because (I suppose) that makes it easy to attach behavior to the type, but it's probably not appropriate to push the type & behavior onto folks in a message-serialization library.

That said, this is a pretty pedantic distinction -- especially now that the functionality has been moved out. If you don't find my argument moving, or if it's just more work than it's worth, I won't block exporting this type. :)

branlwyd (approved):
Thanks for addressing my (rather persnickety) comments! I'm hopeful that these changes on will either (a) be free on `main` due to other message changes or at least (b) be directly translatable to `main`.

tgeoghegan:
I see what you mean about `NonceChecksum`. I prefer to have named types for clarity even if it doesn't technically correspond to a proper protocol structure.

PR #534, opened 2022-09-21
Message changes for DAP-02.

branlwyd:
No functionality changes are implemented; in particular, this commit does not include any changes to implement fixed-size tasks.

While I'm touching all the messages, I did a few more cleanups:
 * Opaque types (e.g. Nonce, TaskId, BatchId, ...) now implement AsRef<[u8]> rather than a custom as_bytes() method.
 * Opaque types now implement From rather than using a custom new() method.
 * Random generation is now implemented by teaching rand's Standard distribution how to generate the type, rather than using a custom random() / generate() method. (The expected way to generate random values is now calling rand::random().)
 * Structural types no longer have pub fields. Getters have been implemented for all fields. I made the opinionated choice to return references even for Copy types, as this allows certain one-liners that otherwise would encounter lifetime issues. (e.g. let bytes = foo.task_id().as_ref())


janus_server/src/task.rs
```diff
@@ -344,6 +344,9 @@ impl Task {
     }
 }
 
+#[cfg(feature = "test-util")]
+impl Task {}
```
tgeoghegan, line 348: Does something break if we don't have an empty `impl`?

janus_server/src/task.rs
```diff
@@ -588,61 +588,63 @@ pub mod test_util {
...
-            aggregator_config_1.aead_id(),
-            aggregator_config_1.public_key().clone(),
-        );
-
-        let vdaf_verify_key = SecretBytes::new(
-            iter::repeat_with(|| thread_rng().gen())
-                .take(vdaf.verify_key_length())
-                .collect(),
-        );
+    impl Task {
```
tgeoghegan, line 591: Perhaps the `#[cfg(feature = "test-util")]` from line 347 is meant to go here? Though if this is needed in non-test configs, I don't mind exporting it.

janus_server/src/aggregator/aggregation_job_driver.rs
```diff
@@ -545,7 +546,7 @@ impl AggregationJobDriver {
         aggregation_job: AggregationJob<L, A>,
         stepped_aggregations: Vec<SteppedAggregation<L, A>>,
         mut report_aggregations_to_write: Vec<ReportAggregation<L, A>>,
```
tgeoghegan, line 548: nit: should these arguments be `&[SteppedAggregation]` and `&[ReportAggregation]` for consistency with `prep_steps`?

janus_server/src/aggregator.rs
```diff
@@ -95,28 +95,28 @@ pub enum Error {
...
     /// Corresponds to `reportTooLate`, §3.1
-    #[error("task {1:?}: stale report: {0}")]
-    ReportTooLate(ReportMetadata, TaskId),
+    #[error("task {1}: stale report: {0}")]
+    ReportTooLate(Nonce, TaskId),
     /// Corresponds to `reportTooEarly`, §3.1. A report was rejected becuase the timestamp is too far in the future, §4.3.4.
-    #[error("task {1:?}: report from the future: {0}")]
-    ReportTooEarly(ReportMetadata, TaskId),
+    #[error("task {1}: report from the future: {0}")]
+    ReportTooEarly(Nonce, TaskId),
```
tgeoghegan, line 102: I think it'd be nice to have the report timestamp in here, which will then show up in logs.

janus_server/src/message.rs
```diff
@@ -182,53 +243,139 @@ impl Decode for AggregationJobId {
...
+impl<Q: QueryType> AggregateInitializeReq<Q> {
     /// The media type associated with this protocol message.
     pub const MEDIA_TYPE: &'static str = "application/dap-aggregate-initialize-req";
+
+    /// Constructs an aggregate initialization request from its components.
+    ///
+    /// This method would typically be used for code which is generic over the query type.
+    /// Query-type specific code will typically call one of [`Self::new_time_interval`] or
+    /// [`Self::new_fixed_size`].
+
```
tgeoghegan, line 273:
```suggestion
```
Delete blank line so that doccomment will be correctly associated with `new()`

janus_core/src/message.rs
```diff
@@ -815,7 +898,7 @@ impl Decode for HpkeCiphertext {
 
         Ok(Self {
             config_id,
-            encapsulated_context,
+            encapsulated_key: encapsulated_context,
```
tgeoghegan, line 901: nit: rename variable `encapsulated_context` -> `encapsulated_key`

janus_core/src/message.rs
```diff
@@ -1093,96 +1149,390 @@ impl Decode for Report {
...
 
-#[cfg(test)]
-mod tests {
-    use super::{
-        CollectReq, CollectResp, Duration, Extension, ExtensionType, HpkeAeadId, HpkeCiphertext,
-        HpkeConfig, HpkeConfigId, HpkeKdfId, HpkeKemId, HpkePublicKey, Interval, Nonce, Report,
-        ReportMetadata, Role, TaskId, Time,
-    };
-    use assert_matches::assert_matches;
+pub mod query_type {
```
tgeoghegan, line 1392: nit/thought: If this is going to be a nested module, we could move it to its own file (`janus_core/src/message/query_type.rs`) just to limit the size of this file. I'm not sure if I care about how big any individual source file gets, though.

tgeoghegan (approved):
A few non-blocking nits but LGTM! Thanks for slogging through this.

janus_server/src/task.rs
branlwyd: Oops, nope, this shouldn't be here -- removed. (way too much context: I was initially going to put the new test-only `impl Test` block here, but I realized I could place it directly in the `test_utils` mod instead)

janus_server/src/task.rs
branlwyd: Indeed, though in this case it's not needed here directly as the `test_util` module it is placed into already requires the `test-util` feature.

divergentdave (approved):

janus_server/src/aggregator.rs
branlwyd: 👍🏻 While I was touching this code again, I moved the `TaskId` parameter to be first in all cases -- it's always rendered first, and conceptually it's the "largest-scoped" piece of information included in these errors (and I generally like specifying identifiers from largest to smallest scope)

janus_core/src/message.rs
branlwyd: 👍🏻 (also highlighted another spot where we need to use `{encode,decode}_u32_items` once we upgrade to a version of libprio that includes it)

janus_core/src/message.rs
branlwyd: Hmm -- I don't have strong feelings here about which way is best. My notion would be to leave this as-is, on the thought that a couple dozen extra lines of code in this file will probably be easier to navigate that an additional file in our codebase. But I'm happy to move if folks think that's the better course?

janus_core/src/hpke.rs
```diff
@@ -42,19 +42,21 @@ pub fn associated_data_for_report_share(
     task_id: TaskId,
     report_metadata: &ReportMetadata,
 ) -> Vec<u8> {
-    let mut associated_data = vec![];
+    let mut associated_data = Vec::new();
     task_id.encode(&mut associated_data);
-    report_metadata.time().encode(&mut associated_data);
-    report_metadata.nonce().encode(&mut associated_data);
-    encode_u16_items(&mut associated_data, &(), report_metadata.extensions());
+    report_metadata.encode(&mut associated_data);
```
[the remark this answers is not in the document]
branlwyd: 👍🏻 I justified that I would wait to add it until `public_share` as populated, but as the field exists there's no reason I can't go ahead and update this function now.

janus_core/src/message.rs
tgeoghegan: I don't have a strong opinion, so let's leave it as-is.

janus_core/src/hpke.rs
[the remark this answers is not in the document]
branlwyd: ... and I noticed I forgot to include `public_share` in the `Report` struct. shame cube

janus_server/src/aggregator/aggregation_job_driver.rs
branlwyd: First, the easy one: `report_aggregations_to_write` should stay as a `Vec`, as it needs to be mutable -- `process_response_from_helper` effectively uses this parameter to allow the caller to specify some report aggregations to write, and it needs to be able to add its own to the list. (I think this level of coupling is OK since this is a private method, which only exists because the "aggregate init" & "aggregate continue" implementations both eventually fall into the same logic.)

Now, the harder one: I changed `stepped_aggregations` as you suggest. This required a few changes as previously we were passing the `Vec` to pass ownership of the included `SteppedAggregation`, and ownership was handy because we were mutating some of the internal fields of the stepped aggregation. However, the mutated internal state is a `ReportAggregation`; I hope to do a pass over `datastore.rs` and remove public fields from structural types similarly to what this PR does to `message.rs`, so soon mutating the `ReportAggregation` won't work anyway. And, other than that, IMO it's cleaner to avoid mutation when possible.

Note this required a couple of new clones, but I think this is an OK tradeoff.

branlwyd: First, the easy one: `report_aggregations_to_write` should stay as a `Vec`, as it needs to be mutable -- `process_response_from_helper` effectively uses this parameter to allow the caller to specify some report aggregations to write, and it needs to be able to add its own to the list. (I think this level of coupling is OK since this is a private method, which only exists because the "aggregate init" & "aggregate continue" implementations both eventually fall into the same logic.)

Now, the harder one: I changed `stepped_aggregations` as you suggest. This required a few changes as previously we were passing the `Vec` to pass ownership of the included `SteppedAggregation`, and ownership was handy because we were mutating some of the internal fields of the stepped aggregation. However, the mutated internal state is a `ReportAggregation`; I hope to do a pass over `datastore.rs` and remove public fields from structural types similarly to what this PR does to `message.rs`, so soon mutating the `ReportAggregation` won't work anyway. And, other than that, IMO it's cleaner to avoid mutation when possible.

Note this required a couple of new clones, but I think this is an OK tradeoff.

PR #536, opened 2022-09-21
`main`: Introduce `janus_messages` crate

tgeoghegan:
Adds a new crate that contains Janus' Distributed Aggregation Protocol
message definitions and support for encoding and decoding them to/from
the wire format. By default, `janus_messages` does not pull in any
cryptography dependencies in order to make it easy to use.

Message definitions are moved from `janus_core::message` and
`janus_server::message` and are now all exported from `janus_messages`.

The DAP message decoding feature is moved from a subcommand on
`janus_server/src/bin/janus_cli.rs` into a new tool, `dap-decode`, built
from the `janus_messages` crate.

The `janus_messages` crate is published to crates.io on Janus releases,
and the CI build verifies that `janus_messages` builds without any
features enabled.

Part of https://github.com/divviup/janus/issues/518


janus_messages/src/lib.rs
```diff
@@ -126,7 +119,7 @@ impl Time {
     /// Construct a [`Time`] representing the instant that is a given number of seconds after
     /// January 1st, 1970, at 0:00:00 UTC (i.e., the instant with the Unix timestamp of
     /// `timestamp`).
-    pub fn from_seconds_since_epoch(timestamp: u64) -> Self {
```
branlwyd, line 129: Matching the `0.1` version: we should drop the `chrono` dependency from `janus_messages`.

tgeoghegan:
Oof, good catches there! I don't know how I missed all that. A commit must have gotten lost along the way while I was rebasing, which is a bit scary.

branlwyd (approved):

PR #538, opened 2022-09-22
Update uses of thread_rng() throughout codebase.

branlwyd:
Specifically, prefer random() for one-off values, and thread_rng().sample_iter() for repeated values. The former is (IMO) both simpler & clearer than what it replaces; the latter is a more opinionated change but avoids us needing to create a mutable variable initialized to some arbitrary value every time we want to create an array of random values.

Also, rename ENCODED_LEN -> LEN for all opaque types. I suppose the opaque types really are just an array of bytes, so it makes sense to talk about their "length".

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,3 +133,38 @@ metrics_config:
 ```
 
 The command line flag `--otlp-metrics-metadata` or environment variable `OTLP_METRICS_METADATA` may alternately be used to supply gRPC metadata for the metrics exporter.
+
+## Code Style
+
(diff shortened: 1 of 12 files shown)


tgeoghegan (approved):
I had a similar thought while working through another change. I think standardizing on one way to randomly generate bytes is a great idea. If we have conventions like this that aren't enforced by tools like `clippy`, maybe we should write them down in the project README?

divergentdave (approved):

PR #546, opened 2022-09-26
`janus_server`: correct errors from `/hpke_config`

### LINKED ISSUE #475: Deal with optional `task_id` parameter to `hpke_config` requests
We might simply choose to require that `task_id` be present, but we should verify that our behavior is consistent with the protocol text.

https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/289
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/303

tgeoghegan:
draft-ietf-ppm-dap-02 allows aggregators to use a single HPKE config for many tasks. In Janus we choose to use one HPKE config per task (for now), but we must use the HTTP statuses and problem document types mandated by DAP.

Resolves #475

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -96,28 +96,32 @@ pub enum Error {
     /// Error handling a message.
     #[error("invalid message: {0}")]
     Message(#[from] janus_messages::Error),
-    /// Corresponds to `reportTooLate`, §3.1
+    /// Corresponds to `reportTooLate`, §3.2
     #[error("task {0}: report {1} too late: {2}")]
     ReportTooLate(TaskId, ReportId, Time),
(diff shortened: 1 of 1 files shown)


tgeoghegan:
Draft for now so I can see whether CI works. I also want to add support for this to `janus_client`.

tgeoghegan:
After some more thought, I don't think we need any `janus_client` support. The task ID is always known in the client (it's needed to fill out a `ReportMetadata`), so we can always set the `task_id` query parameter in `hpke_config` requests. My reading of [the protocol text](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.3.1-2) is that even if an implementation serves a single HPKE config from `/hpke_config`, it still has to answer requests to `/hpke_config?task_id=$TASK_ID` provided it recognizes `$TASK_ID`.

divergentdave (approved):
LGTM modulo updating expected status codes in tests

Issue #344, opened 2022-07-28
DAP compliance: handle HTTP DELETE requests to collect job URIs


branlwyd:
According to [DAP-01 section 4.4.1](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-01.html#name-collection-initialization), the leader "`MUST` retain a collect job's results until the collector sends an HTTP DELETE request to the collect job URI".

We currently don't support the `DELETE` verb at all. We should support deleting results when we receive a `DELETE` request.

tgeoghegan:
We should make sure we are in line with the new text in draft-02: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/292

tgeoghegan:
This is implemented in PRs #576, #579, #580, #581. Collect jobs now get marked deleted persistently in the database. However I haven't actually implemented the deletion of any data related to a collect job (tracked in #313). Doing that isn't actually required for DAP-02 compliance (nobody outside of Janus can tell that we have or have not deleted any data), and I think we have some open design questions around how we would approach GC. The obvious thing to do is to immediately delete all data associated with a collect job upon its deletion, but there's no reason to do that work synchronously while `janus_aggregator` handles a request, or even to do it in `janus_aggregator` at all. We might build a whole new component that runs as a cronjob and scans the datastore looking for stale data to garbage collect.

PR #576, opened 2022-09-27
Introduce `CollectJobState::Deleted`

### LINKED ISSUE #242: Update `{put,update}_collect_job` to match other `put`/`update` operations
Most `put`/`update` operations take a reference to the object being put or updated. The collect job operations instead take the fields that are updated. The collect job operations should be updated to  match the other datastore operations for consistency (and simpler code around the places that use the operations).

tgeoghegan:
For https://github.com/divviup/janus/issues/344, we introduce a "Deleted" state for a collect job. We have to distinguish from the "Abandoned" state so that we can eventually return the right HTTP status if a collector polls the collect job URI.

This commit defines the new state in the SQL schema and adds support for entering it using `Datastore::update_collect_job`. Along the way, we resolve https://github.com/divviup/janus/issues/242: `put_collect_job` and `update_collect_job` now take
`&CollectJob` as an argument instead of just the field values to set.

Part of https://github.com/divviup/janus/issues/344
Resolves https://github.com/divviup/janus/issues/242

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -156,7 +156,8 @@ CREATE TABLE batch_unit_aggregations(
 CREATE TYPE COLLECT_JOB_STATE AS ENUM(
     'START',     -- the aggregator is waiting to run this collect job
     'FINISHED',  -- this collect job has run successfully and is ready for collection
-    'ABANDONED'  -- this collect job has been abandoned & will never be run again
+    'ABANDONED', -- this collect job has been abandoned & will never be run again
+    'DELETED'    -- this collect job has been deleted
 );
 
 -- The leader's view of collect requests from the Collector.
--- a/janus_server/src/datastore.rs
+++ b/janus_server/src/datastore.rs
@@ -1447,13 +1447,20 @@ impl<C: Clock> Transaction<'_, C> {
     /// If a collect job corresponding to the provided values exists, its UUID is returned, which
     /// may then be used to construct a collect job URI. If that collect job does not exist, returns
     /// `Ok(None)`.
-    #[tracing::instrument(skip(self, encoded_aggregation_parameter), err)]
-    pub(crate) async fn get_collect_job_id(
+    #[tracing::instrument(skip(self, aggregation_parameter), err)]
+    pub(crate) async fn get_collect_job_id<const L: usize, A>(
         &self,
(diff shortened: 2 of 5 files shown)


tgeoghegan:
Downgrading to draft while I tweak this further.

divergentdave (approved):

db/schema.sql
```diff
@@ -156,7 +156,8 @@ CREATE TABLE batch_unit_aggregations(
 CREATE TYPE COLLECT_JOB_STATE AS ENUM(
     'START',     -- the aggregator is waiting to run this collect job
     'FINISHED',  -- this collect job has run successfully and is ready for collection
-    'ABANDONED'  -- this collect job has been abandoned & will never be run again
+    'ABANDONED', -- this collect job has been abandoned & will never be run again
+    'DELETED'    -- this collect job has been deleted due to an explicit request by the collector
```
branlwyd, line 160: I think this "deleted" state should be used for GC'ed collect jobs, as well, once we implement GC. WDYT?

If you agree, I think this comment should be updated, either to remove the "due to an explicit request by the collector" or to make it clear that's just one way that a collect job can be deleted.

janus_server/src/datastore.rs
```diff
@@ -2973,14 +2983,34 @@ pub mod models {
...
 
+    impl<const L: usize, A: vdaf::Aggregator<L>> CollectJob<L, A>
+    where
+        for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
+    {
+        /// Create a new `CollectJob` in the `Start` state, with a random collect job ID.
+        pub(crate) fn new(
+            task_id: TaskId,
+            batch_interval: Interval,
+            aggregation_param: &A::AggregationParam,
```
branlwyd, line 2994: nit: this should take `A::AggregationParam` rather than `&A::AggregationParam`; the idea is that this function needs to own an `A::AggregationParam`. It currently does so by `clone`ing a caller-provided reference. But if the caller happens to already own an aggregation param they want to pass in, writing the constructor forces a `clone` where otherwise one would not be necessary. And if the caller only has a reference, they can call `clone` themselves.

branlwyd (approved):

PR #585, opened 2022-09-29
Re-export janus_messages types from janus_core

divergentdave:
This re-adds a `janus_core::message` module, with re-exports of types from `janus_messages`. This will blunt the impact of the refactoring on backwards compatibility, but we'll still have to accept some backwards incompatibility due to methods moving from structs to extension traits. This is a 0.1-only change.

### DIFF
--- a/janus_core/src/lib.rs
+++ b/janus_core/src/lib.rs
@@ -31,3 +31,11 @@ impl Runtime for TokioRuntime {
         tokio::task::spawn(future)
     }
 }
+
+pub mod message {
+    pub use janus_messages::{
+        CollectReq, CollectResp, Duration, Error, Extension, ExtensionType, HpkeAeadId,
+        HpkeCiphertext, HpkeConfig, HpkeConfigId, HpkeKdfId, HpkeKemId, HpkePublicKey, Interval,
+        Nonce, NonceChecksum, Report, Role, RoleParseError, TaskId, Time,
+    };
+}


tgeoghegan (commented):
This change is sound in that it accomplishes what it says it does, but I'm not sure how I feel about `pub use` in Rust crates. I think it has a tendency to obscure where types really come from and can cause some headaches when updating dependencies, because surprising compiler errors can result if your `janus_core` and `janus_messages` are not in lockstep. 

That being said, I don't have a good enough argument against this to block it, and I'm happy to defer to the group consensus, so if you and @branlwyd want to go ahead with this then I'm OK with it.

divergentdave:
Generally speaking, I think re-exports can help with mismatches of types from different crate versions, if the larger library provides enough re-exports that its API can be used by clients without the client re-importing transitive dependencies. (For example, if `opentelemetry-otlp` re-exported a few metadata-related types, we could drop our `tonic` dependency entirely. The re-exports would move in lock-step with the methods we pass metadata objects to during future breaking change upgrades.) The compiler errors are fundamentally due to function signatures that use types originating in different crates. Re-exports only add a layer of indirection that obfuscates their origin.

In this particular case, the re-exports are instead papering over a refactor. Going forward, we can require all users of 0.2 to import `janus_messages` directly, but as of `janus_core` 0.1.13, there was no `janus_messages` crate available. With the re-exports here, I was able to upgrade the Janus dependencies of a load generation command with Cargo.toml changes and imports of the two new `TimeExt` traits.

branlwyd (approved):
I'm OK using `pub use` in this way. It effectively creates a type alias, which is what we want in this case.

Issue #470, opened 2022-09-06
Verify that collect requests cannot collect overlapping batch intervals


tgeoghegan:
See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/277

I think maybe we already implement this in Janus.

branlwyd:
I think we implement this too, but we should certainly double-check.

tgeoghegan:
The checks we have in [`janus_server::aggregate_share::validate_batch_lifetime_for_collect`](https://github.com/divviup/janus/blob/d38cd7b0f7fe405a9506fcd354336cd2a19965d8/janus_server/src/aggregator/aggregate_share.rs#L454) do implement all the conditions required by [DAP-02](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.5.6), except that we don't return the correct problem types.

tgeoghegan:
#586 mostly resolves this, except that `batchOverlap` is only checked in `collect_job_driver` but not while handling a `CollectReq`.

PR #586, opened 2022-09-29
Correct errors for DAP-02 collect, agg share reqs

tgeoghegan:
DAP-02 updates the "Batch Validation" section [1] with several new error types. This commit updates `janus_server` with the new problem document types, returns them as appropriate, and adds tests to verify that correct errors are constructed.

The exception is `batchOverlap`: while the `collect_job_driver` does correctly check for that condition when stepping a collect job, DAP-02 requires that this error be presented to the collector while handling a `CollectReq`. Meeting that requirement will require more significant refactoring that will arrive in a later pull request.

Part of #470

[1]: https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.5.6

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -134,13 +134,14 @@ pub enum Error {
     /// An error from the underlying VDAF library.
     #[error("VDAF error: {0}")]
     Vdaf(#[from] vdaf::VdafError),
-    /// A collect or aggregate share request was rejected because the interval is invalid, per
-    /// §4.5.6
+    /// A collect or aggregate share request was rejected because the interval failed boundary
+    /// checks (§4.5.6).
+    // TODO(#468): Augment this error so it can represent invalid fixed-size queries, too.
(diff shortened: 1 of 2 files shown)


janus_server/src/aggregator.rs
```diff
@@ -2727,6 +2745,8 @@ mod tests {
     use uuid::Uuid;
     use warp::{cors::CorsForbidden, reply::Reply, Rejection};
 
+    const DUMMY_VERIFY_KEY_LENGTH: usize = dummy_vdaf::Vdaf::VERIFY_KEY_LENGTH;
```
branlwyd, line 2748: this should be `#[cfg(test)]`, I think

branlwyd (approved):

janus_server/src/aggregator.rs
tgeoghegan: It's in `mod tests` so it should already be restricted to test builds.

PR #592, opened 2022-09-30
DAP-02: populate CollectResp::report_count.

### LINKED ISSUE #469: Include `report_count` in `CollectResp`
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/163

branlwyd:
I also update the interop collector to return the report count; technically this could have been in another PR, but I need this functionality to augment the integration tests to check that the correct number of reports are returned. (This should be specified; I'm going to do so in a followup PR.)

Closes #469.

### DIFF
--- a/janus_collector/src/lib.rs
+++ b/janus_collector/src/lib.rs
@@ -240,7 +240,7 @@ impl<P> CollectJob<P> {
 /// or indicate that the collection is still being processed.
 enum PollResult<T> {
     /// The collection result from a completed collect request.
-    CollectionResult(#[derivative(Debug = "ignore")] CollectionResult<T>),
+    CollectionResult(#[derivative(Debug = "ignore")] Collection<T>),
     /// The collect request is not yet ready. If present, the [`RetryAfter`] object is the time at
     /// which the leader recommends retrying the request.
     NextAttempt(Option<RetryAfter>),
@@ -248,12 +248,12 @@ enum PollResult<T> {
 
 /// The result of a collection operation.
 #[derive(Debug)]
-pub struct CollectionResult<T> {
+pub struct Collection<T> {
     report_count: u64,
     aggregate_result: T,
 }
 
-impl<T> CollectionResult<T> {
+impl<T> Collection<T> {
     /// Retrieves the number of client reports included in this collection.
     pub fn report_count(&self) -> u64 {
         self.report_count
@@ -266,8 +266,8 @@ impl<T> CollectionResult<T> {
 }
 
 #[cfg(feature = "test-util")]
-impl<T> CollectionResult<T> {
-    /// Creates a new [`CollectionResult`].
+impl<T> Collection<T> {
+    /// Creates a new [`Collection`].
     pub fn new(report_count: u64, aggregate_result: T) -> Self {
         Self {
             report_count,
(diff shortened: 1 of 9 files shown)


janus_collector/src/lib.rs
```diff
@@ -236,16 +236,75 @@ impl<P> CollectJob<P> {
...
+    /// The collection result from a completed collect request.
+    CollectionResult(#[derivative(Debug = "ignore")] CollectionResult<V>),
     /// The collect request is not yet ready. If present, the [`RetryAfter`] object is the time at
     /// which the leader recommends retrying the request.
     NextAttempt(Option<RetryAfter>),
 }
 
+/// The result of a collection operation.
+#[derive(Debug)]
+pub struct CollectionResult<V: vdaf::Collector>
```
branlwyd, line 254: I'm not so sure about this name: it's slightly awkward since we sometimes have to handle types like `Result<CollectionResult>`. Suggestions for better names welcome.

branlwyd:
See also https://github.com/divergentdave/draft-dcook-ppm-dap-interop-test-design/pull/9.

janus_collector/src/lib.rs
```diff
@@ -236,16 +236,75 @@ impl<P> CollectJob<P> {
 
 #[derive(Derivative)]
 #[derivative(Debug)]
-/// The result of a collect request poll operation. This will either provide the aggregate result
+/// The result of a collect request poll operation. This will either provide the collection result
 /// or indicate that the collection is still being processed.
-enum PollResult<T> {
-    /// The aggregate result from a completed collect request.
-    AggregateResult(#[derivative(Debug = "ignore")] T),
+enum PollResult<V: vdaf::Collector>
```
divergentdave, line 241: I prefer keeping the type constraints "pushed down" here, and parametrizing `PollResult` and `CollectionResult` on the aggregate result type, not the VDAF type. This models what they do better, as they only act as a container for this result, and don't otherwise make use of the `Vdaf` trait. It also would allow eliminating many where clauses when working with just these two types.

divergentdave (approved):

janus_collector/src/lib.rs
tgeoghegan: I suppose it represents a completed collect job, but `CompletedCollectJob` isn't much better than what you've got.

tgeoghegan (approved):

janus_collector/src/lib.rs
branlwyd: Hmm, I made this change, but IMO it's clearer to talk about a `PollResult`/`CollectionResult` for a given VDAF -- otherwise, the type system doesn't make it clear which VDAF the concrete result type is associated with. That is, specifying the VDAF (& using the VDAF's associated type to specify the concrete stored type) over specifying the concrete type directly provides more information to both the type system & the programmer, and I think there are no cases where we can't specify the VDAF in use.

The infectious `where` clauses are indeed a pain. I took some time looking to see if an equivalent clause could be specified in a non-infectious way, but as far as I could tell this is a (particularly annoying) rough edge of Rust at the moment.

tgeoghegan: I wonder if we could spare ourselves some `where`s in Janus if we defined a supertrait over `prio::vdaf::Vdaf` that is just `Vdaf` plus the extra bounds we end up writing everywhere. Then we could make everything generic over `JanusVdaf` or `JanusAggregator` instead of `prio::vdaf::Vdaf` or `prio::vdaf::Aggregator`. I wouldn't tackle that here, though.

branlwyd:
I tried naming `CollectResult` as `Collection` instead, I think it reads pretty well. The ultimate result type of the `collect` method is now `Result<Collection<V::AggregateResult>, Error>` which reads fairly easily.

PR #608, opened 2022-10-05
Parse problem types in errors from the helper

### LINKED ISSUE #381: Progressive enhancement of error messages when Janus receives a problem details document
Janus's aggregator and job drivers should attempt to read a problem type URI out of error responses, and include it in error messages if present. This is parallel to #233, which applies to the client.

divergentdave:
This parses the problem type out of error responses from the helper, where available. Closes #381.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -66,6 +66,7 @@ use std::{
     future::Future,
     io::Cursor,
     net::SocketAddr,
+    str::FromStr,
     sync::Arc,
     time::Instant,
 };
@@ -176,8 +177,8 @@ pub enum Error {
     #[error("HTTP client error: {0}")]
     HttpClient(#[from] reqwest::Error),
     /// HTTP server returned an error status code.
-    #[error("HTTP response status {0}")]
-    Http(StatusCode),
+    #[error("HTTP response status {0} (problem type {1:?})")]
+    Http(StatusCode, Option<DapProblemType>),
     /// An error representing a generic internal aggregation error; intended for "impossible"
     /// conditions.
     #[error("internal aggregator error: {0}")]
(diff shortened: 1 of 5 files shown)


janus_server/src/aggregator.rs
```diff
@@ -176,8 +177,8 @@ pub enum Error {
     #[error("HTTP client error: {0}")]
     HttpClient(#[from] reqwest::Error),
     /// HTTP server returned an error status code.
-    #[error("HTTP response status {0}")]
-    Http(StatusCode),
+    #[error("HTTP response status {0} (problem type {1:?})")]
+    Http(StatusCode, Option<DapProblemType>),
```
tgeoghegan, line 181: This works, but let's say that for some error, a helper emitted a problem document with some extra keys in it (maybe some DAP implementation chooses to put a backtrace indicating where the error occurred in its problem documents). Our leader wouldn't preserve that extra information. So I wonder if what we want is `aggregator::Error::Http(StatusCode, Option<Rfc7807ProblemDocument>)` (where of course we are missing a Rust type to represent a document, but there's a crate out there for this), so that we could preserve the helper's entire problem document?

tgeoghegan (approved):

janus_server/src/aggregator.rs
tgeoghegan: All that being said, I'm not sure how to interpret [RFC 7807 section 3.2](https://www.rfc-editor.org/rfc/rfc7807#section-3.2). It could be saying that while DAP is allowed to define new problem document keys, DAP implementations may not add new keys to their documents.

But in any case, I think implementations are free to put whatever they want in the `details` member (RFC 7807 explicitly recommends against machines parsing it), so it'd be nice if our solution here preserved the helper problem document's `details`.

divergentdave: I agree, each type URI defines its own JSON object 'shape', so for all the `urn:ietf:params:ppm:dap:error:...` types, the DAP document gets to define the shape, and adding anything extra would not be forwards-compatible.

I rewrote the error handling to use `http_api_problem::HttpApiProblem`, which captures all fields, and will print the title and detail as part of its `Display` implementation.

tgeoghegan (approved):

PR #611, opened 2022-10-06
`janus_cli`: correctly handle datastore key secret

tgeoghegan:
`janus_cli` takes arguments for the Kubernetes secret where the datastore keys are, as well as the key into the secret. However, there were two bugs. `datastore-keys-secret-data-key` was being ignored completely, and also both its default value and the value hard coded into `fn create_datastore_key` were wrong. They used `datastore_key` instead of `datastore_keys` (note plural), which disagrees with what the `janus_server` components use.


branlwyd (approved):

janus_server/src/bin/janus_cli.rs
```diff
@@ -280,7 +282,7 @@ struct KubernetesSecretOptions {
         env = "DATASTORE_KEYS_SECRET_KEY",
         takes_value = true,
         help = "Key into data of datastore key Kubernetes secret",
-        default_value = "datastore_key"
+        default_value = "datastore_keys"
```
divergentdave, line 285: Same issue: I think this default value is incorrect. Our deployment includes `DATASTORE_KEYS: <set to the key 'datastore_key' in secret 'datastore-key'> Optional: true`, and the Janus component binaries read the environment variable `DATASTORE_KEYS` via clap.

divergentdave (changes requested):

divergentdave (approved):

PR #612, opened 2022-10-06
`janus_cli`: correctly handle datastore key secret

tgeoghegan:
`janus_cli` takes arguments for the Kubernetes secret where the datastore keys are, as well as the key into the secret. However, there were two bugs. `datastore-keys-secret-data-key` was being ignored completely, and also both its default value and the value hard coded into `fn create_datastore_key` were wrong. They used `datastore_key` instead of `datastore_keys` (note plural), which disagrees with what the `janus_server` components use.

### DIFF
--- a/janus_server/src/bin/janus_cli.rs
+++ b/janus_server/src/bin/janus_cli.rs
@@ -128,6 +128,7 @@ impl Command {
                     kube_client,
                     k8s_namespace,
                     &kubernetes_secret_options.datastore_keys_secret_name,
+                    &kubernetes_secret_options.datastore_keys_secret_data_key,
                 )
                 .await
             }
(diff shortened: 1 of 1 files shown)


janus_server/src/bin/janus_cli.rs
```diff
@@ -282,7 +284,7 @@ struct KubernetesSecretOptions {
         env = "DATASTORE_KEYS_SECRET_KEY",
         num_args = 1,
         help = "Key into data of datastore key Kubernetes secret",
-        default_value = "datastore_key"
+        default_value = "datastore_keys"
```
divergentdave, line 287: I think this default value is incorrect, see https://github.com/divviup/janus-ops/blob/148ca62f37758275e70cbe030fa5fb5babffa967/terraform/main_gcp/main.tf#L221

divergentdave (approved):

PR #614, opened 2022-10-06
janus_collector: Store problem details error info

divergentdave:
This adds parsing of problem details to `janus_collector`, similar to #608 and #613. This will be a breaking change for 0.2, as it changes a variant of a public enum.

### DIFF
--- a/janus_collector/src/lib.rs
+++ b/janus_collector/src/lib.rs
@@ -69,7 +69,7 @@ use prio::{
 };
 use reqwest::{
     header::{HeaderValue, ToStrError, CONTENT_TYPE, LOCATION, RETRY_AFTER},
-    StatusCode,
+    Response, StatusCode,
 };
 use retry_after::FromHeaderValueError;
 use retry_after::RetryAfter;
@@ -78,6 +78,7 @@ use std::{
     time::{Duration as StdDuration, SystemTime},
 };
 use tokio::time::{sleep, Instant};
+use tracing::warn;
 use url::Url;
 
 /// Errors that may occur when performing collections.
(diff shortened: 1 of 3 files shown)


janus_collector/src/lib.rs
```diff
@@ -344,16 +345,39 @@ where
...
+        .await;
+
+        let response = match response_res {
+            // Successful response or unretryable error status code:
+            Ok(response) => {
+                let status = response.status();
+                if status == StatusCode::SEE_OTHER {
+                    response
+                } else if status.is_client_error() || status.is_server_error() {
+                    if let Ok(mut problem_details) = response.json::<HttpApiProblem>().await {
```
branlwyd, line 357: This logic is slightly awkward as any failure to parse the body as a problem document will result in (a) the parsing error being silently ignored and (b) a problem document with only the status code populated being returned. My bigger concern is the former; silently dropping errors is not great.

This is complicated by the fact that (by my reading of the DAP spec) the body being a problem document is only a _suggestion_: an implementation is free to populate the body however it likes.

WDYT about the following logic:
1) If the response has content type `application/problem+json`, attempt to parse the body as a problem document. If parsing fails, log the error, and return a problem document based on the status code.
2) Otherwise, return a problem document based on the status code.

But I'm not sure if using the content type is mandatory; perhaps the DAP specification should be updated to say so? (if nothing else, problem documents can also be rendered as XML, and it'd be DAP-compliant to return an XML problem document as `application/problem+xml`; Janus wouldn't know how to handle this response.)

janus_collector/src/lib.rs
```diff
@@ -344,16 +345,39 @@ where
...
+                        return Err(Error::Http(problem_details));
+                    } else {
+                        return Err(Error::Http(HttpApiProblem::new(status)));
+                    }
+                } else {
+                    return Err(Error::Http(HttpApiProblem::new(status)));
+                }
+            }
+            // Retryable error status code, but ran out of retries:
+            Err(Ok(response)) => {
```
branlwyd, line 368: nit: these two match arms are very similar (but not identical), and similar logic appears below as well. Could we introduce a helper function (or something) to avoid repeating the process-body-into-problem-document logic?

branlwyd (changes requested):

branlwyd (approved):

PR #616, opened 2022-10-06
Update Task for DAP-02.

branlwyd:
Specifically:
* Introduce query_type field, determining how batches of reports are determined.
* Add task_expiration field, determining when a task naturally expires.
* Rename min_batch_duration to time_precision.
* Rename max_batch_lifetime to max_batch_query_count.

Also, clean up Task while I'm touching it: make it immutable, with methods returning references. I also introduce a test-only TaskBuilder abstraction which allows building a Task by specifying a few required fields, and using default values for other fields (while allowing those defaults to be overridden).


branlwyd:
Thought: I could get rid of `TaskBuilder` entirely, placing the `with_foo` methods as (`test-util`-only) methods on `Task` itself. What do y'all think?

janus_server/src/task.rs
```diff
@@ -253,43 +273,120 @@ impl Task {
...
             collector_hpke_config,
             aggregator_auth_tokens,
             collector_auth_tokens,
-            hpke_keys: hpke_configs,
+            hpke_keys,
         })
     }
 
+    /// Retrieves the task ID associated with this task.
+    pub fn task_id(&self) -> &TaskId {
```
tgeoghegan, line 304: `task.task_id()` stutters. How about `Task::id(&self) -> &TaskId`? That would also be consistent with `HpkeConfig::id(&self)`.

janus_server/src/aggregator/aggregate_share.rs
```diff
@@ -568,23 +569,26 @@ where
     for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
 {
     // Check how many rows in the relevant table have an intersecting batch interval.
-    // Each such row consumes one unit of batch lifetime (§4.6).
-    let intersecting_intervals: Vec<_> = match task.role {
+    // Each such row consumes one unit of query count (§4.6).
```
tgeoghegan, line 572: nit: I think this should refer to section [4.5.6](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#name-batch-validation) or [4.5.7](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.5.7) now. Though there's lots of incorrect section references in Janus.

tgeoghegan (approved):
> Thought: I could get rid of `TaskBuilder` entirely, placing the `with_foo` methods as (test-util-only) methods on Task itself. What do y'all think?

That approach would work well, too, but I'm not sure I see that it's necessarily better than what you already have, so I would choose the option where you don't have to rewrite a bunch of code. i.e., keep `TaskBuilder` as is. 

If we wanted to invest effort in doing better, then I think we should find a crate that implements the builder pattern with a macro. For instance, [`derive_builder`](https://crates.io/crates/derive_builder), but that's just the first thing that came up when I searched for "rust builder pattern" so I'm not sure if it's the "good" one (though it does seem to have lots of nice features).

janus_server/src/task.rs
branlwyd: I went back and forth on this -- the reason I didn't go for it originally is that e.g. reports aren't actually identified just by `report_id`; their effective ID is actually `(task_id, report_id)`. So an `id()` method would return something like `(&TaskId, &ReportId)`.

Of course, that doesn't apply to tasks (since there is no higher-level entity tasks are conceptually "inside of"), and the stuttering bugs me too. Since you noticed it as well, I fixed it here as well as for ReportMetadata & AggregationJob. (The last ID type, BatchID, doesn't yet correspond directly to a defined data structure.)

branlwyd: I went back and forth on this -- the reason I didn't go for it originally is that e.g. reports aren't actually identified just by `report_id`; their effective ID is actually `(task_id, report_id)`. So an `id()` method would return something like `(&TaskId, &ReportId)`.

Of course, that doesn't apply to tasks (since there is no higher-level entity tasks are conceptually "inside of"), and the stuttering bugs me too. Since you noticed it as well, I fixed it here as well as for ReportMetadata & AggregationJob. (The last ID type, BatchID, doesn't yet correspond directly to a defined data structure.)

branlwyd: I went back and forth on this -- the reason I didn't go for it originally is that e.g. reports aren't actually identified just by `report_id`; their effective ID is actually `(task_id, report_id)`. So an `id()` method would return something like `(&TaskId, &ReportId)`.

Of course, that doesn't apply to tasks (since there is no higher-level entity tasks are conceptually "inside of"), and the stuttering bugs me too. Since you noticed it as well, I fixed it here as well as for ReportMetadata & AggregationJob. (The last ID type, BatchID, doesn't yet correspond directly to a defined data structure.)

janus_server/src/aggregator/aggregate_share.rs
branlwyd: I removed the section reference entirely. I think it's only of marginal value given how fungible the DAP specification still is.

Maybe we can do a pass or something once per major version to update all of the section references appropriately? I don't think a process based on one-off updating section references as we touch the relevant code lines will work well -- even code that doesn't need to change to update the DAP version might well have its underlying spec sections moved around under it.

tgeoghegan: What I've done most recently is include a link to the relevant spec section instead of the section number. Then at least it's a little more obvious that a reference is out of date, because you have the draft revision in the URL. However I'm also OK with not putting in section references until DAP stabilizes.

PR #617, opened 2022-10-07
Replace vec! with Vec::{from,new}.

branlwyd:
Opinionated change: since Rust 1.44 [1], it has been possible to use Vec::from to write vector literals. On the idea that a method call is a simpler tool than a macro, prefer a method call. (Vec::new is also used, if the vector being created is empty.)

[1] https://github.com/rust-lang/rust/pull/68692/

### DIFF
--- a/janus_collector/examples/collect.rs
+++ b/janus_collector/examples/collect.rs
@@ -479,7 +479,7 @@ mod tests {
             ErrorKind::ValueValidation,
         );
 
-        let base_arguments = vec![
+        let base_arguments = Vec::from([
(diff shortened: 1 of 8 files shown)


branlwyd:
(This is a style-only change, I suspect it will have negligible effect on compile times & no effect on run times. I've been experimenting with replacing `vec!` in the last few PRs and it worked well so I cleaned up the last few instances of `vec!`.)

divergentdave (approved):

tgeoghegan (approved):

PR #618, opened 2022-10-07
Log errors using Debug rather than Display implementations.

### LINKED ISSUE #609: Cleanup: log errors using Debug rather than Display
The intent of this issue is to log (e.g. via `info!`, `warn!`, etc) errors via their `Debug` rendering (i.e. `?error`) rather than their `Display` rendering (i.e. `%error`).

An `anyhow::Error`'s `Display` rendering does not render any wrapped errors.

branlwyd:
This is necessary because anyhow error types do not render their internally-wrapped "source" errors with their Display implementations, only their Debug implementations.

While I'm examining all the log lines, also fixup a few things to use Display as needed -- this is most important for ID types, which would previously be logged as e.g. `task_id = TaskId(abcde)` instead of just `task_id = abcde`. The additional type information is not helpful to humans & stutters with the field name.

Closes #609.

### DIFF
--- a/janus_core/src/retries.rs
+++ b/janus_core/src/retries.rs
@@ -94,7 +94,7 @@ where
             }
             Err(error) => {
                 if error.is_timeout() || error.is_connect() {
-                    warn!(%error, "encountered retryable error");
+                    warn!(?error, "encountered retryable error");
                     return Err(backoff::Error::transient(Err(error)));
                 }
 
@@ -103,7 +103,7 @@ where
                     | std::io::ErrorKind::ConnectionReset
                     | std::io::ErrorKind::ConnectionAborted = io_error.kind()
                     {
-                        warn!(%error, "encountered retryable error");
+                        warn!(?error, "encountered retryable error");
                         return Err(backoff::Error::transient(Err(error)));
                     }
                 }
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -1091,7 +1091,7 @@ impl VdafOps {
                 report.public_share(),
             ),
         ) {
-            info!(report.task_id = ?report.task_id(), report.metadata = ?report.metadata(), %error, "Report decryption failed");
+            info!(report.task_id = %report.task_id(), report.metadata = ?report.metadata(), ?error, "Report decryption failed");
             upload_decrypt_failure_counter.add(&Context::current(), 1, &[]);
             return Ok(());
         }
@@ -1194,7 +1194,7 @@ impl VdafOps {
                 .get(report_share.encrypted_input_share().config_id())
                 .ok_or_else(|| {
                     info!(
-                        config_id = ?report_share.encrypted_input_share().config_id(),
+                        config_id = %report_share.encrypted_input_share().config_id(),
                         "Helper encrypted input share references unknown HPKE config ID"
                     );
                     aggregate_step_failure_counter.add(
@@ -1222,7 +1222,7 @@ impl VdafOps {
                     info!(
                         task_id = %task.id(),
                         metadata = ?report_share.metadata(),
-                        %error,
+                        ?error,
                         "Couldn't decrypt helper's report share"
                     );
                     aggregate_step_failure_counter.add(
@@ -1242,14 +1242,14 @@ impl VdafOps {
             let input_share = plaintext.and_then(|plaintext| {
                 A::InputShare::get_decoded_with_param(&(vdaf, Role::Helper.index().unwrap()), &plaintext)
                     .map_err(|error| {
-                        info!(task_id = %task.id(), metadata = ?report_share.metadata(), %error, "Couldn't decode helper's input share");
+                        info!(task_id = %task.id(), metadata = ?report_share.metadata(), ?error, "Couldn't decode helper's input share");
                         aggregate_step_failure_counter.add(&Context::current(), 1, &[KeyValue::new("type", "input_share_decode_failure")]);
                         ReportShareError::VdafPrepError
                     })
             });
 
             let public_share = A::PublicShare::get_decoded_with_param(&vdaf, report_share.public_share()).map_err(|error|{
-                info!(task_id = %task.id(), metadata = ?report_share.metadata(), %error, "Couldn't decode public share");
+                info!(task_id = %task.id(), metadata = ?report_share.metadata(), ?error, "Couldn't decode public share");
                 aggregate_step_failure_counter.add(&Context::current(), 1, &[KeyValue::new("type", "public_share_decode_failure")]);
                 ReportShareError::VdafPrepError
             });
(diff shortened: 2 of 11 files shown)


divergentdave:
Could we limit the change to use `Debug` on errors to only those cases where the error is of type `anyhow::Error`? The first change in the patch, for example, handles a `reqwest::Error`. Generally, most other errors 

Alternatively, we could wrap `anyhow::Error` in a newtype to make it get formatted using `Display` with the "alternate" flag set. Anyhow will then print causes as well, but without newlines. See https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations.

branlwyd:
> Could we limit the change to use `Debug` on errors to only those cases where the error is of type `anyhow::Error`? The first change in the patch, for example, handles a `reqwest::Error`. Generally, most other errors are less readable with `Debug` than `Display`.
> 
> Alternatively, we could wrap `anyhow::Error` in a newtype to make it get formatted using `Display` with the "alternate" flag set. Anyhow will then print causes as well, but without newlines. See https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations.
> 
> Edit: Here's the code for such a newtype
> 
> ```rust
> struct AlternateDisplay<T>(T);
> 
> impl<T: std::fmt::Display> std::fmt::Display for AlternateDisplay<T> {
>     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
>         write!(f, "{:#}", &self.0)
>     }
> }
> ```

I don't want to do this because many of our error types that aren't directly `anyhow::Error` can embed an `anyhow::Error` (typically as a value in an enum variant). I also don't want to introduce a newtype just to get a better formatting result, since then we have our own "special" anyhow error type that is like a normal anyhow error except in this one important way, surprising to any Rust developer that doesn't yet know our codebase. It's also additional "cruft" or "glue" code that I would not like to carry around in our codebase.

Instead, I think if there are any places that our error's Debug renderings are unacceptable, we should fix them up. I think the Debug renderings show "full information" of what is included in the error, at least, which is what I care about. And I would generically expect a Debug implementation to include at least as much information about the value as the Display implementation.

I think this will become especially important once we improve our approach to error handling. I'm still thinking things through before I make a specific suggestion/change, but in summary:
1) I think basically every one of our error types should have an "anyhow variant" (called something like `Other` or `Generic`, perhaps), and other enum variants should be introduced only if & when our code needs to make programmatic logic decisions based on the error variant.
2) Basically everywhere that currently uses the `?` operator to return early should call `.context()` to provide context on what the failure case was.

With these changes, we'll see a large increase in both the number of `anyhow::Error` values we are handling, and the number of `anyhow::Error` values which have relevant wrapped errors that we would want to render alongside the `anyhow::Error` itself.

divergentdave:
Ok, makes sense. (FWIW, I was only thinking of using the newtype at tracing macro callsites, not for passing between functions) This gives me another idea: we could go through any thiserror attributes on error variants with `anyhow::Error` and make sure all those are formatted with "{0:#}" instead of `transparent`.

branlwyd:
I took a look at implementing your suggestion to annotate `anyhow::Error` renderings using `{:#}` rather than `{}` (which would cause `anyhow` to render the wrapped errors). I quickly discovered that none of the relevant error variants were in Janus' codebase!

There were a few in libprio (`CodecError` has one, for example). Some other errors, like `tokio_postgres::Error` & `deadpool_postgres::PoolError`, are ultimately boxed (e.g. `Box<dyn std::error::Error>`), so we don't know what the underlying types will be at runtime, nor what their expectations for rendering will be.

We also have a few boxed errors of our own: for example, `datastore::Error::User` is a wrapper around a boxed error whose type is determined at runtime; sometimes the underlying error is an `anyhow::Error`. But without discriminating the underlying type while rendering at runtime, we don't know if we should use `Display` or `Debug`.

For the reasons above, I think it's pretty unreasonable to try to enforce using one method of error rendering for most errors and a different method of error rendering for one specific kind of error. While we might clean up the Janus codebase, and any libraries we maintain like libprio, this would be quite a lot of work for such a small result; and cleaning up libraries we don't maintain would be even harder -- the maintainers may decline this change.

Thus, I suggest we use `Debug` for error logging. This is supported by the standard library documentation: [`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html) suggests only that "Errors must describe themselves through the [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) and [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) traits." without guidance as to which trait is preferable to use for which purposes. Looking to each of these traits, [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) is "for user-facing output" while [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html) is for "output in a programmer-facing, debugging context". Given that the context of this PR is about rendering errors for logging, `Debug` is IMO more appropriate.

WDYT?

(All that said, I think this is a problem without a good solution, at least from our position of "application/library maintainer" -- I think this is ultimately a problem best solved by the Rust maintainers, perhaps by providing guidance as to when to render errors using `Display` & when to render them using `Debug`. Since there is no such guidance, different libraries have made different and incompatible choices.

Barring a change at that level, I think `anyhow::Error`'s `Display` implementation should render wrapped errors -- it's hard for me not to take the lack of rendering of wrapped errors as a design mistake, since half of the point of `anyhow` is that you can wrap arbitrary other errors.)

tgeoghegan:
I agree with @branlwyd, we should log with the `Debug` impl so that we avoid losing valuable information.

The other nuance I can see between `Display` and `Debug` is that we might expect `Display` output to be stable and hence parseable, whereas `Debug` output implicitly can change at any time. But even that distinction is irrelevant, I think, because if we want to examine errors to see what went wrong and have some code do one thing or another, we should be matching on enum variants or using methods exposed by an error type, and never (if we can help it) looking for substrings in an error string.

We should be careful about how the error output gets rendered, though. In cloud environments, we output JSON log/trace events so they can be ingested and indexed by Google Cloud Logging. An error's `Debug` output could have multiple lines in it, so we should make sure to escape it as needed so that it can safely be put into a JSON string. Otherwise I'm worried we could end up with an event like this:

```
{
  "foo": "bar",
  "baz": "qux",
  "description": "first error, caused by:
second error, caused by:
third error, no cause",
  "more json key": { /* some more JSON in here */ }
}
```

I worry that the `\r\n` inside of the value of `description` might cause Google Cloud Logging to treat this as multiple events that don't parse as valid JSON.

divergentdave (approved):
OK, that makes sense. I agree that I'd prefer anyhow always print causes (or flip the semantics of the "alternate" flag) but this is a sensible compromise barring that.

My understanding is that the output of both Debug and Display shouldn't be treated as stable. (see, for example, https://github.com/serde-rs/json/releases/tag/v1.0.85)

Escaping newlines and quote characters won't be an issue, as tracing-subscriber's JSON format uses `serde_json` under the hood. All this output will be properly encapsulated in a string literal.

branlwyd:
> My understanding is that the output of both Debug and Display shouldn't be treated as stable. (see, for example, https://github.com/serde-rs/json/releases/tag/v1.0.85)

Yup, agreed -- let's do our very best to never discriminate error types based on a string rendering of the error.

PR #619, opened 2022-10-07
Don't import *.

branlwyd:
The reasons for this are as described in
https://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad (which is Java-specific, but the Rust mechanism is very similar & will have similar issues).


branlwyd:
(This is another minor cleanup I noticed while doing recent cleanup PRs. It took me <10 minutes to generate this PR; given that, I think not allowing `*` imports is reasonable as they don't actually save all that much time vs the level of headache they might inflict.)

tgeoghegan (approved):

PR #620, opened 2022-10-10
Apply input validation in TaskBuilder::build()

divergentdave:
Currently, `TaskBuilder` can violate some of `Task`'s invariants, because it modifies its fields directly, and bypasses `Task::new()`. This round-trips the task through the constructor when `build()` is called, so that we still get the same input validation. This will help keep tests closer in line with production operation.


branlwyd (changes requested):
Hooray for invariants :-)

I left things this way because `TaskBuilder` was test-only (indeed, I justified not maintaining invariants as OK due to `TaskBuilder` being test-only), but I definitely don't object to improving the realism of our tests.

branlwyd (approved):

PR #638, opened 2022-10-11
Unify problem details handling, log parse errors

divergentdave:
This is a follow-up to #614, applying the same error logging in the aggregator and client. The relevant helper method is moved from `janus_collector` to `janus_core`.

### DIFF
--- a/janus_core/src/http.rs
+++ b/janus_core/src/http.rs
@@ -4,7 +4,7 @@ use tracing::warn;
 
 /// Turn a [`reqwest::Response`] into a [`HttpApiProblem`]. If applicable, a JSON problem details
 /// document is parsed from the request's body, otherwise it is solely constructed from the
-/// response's status code.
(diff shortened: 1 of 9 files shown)


janus_core/src/http.rs
```diff
@@ -0,0 +1,22 @@
+use http_api_problem::{HttpApiProblem, PROBLEM_JSON_MEDIA_TYPE};
+use reqwest::{header::CONTENT_TYPE, Response};
+use tracing::warn;
+
+/// Turn a [`reqwest::Response`] into a [`HttpApiProblem`]. If applicable, a JSON problem details
+/// document is parsed from the request's body, otherwise it is solely constructed from the
```
tgeoghegan, line 6: nit: include an RFC 7807 reference in the doccomment.

tgeoghegan (approved):

branlwyd (approved):

Issue #610, opened 2022-10-06
Implement `task_expiration`-based task expiration


branlwyd:
`task_expiration` is a new task parameter introduced in DAP-02. In short, it is a time after which the task is considered "invalid" and no further reports should be accepted. We need to implement logic around `task_expiration` in Janus.

tgeoghegan:
I'm not certain if we should do anything here beyond #639. The other things we could do here are to:

- have the helper refuse to service aggregate jobs that reference expired tasks (per the MAY recommendation in [DAP-02 4.4.1.4](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.4.1.4-3.4), 
- have the helper refuse to service aggregate share requests that reference expired tasks,
- have the leader refuse to service collect requests that reference expired tasks.

I'm of the opinion that we shouldn't do any of these. If a report arrived before task expiry, I don't see the harm in Janus agreeing to aggregate it. Any data (reports, prep state, aggregate shares) relevant to an expired task should be a candidate for garbage collection, but that is tracked by #313, since GC is a bigger issue than just task expiration. If a Janus helper is ever asked to aggregate a report that it has garbage collected, it can fail with `report_dropped` as required by [DAP-02](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.4.1.4-3.5).

So I argue that we should do no further work here and close this issue. I should disclose a bias here though: I'm of the opinion that task expiration should be removed from DAP altogether (not forbidden for implementations but I don't think DAP needs to spell it out), and for that reason I'm reluctant to write code to support it.

tgeoghegan:
Closing per the argument above.

PR #639, opened 2022-10-12
Check for task expiration during /upload

divergentdave:
This updates the `/upload` handler to abort if the report's timestamp is past the task's `task_expiration` parameter. This is part of #610.

### DIFF
--- a/janus_server/src/aggregator.rs
+++ b/janus_server/src/aggregator.rs
@@ -1082,9 +1082,7 @@ impl VdafOps {
 
         // Reject reports after a task has expired.
         // https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.3.2
-        if report.metadata().time().is_after(task.task_expiration())
-            || clock.now().is_after(task.task_expiration())
-        {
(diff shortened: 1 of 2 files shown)


tgeoghegan (approved):

janus_server/src/aggregator.rs
```diff
@@ -1077,6 +1080,18 @@ impl VdafOps {
             ));
         }
 
+        // Reject reports after a task has expired.
+        // https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.3.2
+        if report.metadata().time().is_after(task.task_expiration())
+            || clock.now().is_after(task.task_expiration())
```
branlwyd, line 1086: I think we should probably only check the client timestamp here, not the server timestamp.

Otherwise, delayed reports might be arbitrarily denied, even if Collection of the given batch hasn't started yet & we have not yet deleted the task.

(N.B. we might choose to eventually delete all data related to the task, at which point neither upload nor collection would expect to succeed -- at that point we'd need some way to represent the task as "deleted", and that's what we'd check here.)

branlwyd (changes requested):

PR #642, opened 2022-10-12
Update directory layout for 0.2.

branlwyd:
Specifically, directory names change in the following way:
 * `janus_server` becomes `aggregator`.
 * All other directories with a `janus_` prefix drop this prefix -- this stutters with the name of the repository.

A few notes:
 * The `janus_server` package was renamed to `janus_aggregator`. I think this is a safe change as this package is not published.
 * Several other directories already did not include a `janus_` prefix. These packages have been renamed to include a `janus_` prefix. I think this is a safe change as all affected packages are unpublished. With this change, all packages contained in this repository are named with a `janus_` prefix.
 * In all other cases, the package name does not change. So e.g. while `janus_core` now lives in the `core` directory, the package name is still `janus_core`.
 * I updated CI to rename the `janus_server` image to `janus_aggregator`. I think this is a more accurate name, but the renaming will incur some work in our operational configuration. I can roll back this part of the PR if desirable.


branlwyd:
(if other PRs will conflict, I'm happy to let them merge first & eat the merge conflict myself -- it should be easier to resolve in this PR than in others)

divergentdave (approved):

Issue #655, opened 2022-10-19
Update hpke/hpke-dispatch to fix version resolution of zeroize


divergentdave:
`hpke 0.10.0` updated the version of `x25519-dalek` it uses, which in turn gets rid of the over-specific constraint on `zeroize`. `hpke-dispatch 0.5.0` includes a dependency update to pull this in. Our latest Dependabot run failed to create a PR for this, with the following log messages.

```
updater | INFO <job_485312958> Checking if hpke-dispatch 0.4.0 needs updating
updater | INFO <job_485312958> Latest version is 0.5.0
updater | INFO <job_485312958> Requirements to unlock update_not_possible
updater | INFO <job_485312958> Requirements update strategy bump_versions
updater | INFO <job_485312958> No update possible for hpke-dispatch 0.4.0
```

Merely updating our dependency's version doesn't work, as Daphne also has an `hpke` dependency, and this results in the same "error: failed to select a version for `zeroize`." that we are trying to avoid. (although `daphne` is an optional dependency, resolving dependency versions considers all possible dependencies, so it doesn't matter that the relevant feature isn't enabled) I think we'll need to break the `--features=daphne` build for a time by removing the `daphne` dependency entirely, in order to get out of this jam. (Daphne also has a git dependency on Janus, so they would run into the same issue if they moved to update `hpke` first)

Once we get this resolved, we should be able to update our Kubernetes-related dependencies as well.

Issue #493, opened 2022-09-12
Re-enable Daphne integration tests once DAP-02 implementation is complete


branlwyd:
This is done by re-enabling the `daphne` feature on the `integration_tests` package. Reference #491 as the PR which disabled these tests.

This issue should also track any other incidental technical work that needs to occur to make the integration test work with the latest version of Daphne, once we try re-enabling it.

divergentdave:
This is blocked on cloudflare/daphne#100.

divergentdave:
#656 removed our Daphne dependency and the `daphne` feature from `integration_tests/Cargo.toml` in order to unstick our dependency resolution. Before we re-enable this integration test, we'll need to re-add the feature and dependencies, and I think Daphne will need to update to `hpke 0.10.0` as well (I'll open an issue regarding this).

tgeoghegan:
I think we have long since re-enabled Daphne tests for DAP-02, and then turned them back off for DAP-04, and we'll track further integration testing elsewhere.

PR #656, opened 2022-10-19
Zeroize-related upgrades

divergentdave:
This implements the plan in #655, temporarily removing our Daphne dependency in order to upgrade `hpke-dispatch`, `hpke`, `kube`, and `k8s-openapi`. I removed the entire `daphne` Cargo feature from `integration_tests/Cargo.toml` so that our `--all-features` builds wouldn't break. Aside from the tests here, I also confirmed that the `janus_janus` integration test still passes when run in `JANUS_E2E` mode.

The `kube` upgrades included some interesting changes: many of its dependencies have been moved behind off-by-default features, so we may see a reduction in compile time. There is now native support for port forwarding, so we could possibly use that rather than shelling out to `kubectl`.

### DIFF
--- a/integration_tests/README.md
+++ b/integration_tests/README.md
@@ -2,6 +2,9 @@
 
 ## Daphne
 
+Note: The Daphne integration test has been temporarily disabled to accommodate implementation of
+DAP-02 and upgrading [`hpke`](https://crates.io/crates/hpke) to version 0.10.0.
+
 The [Daphne](https://github.com/cloudflare/daphne) testing functionality is implemented by running a
 compiled version of Daphne inside a container. The test container is built by this package's build
 script based on the test Dockerfile in the Daphne repository; the test container is included in
--- a/integration_tests/README.md
+++ b/integration_tests/README.md
@@ -2,8 +2,11 @@
 
 ## Daphne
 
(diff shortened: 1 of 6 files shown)


tgeoghegan (approved):
Is it possible to hoist the `k8s-openapi` and `kube` dependencies into the top-level `Cargo.toml`'s `workspace.dependencies`? That would make it easier to keep the versions in sync. But maybe you can't do that since not all the crates in the workspace use those two dependencies?

divergentdave:
That works, as each crate that uses the dependency pulls it in with `dependencies.name.workspace = true`. Dependabot also recently added support for workspace-inherited dependencies, so we should be OK there.

integration_tests/README.md
```diff
@@ -2,6 +2,9 @@
 
 ## Daphne
 
+Note: The Daphne integration test has been temporarily disabled to accommodate implementation of
+DAP-02 and upgrading [`hpke`](https://crates.io/crates/hpke) to version 0.10.0.
```
branlwyd, line 6: optional: reference https://github.com/divviup/janus/issues/493 here, and maybe add a comment to that issue noting it has been even further disabled to allow dependency updates?

branlwyd (approved):

PR #657, opened 2022-10-21
Bump janus crates to 0.1.18

divergentdave:


tgeoghegan (approved):
This change is fine, but I think we should take the opportunity to adopt the `version` key in  `[workspace.package]` in the workspace `Cargo.toml` (https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) to make these changes easier in the future.

divergentdave:
Workspace inheritance is only supported in toolchains from 1.64 onward, while this branch has a MSRV of 1.63. On the main branch, we have already made use of workspace inheritance for the version field.

Issue #468, opened 2022-09-06
Implement fixed-size aggregations


tgeoghegan:
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/273
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/297
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/308
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/313

PR #658, opened 2022-10-24
Fixed-size task: implement aggregation job creation.

branlwyd:
Part of #468.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_creator.rs
+++ b/aggregator/src/aggregator/aggregation_job_creator.rs
@@ -52,6 +52,7 @@ mod private {
     impl Sealed for janus_core::test_util::dummy_vdaf::Vdaf {}
 }
 
+// TODO(#680): add metrics to aggregation job creator.
 pub struct AggregationJobCreator<C: Clock> {
     // Dependencies.
     datastore: Datastore<C>,
@@ -247,13 +248,13 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
             }
 
             (task::QueryType::FixedSize{max_batch_size}, VdafInstance::Real(janus_core::task::VdafInstance::Prio3Aes128Count)) => {
-                let max_batch_size = usize::try_from(*max_batch_size)?;
+                let max_batch_size = *max_batch_size;
                 self.create_aggregation_jobs_for_fixed_size_task_no_param::<PRIO3_AES128_VERIFY_KEY_LENGTH, Prio3Aes128Count>(task, max_batch_size)
                     .await
             }
 
             (task::QueryType::FixedSize{max_batch_size}, VdafInstance::Real(janus_core::task::VdafInstance::Prio3Aes128CountVec { .. })) => {
-                let max_batch_size = usize::try_from(*max_batch_size)?;
+                let max_batch_size = *max_batch_size;
                 self.create_aggregation_jobs_for_fixed_size_task_no_param::<
                     PRIO3_AES128_VERIFY_KEY_LENGTH,
                     Prio3Aes128CountVecMultithreaded
@@ -261,13 +262,13 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
             }
 
             (task::QueryType::FixedSize{max_batch_size}, VdafInstance::Real(janus_core::task::VdafInstance::Prio3Aes128Sum { .. })) => {
-                let max_batch_size = usize::try_from(*max_batch_size)?;
+                let max_batch_size = *max_batch_size;
                 self.create_aggregation_jobs_for_fixed_size_task_no_param::<PRIO3_AES128_VERIFY_KEY_LENGTH, Prio3Aes128Sum>(task, max_batch_size)
                     .await
             }
 
             (task::QueryType::FixedSize{max_batch_size}, VdafInstance::Real(janus_core::task::VdafInstance::Prio3Aes128Histogram { .. })) => {
-                let max_batch_size = usize::try_from(*max_batch_size)?;
+                let max_batch_size = *max_batch_size;
                 self.create_aggregation_jobs_for_fixed_size_task_no_param::<PRIO3_AES128_VERIFY_KEY_LENGTH, Prio3Aes128Histogram>(task, max_batch_size)
                     .await
             }
@@ -387,7 +388,7 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
     >(
         self: Arc<Self>,
         task: Arc<Task>,
-        max_batch_size: usize,
+        task_max_batch_size: u64,
     ) -> anyhow::Result<()>
     where
         for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
@@ -396,13 +397,15 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
         A::OutputShare: Send + Sync,
         for<'a> &'a A::OutputShare: Into<Vec<u8>>,
     {
+        let (task_min_batch_size, task_max_batch_size) = (
+            usize::try_from(task.min_batch_size())?,
+            usize::try_from(task_max_batch_size)?,
+        );
         Ok(self
             .datastore
             .run_tx(|tx| {
                 let (this, task) = (Arc::clone(&self), Arc::clone(&task));
                 Box::pin(async move {
-                    let min_batch_size = usize::try_from(task.min_batch_size())?;
-
                     // Find unaggregated client reports & existing unfilled batches.
                     let (mut unaggregated_report_ids, outstanding_batches) = try_join!(
                         tx.get_unaggregated_client_report_ids_for_task(task.id()),
@@ -410,7 +413,8 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
                     )?;
 
                     // First attempt to allocate unaggregated reports to existing unfilled batches,
-                    // then generate new batches as necessary.
+                    // then generate new batches as necessary. This iterator has no end and
+                    // therefore it is safe to unwrap the result of a call to `next`.
                     let mut batch_iter = outstanding_batches
                         .into_iter()
                         .map(|outstanding_batch| (false, outstanding_batch))
@@ -424,8 +428,8 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
                     let mut aggregation_jobs = Vec::<AggregationJob<L, FixedSize, A>>::new();
                     let mut report_aggregations = Vec::<ReportAggregation<L, A>>::new();
                     let mut new_batches = Vec::new();
-                    let (mut is_batch_new, mut batch) = batch_iter.next().unwrap();
-                    let mut max_size = batch.max_size();
+                    let (mut is_batch_new, mut batch) = batch_iter.next().unwrap(); // unwrap safety: infinite iterator
+                    let mut batch_max_size = batch.max_size();
                     loop {
                         // Figure out desired aggregation job size:
                         //  * It can't be larger than the number of reports available.
@@ -438,16 +442,15 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
                         let aggregation_job_size = [
                             unaggregated_report_ids.len(),
                             this.max_aggregation_job_size,
-                            max_batch_size - max_size,
+                            task_max_batch_size - batch_max_size,
                         ]
                         .into_iter()
                         .min()
-                        .unwrap();
+                        .unwrap(); // unwrap safety: iterator is non-empty, so result is Some
 
                         if aggregation_job_size < this.min_aggregation_job_size {
-                            if !is_batch_new
-                                && max_size < min_batch_size
-                                && max_size + aggregation_job_size >= min_batch_size
+                            if batch_max_size < task_min_batch_size
+                                && batch_max_size + aggregation_job_size >= task_min_batch_size
                             {
                                 // This batch is short of the minimum batch size, and requires an
                                 // unusually small aggregation job (smaller than the normal minimum
@@ -459,8 +462,8 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
                             } else if !is_batch_new {
                                 // Move on to the next unfilled batch to see if we can allocate
                                 // reports to it.
-                                (is_batch_new, batch) = batch_iter.next().unwrap();
-                                max_size = batch.max_size();
+                                (is_batch_new, batch) = batch_iter.next().unwrap(); // unwrap safety: infinite iterator
+                                batch_max_size = batch.max_size();
                                 continue;
                             } else {
                                 // We have run out of preexisting batches to evaluate adding reports
@@ -502,12 +505,12 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
                                     )
                                 }),
                         );
+
                         if is_batch_new {
                             new_batches.push(*batch.id())
                         }
-
                         is_batch_new = false;
-                        max_size += aggregation_job_size;
+                        batch_max_size += aggregation_job_size;
                     }
 
                     // Write the outstanding batches, aggregation jobs, & report aggregations we
@@ -1095,7 +1098,7 @@ mod tests {
             Box::pin(async move {
                 tx.put_task(&task).await?;
                 for report in &reports {
-                    tx.put_client_report(&report).await?;
+                    tx.put_client_report(report).await?;
                 }
                 Ok(())
             })
--- a/aggregator/src/aggregator/aggregation_job_creator.rs
+++ b/aggregator/src/aggregator/aggregation_job_creator.rs
@@ -2,14 +2,17 @@ use crate::{
     datastore::models::{
         AggregationJob, AggregationJobState, ReportAggregation, ReportAggregationState,
     },
-    datastore::{self, Datastore},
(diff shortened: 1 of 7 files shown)


aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -347,6 +380,164 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
...
+        max_batch_size: usize,
+    ) -> anyhow::Result<()>
+    where
+        for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
+        A::PrepareMessage: Send + Sync,
+        A::PrepareState: Send + Sync + Encode,
+        A::OutputShare: Send + Sync,
+        for<'a> &'a A::OutputShare: Into<Vec<u8>>,
+    {
+        Ok(self
```
tgeoghegan, line 399: nit: if you did
```
let ret = self.datastore.run_tx(...);
Ok(ret)
```
...then you could de-indent the bulk of the method by one tab, which maybe gives `cargo fmt` a little more breathing room to lay out the code, but it's no big deal.

aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -347,6 +380,164 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
...
+                        //    aggregating more than max_batch_size reports together.
+                        // Choose the maximal size meeting all of these requirements.
+                        let aggregation_job_size = [
+                            unaggregated_report_ids.len(),
+                            this.max_aggregation_job_size,
+                            max_batch_size - max_size,
+                        ]
+                        .into_iter()
+                        .min()
+                        .unwrap();
```
tgeoghegan, line 445: It'd be cool if there was some kind of `InfiniteIterator` trait for iterators that are guaranteed to always yield a value so that one could iterate without all the `unwrap()`s... but that's not a review comment on this change!

aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -347,6 +380,164 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
...
+
+                        // Generate an aggregation job, then update batch metadata & continue.
+                        let aggregation_job_id = random();
+                        debug!(
+                            task_id = %task.id(),
+                            batch_id = %batch.id(),
+                            %aggregation_job_id,
+                            report_count = aggregation_job_size,
+                            "Creating aggregation job"
+                        );
```
tgeoghegan, line 482: Have we plumbed a metrics collector into this struct? If so, we should bump a counter for aggregation job creation. If not, let's put a TODO comment and file an issue about complementing `debug!` and `info!` prints with metrics.

aggregator/src/datastore.rs
```diff
@@ -3619,6 +3726,56 @@ pub mod models {
     {
     }
 
+    /// An outstanding batch, which is a batch which has not yet started collection.
```
tgeoghegan, line 3729: Besides not yet having been collected, an `OutstandingBatch` can also still have reports assigned to it, right?

aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -347,6 +380,164 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
...
+                        }));
+
+                    // Main loop: repeatedly consume some of the unaggregated report IDs to generate
+                    // an aggregation job, assigning it to an existing batch which has need of
+                    // reports, or a new batch if no existing batch needs reports.
+                    let mut aggregation_jobs = Vec::<AggregationJob<L, FixedSize, A>>::new();
+                    let mut report_aggregations = Vec::<ReportAggregation<L, A>>::new();
+                    let mut new_batches = Vec::new();
+                    let (mut is_batch_new, mut batch) = batch_iter.next().unwrap();
+                    let mut max_size = batch.max_size();
```
tgeoghegan, line 428: It's a bit confusing resolving the notion of an `OustandingBatch`'s `max_size` against the task's `max_size`. IIUC `OutstandingBatch.max_size` is the biggest batch that could be constructed based on the reports currently assigned to the batch, but in fact the batch could still get bigger if we end up assigning more reports, right? Maybe `OutstandingBatch.potential_size` or (perhaps too wordy) `OutstandingBatch.current_potential_size` is more clear?

tgeoghegan (changes requested):
Looks great, just a handful of suggestions that I don't think are critical.

divergentdave (commented):
Thinking ahead to database performance implications, I think `read_batch_size()`will take out a lot of predicate locks on `report_aggregations` as it touches random pages during index scans. We don't need to address this now, as this is part of a larger known issue. We may want to denormalize the number of report aggregations in different states in an aggregation job into the `aggregation_jobs` table, to trade off a few writes for many more reads.

aggregator/src/datastore.rs
```diff
@@ -2342,6 +2344,111 @@ ORDER BY id DESC
...
+            .prepare_cached(
+                "WITH batch_report_aggregation_statuses AS
+                    (SELECT report_aggregations.state, COUNT(*) AS count FROM report_aggregations
+                     JOIN aggregation_jobs ON report_aggregations.aggregation_job_id = aggregation_jobs.id
+                     WHERE aggregation_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                     AND aggregation_jobs.batch_identifier = $2
+                     GROUP BY report_aggregations.state)
+                SELECT
+                    (SELECT SUM(count)::BIGINT FROM batch_report_aggregation_statuses
+                     WHERE state IN ('FINISHED')) AS min_size,
```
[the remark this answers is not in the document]
branlwyd: Not really -- I was just matching structure with the following few lines. (I would be pretty surprised if the query plan was any different than using `==`.)

aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -23,11 +26,12 @@ use prio::{
     },
 };
 use rand::{random, thread_rng, Rng};
-use std::{collections::HashMap, convert::Infallible, sync::Arc, time::Duration};
+use std::{collections::HashMap, convert::Infallible, iter, sync::Arc, time::Duration};
 use tokio::{
     select,
     sync::oneshot::{self, Receiver, Sender},
     time::{self, Instant, MissedTickBehavior},
+    try_join,
```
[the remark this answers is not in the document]
branlwyd: I can't think of a reason doing so would cause problems, but might as well be consistent -- I mostly-arbitrarily settled on using `tokio::try_join` throughout.

aggregator/src/aggregator/aggregation_job_creator.rs
branlwyd: I tried this out, but unfortunately `rustfmt` didn't actually de-indent anything -- all the lines after the edited line are indented equivalently.

(Since the `Ok` result will always be `()`, I tried returning `Ok(())` instead of assigning to `ret`, but unfortunately the results were the same.)

aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -347,6 +380,164 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
...
+        A::PrepareState: Send + Sync + Encode,
+        A::OutputShare: Send + Sync,
+        for<'a> &'a A::OutputShare: Into<Vec<u8>>,
+    {
+        Ok(self
+            .datastore
+            .run_tx(|tx| {
+                let (this, task) = (Arc::clone(&self), Arc::clone(&task));
+                Box::pin(async move {
+                    let min_batch_size = usize::try_from(task.min_batch_size())?;
```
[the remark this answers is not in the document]
branlwyd: Sure, makes sense. I also pedantically moved the conversions out of the transaction, on the idea that if there is a conversion problem, we're better served noticing it before we open a transaction than after. (pedantic because, on 64-bit archs, I'm pretty sure `u64` to `usize` conversion is infallible)

aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -347,6 +380,164 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
...
+                            (true, OutstandingBatch::new(*task.id(), random(), 0, 0))
+                        }));
+
+                    // Main loop: repeatedly consume some of the unaggregated report IDs to generate
+                    // an aggregation job, assigning it to an existing batch which has need of
+                    // reports, or a new batch if no existing batch needs reports.
+                    let mut aggregation_jobs = Vec::<AggregationJob<L, FixedSize, A>>::new();
+                    let mut report_aggregations = Vec::<ReportAggregation<L, A>>::new();
+                    let mut new_batches = Vec::new();
+                    let (mut is_batch_new, mut batch) = batch_iter.next().unwrap();
```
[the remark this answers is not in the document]
branlwyd: Added comments to both this & the other `unwrap` you mentioned. (I'm in favor of commenting all `unwrap`s as to why they are safe, in the cases they can't be avoided entirely.)

aggregator/src/aggregator/aggregation_job_creator.rs
branlwyd: 👍🏻

aggregator/src/aggregator/aggregation_job_creator.rs
branlwyd: Filed https://github.com/divviup/janus/issues/680.

aggregator/src/datastore.rs
branlwyd: Correct. It also only applies to `fixed-size` tasks. I noted both of these in the doc comment.

aggregator/src/aggregator/aggregation_job_creator.rs
branlwyd: Correct: `min_size` & `max_size` track the potential size of the batch, depending on how currently-incomplete aggregations resolve. Specifically, `min_size` tracks how many reports are included in the batch if all reports currently undergoing aggregation eventually fail (i.e. it is the number of report aggregations in state `FINISHED`). `max_size` tracks how many reports are included in the batch if all reports currently undergoing aggregation eventually succeed (i.e. it is the number of report aggregations in any non-terminal-failure state).

Let's chat about the names:
* I like not including `current` -- the reasoning here is that every model is the "current" state of the underlying data [at some point in time, I suppose], so that word is basically implied for every field, or at least every field that is effectively mutable over time. That is, by the same logic we would justify calling the `state` field contained by many data models `current_state` instead.
* `potential_size` has a couple of issues that I think need to be resolved to get to a good naming scheme:
  * An fair (IMO) interpretation is that every size between `min_size` and `max_size` (inclusive) is a "potential" size of the batch. We could call this `max_potential_size` but then we're kind of back to the original problem, I guess.
  * If we interpret the max size as the "potential" size, we still need a good name for the minimum size.
  * I considered just making a single field `potential_size: Range<usize>`, tracking the entire range of possible batch sizes. Unfortunately, Rust's `Range` type is exclusive, which would make the logic using this field more complex/error-prone (i.e. off-by-one errors) than otherwise necessary. Still, I can go ahead w/ this strategy if you think it wise.

IMO, `min_size` & `max_size` are OK since they are attached to an `OutstandingBatch` object. I did go ahead and update the variable naming in `aggregation_job_creator.rs` to attempt to ensure clarity between a task's max_batch_size & a batch's size.

divergentdave: There is std::ops::RangeInclusive, FWIW

aggregator/src/aggregator/aggregation_job_creator.rs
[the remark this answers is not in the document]
branlwyd: Agh, unfortunately this makes the borrow-checker unhappy(!). Specifically, in the match arms, `max_batch_size` ends up being a borrow from `task`. Then trying to pass both `task` & `max_batch_size` (or `*max_batch_size`) causes the compiler to complain about task being used while borrowed. I think I could fix this by passing a `&Task` rather than an `Arc<Task>`, but the whole point of using `Arc` is that the `create_aggregation_jobs_for_fixed_size_task_no_param` needs one -- I would ultimately have to clone the `Arc`. I'd like to avoid incurring even likely-negligible runtime cost to avoid one line, so I moved the max_batch_size copy back out to `create_aggregation_jobs_for_task`.

aggregator/src/datastore.rs
```diff
@@ -3619,6 +3726,56 @@ pub mod models {
     {
     }
 
+    /// An outstanding batch, which is a batch which has not yet started collection.
+    #[derive(Clone, Debug, PartialEq, Eq)]
+    pub struct OutstandingBatch {
+        /// The task ID for this outstanding task.
+        task_id: TaskId,
+        /// The batch ID for this outstanding task.
```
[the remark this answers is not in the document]
branlwyd: 👍🏻 thanks, nice spot.

aggregator/src/datastore.rs
```diff
@@ -3619,6 +3726,56 @@ pub mod models {
...
+            &self.batch_id
+        }
+
+        /// Gets the minimum possible size of this batch with current aggregation jobs (i.e. the
+        /// count of reports which have successfully completed the aggregation process).
+        pub fn min_size(&self) -> usize {
+            self.min_size
+        }
+
+        /// Gets the maximum possible size of this batch with current aggregatino jobs (i.e. the
```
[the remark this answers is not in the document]
branlwyd: 👍🏻

aggregator/src/datastore.rs
```diff
@@ -7089,6 +7219,154 @@ mod tests {
...
+        let (ds, _db_handle) = ephemeral_datastore(clock.clone()).await;
+
+        let (task_id, batch_id) = ds
+            .run_tx(|tx| {
+                let clock = clock.clone();
+                Box::pin(async move {
+                    let task = TaskBuilder::new(
+                        QueryType::FixedSize { max_batch_size: 10 },
+                        VdafInstance::Fake,
+                        Role::Helper,
```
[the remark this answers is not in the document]
branlwyd: 👍🏻

branlwyd:
> Thinking ahead to database performance implications, I think `read_batch_size()`will take out a lot of predicate locks on `report_aggregations` as it touches random pages during index scans. We don't need to address this now, as this is part of a larger known issue. We may want to denormalize the number of report aggregations in different states in an aggregation job into the `aggregation_jobs` table, to trade off a few writes for many more reads.

Agreed, though I think the fact that updates to aggregation job/report aggregations in the leader happen as "pure writes" (i.e. we have a transaction just for writing back updates to the aggregation jobs when they are stepped, reading happens in a different transaction & the VDAF computations happen outside of a transaction) will help to mitigate this -- the necessary retries will not be overly costly.

Another strategy would be to denormalize the counts into the `outstanding_batches` table (e.g. directly track `{min,max}_size`) -- this would lead to additional contention when an aggregation job is started or completes, but given the above argument about contention not being _too_ bad in this situation, we might be able to get away with it.

In any case, this is a lot of speculation on my part. :) Let's go with this approach for now and improve if we find we need to. This approach, at least, is less error-prone than trying to keep a counter updated at every point an aggregation job or its report aggregations can be updated.

aggregator/src/aggregator/aggregation_job_creator.rs
branlwyd: [I realized you're traveling, so I'm going to submit this PR without waiting to hear back on this issue, but poke this if you have an idea & I'll address in a followup PR.]

branlwyd: I don't know how I missed that, I was looking specifically for it. Thanks, `OutstandingBatch` now has a single `size` method.

branlwyd:
One more stray thought regarding performance: other than denormalization, another strategy would be to split `create_aggregation_jobs_for_fixed_size_task_no_param` into the following pieces:

* A read-only transaction which reads the required data (i.e. unaggregated report IDs, outstanding batches). [This would be another place justifying true read-only transaction support, avoiding any potential for restarts due to contention.]
* A computation of the outstanding batches/aggregation jobs/report aggregations to write, outside of a transaction.
* A write-only transaction which inserts the values computed in the previous step.

This would _sort of_ work out currently, since the current model is that there is a single aggregation job creator running. This model will likely need to evolve as we scale, but I imagine we could justify that there is a single aggregation job creator job responsible for generating aggregation jobs for a given task for quite a while.

I say it _sort of_ works out because, even though there should be only one aggregation job creator running at once, there is nothing enforcing this -- a simple operational misconfiguration causing multiple aggregation job creators to run at once (for a given task) could lead to some very strange behavior. We could mitigate this by implementing some change-detection in the write tx (bailing if it notices an unexpected write on rows it believes it is the only writer to); this may be simpler/more localized to implement than maintaining counters.

tgeoghegan (approved):
The changes LGTM.

re: database performance: I agree with moving forward with the current solution while acknowledging the potential for future changes. Could we capture the discussion of performance risks and solution strategies in an issue so it doesn't get lost when this PR is merged? It seems like we will eventually need to significantly overhaul the DB schema and datastore module, so it'd be nice to be able to quickly refer back to all our concerns and ideas when we tackle that work.

PR #679, opened 2022-10-24
provide `GITHUB_TOKEN` to `setup-protoc` action

tgeoghegan:
We saw at least one build failure ([1]) in CI that is down to the `arduino/setup-protoc` action encountering a rate limit. According to the action's README ([2]), it uses GitHub API to query releases on some `protoc` related repository. We can avoid rate limits by authenticating, which we do by passing the action's `GITHUB_TOKEN` into the `setup-protoc` action.

[1]: https://github.com/divviup/janus/actions/runs/3316231068/jobs/5477782062
[2]: https://github.com/arduino/setup-protoc#usage

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -50,6 +50,8 @@ jobs:
         override: true
     - name: Install Protoc
       uses: arduino/setup-protoc@v1
+      with:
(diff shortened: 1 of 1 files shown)


tgeoghegan:
See also https://github.com/actions/runner-images/issues/602

divergentdave (approved):

PR #681, opened 2022-10-25
Support JANUS_E2E_LOGS_PATH in interop_binaries tests

divergentdave:
This adds support for exporting container logs to the `janus_interop_binaries` tests as well.

### DIFF
--- a/integration_tests/src/janus.rs
+++ b/integration_tests/src/janus.rs
@@ -10,7 +10,8 @@ use janus_core::{
     time::RealClock,
 };
 use janus_interop_binaries::{
-    test_util::await_http_server, testcontainer::Aggregator, AggregatorAddTaskRequest,
+    log_export_path, test_util::await_http_server, testcontainer::Aggregator,
+    AggregatorAddTaskRequest,
 };
 use janus_messages::Role;
 use k8s_openapi::api::core::v1::Secret;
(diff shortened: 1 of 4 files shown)


tgeoghegan (commented):
This change accomplishes what it sets out to do, but I think we should pause and make sure of whether this is the direction we want things to move. Maintaining two container test harnesses in `integration_tests` and `interop_binaries` seems like redundant effort. Didn't we agree some time ago to migrate the tests in `interop_binaries/tests/end_to_end.rs` into `integration_tests`?

branlwyd (approved):
> This change accomplishes what it sets out to do, but I think we should pause and make sure of whether this is the direction we want things to move. Maintaining two container test harnesses in `integration_tests` and `interop_binaries` seems like redundant effort. Didn't we agree some time ago to migrate the tests in `interop_binaries/tests/end_to_end.rs` into `integration_tests`?

Agreed, this is what we decided previously.

My opinion would be:
* DAP behavioral tests should live in `integration_tests`.
* `interop_binaries` should have tests that, at most, focus on testing the test API implementation itself rather than DAP. This is valuable because the tests in `integration_tests` currently use only the parts of the test API they need, i.e. they are not a complete test of our test API implementation. Obviously there is some overlap here, but to give an example of what I'm thinking, e.g. I would be supportive of a change that moved the "aggregate is correct" validation into `integration_tests` (and dropped it from `interop_binaries` entirely).

That said, I'm OK with this change -- it adds minimal complexity, adds information that is plausibly useful for debugging, etc. And for the foreseeable future, we will need to maintain both `integration_tests` and `interop_binaries` tests. If anything, I'd say it would be ideal to implement a single abstraction that lets us "export logs" in both contexts, but I haven't thought through how complicated this would be -- the existing functionality is fine IMO.

tgeoghegan (approved):

PR #683, opened 2022-10-26
CI: Tag and push container images for public repo

divergentdave:
This updates the `push-docker-images-release` workflow to retag and push images to our public-facing repository. Since the two `docker login` executions will stomp on the same registry/username with different passwords, all of the public-facing pushes happen at the end, after a second `docker/login-action`.

### DIFF
--- a/.github/workflows/push-docker-images-release.yml
+++ b/.github/workflows/push-docker-images-release.yml
@@ -18,8 +18,8 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     # See https://github.com/google-github-actions/auth#authenticating-to-container-registry-and-artifact-registry
-    - id: "gcp-auth"
-      name: "Authenticate to GCP"
+    - id: "gcp-auth-private"
+      name: "Authenticate to GCP (private repositories)"
       uses: "google-github-actions/auth@v0"
       with:
         workload_identity_provider: ${{ secrets.GCP_ARTIFACT_PUBISHER_WORKLOAD_IDENTITY_PROVIDER }}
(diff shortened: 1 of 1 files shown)


branlwyd (approved):

.github/workflows/push-docker-images-release.yml
```diff
@@ -74,20 +94,48 @@ jobs:
...
+    - run: docker push us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_collect_job_driver:${{ steps.get_version.outputs.VERSION }}
+    - run: docker push us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_cli:latest
+    - run: docker push us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_cli:${{ steps.get_version.outputs.VERSION }}
+    - run: docker push us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_interop_client:latest
+    - run: docker push us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_interop_client:${{ steps.get_version.outputs.VERSION }}
+    - run: docker push us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_interop_aggregator:latest
+    - run: docker push us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_interop_aggregator:${{ steps.get_version.outputs.VERSION }}
+    - run: docker push us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_interop_collector:latest
+    - run: docker push us-west2-docker.pkg.dev/divviup-artifacts-public/janus/janus_interop_collector:${{ steps.get_version.outputs.VERSION }}
```
tgeoghegan, line 141: I think doing this with a [matrix strategy](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) would reduce some repetition and make this faster overall by running `docker push` in parallel.

```
strategy:
  matrix:
    version: latest, ${{ steps.get_version.outputs.VERSION }}
    image: janus_aggregator, janus_aggregation_job_creator, etc.
```

tgeoghegan (approved):

.github/workflows/push-docker-images-release.yml
divergentdave: My understanding is that only allows you to create multiple jobs, not run multiple steps. I think our options for simplifying are bash loops or writing a reusable actions.yml.

tgeoghegan: Probably not worth it then, unless we find that waiting for all the `docker push`es is painful.

PR #692, opened 2022-10-27
Move aggregate result test coverage to janus_integration_tests

divergentdave:
This PR expands `submit_measurements_and_verify_aggregate` to support other VDAFs, and moves our checks of aggregate result values to `janus_integration_tests`, instead of leaning on `janus_interop_binaries` tests.


integration_tests/tests/common/mod.rs
```diff
@@ -143,14 +125,145 @@ pub async fn submit_measurements_and_verify_aggregate(
...
+                &measurements,
+                &(),
+                &measurements.iter().sum(),
+            )
+            .await;
+        }
+        task::VdafInstance::Real(VdafInstance::Prio3Aes128Histogram { buckets }) => {
+            let vdaf = Prio3::new_aes128_histogram(2, buckets).unwrap();
+
+            let mut aggregate_result = vec![0; buckets.len() + 1];
```
divergentdave, line 209: Review note: I'm using `vec![]` because of the following Clippy error. (`Vec::from([0; size])` also wouldn't work here because the size is not known at compile time)

```
error: slow zero-filling initialization
   --> integration_tests/tests/common/mod.rs:210:13
    |
209 |             let mut aggregate_result = Vec::with_capacity(buckets.len() + 1);
    |                                        ------------------------------------- help: consider replace allocation with: `vec![0; buckets.len() + 1]`
210 |             aggregate_result.resize(buckets.len() + 1, 0);
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `-D clippy::slow-vector-initialization` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#slow_vector_initialization
```

tgeoghegan (approved):

branlwyd (approved):

integration_tests/tests/common/mod.rs
branlwyd: Ah, well this is a good enough reason to use `vec!` still. Hopefully the underlying issue can be fixed eventually.

PR #693, opened 2022-10-27
Unify janus_{aggregator,core}::VdafInstance.

branlwyd:
This cleanup has been possible since we started using "test-util" features to conditionally-compile test-only behavior. I trip over this a lot, and importing either type led to some awkward code, so I think it is worthy of a quick cleanup.

Note that the serialization format of VdafInstance changed, which will affect task serialization in our DB. I think this is OK since there is no expectation of backwards-compatibility between 0.1 & 0.2.


tgeoghegan (approved):
This had been bothering me for ages, thanks for cleaning it up!

divergentdave:
I haven't gamed out under what conditions this will come up in practice, but I have the following hesitation: This change would require that we align whether the janus_core/test-util feature is set and whether the downstream crate's test-util feature is set in order for exhaustive match blocks (without a `_ => {...}` to compile correctly. For example, if we have a particular Cargo build with a diamond dependency structure, with janus_core at the top, and only one of the left or right crates has their own test-util feature set, then we could get compilation errors for this reason. The bottom crate would have to turn on test-util for one of its direct dependencies for the sole purpose of aligning definitions of VdafInstance.

aggregator/src/task.rs
```diff
@@ -44,123 +44,10 @@ pub enum QueryType {
...
-    FakeFailsPrepStep,
-}
-
-impl From<janus_core::task::VdafInstance> for VdafInstance {
-    fn from(vdaf: janus_core::task::VdafInstance) -> Self {
-        VdafInstance::Real(vdaf)
-    }
-}
-
+// XXX: is this needed?
```
divergentdave, line 47: FYI, my pre-commit script flagged this

branlwyd:
> I haven't gamed out under what conditions this will come up in practice, but I have the following hesitation: This change would require that we align whether the janus_core/test-util feature is set and whether the downstream crate's test-util feature is set in order for exhaustive match blocks (without a `_ => {...}` to compile correctly. For example, if we have a particular Cargo build with a diamond dependency structure, with janus_core at the top, and only one of the left or right crates has their own test-util feature set, then we could get compilation errors for this reason. The bottom crate would have to turn on test-util for one of its direct dependencies for the sole purpose of aligning definitions of VdafInstance.

This is a good point, I added `#[non_exhaustive]` to `VdafInstance` to require match arms to include `_`. I think that should effectively mitigate the concern.

The only downside I can think of is that we can no longer introduce a new enum variant and then lean on the compiler to show errors in all of the places we need to add that variant to a match arm. But (a) we have tests and (b) failing tests, developers could temporarily remove `#[non_exhaustive]` to recover this information from the compiler.

divergentdave (approved):
`#[non_exhaustive]` sounds good to me. For what it's worth, it looks like the only serde-visible change was the struct's effective name, and I don't think `serde_json` uses that. (so it's technically not database-breaking even though that doesn't matter yet)

Issue #446, opened 2022-08-30
Interop-test containers should use the same binaries as a production deployment.


branlwyd:
Currently, we have custom aggregator, client, & collector binaries which we embed into our interop test containers.

Instead, I think we should use the same binaries we would release in a production deployment:
* This leads to less code to maintain.
* This is a much more realistic test.

Note that we don't currently have a "production" client or collector binary. (and we have a production client library, but not a collector library, yet; the collector library is tracked in #393)

edit: the interop test API itself would still be served up by its own dedicated binary in this suggestion. So e.g. the test interop aggregator container would end up running the following binaries:
* The production Janus server binary.
* The production aggregation job creator binary.
* The production aggregation job driver binary.
* The production collect job driver binary.
* A dedicated binary serving up the test API.

branlwyd:
#694 implements most of this for `janus_interop_aggregator` (one test-only binary still remains)

PR #694, opened 2022-10-28
Split janus_interop_aggregator container into more subprocesses

divergentdave:
This splits the `janus_interop_aggregator` to use a separate subprocess for each component (managed by supervisord). We now use the same binary target as production for three out of the four Janus components. I also rearranged the directory layout of files in the container a bit to clean up the root directory.


branlwyd (approved):
Very nice!

To ensure the PR can be found from the issue, this is effectively part of #446.

PR #695, opened 2022-10-28
Fixed-size tasks: implement aggregation job driver.

branlwyd:
This effectively implements the leader side of the aggregation process. The functional changes are quite small here: effectively, the only change over what was done as part of e.g. the message updates is to pass along the "partial batch identifer" (i.e. batch ID, for fixed-size tasks) in the aggregate initialization request.


branlwyd:
Part of #468.

divergentdave (approved):

PR #696, opened 2022-10-31
`/collect`: handle invalid aggregation parameter

tgeoghegan:
Sending a `CollectReq` with an aggregation parameter that isn't a valid encoding now yields an HTTP 400 Bad Request response with a problem document of type `unrecognizedMessage`, instead of the 500 Internal Server Error it previously provided.


divergentdave (approved):

tgeoghegan:
I noticed this problem while looking into #470, which I will fix in a follow-up PR.

PR #701, opened 2022-10-31
Add a length prefix to public_share in AAD

divergentdave:
This adds a length prefix in our input share AAD construction, to match Daphne's interpretation. See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/371 for discussion of related spec changes.


tgeoghegan (approved):

branlwyd (changes requested):
(I think this PR should not be merged for the reason above.)

divergentdave:
The DAP specification has been updated to make this length prefix explicitly required, and we're now targeting DAP-03 on the main branch, so I think this should be good to go now.

branlwyd (approved):

PR #709, opened 2022-11-02
Store leader input shares unencrypted

### LINKED ISSUE #442: Store uploaded reports unencrypted
Janus stores uploaded report shares in the database encrypted under either aggregator's HPKE public key. I think we should store report shares unencrypted because HPKE encryption is meant to protect messages in flight, so using it to protect data at rest feels like a mistake. It means we have to keep HPKE private keys around not just for as long as we might receive messages encrypted to that key, but as long as we might want to process stored reports encrypted to that key.

Further I think having the report shares in the clear would make it easier for us as developers and operators to investigate problems. Right now, if I want to look at an uploaded report to help someone debug, I have to go fetch the report out of the DB, then fetch the HPKE private key out of k8s secrets, then wire up some code to do the decryption. So nothing actually stops me from decrypting reports except a little extra toil, so I say: why not store the reports in plaintext?

Even if we do this, the leader still has to store the helper's share encrypted, because of course the leader doesn't have the helper's HPKE private key. However if, as we have long expected, DAP changes to mandate direct uploads to the helper and leader, then both aggregators could store unencrypted report shares.

tgeoghegan:
When handling uploads, the leader now decrypts and decodes its input share and the public share before recording them in the datastore. This allows us to discard immediately reports that can never be aggregated and simplifies the implementation of the aggregation job creator, since it can now fetch well-formed values from the datastore.

See #442 for rationale.

Resolves #442

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -16,7 +16,7 @@ use crate::{
         self,
         models::{
             AggregateShareJob, AggregationJob, AggregationJobState, CollectJob, CollectJobState,
-            ReportAggregation, ReportAggregationState,
+            LeaderStoredReport, ReportAggregation, ReportAggregationState,
         },
         Datastore,
     },
@@ -292,6 +292,8 @@ pub struct Aggregator<C: Clock> {
     // Metrics.
     /// Counter tracking the number of failed decryptions while handling the /upload endpoint.
     upload_decrypt_failure_counter: Counter<u64>,
+    /// Counter tracking the number of failed message decodes while handling the /upload endpoint.
+    upload_decode_failure_counter: Counter<u64>,
     /// Counters tracking the number of failures to step client reports through the aggregation
     /// process.
     aggregate_step_failure_counter: Counter<u64>,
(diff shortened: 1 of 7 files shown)


aggregator/src/datastore.rs
```diff
@@ -2775,10 +2839,101 @@ pub mod models {
...
+    /// Represents a report as it is stored in the leader's database, corresponding to a row in
+    /// `client_reports`, where `leader_input_share` and `helper_encrypted_input_share` are required
+    /// to be populated.
+    #[derive(Clone, Derivative)]
+    #[derivative(Debug)]
+    pub(crate) struct LeaderStoredReport<const L: usize, A>
+    where
+        A: vdaf::Aggregator<L>,
+        A::InputShare: PartialEq,
+        A::PublicShare: PartialEq,
```
divergentdave, line 2855: Should we drop these constraints from the struct, and only state them on the PartialEq/Eq implementations, and anything else that needs them?

aggregator/src/aggregator.rs
```diff
@@ -1115,35 +1126,91 @@ impl VdafOps {
...
+            match A::PublicShare::get_decoded_with_param(&vdaf, report.public_share()) {
+                Ok(public_share) => public_share,
+                Err(err) => {
+                    warn!(
+                        report.task_id = %report.task_id(),
+                        report.metadata = ?report.metadata(),
+                        ?err,
+                        "public share decoding failed",
+                    );
+                    upload_decrypt_failure_counter.add(&Context::current(), 1, &[]);
```
divergentdave, line 1144: I think we should create a separate counter metric for decoding errors, as it will be very useful to track decryption failures separate from other failure cases.

divergentdave (approved):

PR #710, opened 2022-11-03
Add integration tests with divviup-ts

divergentdave:
This refactors some components of the integration tests and adds new integration tests with divviup-ts, running it inside a Docker container managed by `testcontainers`.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -5,6 +5,11 @@ on:
     branches: [ main, release/** ]
   pull_request:
   workflow_dispatch:
+    inputs:
+      divviup_ts_interop_container:
+        description: divviup-ts container image tag for use in integration tests
+        required: false
+        type: string
 
 jobs:
   janus_build:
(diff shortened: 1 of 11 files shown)


.github/workflows/ci-build.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       CARGO_INCREMENTAL: 0
       CARGO_TERM_COLOR: always
       DAPHNE_INTEROP_CONTAINER: prebuilt=${{ secrets.DAPHNE_PREBUILT_IMAGE_NAME_AND_TAG }}
+      DIVVIUP_TS_INTEROP_CONTAINER: ${{ secrets.DIVVIUP_TS_INTEROP_IMAGE_NAME_AND_TAG }}
```
tgeoghegan, line 19: I'm not sure I understand why this is a secret. We're not trying to keep the value confidential, and we don't need to use different values in different environments, so I guess this is just so that we can change the `divviup-ts` version in use without making a change to `janus` sources?

#371 notes that the tag in this value is currently `dap-draft-02`. I definitely think we should pin a version of `divviup_ts_interop_client`. `divviup-ts`' [`docker` action](https://github.com/divviup/divviup-ts/blob/main/.github/workflows/docker.yml) runs on every commit to `main`, which means that two runs of CI using the same `janus` commit could end up using two different `divviup-ts` versions.

We could also make the `ci-build` workflow take a parameter for the image tag. So then, when run automatically on a pull request, it would use the default value, which is some pinned SHA, but when run with `workflow_dispatch`, the developer could pass in some arbitrary container image.

tgeoghegan (changes requested):

.github/workflows/ci-build.yml
divergentdave: The original reasons were to only use the janus-artifacts repository when running from GHA, with a login token derived from the GitHub Actions identity federation, and to follow what's done with the Daphne container image. We could simplify this and get rid of the secret by always using the divviup-artifacts-public repository, I don't think our request quotas should be an issue here.

tgeoghegan (approved):

PR #711, opened 2022-11-03
1.65 clippy fixes

divergentdave:
This fixes some new Clippy warnings with the latest toolchain. As discussed, the `clippy::result_large_err` lints are ignored in each place they occur.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ edition = "2021"
 homepage = "https://divviup.org"
 license = "MPL-2.0"
 repository = "https://github.com/divviup/janus"
-rust-version = "1.64.0"
+rust-version = "1.65.0"
 version = "0.2.0"
 
 [workspace.dependencies]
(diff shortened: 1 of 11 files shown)


tgeoghegan (approved):
This change is good in terms of silencing the accursed Clippy. However when we discussed in Slack, I was only thinking about `prio-server`. For `janus`, I think we should eventually make changes to the error types so they're not quite so big. I think we should take this change as-is, but let's file an issue to do something like use `Box<>` in `aggregator::Error` to clean this up properly.

tgeoghegan:
We should be more conservative with projects like `libprio-rs`, but frankly I don't see a problem with aggressively bumping MSRV on `janus`, except that it can be a hassle for consumers of the `janus_client`, `janus_collector` and `janus_messages` crates.

divergentdave:
OK, I bumped our MSRV, but we'll need to wait on the Docker library `rust` image to be updated before merging.

PR #732, opened 2022-11-09
Clean up dummy report generation

tgeoghegan:
Replaces the test-util item that generated dummy
`janus_messages::Report`s with one that generates dummy `janus_aggregator::datastore::models::LeaderStoredReport`s, which is more useful for testing code past the HTTP message handling layer (`Report` is a wire message, not the internal representation of a report). This allows us to delete some redundant methods on `Datastore` for writing reports, because we now consistently are writing `LeaderStoredReport` values.


tgeoghegan:
I did all this while working on #731 and decided to backport it since it's valuable now, and it will make #731 easier to eventually merge.

divergentdave (approved):

PR #733, opened 2022-11-09
Add missing check to only dump logs on panic

divergentdave:
This PR brings `ContainerLogsDropGuard` in line with the Janus and Daphne integration tests, by only exporting logs when the current thread is panicking. I missed this check when implementing this, so interop_binary tests and divviup-ts tests were writing extra logs out from passing tests if `JANUS_E2E_LOG_PATH` was set.

### DIFF
--- a/interop_binaries/src/lib.rs
+++ b/interop_binaries/src/lib.rs
@@ -333,39 +333,40 @@ impl<'d, I: Image> ContainerLogsDropGuard<'d, I> {
 
 impl<'d, I: Image> Drop for ContainerLogsDropGuard<'d, I> {
     fn drop(&mut self) {
-        if panicking() {
-            if let Some(base_dir) = log_export_path() {
-                create_dir_all(&base_dir).expect("could not create log output directory");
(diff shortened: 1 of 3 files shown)


interop_binaries/src/lib.rs
```diff
@@ -332,37 +333,39 @@ impl<'d, I: Image> ContainerLogsDropGuard<'d, I> {
...
-            let destination = base_dir.join(name);
-
-            let copy_status = Command::new("docker")
-                .arg("cp")
-                .arg(format!("{}:/logs", id))
-                .arg(destination)
-                .status()
-                .expect("running `docker cp` failed");
-            assert!(copy_status.success());
+        if panicking() {
```
tgeoghegan, line 336: You could do
```
if !panicking() {
    return;
}

<rest of method>
```
...to avoid indenting the whole rest of the method by one tab stop.

tgeoghegan (approved):

branlwyd (approved):

PR #734, opened 2022-11-10
Fix propagation of database-related errors

### LINKED ISSUE #730: Failure to retry upon DB serialization errors due to try_join
I investigated flaky failures in our integration tests, and I think I found a mechanism that's causing them. (I'm waiting on a second reproduction with more logging to confirm) In a few places, we use `try_join!` or `try_join_all` to wait on concurrent futures for pipelined database requests. If any future returns an error, these will take the first error they receive, cancel the other futures, and return that first error to the caller. In these situations, if we hit a serialization error in our transaction, only one future will return an error with the "could not serialize access..." message, while some of the others may return an error with "current transaction is aborted...". If the `try_join` internals poll the latter error before the first, then that's the error that will bubble up out of `run_tx_once()`, and our retry loop will fail to recognize that the transaction closure ought to be retried.

The rest of the causal chain is straightforward, I saw this happen within `Accumulator::flush_to_datastore()` in the aggregation job driver, during the second step of an aggregation job. The "transaction is aborted" database error meant that the driver failed to store the result of an `/aggregate` request it made. Due to current protocol limitations, this meant that the leader and helper became out of sync, and subsequent retries of this aggregation job failed, because the helper thought the job should be done already.

Solving this could be tricky, as there are many ways to narrow the race condition window between the first and subsequent errors, but fail to close it entirely. I think we may have to avoid try_join constructs entirely, in favor of join, because we need to inspect errors other than the first. Cancellation poses a problem because it could cause us to lose an error that arrived first over the TCP connection, but is waiting in an async channel in `tokio-postgres`. In addition, we'd need to helpers to propagate multiple errors in a `run_tx()`-friendly way, tracking of serialization errors within our `Transaction` struct instead of via returned errors, or something else to properly gather our database errors.

divergentdave:
This fixes #730 by replacing uses of `futures::future::try_join_all()` and `tokio::try_join!()`. It introduces a free function `gather_errors()` that can be used with `futures::future::join_all()` and a new `try_join!()` macro that wraps `tokio::join!()`. Both of these new code patterns will wait for all futures to complete, and select one error to propagate after seeing all outputs. If any future returns a serialization failure error, that will be returned. If some of the futures return an error indicating the transaction is already in a failed state, those are given a lower priority, because any other error will be more useful for diagnostic purposes.

I tested an earlier version of this change (focused on the aggregation job driver's critical window after talking to the helper) and I was able to run the integration tests for hours in a loop without any failures, so this could fix our last flaky test for now.

I wasn't able to write a straightforward unit test for `datastore::Error::combine()` because `tokio-postgres` doesn't provide any way to synthetically construct its errors. It would be possible to add `#[cfg(test)]`-only variants to `datastore::Error`, check for those in the predicate functions as well, and write a unit test based on that, but I'm not sure if it would be worth the complication.

Please chime in if you have any suggestions for naming improvements or code ergonomics, this could perhaps benefit from some more iteration. I tried writing a replacement `try_join_all()` at first, but I wound up with weird "action at a distance" rustc errors, likely due to a known issue with HRTBs. The `try_join!` macro could perhaps be generalized with some declarative macro tricks like passing around a list of placeholder tokens during recursion, but I decided to go with explicitly writing out cases for two to five arguments just to get things working.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -22,6 +22,7 @@ use crate::{
     },
     messages::TimeExt,
     task::{self, Task, VerifyKey, PRIO3_AES128_VERIFY_KEY_LENGTH},
+    try_join,
 };
 use bytes::Bytes;
 use http::{
@@ -74,7 +75,7 @@ use std::{
     sync::Arc,
     time::Instant,
 };
-use tokio::{sync::Mutex, try_join};
+use tokio::sync::Mutex;
 use tracing::{debug, error, info, warn};
 use url::Url;
 use uuid::Uuid;
--- a/aggregator/src/aggregator/accumulator.rs
+++ b/aggregator/src/aggregator/accumulator.rs
@@ -2,11 +2,11 @@
 
 use super::{query_type::AccumulableQueryType, Error};
 use crate::{
-    datastore::{self, models::BatchAggregation, Transaction},
+    datastore::{self, gather_errors, models::BatchAggregation, Transaction},
     task::Task,
 };
 use derivative::Derivative;
-use futures::future::try_join_all;
+use futures::future::join_all;
 use janus_core::{report_id::ReportIdChecksumExt, time::Clock};
 use janus_messages::{ReportId, ReportIdChecksum, Time};
 use prio::vdaf::{self, Aggregatable};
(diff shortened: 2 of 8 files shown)


tgeoghegan (approved):
This change LGTM, but I am not totally confident in my understanding of the problem, so I think an additional review would be wise.

branlwyd (approved):
Excellent work. 👍🏻

I thought about the problem around inconsistent reads leading to a non-serialization failure before we receive a serialization failure. Your implementation strategy fixes this for `try_join` pipelined usages, but this problem could occur even in "serial" code (i.e. one future being driven at a time). Of course, that is a problem outside the scope of this PR.

PR #737, opened 2022-11-10
Fixed-size tasks: implement leader collection.

branlwyd:
Effectively, this PR finishes out fixed-size support for the collect job driver.

Part of #468.

### DIFF
--- a/aggregator/src/aggregator/collect_job_driver.rs
+++ b/aggregator/src/aggregator/collect_job_driver.rs
@@ -1,5 +1,6 @@
 //! Implements portions of collect sub-protocol for DAP leader and helper.
 
+use super::query_type::CollectableQueryType;
 use crate::{
     aggregator::{post_to_helper, Error},
     datastore::{
@@ -8,7 +9,7 @@ use crate::{
         models::{BatchAggregation, CollectJobState, Lease},
         Datastore, Transaction,
     },
-    task::{Task, PRIO3_AES128_VERIFY_KEY_LENGTH},
+    task::{self, Task, PRIO3_AES128_VERIFY_KEY_LENGTH},
 };
 use derivative::Derivative;
 use futures::future::BoxFuture;
@@ -16,7 +17,7 @@ use futures::future::BoxFuture;
 use janus_core::test_util::dummy_vdaf;
 use janus_core::{report_id::ReportIdChecksumExt, task::VdafInstance, time::Clock};
 use janus_messages::{
-    query_type::{QueryType, TimeInterval},
+    query_type::{FixedSize, QueryType, TimeInterval},
     AggregateShareReq, AggregateShareResp, BatchSelector, Duration, Interval, ReportIdChecksum,
     Role,
 };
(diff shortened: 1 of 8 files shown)


aggregator/src/aggregator/collect_job_driver.rs
```diff
@@ -1,5 +1,6 @@
 //! Implements portions of collect sub-protocol for DAP leader and helper.
```
tgeoghegan, line 1: I don't agree with this file rename because several items from this module are used in the helper's `/aggregate_share` implementation, so it's not exclusive to the leader's `collect_job_driver`. However I concede that the existing name `aggregate_share.rs` is misleading in the other direction, since that resource is exclusive to the DAP helper. I don't have a better idea for naming the file right now, so I'm OK going ahead with this.

db/schema.sql
```diff
@@ -170,7 +170,8 @@ CREATE TABLE collect_jobs(
     id                      BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,  -- artificial ID, internal-only
     collect_job_id          UUID NOT NULL,               -- UUID used by collector to refer to this job
     task_id                 BIGINT NOT NULL,             -- the task ID being collected
-    batch_interval          TSRANGE NOT NULL,            -- the batch interval, as a range of timestamps
+    batch_identifier        BYTEA NOT NULL,              -- encoded query-type-specific batch identifier (corresponds to identifier in BatchSelector)
+    batch_interval          TSRANGE,                     -- batch interval, as a TSRANGE, populated only for time-interval tasks. (will always match batch_identifier)
```
tgeoghegan, line 174: Column `batch_identifier` now occurs in table `collect_jobs` and `aggregate_jobs`, and I would bet you'll need it in `aggregate_share_jobs` when you do the helper part of this work. Should we introduce a `batches` table so that `aggregate_jobs` and `collect_jobs` or `aggregate_share_jobs` (depending on aggregator role) can consistently refer to the same batches?

aggregator/src/datastore.rs
```diff
@@ -3244,6 +3259,11 @@ pub mod models {
             &self.collect_job_id
         }
 
+        /// Returns the query type associated with this acquired aggregation job.
```
tgeoghegan, line 3262:
```suggestion
        /// Returns the query type associated with this acquired collect job.
```

tgeoghegan (approved):

aggregator/src/aggregator/collect_job_driver.rs
branlwyd: I noticed this too. I'm going to split out the pieces that don't fit in the next PR. (the motivation for this change was basically just, the aggregation job creator is `aggregation_job_creator.rs`; the aggregation job driver is `aggregation_job_driver.rs`; the collect job driver is `aggregate_share.rs` -- it'd be better if they matched)

db/schema.sql
branlwyd: Sure -- there's no technical requirement to do so, but I suppose not writing down the batch identifiers over & over again might be nice (we certainly arrange so that other DAP identifiers are typically only written down in one row).

This also allows me to move the batch interval logic to a central location.

aggregator/src/aggregator/collect_job_driver.rs
tgeoghegan: Yes, I agree that harmonizing with the name schemes for everything else has value. Also, and this is rather speculative: if my plans for https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/367 come to fruition, then aggregate share handling in the helper might become asynchronous just like collection handling in the leader, at which point we might want to run something almost exactly like `collect_job_driver` in Janus helpers, so maybe a `<something>_job_driver` (where "something" is a noun that accurately captures both aggregate shares and collections) is the right direction to go.

db/schema.sql
tgeoghegan: SGTM, we'll have a better perspective on all this once the helper aggregate share stuff is in place. Go ahead and merge at your convenience.

PR #756, opened 2022-11-16
Simplify GiST indices

### LINKED ISSUE #729: Check effectiveness of multi-column GiST indexes
We have two similar multi-column GiST indexes, over the `collect_jobs` and `aggregate_share_jobs` tables. Each indexes `(task_id, batch_interval)` in its respective table. Our intent is that these indexes are used by queries in the methods `get_collect_jobs_including_time()`, `get_collect_jobs_jobs_intersecting_interval()`,  `get_aggregate_share_jobs_including_time()`, and `get_aggregate_share_jobs_intersecting_interval()`.

I experimented with query plans for these queries on a database with a small amount of data in it, and fiddled with planner settings, but I was not able to produce plans that made use of the multi-column indexes. We should confirm whether the query planner can use these indexes when row counts are large. If there is a Postgres limitation preventing it from doing so, this could introduce unnecessary scans over every collect flow for a task. `get_collect_jobs_including_time()` is called while handling uploads, so this is hot code.

divergentdave:
Following #729, this removes `btree_gist` columns from our GiST indices. It appears the planner was not able to use these columns, so we can drop them to make the indices more compact as a minor optimization. The `btree_gist` extension is removed from the schema file as well, since we no longer use it.

Closes #729.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -1,7 +1,5 @@
 -- Load pgcrypto for gen_random_bytes.
 CREATE EXTENSION pgcrypto;
--- Load an extension to allow indexing over both BIGINT and TSRANGE in a multicolumn GiST index.
-CREATE EXTENSION btree_gist;
 
 -- Identifies which aggregator role is being played for this task.
 CREATE TYPE AGGREGATOR_ROLE AS ENUM(
@@ -187,7 +185,7 @@ CREATE TABLE collect_jobs(
 );
 -- TODO(#224): verify that this index is optimal for purposes of acquiring collect jobs.
 CREATE INDEX collect_jobs_lease_expiry ON collect_jobs(lease_expiry);
-CREATE INDEX collect_jobs_interval_containment_index ON collect_jobs USING gist (task_id, batch_interval);
+CREATE INDEX collect_jobs_interval_containment_index ON collect_jobs USING gist (batch_interval);
 
 -- The helper's view of aggregate share jobs.
 CREATE TABLE aggregate_share_jobs(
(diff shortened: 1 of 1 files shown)


db/schema.sql
```diff
@@ -202,7 +200,7 @@ CREATE TABLE aggregate_share_jobs(
     CONSTRAINT unique_aggregate_share_job_task_id_interval_aggregation_param UNIQUE(task_id, batch_interval, aggregation_param),
     CONSTRAINT fk_task_id FOREIGN KEY(task_id) REFERENCES tasks(id)
 );
-CREATE INDEX aggregate_share_jobs_interval_containment_index ON aggregate_share_jobs USING gist (task_id, batch_interval);
+CREATE INDEX aggregate_share_jobs_interval_containment_index ON aggregate_share_jobs USING gist (batch_interval);
```
branlwyd, line 203: To make sure I understand this change:
* The planner already wasn't using this index in relevant queries (i.e. queries finding collect jobs/aggregate share jobs by time or interval).
* With this change, the index is used.
* But the index no longer covers task_id, so now we will have to sort through collect jobs/aggregate share jobs from all tasks in order to find the ones we care about.

(correct me if I'm wrong on any of this, please)

If so, I think we can make this change, but this won't scale too well as the number of tasks increases. We may need to do something eventually, but for now this is a good improvement.

branlwyd (approved):

db/schema.sql
divergentdave: Slight correction: The planner was already using this index, but it was only ever putting the `batch_interval` column in the index condition when it did so, not the `task_id`. The `task_id` condition from the query gets put in the "filter" part of the index scan both before and after this index change. I agree regarding the concern at large numbers of tasks. Bitmap index scans are pretty good, but not optimal.

divergentdave:
I take it back, I just saw the following snippet in an unrelated query plan:

```
->  Bitmap Heap Scan on collect_jobs  (cost=4.33..23.61 rows=1 width=39)
      Recheck Cond: ((task_id = tasks.id) AND (batch_interval @> aggregation_jobs.batch_interval))
      Filter: ((lease_expiry <= '2022-11-17 00:00:00'::timestamp without time zone) AND (state = 'START'::collect_job_state) AND (aggregation_jobs.aggregation_param = aggregation_param))
      ->  Bitmap Index Scan on collect_jobs_interval_containment_index  (cost=0.00..4.33 rows=5 width=0)
            Index Cond: ((task_id = tasks.id) AND (batch_interval @> aggregation_jobs.batch_interval))
```

It is possible for the query planner to fully utilize both index columns, but presumably it's unlikely or unpredictable when it does, due to the aforementioned cost model limitations. This particular example came deep within an infeasible plan, where the total cost was 2*10^8. (Perhaps the planner has to really want it)

As such, I'll retract this PR. I don't think there's much for us to do here other than await future query planner improvements in Postgres 16 or later.

PR #759, opened 2022-11-18
 Fixed-size tasks: implement helper collection.

### LINKED ISSUE #468: Implement fixed-size aggregations

branlwyd:
Closes #468.

### DIFF
--- a/aggregator/src/datastore.rs
+++ b/aggregator/src/datastore.rs
@@ -1054,7 +1054,8 @@ impl<C: Clock> Transaction<'_, C> {
         let stmt = self
             .tx
             .prepare_cached(
-                "SELECT COUNT(1) AS count FROM report_aggregations
+                "SELECT COUNT(DISTINCT report_aggregations.client_report_id) AS count
+                FROM report_aggregations
                 JOIN aggregation_jobs ON aggregation_jobs.id = report_aggregations.aggregation_job_id
                 WHERE aggregation_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
                   AND aggregation_jobs.partial_batch_identifier = $2",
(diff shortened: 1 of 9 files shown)


aggregator/src/datastore.rs
```diff
@@ -1032,12 +1033,43 @@ impl<C: Clock> Transaction<'_, C> {
...
+        task_id: &TaskId,
+        batch_id: &BatchId,
+    ) -> Result<u64, Error> {
+        let stmt = self
+            .tx
+            .prepare_cached(
+                "SELECT COUNT(1) AS count FROM report_aggregations
+                JOIN aggregation_jobs ON aggregation_jobs.id = report_aggregations.aggregation_job_id
+                WHERE aggregation_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                  AND aggregation_jobs.partial_batch_identifier = $2",
```
divergentdave, line 1060: I think we need a `GROUP BY report_aggregations.client_report_id` to avoid double-counting in the case of nontrivial aggregation parameters.

divergentdave (approved):

db/schema.sql
```diff
@@ -193,13 +193,14 @@ CREATE INDEX collect_jobs_interval_containment_index ON collect_jobs USING gist
 CREATE TABLE aggregate_share_jobs(
     id                      BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY, -- artificial ID, internal-only
     task_id                 BIGINT NOT NULL,    -- the task ID being collected
-    batch_interval          TSRANGE NOT NULL,   -- the batch interval, as a range of timestamps
+    batch_identifier        BYTEA NOT NULL,     -- encoded query-type-specific batch identifier (corresponds to identifier in BatchSelector)
+    batch_interval          TSRANGE,            -- batch interval, as a TSRANGE, populated only for time-interval tasks. (will always match batch_identifier)
```
tgeoghegan, line 197: If I understand correctly:
- `batch_interval` will either be `NULL` or it will be identical to the value encoded into `batch_identifier`
- the only reason we keep this redundant representation is so that we can use Postgres' range operators within queries, as otherwise we would have to decode a time range from `batch_identifier` and do time interval intersection checks in code

I'm concerned about the possibility of the two columns disagreeing and I wonder if we can make this safer. A couple ideas off the top of my head:
- have columns `batch_interval TSRANGE` and `batch_identifier BIGINT` and use a [CHECK constraint](https://www.postgresql.org/docs/14/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS) on the table to ensure exactly one of them is non-NULL
- Keep the schema as-is, but use some more complex constraint to verify that `batch_identifier` is the correct encoding of the current value of `batch_interval`, if `batch_interval != NULL`
- Use a `TRIGGER` so that on any update to `batch_identifier`, we automatically set `batch_interval` to the correct value (this may not be possible since Postgres doesn't understand the DAP encoding scheme, but maybe we can register a callback to our code?).

I have to admit I don't fully understand the performance implications of any of these approaches.

tgeoghegan (approved):

db/schema.sql
divergentdave: CHECK constraints are only evaluated upon inserts and updates, so I think it should be OK from a performance point of view. One hurdle for implementing constraints is that the query type is not in the tables that have batch identifiers, and the identifier encodings are not self-describing. For now, though, we can distinguish them by length. (`Interval` encodes to 16 bytes, and `BatchId` encodes to 32 bytes) It looks like we'd have to compose our own big-endian integer decoding expressions with `get_byte(bytea, integer)`, which would be gnarly, but feasible.

aggregator/src/datastore.rs
branlwyd: 👍🏻 Nice catch. I expanded the test to test for this, as well.

db/schema.sql
branlwyd: (one more thought: in the longer term, we might decide that time-interval query types are no longer required. that would be a longer-term strategy to resolve the need to keep both of these fields around.)

PR #767, opened 2022-11-23
Add fixed size query support to collector crate

### LINKED ISSUE #764: Fixed-size task support in `janus_collector`
The collector library currently hardcodes the time interval query type in several places. I think adding fixed size task support here is the last thing we'll need for the integration test in #754.

(*narrator voice* it wasn't)

divergentdave:
This implements #764, changing the `janus_collector` crate to take any query as an argument, instead of solely a batch interval. The accompanying CLI is updated as well.

Closes #764.

### DIFF
--- a/collector/examples/collect.rs
+++ b/collector/examples/collect.rs
@@ -294,10 +294,18 @@ struct Options {
     buckets: Option<Buckets>,
 
     /// Start of the collection batch interval, as the number of seconds since the Unix epoch
-    #[clap(long, help_heading = "Collect Request Parameters (Time Interval)")]
+    #[clap(
+        long,
+        requires = "batch_interval_duration",
+        help_heading = "Collect Request Parameters (Time Interval)"
+    )]
     batch_interval_start: Option<u64>,
     /// Duration of the collection batch interval, in seconds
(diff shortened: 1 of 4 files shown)


collector/examples/collect.rs
```diff
@@ -259,84 +294,113 @@ struct Options {
...
+            .into())
+        }
+        (Some(_), Some(_), Some(_)) | (Some(_), None, Some(_)) | (None, Some(_), Some(_)) => {
+            Err(clap::Error::raw(
+                ErrorKind::ArgumentConflict,
+                "time interval and fixed size collect request parameters cannot be combined"
+                    .to_string(),
+            )
+            .into())
+        }
```
tgeoghegan, line 367: If the `batch_interval_start`, `batch_interval_duration` and `batch_id`  args are appropriately marked with [`Arg::requires`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires) and `conflicts`, could we replace these with a single case matching `_` that just returns some generic error or panics, on the premise that `clap` will have caught incorrect arguments before we get here?

tgeoghegan (approved):

collector/examples/collect.rs
divergentdave: That was a nice cleanup, thanks!

Issue #763, opened 2022-11-21
`janus_collector` should have better representation of invalid batch size


tgeoghegan:
If `janus_collector` makes a collect request that can't be serviced because there aren't enough reports in the batch interval, then (by source inspection), `Collector::collect` will fail with [`janus_collector::Error::Http(Box<HttpApiProblem>)`](https://github.com/divviup/janus/blob/578df05dc860ebf7b37341b436aa751b9761c3ee/collector/src/lib.rs#L90), and the boxed problem document will be of type [`invalidBatchSize`](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#name-errors). There is enough information here for a client of `janus_collector` to detect this case and take the appropriate action (generally, try again with a bigger batch interval), but since we expect this to happen in the normal course of collecting aggregate results, we could express this better in the API, by having `janus_collector` parse the `HttpApiProblem`, detect `invalidBatchSize` and return an `Error` variant that expresses exactly that.

tgeoghegan:
The error variant should still contain the `HttpApiProblem` in case the `janus_collector` client wants to log it or tease some metadata out of it.

PR #768, opened 2022-11-23
Add support for fixed size tasks to interop binaries and integration tests

divergentdave:
This adds a more-complete integration test for fixed size tasks. I had to solve the batch ID discovery problem ~3 times, so there is some near-repetition between different settings. We'll be able to rip this out once we move on to DAP-03, thanks to the `current_batch` query sub-type. I successfully tested this against a hacked-up collect job driver. I did notice one test flake, possibly related to #763, which I'll investigate further.

### DIFF
--- a/integration_tests/src/janus.rs
+++ b/integration_tests/src/janus.rs
@@ -1,8 +1,13 @@
 //! Functionality for tests interacting with Janus (<https://github.com/divviup/janus>).
 
+use crate::BatchDiscovery;
+use anyhow::{anyhow, Context};
+use async_trait::async_trait;
+use base64::URL_SAFE_NO_PAD;
 use janus_aggregator::{
     binary_utils::{database_pool, datastore},
     config::DbConfig,
+    datastore::Datastore,
     task::Task,
 };
 use janus_core::{
@@ -13,13 +18,17 @@ use janus_interop_binaries::{
     log_export_path, test_util::await_http_server, testcontainer::Aggregator,
     AggregatorAddTaskRequest,
 };
-use janus_messages::Role;
+use janus_messages::{BatchId, Role, TaskId};
 use k8s_openapi::api::core::v1::Secret;
 use portpicker::pick_unused_port;
+use prio::codec::{Decode, Encode};
+use serde::Deserialize;
+use serde_json::json;
 use std::{
     collections::HashMap,
     path::Path,
     process::{Command, Stdio},
+    sync::Arc,
     thread::panicking,
 };
 use testcontainers::{clients::Cli, Container, RunnableImage};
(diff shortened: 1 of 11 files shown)


divergentdave:
The flakiness I'm running into is due to the following race condition: the aggregation job creator may run when some, but not all reports have been uploaded. At this point, one BatchId has been stored, and there are not enough reports associated with it for collection. If the collect request arrives before the job creator runs again, then the `/collect` request will fail at batch validation with `urn:ietf:params:ppm:dap:error:invalidBatchSize`. (When the aggregation job creator next runs, it will assign the remainder of the reports to the batch)

This race condition didn't show up with time interval tasks because membership in a collect batch was defined implicitly, rather than decided by background processing. Since Janus durably stores reports before `/upload` finishes, the test would always succeed because all reports were counted towards the batch's report count.

integration_tests/src/janus.rs
```diff
@@ -218,3 +258,79 @@ impl<'a> Drop for Janus<'a> {
...
+            .send()
+            .await?
+            .error_for_status()?
+            .json::<InteropFetchBatchIdsResponse>()
+            .await?;
+        if &response.status != "success" {
+            if let Some(error) = &response.error {
+                return Err(anyhow!(
+                    "Error from /internal/test/fetch_batch_ids: {error}"
+                ));
```
tgeoghegan, line 293: Using the `Display` impl on `error` to construct an `anyhow::Error` is lossy. I think [`anyhow::Error::new`](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.new) would wrap `error` while preserving its structure.

integration_tests/src/janus.rs
```diff
@@ -169,18 +203,21 @@ impl Janus<'static> {
...
-            .unwrap()
-            .put_task(task)
-            .await
-            .unwrap();
+        let datastore = datastore(pool, RealClock::default(), &[datastore_key]).unwrap();
+        datastore.put_task(task).await.unwrap();
+
+        let batch_discovery = Arc::new(JanusClusterBatchFetch {
+            datastore,
+            _datastore_port_forward,
```
tgeoghegan, line 211: Does this line "use" `_datastore_port_forward` enough that we can remove the leading `_` without earning an unused variable warning?

tgeoghegan (approved):

integration_tests/src/janus.rs
divergentdave: The local variable: yes, the struct member: no. We store the `PortForward` in the struct to manage when `kubectl port-forward` is torn down as a side effect of dropping it, but we don't read from the struct member at all.

integration_tests/src/janus.rs
divergentdave: This variable is a `String` out of a `Deserialize`-enabled struct, so we aren't losing anything. The error was necessarily flattened into a string in another process, to be passed over via JSON.

branlwyd (approved):

PR #769, opened 2022-11-26
Fixed-size tasks: fix collect job driver.

### LINKED ISSUE #766: acquire_incomplete_collect_jobs() only supports time interval tasks
The collect job driver picks up incomplete jobs using `Transaction::acquire_incomplete_collect_jobs()`, but this only supports time interval tasks, due to its reliance on the `collect_jobs.batch_interval` column. A separate query should be introduced for fixed size task collect jobs.

branlwyd:
Previously, the collect job driver would only work for time-interval tasks because the underlying datastore operation only worked with time-interval tasks.

Closes #766.

### DIFF
--- a/aggregator/src/aggregator/collect_job_driver.rs
+++ b/aggregator/src/aggregator/collect_job_driver.rs
@@ -465,7 +465,7 @@ impl CollectJobDriver {
                 datastore
                     .run_tx(|tx| {
                         Box::pin(async move {
-                            let (mut time_interval_jobs, mut fixed_size_jobs) = try_join!(
+                            let (time_interval_jobs, fixed_size_jobs) = try_join!(
                                 tx.acquire_incomplete_time_interval_collect_jobs(
                                     &lease_duration,
                                     maximum_acquire_count_per_query_type,
@@ -475,8 +475,10 @@ impl CollectJobDriver {
                                     maximum_acquire_count_per_query_type
                                 ),
                             )?;
-                            time_interval_jobs.append(&mut fixed_size_jobs);
-                            Ok(time_interval_jobs)
+                            Ok(time_interval_jobs
+                                .into_iter()
+                                .chain(fixed_size_jobs.into_iter())
+                                .collect())
                         })
                     })
                     .await
--- a/aggregator/src/aggregator/collect_job_driver.rs
+++ b/aggregator/src/aggregator/collect_job_driver.rs
@@ -459,17 +459,24 @@ impl CollectJobDriver {
         lease_duration: Duration,
     ) -> impl Fn(usize) -> BoxFuture<'static, Result<Vec<Lease<AcquiredCollectJob>>, datastore::Error>>
     {
-        move |maximum_acquire_count| {
+        move |maximum_acquire_count_per_query_type| {
             let datastore = Arc::clone(&datastore);
             Box::pin(async move {
                 datastore
                     .run_tx(|tx| {
                         Box::pin(async move {
-                            tx.acquire_incomplete_collect_jobs(
-                                &lease_duration,
-                                maximum_acquire_count,
-                            )
-                            .await
+                            let (mut time_interval_jobs, mut fixed_size_jobs) = try_join!(
+                                tx.acquire_incomplete_time_interval_collect_jobs(
+                                    &lease_duration,
+                                    maximum_acquire_count_per_query_type,
+                                ),
+                                tx.acquire_incomplete_fixed_size_collect_jobs(
+                                    &lease_duration,
+                                    maximum_acquire_count_per_query_type
+                                ),
+                            )?;
+                            time_interval_jobs.append(&mut fixed_size_jobs);
+                            Ok(time_interval_jobs)
                         })
                     })
                     .await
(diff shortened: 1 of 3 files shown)


aggregator/src/aggregator/collect_job_driver.rs
```diff
@@ -459,17 +459,24 @@ impl CollectJobDriver {
...
+                                tx.acquire_incomplete_time_interval_collect_jobs(
+                                    &lease_duration,
+                                    maximum_acquire_count_per_query_type,
+                                ),
+                                tx.acquire_incomplete_fixed_size_collect_jobs(
+                                    &lease_duration,
+                                    maximum_acquire_count_per_query_type
+                                ),
+                            )?;
+                            time_interval_jobs.append(&mut fixed_size_jobs);
```
tgeoghegan, line 478: nit: using [`Vec::extend`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.extend) would not require `fixed_size_jobs` to be `mut`.

aggregator/src/datastore.rs
```diff
@@ -2149,6 +2149,92 @@ WITH updated as (
         WHERE
             -- Constraint for tasks table in FROM position
             tasks.id = collect_jobs.task_id
+            -- Only return
```
tgeoghegan, line 2152: Are there some words missing from this sentence?
```suggestion
            -- Only return time interval collect jobs.
```

aggregator/src/datastore.rs
```diff
@@ -2149,6 +2149,92 @@ WITH updated as (
...
+            AND tasks.query_type ? 'TimeInterval'
```
tgeoghegan, line 2153: Why do we use the `?` operator here? If I'm reading [the docs](https://www.postgresql.org/docs/14/functions-json.html) right, that will test whether `TimeInterval` appears as a "top-level key or array element", but don't we want to do simple string comparison between `query_type` and `TimeInterval`?

aggregator/src/datastore.rs
```diff
@@ -2149,6 +2149,92 @@ WITH updated as (
...
+        let now = self.clock.now();
+        let lease_expiry_time = now.add(lease_duration)?;
+        let maximum_acquire_count: i64 = maximum_acquire_count.try_into()?;
+
+        let stmt = self
+            .tx
+            .prepare_cached(
+                r#"
+WITH updated as (
+    UPDATE collect_jobs SET lease_expiry = $1, lease_token = gen_random_bytes(16), lease_attempts = lease_attempts + 1
```
tgeoghegan, line 2223: I'd like to see logic like this re-used between the fixed size and time interval queries, since the main difference between the two seems to be the `INNER JOIN` against `aggregation_jobs`. However, I'm not sure what's a good way to reuse code between prepared statements so I'm OK going ahead with this.

aggregator/src/datastore.rs
```diff
@@ -2149,6 +2149,92 @@ WITH updated as (
...
+        -- Join on aggregation jobs with matching task ID, matching aggregation parameter, and
+        -- matching batch identifier.
+        INNER JOIN aggregation_jobs
+            ON collect_jobs.aggregation_param = aggregation_jobs.aggregation_param
+            AND collect_jobs.task_id = aggregation_jobs.task_id
+            AND collect_jobs.batch_identifier = aggregation_jobs.batch_identifier
+        WHERE
+            -- Constraint for tasks table in FROM position
+            tasks.id = collect_jobs.task_id
+            -- Only return
```
tgeoghegan, line 2236: Same comment as on L2152

tgeoghegan (approved):

aggregator/src/datastore.rs
divergentdave: I tried this out, and there is no `jsonb = text` operator defined.

```
postgres=# SELECT '"TimeInterval"'::jsonb = 'TimeInterval'::text;
ERROR:  operator does not exist: jsonb = text
```

While that doc page doesn't mention it, using `?` on a JSON value that's a string is equivalent to unwrapping that string and comparing it to the `text` operand.

divergentdave (approved):

aggregator/src/aggregator/collect_job_driver.rs
branlwyd: SG, I went with an approach that chained two iterators to avoid making either variable `mut`.

aggregator/src/datastore.rs
branlwyd: Yeah, I played with this for a bit. We could do literal string concatenation, but that doesn't feel particularly maintainable. More than that would require an ORM (or the moral equivalent), and I don't think we want to go that way.

aggregator/src/datastore.rs
divergentdave: Ah, here we go, this is mentioned in an example on this page: https://www.postgresql.org/docs/14/datatype-json.html#JSON-CONTAINMENT

```
-- A string is considered to exist if it matches a primitive JSON string:
SELECT '"foo"'::jsonb ? 'foo';
```

I submitted a comment to the pgsql-docs mailing list about fleshing this out.

Issue #705, opened 2022-11-01
DAP-03: unrecognizedTask should use 400 Bad Request


divergentdave:
The next draft of the DAP spec will remove language specifying 404 Not Found for the `urn:ietf:params:ppm:dap:error:unrecognizedTask` error. Instead, we should respond with "400 Bad Request" for these errors.

tgeoghegan:
I don't know that I agree. The specification should avoid being more specific than HTTP 400 or 500 per [RFC 9205](https://www.rfc-editor.org/rfc/rfc9205.html#name-using-http-status-codes), but an implementation like Janus is free to use a more specific error code in the 400 range, and for this case, 404 feels more appropriate and helpful to clients than a generic Bad Request.

divergentdave:
There's a new sentence in [this paragraph in section 3.2](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-3.2-6) that says 'abort' "implies HTTP status code 400 Bad Request unless explicitly specified otherwise." I agree it would be better if the spec were flexible within 4xx/5xx categories.

tgeoghegan:
The good news is that we can still change the protocol text for DAP-03!

branlwyd:
I agree the specification text here should be more flexible; I'm going to hold off on implementing this change for now.

PR #795, opened 2022-11-29
Improve collector error handling

### LINKED ISSUE #763: `janus_collector` should have better representation of invalid batch size

divergentdave:
This PR makes the following changes:

- Moves the enum `DapProblemType` from `janus_aggregator` to `janus_messages` so it can be reused. (One method is left behind in an extension trait, but this will go away as part of DAP-03, see #705)
- Stores a parsed `DapProblemType` in the collector library's `Error::Http` variant, similar to that in the aggregator module.

Closes #763.

### DIFF
--- a/messages/src/lib.rs
+++ b/messages/src/lib.rs
@@ -1961,6 +1961,133 @@ impl Decode for AggregateShareResp {
     }
 }
 
+/// Representation of the different problem types defined in Table 1 in §3.2.
+#[derive(Debug, PartialEq, Eq)]
+pub enum DapProblemType {
+    UnrecognizedMessage,
+    UnrecognizedTask,
+    MissingTaskId,
+    UnrecognizedAggregationJob,
+    OutdatedConfig,
+    ReportTooLate,
+    ReportTooEarly,
+    BatchInvalid,
+    InvalidBatchSize,
+    BatchQueriedTooManyTimes,
+    BatchMismatch,
+    UnauthorizedRequest,
+    BatchOverlap,
+}
+
+impl DapProblemType {
+    /// Returns the problem type URI for a particular kind of error.
+    pub fn type_uri(&self) -> &'static str {
+        match self {
+            DapProblemType::UnrecognizedMessage => {
+                "urn:ietf:params:ppm:dap:error:unrecognizedMessage"
+            }
+            DapProblemType::UnrecognizedTask => "urn:ietf:params:ppm:dap:error:unrecognizedTask",
+            DapProblemType::MissingTaskId => "urn:ietf:params:ppm:dap:error:missingTaskID",
+            DapProblemType::UnrecognizedAggregationJob => {
+                "urn:ietf:params:ppm:dap:error:unrecognizedAggregationJob"
+            }
+            DapProblemType::OutdatedConfig => "urn:ietf:params:ppm:dap:error:outdatedConfig",
+            DapProblemType::ReportTooLate => "urn:ietf:params:ppm:dap:error:reportTooLate",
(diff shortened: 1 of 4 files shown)


branlwyd:
> * Adds a new set of configurable backoff parameters to control whether and how to retry upon receiving urn:ietf:params:ppm:dap:error:invalidBatchSize. By default, these errors are not retried at all. (Note that the http_request_retry_parameters retry loop will not retry these errors because they come with a 400 status code) For fixed-size tasks, users can set this parameter and let it run until it returns a result. For time interval tasks, users should be changing their query instead. (issuing collect requests for a time interval that is expected to receive more reports is inadvisable in the first place, as that will cause late reports to be rejected)

I don't object to keeping this, but is it necessary?

* For time-interval tasks, by the time the batch is collected it should be receiving ~no new reports (since it will be in the past)
* For fixed-size tasks, eventually we'll be using `current-batch`, which will be responsible for waiting for a batch to have enough reports before returning it.

branlwyd (approved):

divergentdave:
Yeah, I might peel that commit off to keep the library simpler, and just deal with these retries in tests.

Looking at [the relevant batch validation section](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.5.6.2.2), I don't think that `current_batch` is treated any differently than `by_batch_id`. I think that means the POST to `/collect` would return the same `invalidBatchSize` error, rather than return a URI and wait for reports before completing it.

branlwyd:
> Looking at [the relevant batch validation section](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.5.6.2.2), I don't think that `current_batch` is treated any differently than `by_batch_id`. I think that means the POST to `/collect` would return the same `invalidBatchSize` error, rather than return a URI and wait for reports before completing it.

FWIW, I think the idea is that `current_batch` would fail in some way if there is no batch ready to begin collection. That said, a quick glance at the [relevant section](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#name-fixed-size-queries) doesn't make clear exactly how this failure should occur. We may need to expand on the expected behavior (though we can technically leave it up to implementations, so we should consider carefully if additional specification language is justified).

messages/src/lib.rs
```diff
@@ -1961,6 +1961,133 @@ impl Decode for AggregateShareResp {
     }
 }
 
+/// Representation of the different problem types defined in Table 1 in §3.2.
+#[derive(Debug, PartialEq, Eq)]
+pub enum DapProblemType {
```
tgeoghegan, line 1966: Could we move this to a new file like `messages/src/problem_type.rs` just to stop this file from getting too big? It can be a little frustrating to navigate `janus` when you have 4-5 different files named `lib.rs` open and they're all very long.

tgeoghegan (approved):

PR #796, opened 2022-11-29
Bump Janus to version 0.3.

branlwyd:
Version 0.3 will implement the DAP-03 specification.


divergentdave (approved):
We should also bump the versions of inter-crate dependencies (or hoist the version specifiers into `workspace.dependencies` in the top level file, and bump them there)

tgeoghegan (approved):

Issue #507, opened 2022-09-14
Release actions for containers, SQL schema stomp on "latest"


tgeoghegan:
We now release Janus from multiple release branches. The actions for publishing the SQL schema file and the Docker containers posts artifacts versioned by the git tag, but also "latest". So if we did a release from `main`, and then a release from `release/0.1`, then the "latest" version of those artifacts would be the one from the tip `release/0.1`, and it's not clear that we want that behavior.

branlwyd:
I think we want to not use, and ideally not even publish (though I think it may be applied automatically?), a `latest` tag.

Maybe what we want is a tag like `0.1` that points to the latest `0.1.x` release? (and an eventual `0.2` tag that points to the latest `0.2.x` release, etc). Then a release `0.1.12` could be tagged with both `0.1` & `0.1.12` tags.

divergentdave:
I cleaned up the `latest` tags from both GAR repositories. With that, I think we can close this issue.

PR #800, opened 2022-12-01
0.1: Stop pushing images and schema as 'latest'

divergentdave:
Now that we have published 0.2.0, we shouldn't be marking release artifacts from this branch as "latest" anymore. See also #507.


tgeoghegan (approved):
LGTM, but here's a half baked idea: what if we pushed to a tag `0.1-latest`, and correspondingly, `0.2-latest` from `release/0.2`?

That being said, I don't think anyone uses our `latest` tag, so for that reason I think we should simply get rid of it and re-introduce it when we have a clear use case for it.

divergentdave:
As far as naming, I think it would be more conventional to name the tags that follow updates `0.1` or `0.2`. With that said, axing `latest` until we have a need for it sounds good to me.

Issue #761, opened 2022-11-18
DAP-03: Move extensions to ciphertext, reject unknown extensions


divergentdave:
In ietf-wg-ppm/draft-ietf-ppm-dap#373, extensions moved from the ReportMetadata to inside the HPKE ciphertext. Additionally, we must reject reports with unrecognized extensions, (i.e. any report with an extension, currently) and check for repetition of extension types in the list of extensions (ietf-wg-ppm/draft-ietf-ppm-dap#379).

PR #804, opened 2022-12-02
DAP-03: move extensions to ciphertext.

branlwyd:
This commit introduces the PlaintextInputShare structure used for input shares in DAP-03.

Part of #761.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1525,13 +1525,21 @@ impl VdafOps {
                 })
             });
 
+            let plaintext_input_share = plaintext.and_then(|plaintext| {
+                PlaintextInputShare::get_decoded(&plaintext).map_err(|error| {
+                    info!(task_id = %task.id(), metadata = ?report_share.metadata(), ?error, "Couldn't decode helper's plaintext input share");
+                    aggregate_step_failure_counter.add(&Context::current(), 1, &[KeyValue::new("type", "plaintext_input_share_decode_failure")]);
+                    ReportShareError::VdafPrepError
+                })
+            });
+
             // `vdaf-prep-error` probably isn't the right code, but there is no better one & we
             // don't want to fail the entire aggregation job with an UnrecognizedMessage error
             // because a single client sent bad data.
             // TODO(https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/255): agree on/standardize
             // an error code for "client report data can't be decoded" & use it here.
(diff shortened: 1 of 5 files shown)


aggregator/src/aggregator.rs
```diff
@@ -1249,6 +1249,8 @@ impl VdafOps {
         for<'a> &'a A::AggregateShare: Into<Vec<u8>>,
         C: Clock,
     {
+        // XXX: update
+
```
tgeoghegan, line 1253:
```suggestion
```

messages/src/lib.rs
```diff
@@ -853,31 +848,66 @@ impl ReportMetadata {
...
     fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
         let report_id = ReportId::decode(bytes)?;
         let time = Time::decode(bytes)?;
+
+        Ok(Self { report_id, time })
+    }
+}
+
+// XXX: documentation
+// XXX: tests
```
tgeoghegan, line 870: Indeed, this needs doccomment and tests. Same goes on L877, L885, L890.

tgeoghegan (changes requested):
Just needs some comment cleanup and some tests.

messages/src/lib.rs
branlwyd: Argh, one day I will remember to `git grep XXX` consistently before sending a PR for review. Sorry about that!

divergentdave (approved):

messages/src/lib.rs
tgeoghegan: No worries, I do it all the time myself. Maybe we need some kind of git pre-push hook that does that grep for us.

tgeoghegan (approved):

aggregator/src/aggregator.rs
```diff
@@ -1519,13 +1523,21 @@ impl VdafOps {
                 })
             });
 
+            let plaintext_input_share = plaintext.and_then(|plaintext| {
+                PlaintextInputShare::get_decoded(&plaintext).map_err(|error| {
+                    info!(task_id = %task.id(), metadata = ?report_share.metadata(), ?error, "Couldn't decode helper's plaintext input share");
+                    aggregate_step_failure_counter.add(&Context::current(), 1, &[KeyValue::new("type", "plaintext_input_share_decode_failure")]);
```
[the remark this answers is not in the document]
branlwyd: (I'm going to do this in the follow-on PR I'm about to send for review, to avoid needing to re-run CI.)

Issue #801, opened 2022-12-01
Key generation tool


divergentdave:
I think we should write a small CLI tool to generate HPKE keypairs, and output the HPKE config and private key material in the formats Janus expects them. This would simplify the process of manual task provisioning, especially with regards to the collector keypair. Previously, we have written ad-hoc code to help construct task definitions. Providing a keygen tool allows us to ask early partners to generate their own collector private key, and send us the public key/HPKE config. The `janus_messages` crate would be a good home for this, as another binary alongside `dap_decode`.

This should take in command line arguments (via clap) for the algorithm choices and HPKE config ID, and output the resulting `HpkeConfig` and `HpkePrivateKey` as both YAML documents, for use in Janus task definition files, and as encoded DAP messages, in base64url, for use with the `collect` CLI. We could write all four representations to standard output, separated by `---` YAML document separators, and annotated with comments.

tgeoghegan:
I previously wrote up https://github.com/divviup/janus/issues/528 which addresses one of the use cases you describe here, but I like how your proposal helps with the case of distributing keys to collectors.

divergentdave:
Done in #1116, follow-up improvements are tracked in #1182.

Issue #718, opened 2022-11-06
Documentation about how to deploy janus


MxmUrw:
For our [dspa-project](https://github.com/dpsa-project/overview), we need to run a janus instance in order to be able to test our type and pipeline.

It seems that setting up a janus deployment is not trivial, and I cannot find enough documentation to figure out the process. The end-to-end test in `interop_binaries/tests` has a very test-specific flavor, and it seems like a proper production (or even testing) deployment would look very different.

It could be that this is exactly the purpose of [daphne](https://github.com/cloudflare/daphne) (I just saw about it), but from the README it remains unclear to me what the exact relationship between daphne and janus is (both just say they are "implementations of DAP").

My current main goal is to setup such a testing environment, and I think I should continue at (https://github.com/cloudflare/daphne/issues/124). But it seems that it would be useful if there existed a description of how janus is supposed to be deployed (or the various options therefor).

MxmUrw:
I can try to ask directly, then it's probably faster to answer:

1. Is the following correct? Each (of the two) aggregation servers consists of:
   - a postgres database
   - four running janus processes (`aggregator`, `aggregation_job_creator`, `aggregation_job_driver`, `collect_job_driver`)

2. It seems that the collector and the clients only communicate with the `aggregator` process. The others are background processes that just need to be started once and then do their work (as they read from the database) automatically, correct?

branlwyd:
The above is correct (including the edit).

divergentdave:
I started work on configuration for `docker-compose` in #806.

MxmUrw:
I think this issue can be closed now that this [DEPLOYING.md](https://github.com/divviup/janus/blob/main/docs/DEPLOYING.md) is available.

PR #806, opened 2022-12-02
Docker Compose

divergentdave:
This is a start at a docker-compose configuration for development use. (as suggested in #718) Currently, this spins up two aggregators. I want it to also provision some sample tasks, but that will have to wait for a couple blockers to be resolved first. I primarily envision this as being useful for external developers, but internally, it could also serve as a more convenient substitute for a local Kind cluster in certain cases. (The docker-compose.yaml file captures how to build each container, and obviates the need for loading images into kind) The way that the `docker-compose up` terminal output weaves together log output from every component is pretty nice too, it's like having `stern` running without any extra steps.

Blockers:

- `janus_cli provision_tasks` currently requires a connection to a Kubernetes API server, even if datastore keys are provided directly.
- I need to write a small handful of task parameters, and I figure I may as well do #801 first, so I can use that.


tgeoghegan:
I'm on the fence about this: on the one hand, it's really nice to have something in the public Janus repo that makes it easy to spin up a functioning Janus instance. On the other, we have made an investment over in the private janus-ops repo to maintain some Terraform that lets us deploy Janus into a `kind` cluster. I want to avoid maintaining two different orchestration stories. If we commit to supporting a `docker-compose.yaml`, then I think we should get rid of the `kind` stuff and use `docker compose` in our integration tests.

PR #816, opened 2022-12-06
Refactor base64url-encoded fields in serde implementations

divergentdave:
This refactors parts of our task YAML serialization code out into serde trait implementations on individual types. Hoisting HPKE key YAML serialization from `janus_aggregator` to `janus_core` is a prerequisite for #801.  This PR makes the following changes:

- Move base64url-encoding of `TaskId`, `HpkePublicKey`, and `HpkePrivateKey` into custom serde trait implementations on the types themselves. (out of implementations on various serialization helper types)
- Remove `SerializedHpkeConfig`, and use `HpkeConfig` directly in types with serde derive macros applied.
- Replace `(HpkeConfig, HpkePrivateKey)` tuples with a new struct type, `HpkeKeypair`.
- Remove `SerializedHpkeKeypair`, and derive serde traits on `HpkeKeypair` directly.
- Add a `serde_test` test for the `Task` serialization format. The only changes to this test throughout the above refactor were struct names, which don't appear in the output when using `serde_yaml`.

### DIFF
--- a/core/src/hpke.rs
+++ b/core/src/hpke.rs
@@ -253,6 +253,7 @@ pub fn generate_hpke_config_and_private_key(
     )
 }
 
+/// An HPKE configuration and its corresponding private key.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct HpkeKeypair {
     config: HpkeConfig,
@@ -260,6 +261,7 @@ pub struct HpkeKeypair {
 }
 
 impl HpkeKeypair {
+    /// Construct a keypair from its two halves.
     pub fn new(config: HpkeConfig, private_key: HpkePrivateKey) -> HpkeKeypair {
         HpkeKeypair {
             config,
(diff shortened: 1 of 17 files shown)


core/src/hpke.rs
```diff
@@ -203,13 +253,36 @@ pub fn generate_hpke_config_and_private_key(
...
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct HpkeKeypair {
+    config: HpkeConfig,
+    private_key: HpkePrivateKey, // uses unpadded base64url
+}
+
+impl HpkeKeypair {
+    pub fn new(config: HpkeConfig, private_key: HpkePrivateKey) -> HpkeKeypair {
```
branlwyd, line 263: nit: doccomments for these methods

branlwyd (approved):

tgeoghegan (approved):

PR #818, opened 2022-12-06
Add per-task configuration for Daphne compatibility

divergentdave:
This adds a new task parameter, `input_share_aad_tweak` which changes how we build the input share AAD. This affects both the client and the aggregator. This is a 0.2-only change, since this compatibility issue is only relevant when targeting DAP-02. I added an integration test to confirm that aggregation works with this new mode.

We will likely need to do a manual schema migration in our dap-02 deployment. The following statement will suffice:

```sql
ALTER TABLE tasks ADD COLUMN input_share_aad_tweak BOOLEAN NOT NULL DEFAULT false;
```

### DIFF
--- a/aggregator/src/task.rs
+++ b/aggregator/src/task.rs
@@ -111,6 +111,8 @@ pub struct Task {
     collector_auth_tokens: Vec<AuthenticationToken>,
     /// HPKE configurations & private keys used by this aggregator to decrypt client reports.
     hpke_keys: HashMap<HpkeConfigId, (HpkeConfig, HpkePrivateKey)>,
+    /// Configuration option to add a length prefix for the public share in the input share AAD.
+    input_share_aad_tweak: bool,
 }
 
 impl Task {
@@ -131,6 +133,7 @@ impl Task {
         aggregator_auth_tokens: Vec<AuthenticationToken>,
         collector_auth_tokens: Vec<AuthenticationToken>,
         hpke_keys: I,
+        input_share_aad_tweak: bool,
     ) -> Result<Self, Error> {
         // Ensure provided aggregator endpoints end with a slash, as we will be joining additional
         // path segments into these endpoints & the Url::join implementation is persnickety about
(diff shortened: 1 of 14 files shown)


tgeoghegan (approved):

aggregator/src/task.rs
```diff
@@ -111,6 +111,8 @@ pub struct Task {
     collector_auth_tokens: Vec<AuthenticationToken>,
     /// HPKE configurations & private keys used by this aggregator to decrypt client reports.
     hpke_keys: HashMap<HpkeConfigId, (HpkeConfig, HpkePrivateKey)>,
+    /// Configuration option to add a length prefix for the public share in the input share AAD.
+    input_share_aad_tweak: bool,
```
branlwyd, line 115: nit: perhaps name this something more descriptive, like `input_share_include_public_share_length_prefix` (though that's a little wordy). The doc comment makes it clear what this flag controls, but the field name itself is a little too generic IMO.

db/schema.sql
```diff
@@ -22,7 +22,8 @@ CREATE TABLE tasks(
     min_batch_size         BIGINT NOT NULL,           -- the minimum number of reports in a batch to allow it to be collected
     time_precision         BIGINT NOT NULL,           -- the duration to which clients are expected to round their report timestamps, in seconds
     tolerable_clock_skew   BIGINT NOT NULL,           -- the maximum acceptable clock skew to allow between client and aggregator, in seconds
-    collector_hpke_config  BYTEA NOT NULL             -- the HPKE config of the collector (encoded HpkeConfig message)
+    collector_hpke_config  BYTEA NOT NULL,            -- the HPKE config of the collector (encoded HpkeConfig message)
+    input_share_aad_tweak  BOOLEAN NOT NULL DEFAULT false -- selects which input share AAD format should be used (true: includes a length prefix for the public share, akin to DAP-03, false: no length prefix for the public share)
```
branlwyd, line 26: nit: keep field comments at the same indentation level

branlwyd (approved):

PR #831, opened 2022-12-08
`janus_cli`: task parameter generation

### LINKED ISSUE #528: `janus_cli provision-tasks` should generate the aggregator's HPKE config and private key
`janus_cli provision-tasks` expects to be passed a file containing the YAML encoding of the tasks to provision into the Janus instance's database. This includes an HPKE configuration and private key for the aggregator to use, so the operator has to generate an HPKE config and encode it correctly. The tool should support generating the HPKE parameters if they are not present in the YAML task.

tgeoghegan:
`janus_cli provision-tasks` can now generate task IDs, VDAF verify keys, authentication tokens for aggregators and collectors and aggregator HPKE keypairs, if those values are not present in the task YAML. We also add an `--echo-tasks` flag so that the tool prints the YAML encoding of provisioned tasks to stdout.

Resolves #528

### DIFF
--- a/aggregator/src/bin/janus_cli.rs
+++ b/aggregator/src/bin/janus_cli.rs
@@ -10,7 +10,7 @@ use janus_aggregator::{
     config::{BinaryConfig, CommonConfig},
     datastore::{self, Datastore},
     metrics::install_metrics_exporter,
-    task::Task,
+    task::{SerializedTask, Task},
     trace::install_trace_subscriber,
 };
 use janus_core::time::{Clock, RealClock};
@@ -58,6 +58,14 @@ enum Command {
         /// A YAML file containing a list of tasks to be written. Existing tasks (matching by task
         /// ID) will be overwritten.
         tasks_file: PathBuf,
+
+        /// If true, task parameters omitted from the YAML tasks file will be randomly generated.
+        #[clap(long, default_value = "false")]
+        generate_missing_parameters: bool,
+
+        /// Write the YAML representation of the tasks that are written to stdout.
+        #[clap(long, default_value = "false")]
+        echo_tasks: bool,
     },
 
     /// Create a datastore key and write it to a Kubernetes secret.
@@ -91,6 +99,8 @@ impl Command {
                 common_options,
                 kubernetes_secret_options,
                 tasks_file,
+                generate_missing_parameters,
+                echo_tasks,
             } => {
                 let kube_client = kube::Client::try_default()
                     .await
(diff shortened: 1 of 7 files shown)


aggregator/src/datastore.rs
```diff
@@ -838,6 +838,30 @@ impl<C: Clock> Transaction<'_, C> {
         )?)
     }
 
+    /// Retrieves the HPKE config IDs currently in use for the specified task.
+    pub async fn get_hpke_config_ids_for_task(
```
branlwyd, line 842: Could we use `get_task`, then read the HPKE config IDs out of the full structure? That would save a little bit of extra code here, at a probably-negligible cost to performance.

aggregator/src/task.rs
```diff
@@ -373,9 +376,89 @@ struct SerializedTask {
...
+
+        if self.vdaf_verify_keys.is_empty() {
+            let vdaf_verify_key = SecretBytes::new(
+                thread_rng()
+                    .sample_iter(Standard)
+                    .take(self.vdaf.verify_key_length())
+                    .collect(),
+            );
+
+            self.vdaf_verify_keys = vec![base64::encode_config(
```
branlwyd, line 413: nit: prefer `Vec::from` for new code (as long as Clippy doesn't complain -- I think David found one corner case where Clippy reports that `vec!` is still more efficient for now)

aggregator/src/bin/janus_cli.rs
```diff
@@ -167,27 +188,47 @@ async fn provision_tasks<C: Clock>(datastore: &Datastore<C>, tasks_file: &Path)
...
-    datastore
+    let written_tasks = datastore
         .run_tx(|tx| {
-            let tasks = Arc::clone(&tasks);
+            let tasks = tasks.clone();
             Box::pin(async move {
-                for task in tasks.iter() {
+                let mut written_tasks = Vec::new();
+                for mut serialized_task in tasks.into_iter() {
+                    // Check whether any HPKE configs already exist for this task so we can ensure
```
branlwyd, line 198: We delete the existing task before creating a new one -- why do we need to care about collisions?

core/src/task.rs
```diff
@@ -47,6 +68,14 @@ impl AuthenticationToken {
     pub fn as_bytes(&self) -> &[u8] {
         &self.0
     }
+
+    /// Generate a random authentication token.
+    pub fn random() -> Self {
```
branlwyd, line 73: Instead, `impl Distribution<AuthenticationToken> for Standard`. There are a few examples in the messages crate, among other places in the codebase -- the function body will stay much the same except that you'll end up calling `rng.gen()` instead of `random()` to generate underlying random values.

branlwyd (changes requested):

aggregator/src/bin/janus_cli.rs
tgeoghegan: ...you're right, we don't. I really wish I had thought of that before writing the code to check for and handle collisions. Oh well!

aggregator/src/datastore.rs
tgeoghegan: I deleted this because we don't need to query existing HPKE config IDs per your observation in `janus_cli.rs`, but I'll remember your remark about trading off maintenance burden vs. hypothetical performance win.

divergentdave (approved):

branlwyd (approved):

Issue #706, opened 2022-11-01
DAP-03: Current batch queries for fixed size tasks


divergentdave:
The next draft of the DAP spec will introduce a new query sub-type, within fixed-size queries, to get the "current batch". New sub-messages `FixedSizeQuery` and `FixedSizeQueryType` are introduced, and included as part of `Query`. Leader aggregators will need new logic to support this kind of query, selecting a batch that has not yet been collected. The query validation procedures for leaders and helpers have been updated accordingly.

divergentdave:
I noticed that we `INSERT INTO outstanding_batches`, but don't yet `DELETE` from it. From its description, we should be deleting rows from this table once the corresponding batches have started collection. I think cleanup of these rows should probably be done when handling a `current_batch` query, in the same database transaction that writes the new collect job.

PR #832, opened 2022-12-11
DAP-03: implement current-batch functionality in Aggregator.

branlwyd:
The logic for deciding which batch to return for current-batch is currently very simple: we let Postgres decide which of the eligible batches to return. We may well need to tweak this approach later (e.g. to always return the least-recently-returned eligible batch), but this should be fine for a first pass.

Part of #706.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -24,6 +24,7 @@ use crate::{
     task::{self, Task, VerifyKey, PRIO3_AES128_VERIFY_KEY_LENGTH},
     try_join,
 };
+use anyhow::anyhow;
 use bytes::Bytes;
 use http::{
     header::{CACHE_CONTROL, CONTENT_TYPE, LOCATION},
@@ -38,7 +39,7 @@ use janus_core::{
 };
 use janus_messages::{
     problem_type::DapProblemType,
-    query_type::{FixedSize, QueryType, TimeInterval},
+    query_type::{FixedSize, TimeInterval},
     AggregateContinueReq, AggregateContinueResp, AggregateInitializeReq, AggregateInitializeResp,
     AggregateShareAad, AggregateShareReq, AggregateShareResp, AggregationJobId, BatchSelector,
     CollectReq, CollectResp, HpkeCiphertext, HpkeConfig, HpkeConfigId, InputShareAad, Interval,
@@ -758,7 +759,7 @@ impl TaskAggregator {
         collect_job_id: Uuid,
     ) -> Result<Option<Vec<u8>>, Error> {
         self.vdaf_ops
-            .handle_get_collect_job(datastore, &self.task, Arc::new(collect_job_id))
+            .handle_get_collect_job(datastore, Arc::clone(&self.task), Arc::new(collect_job_id))
             .await
     }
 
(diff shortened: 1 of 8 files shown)


aggregator/src/aggregator.rs
```diff
@@ -2427,16 +2438,18 @@ impl VdafOps {
...
                         .await?
                         .ok_or_else(|| {
                             datastore::Error::User(
-                                Error::UnrecognizedCollectJob(collect_job_id).into(),
+                                Error::UnrecognizedCollectJob(*collect_job_id).into(),
                             )
                         })?;
-
+                    Q::acknowledge_collection(tx, task.id(), collect_job.batch_identifier())
+                        .await?;
```
divergentdave, line 2452: FWIW this will make data inaccessible if the collector issues a DELETE request to a `current_batch` job it hasn't received yet, but I think that's a straightforward consequence of the DELETE semantics. (i.e. this is OK, and a collector should avoid doing that)

messages/src/lib.rs
```diff
@@ -995,11 +995,46 @@ impl Decode for Report {
...
+    fn decode(bytes: &mut Cursor<&[u8]>) -> Result<Self, CodecError> {
+        let query_type = u8::decode(bytes)?;
+        match query_type {
+            0 => {
+                let batch_id = BatchId::decode(bytes)?;
+                Ok(FixedSizeQuery::ByBatchId { batch_id })
+            }
+            1 => Ok(FixedSizeQuery::CurrentBatch),
+            _ => Err(CodecError::Other(
+                anyhow!("unexpected FixedSizeQueryTypevalue {}", query_type).into(),
```
divergentdave, line 1027: nit: typo
```suggestion
                anyhow!("unexpected FixedSizeQueryType value {}", query_type).into(),
```

aggregator/src/aggregator.rs
```diff
@@ -2024,29 +2025,36 @@ impl VdafOps {
...
+                let (task, req, aggregation_param) = (
+                    Arc::clone(&task),
+                    Arc::clone(&req),
+                    Arc::clone(&aggregation_param),
+                );
                 Box::pin(async move {
+                    let batch_identifier = Q::batch_identifier_for_query(tx, &task, req.query())
+                        .await?
+                        .ok_or_else(|| {
+                            datastore::Error::User(anyhow!("No batch ready for collection").into())
```
tgeoghegan, line 2043: I'd prefer not to use `anyhow` here. Something like `aggregator::Error::BatchInvalid(*task.id(), "no batch ready for collection".to_string())` should work.

aggregator/src/aggregator/query_type.rs
```diff
@@ -240,12 +247,26 @@ pub trait CollectableQueryType: AccumulableQueryType {
         )?;
         Ok(batch_aggregations.into_iter().flatten().collect::<Vec<_>>())
     }
+
+    async fn acknowledge_collection<C: Clock>(
```
tgeoghegan, line 251: I think this needs a doccomment. In particular, to explain when in the collection state machine this call would occur. If I was just reading this trait, I might think that this method is an appropriate place to consume max query count, but that needs to happen during the computation of the aggregate share, not here.

divergentdave (approved):

tgeoghegan (approved):

aggregator/src/aggregator.rs
branlwyd: That's a good point. I agree this is on-spec-but-funky behavior -- we could simply fail `DELETE` requests for a current batch, I can't think of a good reason to allow a collector to say "please nuke an arbitrary batch". I filed #846.

PR #833, opened 2022-12-12
CI debugging: Dump Kubernetes configurations

divergentdave:
It seems like either the certificate-authority-data field in the kubeconfig file is not valid (it should be base64-encoded PEM, i.e. a base64 turducken) or there's some other source of configuration stomping on it. Hopefully these additional diagnostics will help determine what's happening.


divergentdave:
Result: Cached build script outputs in our target directory caused incorrect compilations after our runners upgraded from Ubuntu 20.04 (with OpenSSL 1.1.1) to Ubuntu 22.04 (with OpenSSL 3.0.2). The function we were getting errors from, [`X509::stack_from_pem()`](https://github.com/sfackler/rust-openssl/blob/5aadcab921fb277a3cc620d822cb7f7d5565d5a8/openssl/src/x509/mod.rs#L670-L697), depends on being able to parse OpenSSL error codes for its correct functioning. (In this case, it's expected to skip a "no start line" error to pass over a PEM-encoded private key that's ahead of the corresponding PEM-encoded certificate) The error code comes from [`ERR_GET_LIB()`](https://github.com/sfackler/rust-openssl/blob/5aadcab921fb277a3cc620d822cb7f7d5565d5a8/openssl-sys/src/err.rs#L10-L70), which has an entirely different definition for the two relevant OpenSSL versions. When using a dynamically linked OpenSSL, `openssl-sys`'s build script is supposed to detect what version `libssl.so` is, and pass that information via `--cfg 300`, etc. For these CI runs, Cargo was reusing the build script output and `.rlib`s from the older operating system, and the resulting executables failed to interpret error codes correctly when run. (When I tried compiling a narrower set of crates, Cargo's metadata hash for openssl-sys wound up being different, for reasons I don't understand, and thus Cargo recompiled the crate and got a working test binary)

We need to add the operating system version to the cache key used by the rust-cache action, to avoid recycling caches across ABI-incompatible OS upgrades again.

PR #836, opened 2022-12-12
`janus_cli`: task parameter generation

### LINKED ISSUE #528: `janus_cli provision-tasks` should generate the aggregator's HPKE config and private key

tgeoghegan:
`janus_cli provision-tasks` can now generate task IDs, VDAF verify keys, authentication tokens for aggregators and collectors and aggregator HPKE keypairs, if those values are not present in the task YAML. We also add an `--echo-tasks` flag so that the tool prints the YAML encoding of provisioned tasks to stdout.

Resolves #528

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -3302,8 +3302,7 @@ mod tests {
         },
         messages::{DurationExt, TimeExt},
         task::{
-            test_util::{generate_auth_token, TaskBuilder},
-            QueryType, Task, VerifyKey, PRIO3_AES128_VERIFY_KEY_LENGTH,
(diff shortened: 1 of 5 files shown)


tgeoghegan:
This is pretty much the same as #831, but targeting `main` instead of `release/0.2`. Some improvements to `Task/TaskBuilder/SerializedTask` landed on `main` that make this change a bit simpler.

branlwyd (changes requested):
(same comments as in `0.2` version)

branlwyd (approved):

PR #848, opened 2022-12-14
`janus_cli`: add dry-run mode

### LINKED ISSUE #529: `janus_cli` should have a `--dry-run` option
Tools like `janus_cli` that operate on live databases should have a `--dry-run` option that makes it print out what it would do without actually doing it so that operators can get confidence that they are setting all the right arguments and constructing config files correctly.

tgeoghegan:
Building on the previously introduced dry-run support in `Datastore`, `janus_cli` can now be run with a `--dry-run` flag. In this mode, no real modifications to the datastore (in the case of `write-schema` or `provision-tasks`) or Kubernetes secrets (in the case of `create-datastore-key`) are made.

We also refactor some of the command line option and configuration file handling to reduce code duplication in `Command::execute`.

Resolves #529

### DIFF
--- a/aggregator/src/bin/janus_cli.rs
+++ b/aggregator/src/bin/janus_cli.rs
@@ -31,27 +31,31 @@ static SCHEMA: &str = include_str!("../../../db/schema.sql");
 
 #[tokio::main]
 async fn main() -> Result<()> {
-    // Parse options, then read & parse config.
-    let options = Options::parse();
+    // Parse command-line options, then read & parse config.
+    let command_line_options = CommandLineOptions::parse();
+    let config_file: ConfigFile = read_config(&command_line_options.common_options)?;
 
-    debug!(?options, "Starting up");
+    install_tracing_and_metrics_handlers(config_file.common_config())?;
 
-    options.cmd.execute().await
+    debug!(?command_line_options, ?config_file, "Starting up");
+
+    if command_line_options.dry_run {
+        info!("DRY RUN: no persistent changes will be made")
+    }
+
+    command_line_options
+        .cmd
+        .execute(&command_line_options, &config_file)
+        .await
 }
 
 #[derive(Debug, Parser)]
 enum Command {
     /// Write the Janus database schema to the database.
-    WriteSchema {
-        #[clap(flatten)]
-        common_options: CommonBinaryOptions,
-    },
+    WriteSchema,
 
     /// Write a set of tasks identified in a file to the datastore.
     ProvisionTasks {
-        #[clap(flatten)]
-        common_options: CommonBinaryOptions,
-
         #[clap(flatten)]
         kubernetes_secret_options: KubernetesSecretOptions,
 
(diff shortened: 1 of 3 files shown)


aggregator/src/bin/janus_cli.rs
```diff
@@ -168,14 +142,12 @@ fn install_tracing_and_metrics_handlers(config: &CommonConfig) -> Result<()> {
...
 }
 
-async fn write_schema(pool: &Pool) -> Result<()> {
+async fn write_schema<C: Clock>(datastore: &Datastore<C>) -> Result<()> {
     info!("Writing database schema");
-    let db_client = pool.get().await.context("couldn't get database client")?;
-    db_client
-        .batch_execute(SCHEMA)
+    datastore
+        .run_tx(|tx| Box::pin(async move { tx.batch_execute(SCHEMA).await }))
```
divergentdave, line 148: Going through the datastore to set up the schema imposes a hard requirement that we provide a datastore key (or fake datastore key) that won't be used. I'd prefer we use a database connection directly, in the name of ease-of-use when setting up a database, but this isn't strongly held.

divergentdave (approved):

aggregator/src/bin/janus_cli.rs
branlwyd: I agree; we made a purposeful choice to not require a datastore key when writing the schema, let's not regress on that choice.

branlwyd (approved):

aggregator/src/bin/janus_cli.rs
tgeoghegan: I see your point, but the reason I did it this way was so that I could use `Datastore`'s support for rolling back transactions to implement a dry-run mode for this subcommand. I'm not sure if it'll be worth the effort and ongoing maintenance burden to set up a means of running a DB transaction that we can roll back without using `Datastore`, or making the `Datastore`'s `Crypter` optional. Do we actually ever want to provision a DB schema without access to the datastore key? My expectation is that anyone running this would run `janus_cli create-datastore-key` (to write the key into Kubernetes secrets) and then `janus_cli write-schema` (which would fetch the data store key from k8s secrets, meaning the operator doesn't actually have to provide it).

Maybe what we should do here is collapse `create-datastore-key` and `write-schema` into a single `provision-database` subcommand that would ensure that the datastore is ready to be used. Does it ever make sense to create a datastore key but _not_ write the schema, or vice-versa?

aggregator/src/datastore.rs
```diff
@@ -293,6 +293,12 @@ pub struct Transaction<'a, C: Clock> {
 }
 
 impl<C: Clock> Transaction<'_, C> {
+    /// Execute a sequence of SQL statements.
+    #[tracing::instrument(skip(self), err)]
+    pub async fn batch_execute(&self, query: &str) -> Result<(), Error> {
```
branlwyd, line 298: This is a very non-portable operation; I think we should not include it in the `Datastore` struct. (the purpose of the datastore layer is to abstract the operations from the raw SQL that implements them, to allow other storage backends to more easily be written)

aggregator/src/bin/janus_cli.rs
branlwyd: We might want to provision a DB without access to the datastore key: in a production environment, the key should be generated in a secure location, i.e. not on an operator workstation, to avoid exposing the key to an untrusted machine. (The general principle is that secrets should be exposed only to trusted machines serving production workloads, on an as-needed basis.)

OTOH, we might decide to do schema deployments (and, eventually, updates) from operator workstations. We might also introduce other operational commands which touch the DB without needing a datastore.

Maybe we need to rethink the way that dry-run is implemented -- the current approach has a few problems already, such as (a) reverting a transaction won't necessarily reveal all of the issues that would occur on commit and (b) it expects that all operations be done in a single transaction since otherwise operations from the first transaction won't be visible in the following transaction. Instead of adding dry-run to the entire datastore, we could instead introduce a CLI-specific trait which provides the necessary datastore operations, with a "real" implementation on `Datastore` and a "dry run" implementation that does nothing (except perhaps print what would be done).

branlwyd (changes requested):

aggregator/src/bin/janus_cli.rs
```diff
@@ -497,20 +556,57 @@ mod tests {
...
+        ds.run_tx(|tx| Box::pin(async move { tx.get_tasks().await }))
+            .await
+            .unwrap_err();
+    }
+
     fn task_hashmap_from_slice(tasks: Vec<Task>) -> HashMap<TaskId, Task> {
         tasks.into_iter().map(|task| (*task.id(), task)).collect()
     }
 
+    async fn setup_provision_tasks_testcase(
```
divergentdave, line 588: nit: I think `run_provision_tasks_testcase()` might be a better name, since this does run `provision_tasks()` after setting up the temporary file.

divergentdave (approved):

branlwyd (approved):

PR #849, opened 2022-12-14
Run CI with MSRV and stable Rust toolchains

### LINKED ISSUE #540: Run CI with MSRV toolchain as well as the latest
Janus crates advertise a minimum supported Rust version, but our CI jobs only build and test with the current stable toolchain. We should do [what libprio-rs does](https://github.com/divviup/libprio-rs/blob/64415d0039841d860f2201d79b9661d01a6a4239/.github/workflows/ci-build.yml#L31) and use a matrix to build with both the MSRV and current stable.

tgeoghegan:
Resolves #540

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -13,6 +13,13 @@ on:
 
 jobs:
   janus_build:
+    strategy:
+      matrix:
+        rust-toolchain: [
+          # MSRV from Cargo.toml
+          "1.65",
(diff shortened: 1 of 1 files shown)


tgeoghegan:
At the moment, our MSRV is the same as stable, but it looks like [1.66 is arriving tomorrow](https://releases.rs/docs/unreleased/1.66.0/).

branlwyd (approved):

PR #850, opened 2022-12-14
Document `test-util` items

### LINKED ISSUE #309: Annotate `test-util` items so they show up on docs.rs
We hide quite a few items behind feature `test-util` in the different Janus crates. To get those items to show up in the docs hosted on docs.rs, we have to [annotate them appropriately](https://github.com/divviup/libprio-rs/blob/290b7c74398ee570827f90cca9efa5ac07041544/src/vdaf/prio3.rs#L92) and then [run `cargo doc` with the right flags](https://github.com/divviup/libprio-rs/blob/main/Cargo.toml#L62), just like we do in libprio-rs.

tgeoghegan:
Annotates items across the `janus_core`, `janus_collector` and `janus_aggregator` crates that are gated on the `test-util` feature so that they get documented by `cargo doc`.

Resolves #309

### DIFF
--- a/aggregator/Cargo.toml
+++ b/aggregator/Cargo.toml
@@ -8,6 +8,10 @@ repository.workspace = true
 rust-version.workspace = true
 version.workspace = true
 
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+
 [features]
 default = ["kube-rustls"]
 tokio-console = ["dep:console-subscriber"]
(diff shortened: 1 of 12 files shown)


divergentdave (approved):
While we don't publish `janus_aggregator`, I wanted to flag that the `[package.metadata.docs.rs]` stanza for it won't work as written (setting aside publish=false) because `all-features` turns on `tokio-console`, and compilation with that feature also requires setting `--cfg=tokio_unstable`. The conditional annotations in that crate's code are worth keeping, as they could show up in locally-rendered rustdocs. (with the right command line)

PR #851, opened 2022-12-15
Fixes for Rust 1.66

tgeoghegan:
- Fix `clippy` warnings for unnecessary dereferences
 - Fix `clippy` warnings for unnecessary `u128 -> u128` cast
 - [`cargo publish` now blocks until it sees the published package in the index][1], so we can remove the retry loop from our crate publishing action.

[1]: https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-166-2022-12-15

### DIFF
--- a/.github/workflows/publish-crates.yml
+++ b/.github/workflows/publish-crates.yml
@@ -13,33 +13,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


tgeoghegan:
This change will fix our CI, which is currently broken (e.g. https://github.com/divviup/janus/actions/runs/3707153921/jobs/6283229661).

- [ ] We also need this on `release/0.2`

divergentdave (approved):

PR #854, opened 2022-12-15
DAP-03: Update HTTP problem documents

### LINKED ISSUE #545: Janus problem documents should allow omitting `task_id` key
Some problem documents cannot contain a task ID, such as when no task ID is provided to the `hpke_config` endpoint. [`build_problem_details_response`](https://github.com/divviup/janus/blob/f9eaa0b358a0e7913ed9322dcabf13695130864e/janus_server/src/aggregator.rs#L2031) doesn't handle this particularly well. We end up with problem documents that contain `"taskid" : null`. We should change how we construct those so that we can omit `taskid` when applicable.

tgeoghegan:
DAP-03 has [new language][1] on HTTP problem documents. The fields `taskid` and `instance` are no longer mandatory, so we don't have to include the former when it's not known and we don't have to always set the latter to `".."`.

[1]: https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-03#name-errors

Resolves #545

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -2762,7 +2762,6 @@ fn build_problem_details_response(error_type: DapProblemType, task_id: Option<Ta
             ("type", serde_json::Value::from(error_type.type_uri())),
             ("title", serde_json::Value::from(error_type.description())),
             ("status", serde_json::Value::from(status.as_u16())),
-            ("detail", serde_json::Value::from(error_type.description())),
         ]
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
One suggestion: I think we should remove the `detail` field from our responses for now, as it is duplicative. We could add it back later on a case-by-case basis with instance-specific information. (RFC 7807 says the `title` field is a human-readable description of the `type` URI, and the `detail` field is a human-readable explanation of a particular instance/occurrence of the error)

PR #858, opened 2022-12-15
DAP-03: Aggregators must advertise `HpkeConfigList`

tgeoghegan:
In DAP-03, aggregators may advertise multiple HPKE configs in order of preference, and clients may choose any of them for encrypting reports ([1]). In this change, we introduce the `HpkeConfigList` structure and handle it in `janus_client`, but `janus_aggregator` still only ever advertises a single HPKE configuration.

[1]: https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-03#name-hpke-configuration-request

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -12,8 +12,8 @@ use janus_core::{
     time::{Clock, TimeExt},
 };
 use janus_messages::{
-    Duration, HpkeCiphertext, HpkeConfig, InputShareAad, PlaintextInputShare, Report,
-    ReportMetadata, Role, TaskId,
+    Duration, HpkeCiphertext, HpkeConfig, HpkeConfigList, InputShareAad, PlaintextInputShare,
+    Report, ReportMetadata, Role, TaskId,
 };
 use prio::{
     codec::{Decode, Encode},
@@ -42,6 +42,8 @@ pub enum Error {
     Vdaf(#[from] prio::vdaf::VdafError),
     #[error("HPKE error: {0}")]
     Hpke(#[from] janus_core::hpke::Error),
+    #[error("unexpected server response {0}")]
+    UnexpectedServerResponse(&'static str),
 }
 
 static CLIENT_USER_AGENT: &str = concat!(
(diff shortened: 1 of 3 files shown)


messages/src/lib.rs
```diff
@@ -926,6 +929,40 @@ impl Decode for HpkeConfig {
...
+    /// Construct an HPKE configuration list.
+    pub fn new(hpke_configs: Vec<HpkeConfig>) -> Self {
+        Self(hpke_configs)
+    }
+
+    pub fn hpke_configs(&self) -> &[HpkeConfig] {
+        &self.0
+    }
+
+    pub fn as_vec(self) -> Vec<HpkeConfig> {
```
branlwyd, line 949: nit: I think this can be removed once the other comment is taken.

client/src/lib.rs
```diff
@@ -155,9 +158,17 @@ pub async fn aggregator_hpke_config(
...
 
-    Ok(HpkeConfig::decode(&mut Cursor::new(
+    let hpke_configs = HpkeConfigList::decode(&mut Cursor::new(
         hpke_config_response.bytes().await?.as_ref(),
-    ))?)
+    ))?;
+
+    // TODO(#857): Pick one of the advertised HPKE configs. For now, just take the first one, since
+    // we support any HpkeConfig we can decode, and it should be the server's preferred one.
+    let hpke_config = VecDeque::from(hpke_configs.as_vec()).pop_front().ok_or(
```
branlwyd, line 167: Is the goal to just get the first `hpke_config`? If so, I think you could state this more simply with something like `hpke_configs.hpke_configs().get(0)` (and then `clone` the result, I think).

branlwyd (approved):

client/src/lib.rs
divergentdave: (alternately: `hpke_configs.as_vec().into_iter().next().ok_or(...)?`

messages/src/lib.rs
divergentdave: Per typical [naming conventions](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv), `into_vec()` would be better for this method.

client/src/lib.rs
tgeoghegan: Yes, I was trying to avoid the `clone()`, but I suspect it's the sort of thing that gets optimized away, and it's not worth carrying around the extra method on `HpkeConfigList`.

divergentdave (approved):

Issue #464, opened 2022-09-06
Integrating our custom prio type into janus


ooovi:
Hi!
We just forked with the intention of integrating our new [prio type for aggregating fixed-point vectors](https://github.com/divviup/libprio-rs/pull/283) with janus. Ultimately, we wish to use this as a building block of [federated learning with global differential privacy](https://github.com/dpsa-project/overview).

Do you have some tips or suggestions on how to best proceed for using janus with a custom prio3 type? It seems that it is required to integrate references to our type into many files (e.g. in `janus_server/aggregator.rs`). From our point of view this looks like an opportunity to refactor and make it easier to add new types to janus. Are there any existing plans or thoughts regarding this?

If such a refactoring is deemed useful to janus, we would like to give it a go!

tgeoghegan:
Janus is designed to work generically over implementations of the [`Aggregator`, `Client`, `Collector` and `Vdaf` traits in `prio::vdaf`](https://docs.rs/prio/latest/prio/vdaf/index.html#traits), so no new logic should be needed to support new VDAFs. The exception is that Janus needs to dynamically figure out which implementation of traits `Aggregator` and/or `Vdaf` to instantiate based on a network message it receives (i.e., an upload request, which contains a task ID, which refers to a row in the database, which contains a description of the VDAF in use).

I think you will find yourself needing to add entries for the fixed point vector type to various enums throughout Janus, like [this one](https://github.com/divviup/janus/blob/106fa81991c83355d56fab2f684f59c76e6f7d1b/janus_core/src/task.rs#L6) or [this one](https://github.com/divviup/janus/blob/106fa81991c83355d56fab2f684f59c76e6f7d1b/janus_server/src/task.rs#L127). You will also have to update `match` blocks where we dynamically dispatch to a specialized impl of a generic method like [here](https://github.com/divviup/janus/blob/106fa81991c83355d56fab2f684f59c76e6f7d1b/janus_server/src/aggregator/aggregation_job_creator.rs#L216) or [here](https://github.com/divviup/janus/blob/106fa81991c83355d56fab2f684f59c76e6f7d1b/janus_server/src/aggregator.rs#L727).

It'd be interesting if we could use macros to stamp out the dispatch tables and even enum variants, but given the small number of supported VDAFs, none of us have yet had the courage to try to write such a macro.

ooovi:
[...]

PR #893, opened 2023-01-08
integrating bounded l2 norm fixed point vector sum types.

ooovi:
As discussed [here](https://github.com/divviup/janus/issues/464), this PR enables using our [prio3 type for aggregating vectors of fixed-point numbers](https://github.com/divviup/libprio-rs/pull/283) with janus. It mainly adds the enum cases and match branches required to dispatch on the 16-, 32- and 64-bit variants of fixed-point vector aggregation.

One notable thing I did is adding a [helper enum to enable tagging the JSON from the upload request with the fixedpoint type in the end-to-end interop test](https://github.com/dpsa-project/janus/blob/abda6a06637014808cdfead37912d88b2ca972bb/interop_binaries/src/bin/janus_interop_client.rs#L30-L43) only for our vectors, as the standard string representation of fixedpoint numbers does not carry that information.

Looking forward to your comments :)

### DIFF
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -652,6 +652,8 @@ mod tests {
     };
     use assert_matches::assert_matches;
     use chrono::{TimeZone, Utc};
+    #[cfg(feature = "fpvec_bounded_l2")]
+    use fixed_macro::fixed;
     use janus_core::{
         hpke::{
             self, test_util::generate_test_hpke_config_and_private_key, HpkeApplicationInfo, Label,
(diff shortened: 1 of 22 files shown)


tgeoghegan (commented):
This change looks sound to me, but the Janus team is debating how we want to handle this. For one thing, we may want to gate this VDAF behind a feature (much as the fixed point implementation in `libprio-rs` is gated behind feature `experimental`).

collector/src/lib.rs
```diff
@@ -822,7 +823,7 @@ mod tests {
         let collector = setup_collector(vdaf);
 
         let batch_interval = Interval::new(
-            Time::from_seconds_since_epoch(1_000_000),
+            Time::from_seconds_since_epoch(2_000_000), // TODO why???
```
branlwyd, line 826: Not sure offhand -- if you don't know why this change is necessary, I think it's OK to just drop the TODO.

branlwyd (changes requested):
As Tim mentioned, let's gate this with crate features, disabled by default. (Tests can enable the features in order to allow testing this functionality by default.)

collector/src/lib.rs
divergentdave: I saw a test flake in this module yesterday, so it was probably random chance, unconnected to this timestamp. (I suspect we may be running into a `backoff` timer we need to tweak, but I haven't had a chance to dig into it yet)

ooovi:
we are now hiding behind a feature.

> Tests can enable the features in order to allow testing this functionality by default
we had to adjust the `Dockerfile.interop` and `Dockerfile.interop_aggregator` to enable the feature during tests. is that what you meant, or should we do something more?

branlwyd (commented):
> we had to adjust the Dockerfile.interop and Dockerfile.interop_aggregator to enable the feature during tests. is that what you meant, or should we do something more?

I meant to modify the dependency in the `dev-dependencies` so that it enables the new feature, allowing the relevant tests to _not_ be gated by the feature & thus tested by `cargo test` without additional flags. (My goal is to make a release binary not include these "experimental"/non-standardized VDAFs by default, but I think it is wise to test them by default.)

See https://github.com/divviup/janus/blob/f8ad51953c0a9d31e6eaf973c882c62287844caf/aggregator/Cargo.toml#L98 for an example of the change to `Cargo.toml` -- there are a number of `dev-dependencies` throughout the workspace that enable the `test-util` feature.

ooovi:
> I meant to modify the dependency in the `dev-dependencies` so that it enables the new feature, allowing the relevant tests to _not_ be gated by the feature & thus tested by `cargo test` without additional flags. (My goal is to make a release binary not include these "experimental"/non-standardized VDAFs by default, but I think it is wise to test them by default.)

I did this (and the other changes you suggested) in the latest commit. Is there another way of enabling the feature on the containers used in the interop end-to-end test except passing the feature flag to the build command in the `Dockerfile.interop` and `Dockerfile.interop_aggregator`?

collector/src/lib.rs
```diff
@@ -822,7 +824,7 @@ mod tests {
         let collector = setup_collector(vdaf);
 
         let batch_interval = Interval::new(
-            Time::from_seconds_since_epoch(1_000_000),
+            Time::from_seconds_since_epoch(2_000_000),
```
divergentdave, line 827: This change can be backed out, #916 properly fixes this test.

branlwyd (approved):

ooovi:
I merged the current master, as there were conflicting changes. Also squashed into one commit. Could you run the workflow again? :)

ooovi:
the workflow cannot be run on forks because they have no access to your github secrets. there seems to be an event that gives them access:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target
but i don't know if that's what you want to be doing in general:
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

branlwyd:
I ran the tests locally; I am OK merging with that level of testing. I would prefer not to enable running with secrets against forks; perhaps if you will be creating many PRs, we can flip the appropriate bits so that you can create a branch on the main repository, and then we can enable you to run CI directly.

PR #894, opened 2023-01-09
Implement garbage collection. [0.2]

branlwyd:
The 0.2 version of this functionality only supports leader tasks with time-interval query types. This is because supporting helper tasks would require a schema change & backfill, and because fixed-size query types are not deployable in DAP-02 due to the lack of current-batch support.

### DIFF
--- a/aggregator/src/aggregator/garbage_collector.rs
+++ b/aggregator/src/aggregator/garbage_collector.rs
@@ -1,0 +1,240 @@
+use std::sync::Arc;
+
+use crate::{
+    datastore::Datastore,
+    messages::TimeExt,
+    task::{QueryType, Task},
+};
+use anyhow::{anyhow, Context, Result};
+use futures::future::join_all;
+use janus_core::time::Clock;
+use janus_messages::Role;
+use tracing::error;
+
+pub struct GarbageCollector<C: Clock> {
+    // Dependencies.
+    datastore: Arc<Datastore<C>>,
+    clock: C,
+}
+
+impl<C: Clock> GarbageCollector<C> {
+    pub fn new(datastore: Arc<Datastore<C>>, clock: C) -> Self {
+        Self { datastore, clock }
+    }
+
+    #[tracing::instrument(skip(self))]
+    pub async fn run(&self) -> Result<()> {
+        // TODO(#224): add support for handling only a subset of tasks in a single job (i.e. sharding).
+
+        // Retrieve tasks.
+        let tasks = self
+            .datastore
+            .run_tx(|tx| Box::pin(async move { tx.get_tasks().await }))
+            .await;
+        let tasks = match tasks {
+            Ok(tasks) => tasks,
+            Err(error) => return Err(error).context("couldn't retrieve tasks"),
+        };
+
+        // Run GC for each task.
+        join_all(tasks.into_iter().map(|task| async move {
+            let task = Arc::new(task);
+            if let Err(err) = self.gc_task(Arc::clone(&task)).await {
+                error!(task_id = ?task.id(), ?err, "Couldn't GC task");
+            }
+        }))
+        .await;
+        Ok(())
+    }
+
+    async fn gc_task(&self, task: Arc<Task>) -> Result<()> {
+        let oldest_allowed_report_timestamp =
+            if let Some(report_expiry_age) = task.report_expiry_age() {
+                if task.role() != &Role::Leader || task.query_type() != &QueryType::TimeInterval {
+                    return Err(anyhow!(
+                        "garbage collection is implemented only for leader, time-interval tasks"
+                    ));
+                }
+                self.clock.now().sub(report_expiry_age)?
+            } else {
+                // No configured report expiry age -- nothing to GC.
+                return Ok(());
+            };
+
+        self.datastore
+            .run_tx(|tx| {
+                let task = Arc::clone(&task);
+                Box::pin(async move {
+                    // Find and delete old collect jobs.
+                    tx.delete_old_collect_artifacts(task.id(), oldest_allowed_report_timestamp)
+                        .await?;
+
+                    // Find and delete old aggregation jobs/report aggregations/batch aggregations.
+                    tx.delete_old_aggregation_artifacts(task.id(), oldest_allowed_report_timestamp)
+                        .await?;
+
+                    // Find and delete old client reports.
+                    tx.delete_old_client_reports(task.id(), oldest_allowed_report_timestamp)
+                        .await?;
+
+                    Ok(())
+                })
+            })
+            .await?;
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::sync::Arc;
+
+    use super::GarbageCollector;
+    use crate::{
+        datastore::{
+            models::{
+                AggregationJob, AggregationJobState, CollectJob, CollectJobState,
+                LeaderStoredReport, ReportAggregation, ReportAggregationState,
+            },
+            test_util::ephemeral_datastore,
+        },
+        messages::TimeExt,
+        task::{self, test_util::TaskBuilder},
+    };
+    use janus_core::{
+        task::VdafInstance,
+        test_util::{
+            dummy_vdaf::{self, AggregationParam},
+            install_test_trace_subscriber,
+        },
+        time::{Clock, MockClock},
+    };
+    use janus_messages::{query_type::TimeInterval, Duration, Interval, Role, Time};
+    use rand::random;
+    use uuid::Uuid;
+
+    #[tokio::test]
+    async fn gc_task() {
+        install_test_trace_subscriber();
+
+        let clock = MockClock::default();
+        let (ds, _db_handle) = ephemeral_datastore(clock.clone()).await;
+        let ds = Arc::new(ds);
+        let vdaf = dummy_vdaf::Vdaf::new();
+
+        // Setup.
+        let task = ds
+            .run_tx(|tx| {
+                let clock = clock.clone();
+                Box::pin(async move {
+                    const REPORT_EXPIRY_AGE: Duration = Duration::from_seconds(3600);
+                    let task = TaskBuilder::new(
+                        task::QueryType::TimeInterval,
+                        VdafInstance::Fake,
+                        Role::Leader,
+                    )
+                    .with_report_expiry_age(Some(REPORT_EXPIRY_AGE))
+                    .build();
+                    tx.put_task(&task).await?;
+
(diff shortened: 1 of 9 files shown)


aggregator/src/datastore.rs
```diff
@@ -875,36 +885,91 @@ impl<C: Clock> Transaction<'_, C> {
...
+            .await?
+            .into_iter()
             .map(|row| {
-                let time = Time::from_naive_date_time(&row.get("client_timestamp"));
+                let report_id_bytes: [u8; ReportId::LEN] = row
+                    .get::<_, Vec<u8>>("report_id")
+                    .try_into()
+                    .map_err(|err| {
+                        Error::DbState(format!("couldn't convert report_id value: {err:?}"))
+                    })?;
```
tgeoghegan, line 928: nit: I think you could use `RowExt::get_bytea_and_convert` here, but you might need some more `From` or `TryFrom` implementations on `ReportId`, so it might not be worth the trouble.

aggregator/src/datastore.rs
```diff
@@ -1608,6 +1673,70 @@ impl<C: Clock> Transaction<'_, C> {
...
+                let aggregation_job_id = AggregationJobId::from(aggregation_job_id_bytes);
+
+                let report_id_bytes: [u8; ReportId::LEN] = row
+                    .get::<_, Vec<u8>>("report_id")
+                    .try_into()
+                    .map_err(|err| {
+                        Error::DbState(format!("couldn't convert report_id value: {err:?}"))
+                    })?;
+                let report_id = ReportId::from(report_id_bytes);
```
tgeoghegan, line 1726: As in `get_client_reports_for_task`, `RowExt::get_bytea_and_convert` could help here.

aggregator/src/datastore.rs
```diff
@@ -2885,6 +3070,155 @@ ORDER BY id DESC
                 .try_into()?,
         ))
     }
+
+    /// Deletes old client reports for a given task, that is, client reports whose timestamp is
+    /// older than a given timestamp which are not included in any report aggregations.
+    #[tracing::instrument(skip(self), err)]
+    pub async fn delete_old_client_reports(
```
tgeoghegan, line 3077: The schema has `report_expiry_age`, so I think we should be consistent:
```suggestion
    pub async fn delete_expired_client_reports(
```

aggregator/src/datastore.rs
```diff
@@ -2885,6 +3070,155 @@ ORDER BY id DESC
...
+                        GROUP BY report_aggregations.aggregation_job_id
+                    ) report_max_timestamps
+                        ON report_max_timestamps.aggregation_job_id = aggregation_jobs.id
+                    WHERE aggregation_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                      AND report_max_timestamps.max_timestamp < $2
+                      AND NOT EXISTS (
+                          SELECT id FROM collect_jobs
+                          WHERE aggregation_jobs.task_id = collect_jobs.task_id
+                            AND aggregation_jobs.batch_interval <@ collect_jobs.batch_interval)
+                ),
```
tgeoghegan, line 3141: Would this query be simpler if the aggregation jobs table had a column for the most recent timestamp of any report in the aggregation job? Then we would just need to check if that column's value is farther in the past than `oldest_allowed_report_timestamp`, right?

aggregator/src/datastore.rs
```diff
@@ -2885,6 +3070,155 @@ ORDER BY id DESC
...
+                    DELETE FROM report_aggregations
+                    WHERE report_aggregations.aggregation_job_id IN (
+                        SELECT id FROM aggregation_jobs_to_delete)
+                ),
+                deleted_batch_aggregations AS (
+                    DELETE FROM batch_aggregations
+                    WHERE batch_aggregations.batch_identifier IN (
+                        SELECT batch_identifier FROM aggregation_jobs
+                        WHERE aggregation_jobs.id IN (SELECT id FROM aggregation_jobs_to_delete))
+                )
```
tgeoghegan, line 3152: Neither `deleted_report_aggregations` or `deleted_batch_aggregations` are used elsewhere in the query. Would there be a downside to running these two statements as separate calls to `self.tx.prepare_cached` and `self.tx.execute` in this method, run after deleting the aggregation jobs? I think it would be a little more clear and we should get the same transaction semantics, right?

Or perhaps doing all of this in one query allows Postgres to run it more efficiently?

aggregator/src/datastore.rs
```diff
@@ -2885,6 +3070,155 @@ ORDER BY id DESC
...
+                "WITH collect_jobs_to_delete AS (
+                    SELECT id FROM collect_jobs
+                    JOIN (
+                        SELECT
+                            collect_jobs.id AS collect_job_id,
+                            MAX(client_reports.client_timestamp) AS max_timestamp
+                        FROM collect_jobs
+                        JOIN aggregation_jobs
+                            ON aggregation_jobs.task_id = collect_jobs.task_id
+                           AND aggregation_jobs.batch_interval <@ collect_jobs.batch_interval
```
tgeoghegan, line 3196: Continuing from my suggestion in `delete_old_aggregation_artifacts`, it seems like if `aggregation_jobs` had a `max_report_timestamp` column, we could avoid two JOINs here.

aggregator/src/aggregator/garbage_collector.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        // Retrieve tasks.
+        let tasks = self
+            .datastore
+            .run_tx(|tx| Box::pin(async move { tx.get_tasks().await }))
+            .await;
+        let tasks = match tasks {
+            Ok(tasks) => tasks,
+            Err(error) => return Err(error).context("couldn't retrieve tasks"),
+        };
```
tgeoghegan, line 37:
```suggestion
            .await.context("couldn't retrieve tasks")?;
```

db/schema.sql
```diff
@@ -19,6 +19,7 @@ CREATE TABLE tasks(
     vdaf                   JSON NOT NULL,                                      -- the VDAF instance in use for this task, along with its parameters
     max_batch_query_count  BIGINT NOT NULL,                                    -- the maximum number of times a given batch may be collected
     task_expiration        TIMESTAMP NOT NULL,                                 -- the time after which client reports are no longer accepted
+    report_expiry_age      BIGINT,                                             -- the maximum age of a report before it is considered expired (and acceptable for garbage collection), in seconds
```
tgeoghegan, line 22: Maybe explain here that NULL means GC is disabled for the task.

aggregator/src/aggregator/garbage_collector.rs
```diff
@@ -0,0 +1,240 @@
...
+        .await;
+        Ok(())
+    }
+
+    async fn gc_task(&self, task: Arc<Task>) -> Result<()> {
+        let oldest_allowed_report_timestamp =
+            if let Some(report_expiry_age) = task.report_expiry_age() {
+                if task.role() != &Role::Leader || task.query_type() != &QueryType::TimeInterval {
+                    return Err(anyhow!(
+                        "garbage collection is implemented only for leader, time-interval tasks"
```
tgeoghegan, line 55: I get why we can't support fixed size tasks in the 0.2 branch, but is there some reason GC is impossible to implement for the helper or is that a follow up task?

tgeoghegan (changes requested):

tgeoghegan:
A more high level concern: what's the risk here of any data (reports, report aggregations, aggregation jobs or collections) being deleted by the garbage collector while they're being worked on by another component like `aggregation_job_driver`? We can't rely on the DB's transactions for this, I think. It seems like you would either need the garbage collector to check if there is an acquired lease on aggregation jobs or collect jobs (I don't think it does, but I may have missed that part of a SQL query) before considering them and related resources eligible for GC.

aggregator/src/datastore.rs
branlwyd: Sure, that's a little cleaner & it's a pattern repeated in a few places.

aggregator/src/datastore.rs
branlwyd: I updated all instances of this pattern throughout `datastore.rs`.

Interestingly, the type parameter must be explicitly stated, even if it can be inferred. It took me a while to figure this out.

aggregator/src/datastore.rs
branlwyd: It would, but this would incur a schema change & data migration. I'd also prefer to await denormalization until we know which queries are slow.

aggregator/src/datastore.rs
branlwyd: I think the only way this is more efficient is in that it takes only one network round-trip rather than three. This is probably marginal, but I can't think of a good reason to split it out once it's written this way -- we will always want to fuse these operations together.

aggregator/src/datastore.rs
branlwyd: Same as in that comment -- this would imply a schema change/data backfill, and I'd prefer to do denormalizations once we have data showing that they will be helpful.

aggregator/src/aggregator/garbage_collector.rs
branlwyd: Yes: helper GC would imply cleaning up the `aggregation_jobs` table for helper tasks. Doing so requires knowing which aggregation jobs are covered by a `collect_jobs` row. But this isn't easy to determine, since (a) a helper `aggregation_jobs` row has no `batch_interval`/`batch_identifier`, and (b) a leader is free to create an aggregation job containing reports spanning an arbitrary interval of time, not necessarily one that fits into a given collect job interval. (Janus won't create such aggregation jobs.)

branlwyd:
> A more high level concern: what's the risk here of any data (reports, report aggregations, aggregation jobs or collections) being deleted by the garbage collector while they're being worked on by another component like `aggregation_job_driver`? We can't rely on the DB's transactions for this, I think. It seems like you would either need the garbage collector to check if there is an acquired lease on aggregation jobs or collect jobs (I don't think it does, but I may have missed that part of a SQL query) before considering them and related resources eligible for GC. And then I think you'd need the `_driver` components to refuse to acquire leases on aggregation or collect jobs that are old enough to be GC candidates.

There's no risk here: leases already check whether the leased data has changed; in this case, the lease-release logic will notice the missing row and cause the write to fail (assuming other updates don't fail due to missing rows already).

aggregator/src/datastore.rs
tgeoghegan: Oh, yes, good point about backfills.

tgeoghegan (approved):

PR #909, opened 2023-01-12
Re-enable divviup-ts tests for DAP-03.

### LINKED ISSUE #799: DAP-03: re-enable `divviup-ts` tests
https://github.com/divviup/janus/pull/797 disabled the `divviup-ts` tests, since they were broken by the difference in application info strings.

Once `divviup-ts` is updated for DAP-03, we should re-enable these tests.

branlwyd:
Closes #799.

### DIFF
--- a/integration_tests/src/client.rs
+++ b/integration_tests/src/client.rs
@@ -114,7 +114,7 @@ impl InteropClient {
         } else {
             InteropClient {
                 name: "us-west2-docker.pkg.dev/divviup-artifacts-public/divviup-ts/divviup_ts_interop_client".to_string(),
-                tag: "dap-draft-02@sha256:dc4c111adce5719fef9d231f79aecb842c3f626d51363b04355e987d6a40aadc".to_string(),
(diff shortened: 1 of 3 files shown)


integration_tests/src/client.rs
```diff
@@ -114,7 +114,7 @@ impl InteropClient {
         } else {
             InteropClient {
                 name: "us-west2-docker.pkg.dev/divviup-artifacts-public/divviup-ts/divviup_ts_interop_client".to_string(),
-                tag: "dap-draft-02@sha256:dc4c111adce5719fef9d231f79aecb842c3f626d51363b04355e987d6a40aadc".to_string(),
+                tag: "dap-draft-03@sha256:b03102a4e8eb9e1abf879940aa89d6a04e4a8229a215c715a4c92a0aec4b3be7".to_string(),
```
divergentdave, line 117: This image also appears in `.github/workflows/ci-build.yml`, and should be updated there. (We specify this default both here, for developer use, and in the workflow, because we wanted to be able to override the image when using a `workflow_dispatch` trigger)

tgeoghegan (approved):
LGTM (assuming tests pass)

PR #911, opened 2023-01-14
store helper's `A::PrepareShare` in the database

tgeoghegan:
DAP-04 introduces protocol affordances for recovering from round skew. To that end, the helper needs to store the prepare shares from its current round so that it can service replayed aggregation job continuation requests from the leader.

---

I'm working on the changes needed to implement DAP-04 in https://github.com/divviup/janus/pull/731. Since DAP-04 isn't ready yet and in order to reduce the size of the eventual PR wiring up the new HTTP API, I'm teasing out some independent pieces into more digestible PRs.

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -136,7 +136,9 @@ CREATE TABLE report_aggregations(
     ord                 BIGINT NOT NULL,                    -- a value used to specify the ordering of client reports in the aggregation job
     state               REPORT_AGGREGATION_STATE NOT NULL,  -- the current state of this report aggregation
     prep_state          BYTEA,                              -- the current preparation state (opaque VDAF message, only if in state WAITING)
-    prep_msg            BYTEA,                              -- the next preparation message to be sent to the helper (opaque VDAF message, only if in state WAITING if this aggregator is the leader)
+    prep_msg            BYTEA,                              -- for the leader, the next preparation message to be sent to the helper (opaque VDAF message)
+                                                            -- for the helper, the next preparation share to be sent to the leader (opaque VDAF message)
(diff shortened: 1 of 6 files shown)


branlwyd (changes requested):
cjpatton requested that the recovery functionality be pulled out to a different spec PR; due to that, and the fact that we don't need this functionality until DAP-04, let's wait to merge.

tgeoghegan:
Downgrading to draft while we sort out the corresponding protocol changes.

tgeoghegan:
I've resumed working on #994, so it's time to dust off this change

branlwyd (approved):

db/schema.sql
```diff
@@ -136,7 +136,8 @@ CREATE TABLE report_aggregations(
     ord                 BIGINT NOT NULL,                    -- a value used to specify the ordering of client reports in the aggregation job
     state               REPORT_AGGREGATION_STATE NOT NULL,  -- the current state of this report aggregation
     prep_state          BYTEA,                              -- the current preparation state (opaque VDAF message, only if in state WAITING)
-    prep_msg            BYTEA,                              -- the next preparation message to be sent to the helper (opaque VDAF message, only if in state WAITING if this aggregator is the leader)
+    prep_msg            BYTEA,                              -- for the leader, the next preparation message to be sent to the helper (opaque VDAF message)
+                                                            -- for the helper, the next preparation share to be sent to the leader (opaque VDAF message)
```
divergentdave, line 140: Could we spell out in which states this should be non-NULL? (is it when in the `WAITING` state for both leader and helper cases?)

divergentdave (approved):

PR #918, opened 2023-01-17
Miscellaneous query type-related cleanups

divergentdave:
I reviewed the codebase to check that we have parity between the two query types, and I came up with three small cleanups to make.

- `/aggregate` was broken for fixed size queries with the dummy VDAF.
- `test_task_builders()` can be slightly simplified now that `TaskBuilder::new()` takes the query type.
- `dap_decode` would have output an extra `Ok(...)` around one particular message type.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1056,7 +1056,7 @@ impl VdafOps {
 
             #[cfg(feature = "test-util")]
             (task::QueryType::FixedSize { .. }, VdafOps::Fake(vdaf)) => {
-                Self::handle_aggregate_init_generic::<0, TimeInterval, dummy_vdaf::Vdaf, _>(
+                Self::handle_aggregate_init_generic::<0, FixedSize, dummy_vdaf::Vdaf, _>(
                     datastore,
                     vdaf,
                     aggregate_step_failure_counter,
--- a/integration_tests/tests/common/mod.rs
+++ b/integration_tests/tests/common/mod.rs
@@ -29,12 +29,11 @@ pub fn test_task_builders(
 ) -> (HpkePrivateKey, TaskBuilder, TaskBuilder) {
     let endpoint_random_value = hex::encode(random::<[u8; 4]>());
     let collector_keypair = generate_test_hpke_config_and_private_key();
-    let leader_task = TaskBuilder::new(QueryType::TimeInterval, vdaf, Role::Leader)
+    let leader_task = TaskBuilder::new(query_type, vdaf, Role::Leader)
         .with_aggregator_endpoints(Vec::from([
             Url::parse(&format!("http://leader-{endpoint_random_value}:8080/")).unwrap(),
             Url::parse(&format!("http://helper-{endpoint_random_value}:8080/")).unwrap(),
         ]))
-        .with_query_type(query_type)
         .with_min_batch_size(46)
         .with_collector_hpke_config(collector_keypair.config().clone());
     let helper_task = leader_task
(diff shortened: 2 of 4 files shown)


messages/src/bin/dap_decode.rs
```diff
@@ -58,7 +58,7 @@ fn decode_dap_message(message_file: &str, media_type: &MediaType) -> Result<Box<
             if let Ok(decoded) = AggregateShareReq::<TimeInterval>::get_decoded(&message_buf) {
                 Box::new(decoded) as Box<dyn Debug>
             } else {
-                Box::new(AggregateShareReq::<FixedSize>::get_decoded(&message_buf))
+                Box::new(AggregateShareReq::<FixedSize>::get_decoded(&message_buf)?)
                     as Box<dyn Debug>
```
tgeoghegan, line 62: I don't think this is right. It works because `Result<AggregateShareReq::<FixedSize>, Error>` must implement `Debug`, but we want to apply `?` here so that we can yield an obvious error should decoding the message fail.

integration_tests/tests/common/mod.rs
```diff
@@ -29,12 +29,11 @@ pub fn test_task_builders(
 ) -> (HpkePrivateKey, TaskBuilder, TaskBuilder) {
     let endpoint_random_value = hex::encode(random::<[u8; 4]>());
     let collector_keypair = generate_test_hpke_config_and_private_key();
-    let leader_task = TaskBuilder::new(QueryType::TimeInterval, vdaf, Role::Leader)
+    let leader_task = TaskBuilder::new(query_type, vdaf, Role::Leader)
         .with_aggregator_endpoints(Vec::from([
             Url::parse(&format!("http://leader-{endpoint_random_value}:8080/")).unwrap(),
             Url::parse(&format!("http://helper-{endpoint_random_value}:8080/")).unwrap(),
         ]))
-        .with_query_type(query_type)
```
tgeoghegan, line 37: It seems like we should either have an argument in `TaskBuilder::new` _or_ a `with_whatever` method, but not both.

tgeoghegan (changes requested):

messages/src/bin/dap_decode.rs
divergentdave: My patch is adding the `?` operator, so that this branch matches the `AggregateInitializeReq` branch above. However, I think I'm going to rewrite this with let bindings and ascribed types, before everything gets turned into trait objects, so that it's clearer and we can avoid similar problems with future changes.

integration_tests/tests/common/mod.rs
branlwyd: The idea is that `new` takes required parameters for which there is no sensible default. Everything else gets a sensible default and a `with_foo` method to allow the default to be overridden. Query type is more the former, so it goes into the `new` "constructor" method.

aggregator/src/aggregator.rs
```diff
@@ -1056,7 +1056,7 @@ impl VdafOps {
 
             #[cfg(feature = "test-util")]
             (task::QueryType::FixedSize { .. }, VdafOps::Fake(vdaf)) => {
-                Self::handle_aggregate_init_generic::<0, TimeInterval, dummy_vdaf::Vdaf, _>(
+                Self::handle_aggregate_init_generic::<0, FixedSize, dummy_vdaf::Vdaf, _>(
```
branlwyd, line 1059: Good catch...

branlwyd (approved):

messages/src/bin/dap_decode.rs
tgeoghegan: I'm sorry, I read the diff exactly backward. And I put this bug in the program to begin with.

tgeoghegan (approved):

PR #920, opened 2023-01-19
Reject uploaded reports past expiry age (0.2)

divergentdave:
This is a follow-up to #894, changing `/upload` to reject reports with `...:reportTooLate` if the given task has a report expiry age configured, and the report is older than that age. We need to check for this because otherwise an attacker could re-upload the same report multiple times once it's old enough to be garbage collected, and race collections against the garbage collector to get around anti-replay protections.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1297,6 +1297,18 @@ impl VdafOps {
             ));
         }
 
+        // Reject reports that would be eligible for garbage collection, to prevent replay attacks.
+        if let Some(report_expiry_age) = task.report_expiry_age() {
+            let report_expiry_time = report.metadata().time().add(report_expiry_age)?;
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):

aggregator/src/aggregator.rs
```diff
@@ -1297,6 +1297,18 @@ impl VdafOps {
             ));
         }
 
+        // Reject reports that would be eligible for garbage collection, to prevent replay attacks.
+        if let Some(report_expiry_age) = task.report_expiry_age() {
+            let report_expiry_time = report.metadata().time().add(report_expiry_age)?;
+            if clock.now().is_after(&report_expiry_time) {
```
branlwyd, line 1303: We should ideally sample `clock.now()` once, to avoid weird "impossible" bugs around things right at the edge.

branlwyd (approved):

PR #936, opened 2023-01-25
Implement DAP-04 resource oriented API

tgeoghegan:
Implements the new API, as specified in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/398.

This change is only for DAP-04, so we will want to branch for `release/0.3` / DAP-03 before taking this.


aggregator/src/aggregator.rs
```diff
@@ -3687,6 +3700,23 @@ where
...
+    filter: F,
+    cors: Cors,
+    response_time_histogram: Histogram<f64>,
+    name: &'static str,
+) -> BoxedFilter<(impl Reply,)>
+where
+    F: Filter<Extract = (Result<T, Error>,), Error = Rejection> + Clone + Send + Sync + 'static,
+    T: Reply + 'static,
+{
+    filter
```
tgeoghegan, line 3713: I'm not sure how to proceed here. The existing `compose_common_wrappers` is very carefully designed to only apply the response filter (and especially the response time histogram) _after_ the routing filter has been applied. But our routing filter now extracts some values, so we can't do that quite as easily. I'm getting nowhere trying to convince warp to make this work, and I'm wondering if it's worth the effort if we want to move to Trillium anyway.

tgeoghegan:
Question for reviewers: this change collapses `AggregateInitResp`, `AggregateContinueReq` and `AggregateContinueResp` into a single `AggregationJob` message, which is what landed in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/398. However, https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/400 brings back `AggregationJobContinueReq` to avoid unnecessarily transmitting a `round` field in certain messages. This PR and the subsequent one that syncs Janus with DAP PR #400 would be smaller if I jump ahead to #400's message definitions, which I think makes the review easier, at the cost of Janus PRs no longer neatly mapping to DAP PRs. Which way would you prefer to review?

aggregator/src/aggregator.rs
divergentdave: Hard approach: We could partly get around this if we rewrite our higher order wrappers in terms of `warp::generic::Func` instead of `Filter`, so we can glue things together with `.then()` instead of `.and()`, which means we could line up the routing filter's `Extract` type with the response `Func`'s argument. However, that would then prevent us from using filters in the response half, which is a huge impediment. Maybe the only thing that would work is a higher order combiner that takes a routing filter, a responding filter for extracting additional data, and a responding `Func` that takes a big tuple as an argument, combining both the routing filter's and the responding filter's `Extract` types. We're once again running up against the limitations of the sealed `Filter` trait.

The `http` crate has a feature called "extensions", which lets you store extra data hanging off of a `Request` and/or `Response`, indexed by its type. Warp does have a filter to fetch items out of `Extensions`, but it appears there's no filter to set them, and no way to directly access the `Request` to store them, so that's a dead end.

Easier approach: We could resolve the double-counting in metrics and tracing spans by duplicating our routing filters. The first copy would just match URL patterns, but not extract anything, and be used on the outside of the wrapping filters. (we might have to use `warp::filters::path::peek` for this) The second copy would be integrated with the rest of the responding filter, and work as usual.

tgeoghegan:
> I'm fine with combinnig 398+400 into one PR

The implementation of _all_ of 400 will be in at least one PR beyond this one no matter what, because the round skew recovery stuff requires storage changes and some subtle state tracking. But I want to minimize the PRs that change message definitions because those tend to be disproportionately noisy.

aggregator/src/aggregator.rs
tgeoghegan: I think `warp::path::peek` can work, at the cost of some awkward path matching code we have to write. Thanks for the suggestions!

tgeoghegan:
The integration tests pass when run against a `divviup-ts` with https://github.com/divviup/divviup-ts/pull/199, so we should be able to avoid disabling the `divviup-ts` integration tests if we land divviup-ts/#199 and ship an interop container from it.

- [ ] configure divviup-ts tests to use a dap-draft-04 container and reenable them

db/schema.sql
```diff
@@ -172,7 +172,7 @@ CREATE TYPE COLLECT_JOB_STATE AS ENUM(
 -- The leader's view of collect requests from the Collector.
 CREATE TABLE collect_jobs(
```
branlwyd, line 173: Since the standard now refers to this concept as a "collection job" rather than a "collect job", I think we should make the corresponding rename in our codebase. That is, I think this table should be renamed to "collection_jobs" & all variables named something like "collect_job_id" should be renamed to something like "collection_job_id".

(This sounds like a lot of work, and it will impact a bunch of lines, but I think it's the equivalent of doing `s/collect_job/collection_job/g` across the codebase.)

messages/src/lib.rs
```diff
@@ -2353,16 +2320,7 @@ impl Decode for AggregateShareResp {
...
-        query_type::{self, FixedSize, TimeInterval},
-        AggregateContinueReq, AggregateContinueResp, AggregateInitializeReq,
-        AggregateInitializeResp, AggregateShareAad, AggregateShareReq, AggregateShareResp,
-        AggregationJobId, BatchId, BatchSelector, CollectReq, CollectResp, Duration, Extension,
-        ExtensionType, FixedSizeQuery, HpkeAeadId, HpkeCiphertext, HpkeConfig, HpkeConfigId,
-        HpkeKdfId, HpkeKemId, HpkePublicKey, InputShareAad, Interval, PartialBatchSelector,
-        PlaintextInputShare, PrepareStep, PrepareStepResult, Query, Report, ReportId,
-        ReportIdChecksum, ReportMetadata, ReportShare, ReportShareError, Role, TaskId, Time,
-    };
+    use super::*;
```
branlwyd, line 2323: Please don't import `*` from a module, for the same reason that one shouldn't do that in e.g. Java or other languages with similar import mechanisms -- doing so makes fixing naming conflicts unnecessarily painful, and can make things more confusing to folks trying to read the code (is type `T` from _this_ star-import or _that_ star-import?).

branlwyd, line 2323: nit: many of the test names in this module need to be updated to match the renamed type they are roundtripping.

collector/src/lib.rs
```diff
@@ -703,19 +693,25 @@ mod tests {
         random()
     }
 
+    fn collection_uri_regex_matcher(task_id: &TaskId) -> Matcher {
```
branlwyd, line 696: It appears this function is only used in tests -- IMO move it into the `tests` module or otherwise mark it `#[cfg(test)]`.

aggregator/src/aggregator.rs
```diff
@@ -2524,19 +2536,26 @@ impl VdafOps {
...
+                        {
+                            debug!(
+                                collection_id = %collection_id,
+                                collect_request = ?req,
+                                "Collect job already exists"
+                            );
+                            return Ok(());
+                        } else {
+                            return Err(datastore::Error::User(
+                                Error::UnrecognizedCollectJob(collection_id).into(),
```
branlwyd, line 2553: I'm not sure `UnrecognizedCollectJob` is appropriate here -- IIUC this error case correctly, the collect job exists, the caller is attempting to modify/assert a different state for the existing collect job. That is, this codepath identifies an idempotency problem.

We could introduce a new error type, which is reflected back to the caller as an HTTP 409 Conflict for now.

branlwyd (changes requested):

collector/src/lib.rs
tgeoghegan: It's already in `mod tests` and so should not be defined outside `#[cfg(test)]`.

aggregator/src/aggregator.rs
tgeoghegan: Good catch. I wrote up https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/406 to discuss this in DAP, but HTTP 409 sounds good for now.

tgeoghegan: So, it's easy enough to define the new error type and return it from this point in the collection request handler. However, I noticed that we already had this problem when creating aggregation jobs and we don't handle it very gracefully: if we try to PUT the same aggregation job ID twice, it fails with a postgres error about violating a uniqueness constraint, and that materializes as an HTTP 500. I don't think that's hard to fix, but this PR is already quite large, so I think I'd like to deal with adding tests and bugfixes for idempotence of PUTting collection jobs and aggregation jobs in a subsequent PR.

db/schema.sql
tgeoghegan: I gave this a try, let's see what breaks! I think changing the container names might cause some minor havoc.

tgeoghegan:
I'm a little surprised, but just replacing "collect job", "collect_job" and "CollectJob" with their -ion equivalents did work. One thing we should discuss is whether we want to rename the `collect_job_driver` containers to `collection_job_driver`. We'd have to make a corresponding change in `janus-ops` but that's hardly difficult.

divergentdave:
I'm okay with changing the container name, in fact I'd prefer that we keep the binary name and container name in sync. We currently use the container name to guess RUST_LOG targets, so keeping them lined up means we won't need any special-case handling.

aggregator/src/aggregator.rs
```diff
@@ -3692,6 +3720,33 @@ where
 /// See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age
 const CORS_PREFLIGHT_CACHE_AGE: u32 = 24 * 60 * 60;
 
+/// Constructs a Warp filter that matches the request path for tasks, as well as whatever regex is
+/// provided, but without consuming the path or any values encoded into it, such that later filters
+/// can still extract path components.
+fn path_matching_filter(trailing_fragment_regex: &'static str) -> BoxedFilter<()> {
+    warp::path::peek()
+        .and_then(move |peek: warp::path::Peek| async move {
+            let regex = Regex::new(&format!(
```
divergentdave, line 3729: I'd recommend that we compile this regex at filter construction time, and return `Result<BoxedFilter<()>, regex::Error>` out instead of rejecting requests, so that we can catch mistakes in regexes early. Then, the compiled regex can be moved into the closure. (Also, I think we can swap `.and_then()` for `.map()`, since the async block here doesn't await anything)

aggregator/src/aggregator.rs
```diff
@@ -3734,150 +3789,204 @@ pub fn aggregator_filter<C: Clock>(
...
+                    )
+                    .map_err(|err| Error::Internal(format!("couldn't produce response: {err}")))
             },
         );
-    let aggregate_endpoint = compose_common_wrappers(
-        aggregate_routing,
-        aggregate_responding,
+
+    let aggregation_jobs_put_endpoint = compose_common_wrappers(
+        path_matching_filter(AGGREGATION_JOB_PATH_REGEX),
```
divergentdave, line 3849: We should chain another filter (`warp::filters::method::put()`, etc.) on any cases where we have identical `path_matching_filter(...)` calls. Checking the method in these routing filters will keep us from double-counting DELETE requests in both the PUT and DELETE metrics.

divergentdave (changes requested):

aggregator/src/aggregator.rs
tgeoghegan: - [ ] add test proving that PUT collection job is idempotent
- [ ] add test proving that mutating a collection job with PUT fails with expected error
- [ ] add test proving that PUT aggregation job is idempotent
- [ ] add test proving that mutating an aggregation job with PUT fails with expected error
- [ ] add test proving that POST aggregation share is idempotent

...or file an issue to do that in a subsequent PR, depending on consensus we reach in this thread

messages/src/lib.rs
```diff
@@ -1389,18 +1390,76 @@ impl<Q: QueryType> Decode for PartialBatchSelector<Q> {
     }
 }
 
+/// DAP protocol message representing an identifier for a collection.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct CollectionId([u8; Self::LEN]);
+
+impl CollectionId {
```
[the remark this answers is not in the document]
tgeoghegan: Yes, also for consistency with DAP: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/main/draft-ietf-ppm-dap.md#collection-job-initialization-collect-init

aggregator/src/aggregator.rs
tgeoghegan: Good point about the regex. I used `and_then` because it's obvious how to return a rejection from it. It may be possible to do this with `Filter::map` but I am disinclined to spend further time learning `warp`'s idiosyncrasies since we plan to migrate off of it anyway.

aggregator/src/aggregator.rs
tgeoghegan: I believe we do: every `_responding` filter starts with a method filter and gets `and()`ed with the path matching filter in `compose_common_wrappers`.

aggregator/src/aggregator.rs
divergentdave: It might work if the closure returns `Result<(), Rejection>`, but I did not test this.

aggregator/src/aggregator.rs
divergentdave: We need to move (or copy) the method filter from the responding filter to the router filter, or else requests will enter and leave wrapping filters they shouldn't, only to get a rejection from the responding filter.

aggregator/src/aggregator.rs
tgeoghegan: Filed #998 for the items above.

aggregator/src/task.rs
```diff
@@ -342,6 +343,57 @@ impl Task {
         let secret_bytes = self.vdaf_verify_keys.first().unwrap();
         VerifyKey::try_from(secret_bytes).map_err(|_| Error::AggregatorVerifyKeySize)
     }
+
+    /// Returns the relative path for tasks, relative to which other API endpoints are defined.
+    fn tasks_path(&self) -> String {
+        format!("/tasks/{}", self.id())
```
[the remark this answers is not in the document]
tgeoghegan: Good catch! The nuance here is that we _don't_ want a leading slash when using `Url::join` but we _do_ want one to pass a path into `mockito::mock`, so we need a bit more scaffolding on `struct Task` to distinguish the cases (which actually should help me tighten up the visibility of test-only methods).

aggregator/src/aggregator.rs
tgeoghegan: Doing this broke the CORS stuff because if you put the method in the routing filter, you can send OPTIONS anymore. I'll have to pick this up on Monday. It's remarkably hard to get right.

aggregator/src/aggregator.rs
```diff
@@ -2524,95 +2543,107 @@ impl VdafOps {
...
-                            &aggregation_param,
-                        )
-                        .await?
+                    // Check if this collection job already exists, ensuring that all parameters match.
+                    if let Some(collection_job) =
+                        tx.get_collection_job::<L, Q, A>(&collection_id).await?
                     {
-                        debug!(collect_request = ?req, "Serving existing collect job UUID");
-                        return Ok(collect_job_id);
+                        if collection_job.batch_identifier() == &batch_identifier
```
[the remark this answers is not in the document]
branlwyd: If the checks are different enough (and it sounds like they are), this might be a good place to introduce another method on `CollectableQueryType` which performs this validation.

(also, I believe this isn't caused by this PR; maybe it's better to write up an issue & address in a later PR since this PR is already huge?)

aggregator/src/datastore.rs
```diff
@@ -4217,11 +4176,11 @@ pub mod models {
         }
     }
 
-    /// AcquiredCollectJob represents an incomplete collect job whose lease has been acquired.
+    /// AcquiredCollectJob represents an incomplete collection job whose lease has been acquired.
     #[derive(Clone, Debug, PartialOrd, Ord, Eq, PartialEq)]
     pub struct AcquiredCollectJob {
```
branlwyd, line 4181: nit: `AcquiredCollectJob` ->  `AcquiredCollectionJob`?

(and similarly below, `CollectJob`, `CollectJobState`, & perhaps others should be updated. I think `s/CollectJob/CollectionJob/g` would be a good change to make, too.)

db/schema.sql
```diff
@@ -163,37 +163,37 @@ CREATE TABLE batch_aggregations(
...
 );
 
--- Specifies the possible state of a collect job.
-CREATE TYPE COLLECT_JOB_STATE AS ENUM(
-    'START',     -- the aggregator is waiting to run this collect job
-    'FINISHED',  -- this collect job has run successfully and is ready for collection
-    'ABANDONED', -- this collect job has been abandoned & will never be run again
-    'DELETED'    -- this collect job has been deleted
+-- Specifies the possible state of a collection job.
+CREATE TYPE collection_job_STATE AS ENUM(
```
branlwyd, line 167: nit: capitalization

db/schema.sql
```diff
@@ -163,37 +163,37 @@ CREATE TABLE batch_aggregations(
...
+    'ABANDONED', -- this collection job has been abandoned & will never be run again
+    'DELETED'    -- this collection job has been deleted
 );
 
 -- The leader's view of collect requests from the Collector.
-CREATE TABLE collect_jobs(
+CREATE TABLE collection_jobs(
     id                      BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,  -- artificial ID, internal-only
-    collect_job_id          UUID NOT NULL,               -- UUID used by collector to refer to this job
+    collection_job_id          BYTEA NOT NULL,              -- 16 byte identifier used by collector to refer to this job
```
branlwyd, line 177: nit: fixup spacing on comments

interop_binaries/src/bin/janus_interop_collector.rs
```diff
@@ -675,40 +675,39 @@ fn make_filter() -> anyhow::Result<impl Filter<Extract = (Response,)> + Clone> {
             }
         }
     });
-    let collect_start_filter =
-        warp::path!("collect_start").and(warp::body::json()).then({
+    let collect_start_filter = warp::path!("collect_start").and(warp::body::json()).then({
```
branlwyd, line 678: Q: should these be renamed? (to e.g. `collection_start`)

Doesn't need to happen in this PR, could be included as part of test API updates for DAP-04.

messages/src/lib.rs
```diff
@@ -1389,18 +1390,76 @@ impl<Q: QueryType> Decode for PartialBatchSelector<Q> {
...
+
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+        Ok(Self(value.try_into().map_err(|_| {
+            "byte slice has incorrect length for CollectionId"
+        })?))
+    }
+}
+
+impl FromStr for CollectionJobId {
+    type Err = Box<dyn Debug>;
```
branlwyd, line 1419: WDYT about using `anyhow::Error` as our "don't care about the type" error? It's about as expressive as a `Box<dyn Debug>`, we don't need to fiddle with boxing (ourselves), and it allows wrapping (so e.g. we could wrap each error condition with a helpful human-readable error message, if we wanted to).

branlwyd (approved):
(approving, as remaining review comments are trivial, I think. feel free to send back for another round if the `anyhow::Error` suggestion turns out to be larger than expected)

interop_binaries/src/bin/janus_interop_collector.rs
tgeoghegan: I think it's reasonable for this to be the verb to collect rather than the noun collection. If anyone disagrees the place to start is a change to the interop test spec.

messages/src/lib.rs
tgeoghegan: That would indeed work, but it brings its own awkwardness (`anyhow::Context::context` refuses to wrap `&str` so we'd have to mess with more error types) and it's not clear to me that it's sufficiently better than `Box<dyn Debug>` to justify the extra gymnastics.

aggregator/src/aggregator.rs
divergentdave: Looking at things closer, I think I now see why the histograms don't get double-counted. (the relevant `map()` that gets added after the wrapped filter is skipped if there's a rejection, so it actually can't double count) I think the only downside left is spurious tracing spans, and we can live with that in the meantime.

divergentdave (approved):

PR #937, opened 2023-01-26
Add a name to transactions for metrics

divergentdave:
This adds a name parameter to `Datastore::run_tx()`, and includes the transaction name in the `janus_database_transactions_total` metric. Here are the transaction names I used, grouped by component.

- aggregator
  - task_aggregator_get_task
  - upload
  - aggregate_init
  - aggregate_continue
  - collect
  - aggregate_share
  - get_collect_job_get_task
  - get_collect_job
  - delete_collect_job_get_task
  - delete_collect_job
- aggregation job creator
  - aggregation_job_creator_get_tasks
  - aggregation_job_creator_time_with_param
  - aggregation_job_creator_fixed_no_param
  - aggregation_job_creator_time_no_param
- aggregation_job_driver
  - step_aggregation_job_2
  - step_aggregation_job_1
  - cancel_aggregation_job
  - acquire_aggregation_jobs
- collect_job_driver
  - step_collect_job_2
  - step_collect_job_1
  - abandon_collect_job
  - acquire_collect_jobs

Future work: this would enable adding a histogram metric to record the elapsed time in transactions. (without a transaction name label, heterogeneity would make such a histogram hard to interpret)

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -485,7 +485,9 @@ impl<C: Clock> Aggregator<C> {
     ) -> Result<Option<Vec<u8>>, Error> {
         let task_id = self
             .datastore
-            .run_tx(|tx| Box::pin(async move { tx.get_collect_job_task_id(&collect_job_id).await }))
+            .run_tx("get_collect_job_get_task", |tx| {
+                Box::pin(async move { tx.get_collect_job_task_id(&collect_job_id).await })
+            })
             .await?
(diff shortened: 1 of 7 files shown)


tgeoghegan (approved):
I'm onboard with this strategy for metrics, but I wish we didn't have to update every single call to `run_tx` in every test. What if we add a new `Datastore::run_tx_with_name`, and then have the existing `Datastore::run_tx` call it with `name: "default"`?

branlwyd (approved):

PR #945, opened 2023-01-27
Remove janus_message time types from job drivers

divergentdave:
This changes job driver delays, lease durations, and lease expirations to use `std::time::Duration` and `chrono::NaiveDateTime` instead.

This is part of #445.

### DIFF
--- a/aggregator/src/datastore.rs
+++ b/aggregator/src/datastore.rs
@@ -15,6 +15,7 @@ use crate::{
     SecretBytes,
 };
 use anyhow::anyhow;
+use chrono::NaiveDateTime;
 use futures::future::join_all;
 use janus_core::{
     hpke::{HpkeKeypair, HpkePrivateKey},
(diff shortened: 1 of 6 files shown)


aggregator/src/datastore.rs
```diff
@@ -2959,6 +2967,18 @@ fn check_single_row_mutation(row_count: u64) -> Result<(), Error> {
...
+fn add_naive_date_time_duration(
+    time: &NaiveDateTime,
+    duration: &StdDuration,
+) -> Result<NaiveDateTime, Error> {
+    time.checked_add_signed(
+        chrono::Duration::from_std(*duration)
+            .map_err(|_| Error::TimeOverflow("overflow converting duration to signed duration"))?,
+    )
+    .ok_or(Error::TimeOverflow("overflow adding duration to time"))
+}
```
tgeoghegan, line 2980: I can't think of a better place to put this, especially since it's not currently used outside of `datastore.rs`, but I don't love adding unrelated functionality to the already too-large file. However I don't think we can do any better until `chrono` adds better support for `std::time::Duration`. Ah well.

tgeoghegan (approved):

PR #961, opened 2023-01-30
Implement garbage collection.

### LINKED ISSUE #313: Garbage-collection of datastore data
At time of writing, there is no garbage-collection process which deletes old data in a Janus datastore that is no longer useful. This means that, given a constant level of data ingestion, our data storage will grow linearly over time.

We should devise and implement a policy for deletion of data once it is no longer useful. This should cover all existing tables (or, more generally, all data stored by the Janus system). It will be important to make sure we remain DAP-compliant; to my knowledge the DAP spec is well-designed to allow this kind of cleanup, with one open issue that must be resolved (collect results currently `MUST` be stored until manual deletion by the collector).

branlwyd:
This PR is huge, but is mostly straightforward. Over-and-above the 0.2 version of this functionality, this PR includes:

* Support for fixed-size tasks. This is relatively straightforward--no new data needs to be stored, only the relevant expiry SQL queries needed to be updated.
* Support for helper tasks. This is trickier: to support time-interval helper tasks, we need to compute & store the interval of contained reports, as otherwise we do not have a fast (indexed) way to associate collection artifacts with the relevant aggregation jobs since time-interval helper tasks do not store the interval of the aggregation job (nor could they, as this information is not transmitted from the leader). However, we have to be careful to only use the client timestamp interval for time-interval tasks, as fixed-size tasks do not correlate collection artifacts to aggregation jobs in this way.

Please review the tests carefully; I made them as realistic as possible given that this functionality involves the deletion of data. The `garbage_collector.rs` tests are higher-level tests that the individual datastore operations interact as expected; the `datastore.rs` tests verify the various corner cases of the individual datastore operations.

Closes #313.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -21,7 +21,7 @@ use crate::{
         },
         Datastore,
     },
-    messages::TimeExt,
+    messages::{DurationExt, IntervalExt, TimeExt},
     task::{self, Task, VerifyKey, PRIO3_AES128_VERIFY_KEY_LENGTH},
     try_join,
 };
@@ -43,9 +43,9 @@ use janus_messages::{
     query_type::{FixedSize, TimeInterval},
     AggregateContinueReq, AggregateContinueResp, AggregateInitializeReq, AggregateInitializeResp,
     AggregateShareAad, AggregateShareReq, AggregateShareResp, AggregationJobId, BatchSelector,
-    CollectReq, CollectResp, HpkeCiphertext, HpkeConfigId, HpkeConfigList, InputShareAad, Interval,
-    PartialBatchSelector, PlaintextInputShare, PrepareStep, PrepareStepResult, Report, ReportId,
-    ReportIdChecksum, ReportShare, ReportShareError, Role, TaskId, Time,
+    CollectReq, CollectResp, Duration, HpkeCiphertext, HpkeConfigId, HpkeConfigList, InputShareAad,
+    Interval, PartialBatchSelector, PlaintextInputShare, PrepareStep, PrepareStepResult, Report,
+    ReportId, ReportIdChecksum, ReportShare, ReportShareError, Role, TaskId, Time,
 };
 use opentelemetry::{
     metrics::{Counter, Histogram, Meter, Unit},
@@ -176,6 +176,12 @@ pub enum Error {
         problem_details: Box<HttpApiProblem>,
         dap_problem_type: Option<DapProblemType>,
     },
+    /// An aggregate share request was rejected.
+    #[error("task {0}: {1}")]
+    AggregateShareRequestRejected(TaskId, String),
+    /// An empty aggregation (no report shares) was attempted.
+    #[error("task {0}: empty aggregation")]
+    EmptyAggregation(TaskId),
     /// An error representing a generic internal aggregation error; intended for "impossible"
     /// conditions.
     #[error("internal aggregator error: {0}")]
@@ -211,6 +217,8 @@ impl Error {
             Error::TaskParameters(_) => "task_parameters",
             Error::HttpClient(_) => "http_client",
             Error::Http { .. } => "http",
+            Error::AggregateShareRequestRejected(_, _) => "aggregate_share_request_rejected",
+            Error::EmptyAggregation(_) => "empty_aggregation",
             Error::Internal(_) => "internal",
         }
     }
@@ -565,7 +573,7 @@ impl<C: Clock> Aggregator<C> {
         }
 
         let resp = task_aggregator
-            .handle_aggregate_share(&self.datastore, req_bytes)
+            .handle_aggregate_share(&self.datastore, &self.clock, req_bytes)
             .await?;
         Ok(resp.get_encoded())
     }
@@ -778,10 +786,11 @@ impl TaskAggregator {
     async fn handle_aggregate_share<C: Clock>(
         &self,
         datastore: &Datastore<C>,
+        clock: &C,
         req_bytes: &[u8],
     ) -> Result<AggregateShareResp, Error> {
         self.vdaf_ops
-            .handle_aggregate_share(datastore, Arc::clone(&self.task), req_bytes)
+            .handle_aggregate_share(datastore, clock, Arc::clone(&self.task), req_bytes)
             .await
     }
 }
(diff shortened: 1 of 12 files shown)


aggregator/src/aggregator.rs
```diff
@@ -3568,6 +3627,16 @@ where
                     Err(Error::Message(_)) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
                     Err(Error::HttpClient(_)) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
                     Err(Error::Http { .. }) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
+                    Err(Error::AggregateShareRequestRejected(task_id, _)) => {
+                        build_problem_details_response(
+                            DapProblemType::UnrecognizedMessage,
```
branlwyd, line 3632: Q. None of the existing DAP problem types really fit (I somewhat-arbitrarily chose `unrecognizedMessage`), but I think it's still useful to return a problem details response here. Thoughts on defining a custom error type here?

tgeoghegan: We generally use `unrecognizedMessage` for problems parsing messages, whereas this case is a well-encoded message that contains a value we didn't like.

If the leader encounters this error, we don't expect it to do anything but give up on the aggregate share request, right? If so, then nothing more specific than HTTP 400 is needed. Maybe what we should do is change DAP to generalize its `reportRejected` error type into a broader `requestRejected` error type, and eventually use that. In the meamtime, I think `StatusCode::CLIENT_ERROR.into_response()` should suffice here (DAP-03 allows not providing an error document; https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-03#section-3.2-2)

tgeoghegan (commented):
Posting a first set of comments. I'd like to take another pass over the tests though.

branlwyd (commented):
(I'm pretty sure the CI failure is spurious, as it's related to the build only, I can't reproduce it locally, and the PR did not touch the build -- sending back for another round of review, but I'm actively investigating the CI issue.)

tgeoghegan (changes requested):

aggregator/src/datastore.rs
```diff
@@ -2955,22 +3217,229 @@ ORDER BY id DESC
...
+                    DELETE FROM batch_aggregations
+                    WHERE
+                        (
+                            batch_aggregations.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                            AND UPPER(batch_aggregations.batch_interval) < $2
+                            AND NOT EXISTS (
+                                SELECT id FROM collect_jobs
+                                WHERE batch_aggregations.task_id = collect_jobs.task_id
+                                  AND (batch_aggregations.batch_identifier = batch_aggregations.batch_identifier
+                                    OR batch_aggregations.batch_interval <@ batch_aggregations.batch_interval))
```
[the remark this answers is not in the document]
tgeoghegan: Oh, looks like you got to this before I posted the review. Still, there has to be a missing test for this case.

aggregator/src/datastore.rs
```diff
@@ -8734,6 +9279,1440 @@ mod tests {
...
+                                OLDEST_ALLOWED_REPORT_TIMESTAMP
+                                    .sub(&Duration::from_seconds(9))
+                                    .unwrap(),
+                                OLDEST_ALLOWED_REPORT_TIMESTAMP
+                                    .sub(&Duration::from_seconds(8))
+                                    .unwrap(),
+                            ],
+                        )
+                        .await;
+                    tx.put_collect_job(&CollectJob::<0, TimeInterval, dummy_vdaf::Vdaf>::new(
```
tgeoghegan, line 9667: I think you need to add some collect jobs for this task that are related to reports that are not eligible for GC, and then check that those collect jobs don't get garbage collected. That would have caught the bug fixed in eb86863.

aggregator/src/datastore.rs
```diff
@@ -2955,22 +3217,229 @@ ORDER BY id DESC
...
+                      AND NOT EXISTS (
+                          SELECT id FROM collect_jobs
+                          WHERE aggregation_jobs.task_id = collect_jobs.task_id
+                            AND (aggregation_jobs.batch_identifier = collect_jobs.batch_identifier
+                              OR aggregation_jobs.batch_interval <@ collect_jobs.batch_interval))
+                      AND NOT EXISTS (
+                          SELECT id FROM aggregate_share_jobs
+                          WHERE aggregation_jobs.task_id = aggregate_share_jobs.task_id
+                            AND (aggregation_jobs.batch_identifier = aggregate_share_jobs.batch_identifier
+                             OR ('"TimeInterval"'::jsonb IN (SELECT query_type FROM tasks WHERE task_id = $1)
```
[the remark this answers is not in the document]
branlwyd: Checking `aggregation_jobs.batch_identifier = NULL` is not sufficient to conclude that the task is a `TimeInterval` task: specifically, `batch_identifier` will only be `NULL` if the task is _also_ being run in the Helper role.

However, I think that in every case we're checking against at least one `batch_interval` field, which _will_ be non-`NULL` if & only if the task is `TimeInterval`. So I dropped the explicit `TimeInterval` check throughout. (I kind of liked having it, as I don't like relying on implicit reasoning like this, but I didn't apply explicit checks everywhere so I can't complain too much.)

aggregator/src/datastore.rs
branlwyd: This task already had an un-GC'ed collect job (see the "Leader, time-interval aggregation job with attached collect job [not GC'ed]" case).

The problem was that the batch aggregations were being GC'ed anyway due to this clause, because the batch identifiers (i.e. intervals) were identical between the aggregation job & the batch aggregation:

```
                        OR batch_aggregations.batch_identifier IN (
                            SELECT batch_identifier FROM aggregation_jobs
                            WHERE aggregation_jobs.id IN (SELECT id FROM aggregation_jobs_to_delete)
                        )
```

I added a little code to "shorten" the interval used in the batch aggregation, and verified that this change detects regression.

branlwyd: This task already had an un-GC'ed collect job (see the "Leader, time-interval aggregation job with attached collect job [not GC'ed]" case).

The problem was that the batch aggregations were being GC'ed anyway due to this clause, because the batch identifiers (i.e. intervals) were identical between the aggregation job & the batch aggregation:

```
                        OR batch_aggregations.batch_identifier IN (
                            SELECT batch_identifier FROM aggregation_jobs
                            WHERE aggregation_jobs.id IN (SELECT id FROM aggregation_jobs_to_delete)
                        )
```

I added a little code to "shorten" the interval used in the batch aggregation, and verified that this change detects regression.

tgeoghegan (approved):
LGTM!

PR #967, opened 2023-02-02
Create aggregation jobs with arbitrary reports.

branlwyd:
Aggregation jobs for time-interval tasks were previously created with reports in a single "batch unit" (i.e. an interval of time aligned with & as long as the time precision of the task). Now, aggregation jobs for time-interval tasks are created with arbitrary reports.

This allows us to stop ordering the reports we read, which will hopefully improve DB performance as we no longer need to read (and therefore predicate-lock) all reports for a given task.

This implied a schema change: previously, aggregation_jobs tracked the batch_identifier. Now, for time-interval tasks there is no batch_identifier to record. This column (and the equivalent batch_interval) are now removed; instead, client_timestamp_interval & the new batch_id column are used where batch_identifier previously would have been used.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_creator.rs
+++ b/aggregator/src/aggregator/aggregation_job_creator.rs
@@ -2,22 +2,17 @@ use crate::{
     datastore::models::{
         AggregationJob, AggregationJobState, ReportAggregation, ReportAggregationState,
     },
-    datastore::{self, gather_errors, models::OutstandingBatch, Datastore},
+    datastore::{gather_errors, models::OutstandingBatch, Datastore},
     messages::{DurationExt as _, TimeExt as _},
     task::{self, Task, PRIO3_AES128_VERIFY_KEY_LENGTH},
     try_join,
 };
-use anyhow::Result;
 #[cfg(feature = "fpvec_bounded_l2")]
 use fixed::types::extra::{U15, U31, U63};
 #[cfg(feature = "fpvec_bounded_l2")]
 use fixed::{FixedI16, FixedI32, FixedI64};
 use futures::{future::join_all, FutureExt};
-use itertools::Itertools;
-use janus_core::{
-    task::VdafInstance,
-    time::{Clock, TimeExt as _},
-};
+use janus_core::{task::VdafInstance, time::Clock};
 use janus_messages::{
     query_type::{FixedSize, TimeInterval},
     Duration as DurationMsg, Interval, Role, TaskId,
@@ -76,7 +71,6 @@ mod private {
 pub struct AggregationJobCreator<C: Clock> {
     // Dependencies.
     datastore: Datastore<C>,
-    clock: C,
 
     // Configuration values.
     /// How frequently we look for new tasks to start creating aggregation jobs for.
@@ -96,7 +90,6 @@ pub struct AggregationJobCreator<C: Clock> {
 impl<C: Clock + 'static> AggregationJobCreator<C> {
     pub fn new(
         datastore: Datastore<C>,
-        clock: C,
         tasks_update_frequency: Duration,
         aggregation_job_creation_interval: Duration,
         min_aggregation_job_size: usize,
@@ -104,7 +97,6 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
     ) -> AggregationJobCreator<C> {
         AggregationJobCreator {
             datastore,
-            clock,
             tasks_update_frequency,
             aggregation_job_creation_interval,
             min_aggregation_job_size,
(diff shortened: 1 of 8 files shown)


aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -361,77 +353,58 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
...
-                        {
-                            if batch_start >= current_batch_start
-                                && agg_job_reports.len() < this.min_aggregation_job_size
-                            {
-                                continue;
-                            }
+                    for agg_job_reports in
+                        report_ids_and_times.chunks(this.max_aggregation_job_size)
+                    {
+                        if agg_job_reports.len() < this.min_aggregation_job_size {
```
divergentdave, line 367: This is a bit stricter than the previous behavior about minimum aggregation job sizes, since there's no longer an exception for the current batch. It seems our integration tests are okay since they run this with `min_aggregation_job_size: 1`. I think the overhead benefits of this are good enough that we should probably keep it, and accept more tail latency on aggregation of low-volume tasks.

aggregator/src/datastore.rs
```diff
@@ -3265,25 +3223,25 @@ ORDER BY id DESC
...
                     SELECT id FROM aggregation_jobs
                     WHERE task_id = (SELECT id FROM tasks WHERE task_id = $1)
                       AND UPPER(client_timestamp_interval) <= $2
                       AND NOT EXISTS (
                           SELECT id FROM collect_jobs
                           WHERE aggregation_jobs.task_id = collect_jobs.task_id
-                            AND (aggregation_jobs.batch_identifier = collect_jobs.batch_identifier
-                              OR aggregation_jobs.batch_interval <@ collect_jobs.batch_interval))
+                            AND (aggregation_jobs.batch_id = collect_jobs.batch_identifier
+                             OR aggregation_jobs.client_timestamp_interval <@ collect_jobs.batch_interval))
```
divergentdave, line 3234: Should this switch to && as well?

aggregator/src/datastore.rs
```diff
@@ -3265,25 +3223,25 @@ ORDER BY id DESC
...
                       AND NOT EXISTS (
                           SELECT id FROM aggregate_share_jobs
                           WHERE aggregation_jobs.task_id = aggregate_share_jobs.task_id
-                            AND (aggregation_jobs.batch_identifier = aggregate_share_jobs.batch_identifier
+                            AND (aggregation_jobs.batch_id = aggregate_share_jobs.batch_identifier
                              OR aggregation_jobs.client_timestamp_interval <@ aggregate_share_jobs.batch_interval)
```
divergentdave, line 3239: (same question as above)

aggregator/src/datastore.rs
```diff
@@ -3358,8 +3315,8 @@ ORDER BY id DESC
                         FROM collect_jobs
                         JOIN aggregation_jobs
                             ON aggregation_jobs.task_id = collect_jobs.task_id
-                            AND (aggregation_jobs.batch_identifier = collect_jobs.batch_identifier
-                              OR aggregation_jobs.batch_interval <@ collect_jobs.batch_interval)
+                            AND (aggregation_jobs.batch_id = collect_jobs.batch_identifier
+                                OR aggregation_jobs.client_timestamp_interval <@ collect_jobs.batch_interval)
```
divergentdave, line 3319: I think this probably needs to use && as well, since we're dropping the assumption that the aggregation job's timestamp interval is minimally sized.

aggregator/src/datastore.rs
```diff
@@ -3375,7 +3332,7 @@ ORDER BY id DESC
                         FROM aggregate_share_jobs
                         JOIN aggregation_jobs
                             ON aggregation_jobs.task_id = aggregate_share_jobs.task_id
-                            AND (aggregation_jobs.batch_identifier = aggregate_share_jobs.batch_identifier
+                            AND (aggregation_jobs.batch_id = aggregate_share_jobs.batch_identifier
                               OR aggregation_jobs.client_timestamp_interval <@ aggregate_share_jobs.batch_interval)
```
divergentdave, line 3336: I think this may have to change to && as well. If all overlapping aggregation jobs end up with wider intervals than the interval of an aggregate share job, then that aggregate share job would always get filtered out.

aggregator/src/aggregator/aggregation_job_creator.rs
branlwyd: I agree; the new behavior is that we'll wait until we have at least `min_aggregation_job_size` reports available before we generate an aggregation job. This means that low-volume tasks will see increased latency. I agree that we should keep it for now -- if this behavior ends up being problematic, we could eventually add logic to notice if reports are "too old" & generate a smaller-than-min-size aggregation job in this case.

aggregator/src/datastore.rs
branlwyd: Good catch. To test this, I updated the unit tests to artificially extend the `client_timestamp_interval` by one second in each direction. I also spaced out the client timestamps to make sure there were no "false overlaps".

db/schema.sql
```diff
@@ -99,9 +99,8 @@ CREATE TABLE aggregation_jobs(
     id                        BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY, -- artificial ID, internal-only
     task_id                   BIGINT NOT NULL,                 -- ID of related task
     aggregation_job_id        BYTEA NOT NULL,                  -- 32-byte AggregationJobID as defined by the DAP specification
-    batch_identifier          BYTEA,                           -- encoded query-type-specific batch identifier (corresponds to identifier in BatchSelector; present for leader tasks and fixed size tasks, NULL otherwise)
-    batch_interval            TSRANGE,                         -- batch interval, as a TSRANGE, populated only for leader time-interval tasks. (will match batch_identifier when present)
     aggregation_param         BYTEA NOT NULL,                  -- encoded aggregation parameter (opaque VDAF message)
+    batch_id                  BYTEA NOT NULL,                  -- batch ID (fixed-size only; corresponds to identifier in BatchSelector)
```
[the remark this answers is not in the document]
divergentdave: I think it is an empty BYTEA in that case, the encoding of the unit type inside the partial batch selector.

divergentdave (approved):
LGTM. I tested this out, and compared to the last run of the 0.2.6 release, this handled 100 reports per second roughly similarly, with job creation slowing down and finally stalling out after ten hours. (I confirmed that the new `LIMIT 5000` query is in use at the database logs) The major difference I saw was that the aggregation job creator transaction now has a 100% rollback rate in the live lock regime, whereas previously I think it got stuck in an interminably long transaction, and there were no success or error statuses. As the error rate was increasing, there was an hour-long regime where the error rate for this was ~99.5%, and during this time, collections were successfully finishing at a lower rate, falling behind incoming reports. Once the aggregation job creator error rate pegged at 100%, there were no further successful collections.

branlwyd:
> LGTM. I tested this out, and compared to the last run of the 0.2.6 release, this handled 100 reports per second roughly similarly, with job creation slowing down and finally stalling out after ten hours. (I confirmed that the new `LIMIT 5000` query is in use at the database logs) The major difference I saw was that the aggregation job creator transaction now has a 100% rollback rate in the live lock regime, whereas previously I think it got stuck in an interminably long transaction, and there were no success or error statuses. As the error rate was increasing, there was an hour-long regime where the error rate for this was ~99.5%, and during this time, collections were successfully finishing at a lower rate, falling behind incoming reports. Once the aggregation job creator error rate pegged at 100%, there were no further successful collections.

Did you happen to catch, or could you easily get, a query plan for the queries used by the aggregation job creator? Perhaps one during each of the regimes. Do we know what we are conflicting with?

divergentdave:
I'm not tooled up to collect query plans over time yet, but here's a query plan from the live lock regime. This came from `get_unaggregated_client_report_ids_for_task()`, and just happened to make the cut for my `auto_explain.log_min_duration` setting.

```
2023-02-03 19:24:51.423 GMT [141139] LOG:  duration: 1049.072 ms  plan:
        Query Text: SELECT report_id, client_timestamp FROM client_reports
                        LEFT JOIN report_aggregations ON report_aggregations.client_report_id = client_reports.id
                        WHERE task_id = (SELECT id FROM tasks WHERE task_id = $1)
                        AND report_aggregations.id IS NULL
                        LIMIT 5000
        Limit  (cost=79873.90..308249.48 rows=1 width=25)
          InitPlan 1 (returns $0)
            ->  Index Scan using tasks_task_id_key on tasks  (cost=0.15..8.17 rows=1 width=8)
                  Index Cond: (task_id = '\xfd7d76707c1c36d09c1ec2c69df372fea86ce89245c0405a31d2ffb14e3637c4'::bytea)
          ->  Gather  (cost=79865.73..308241.32 rows=1 width=25)
                Workers Planned: 2
                Params Evaluated: $0
                ->  Parallel Hash Left Join  (cost=78865.73..307241.22 rows=1 width=25)
                      Hash Cond: (client_reports.id = report_aggregations.client_report_id)
                      Filter: (report_aggregations.id IS NULL)
                      ->  Parallel Seq Scan on client_reports  (cost=0.00..174740.28 rows=2154822 width=33)
                            Filter: (task_id = $0)
                      ->  Parallel Hash  (cost=56037.88..56037.88 rows=1313188 width=16)
                            ->  Parallel Seq Scan on report_aggregations  (cost=0.00..56037.88 rows=1313188 width=16)
```

Looking at the surrounding parts of the log tells a straightforward story. Looking prior to this message, another similar transaction aborted 1.1 seconds ago. After the slow `get_unaggregated_client_report_ids_for_task()` query executes, I see a serialization error on `INSERT INTO aggregation_jobs ...`. Then, there are many messages from similar inserts being ignored due to the transaction having aborted. Following that is another slow query's plan, and the cycle repeats.

I don't have any direct evidence on what other transactions were part of the conflict, i.e. what this was the pivot for, but given that the hash join scanned all of client_reports and report_aggregations, it was most likely upload transactions, or perhaps aggregation job driver transactions. (I do also see `INSERT INTO report_aggregations` queries being ignored due to cancelled transactions)

tgeoghegan (approved):

branlwyd:
Thanks. I'm going to merge this, but I think it's clear additional work is needed here. I do suspect that client uploads are one cause of serialization failures--so the upload batching you are working on will help a lot, just by lieu of there being fewer transactions to conflict with. But I do wonder how much aggregation job driver transactions interfere, as well (they'll never update `report_aggregations.client_report_id`, but I'm not sure if Postgres is smart enough to take that into consideration).

I'm going to explore breaking the join against `report_aggregations` by denormalizing: likely by adding an `aggregated` flag to `client_reports` which starts false and is set to true by the transaction that creates an aggregation job including the report. Hopefully, then, (I need to think about how VDAFs with an aggregation parameter will work, but I don't think they'll break this approach.)

PR #986, opened 2023-02-06
Update release branches table for 0.3

tgeoghegan:
Note that we don't update `.github/workflows/dependabot.yml` here as we have agreed not to maintain `release/0.3`, since we are not aware of any deployments of `draft-ietf-ppm-dap-03`.


tgeoghegan:
Once this is merged, I still need to:

- [ ] branch `release/0.3` from `main`

divergentdave (approved):

PR #988, opened 2023-02-06
Introduce client_reports.aggregation_started flag.

branlwyd:
This new flag is used to track which client reports have been attached to an aggregation, allowing us to create aggregation jobs without needing to consult the report_aggregations table. Hopefully, this will reduce the amount of contention between aggregation job creation & other datastore operations in the system.

Note that we set the flag for both VDAFs without an aggregation parameter, and VDAFs with an aggregation parameter. However, it's only effectively used by VDAFs without an aggregation parameter. Still, the flag is set for consistency.

### DIFF
--- a/aggregator/src/datastore.rs
+++ b/aggregator/src/datastore.rs
@@ -1004,10 +1004,13 @@ impl<C: Clock> Transaction<'_, C> {
     }
 
     /// `get_unaggregated_client_report_ids_for_task` returns some report IDs corresponding to
-    /// unaggregated client reports for the task identified by the given task ID.
+    /// unaggregated client reports for the task identified by the given task ID. Returned client
+    /// reports are marked as aggregation-started: the caller must either create an aggregation job
+    /// with, or call `mark_reports_unaggregated` on, each returned report as part of the same
+    /// transaction.
     ///
-    /// This should only be used with VDAFs that have an aggregation parameter of the unit type.
-    /// It relies on this assumption to find relevant reports without consulting collect jobs. For
+    /// This should only be used with VDAFs that have an aggregation parameter of the unit type. It
+    /// relies on this assumption to find relevant reports without consulting collect jobs. For
     /// VDAFs that do have a different aggregation parameter,
     /// `get_unaggregated_client_report_ids_by_collect_for_task` should be used instead.
     #[tracing::instrument(skip(self), err)]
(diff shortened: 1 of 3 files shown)


branlwyd:
Assuming no bugs, I think this change won't modify the system's normal behavior. However, some manual-operational workflows might change: for example, if we ever were to delete an aggregation job & related report aggregations, we'd no longer expect an equivalent aggregation job to be recreated for the affected reports. I think the impact of this is OK but please think things through as part of review.

aggregator/src/datastore.rs
```diff
@@ -1004,10 +1004,13 @@ impl<C: Clock> Transaction<'_, C> {
     }
 
     /// `get_unaggregated_client_report_ids_for_task` returns some report IDs corresponding to
-    /// unaggregated client reports for the task identified by the given task ID.
+    /// unaggregated client reports for the task identified by the given task ID. Returned client
+    /// reports are marked as aggregation-started: the caller must either create an aggregation job
+    /// with, or call `mark_reports_unaggregated` on, each returned report as part of the same
```
tgeoghegan, line 1009:
```suggestion
    /// with, or call `mark_reports_unaggregated` on each returned report as part of the same
```

tgeoghegan (approved):
This change looks sound. The only half-baked thought I have is:

This change only helps for VDAFs with no aggregation parameter. Where there is an aggregation parameter, you still have to join across the `report_aggregations` table to deal with the aggregation parameter. I guess the problem there is that you can have multiple aggregation jobs for a single report, with different aggregation parameters, so a single column in `client_reports` can't help. You'd need a new table mapping report ID to aggregation job ID, or maybe just a view on `report_aggregations`.

In any case, that's all beyond the scope of this change, which should help the no-agg-param case.

branlwyd:
> This change looks sound. The only half-baked thought I have is:
> 
> This change only helps for VDAFs with no aggregation parameter. Where there is an aggregation parameter, you still have to join across the `report_aggregations` table to deal with the aggregation parameter. I guess the problem there is that you can have multiple aggregation jobs for a single report, with different aggregation parameters, so a single column in `client_reports` can't help. You'd need a new table mapping report ID to aggregation job ID, or maybe just a view on `report_aggregations`.
> 
> In any case, that's all beyond the scope of this change, which should help the no-agg-param case.

Yep -- the `aggregation_started` flag is appropriately written for the VDAF-with-aggregation-parameter case, but it's not read, nor do I think it would be conceptually useful for the reason you note. However, I think introducing additional tables might end up creating equivalent contention to that which would already exist with `collect_jobs` -- these tables would conceptually need to be read/written at the same time as `collect_jobs`.

In any case, this is all marked `#[cfg(test)]` for now, so I'm OK with only improving the no-aggregation-parameter case for now.

(Off-topic, but while reviewing just now I also noticed that the with-aggregation-parameter codepath actually only works for time-interval tasks, due to the way `get_unaggregated_client_report_ids_by_collect_for_task` joins `client_reports` to `collect_jobs` based on `batch_interval`.)

PR #1012, opened 2023-02-14
Update divviup-ts container used in local tests

divergentdave:
This is a follow-up to #936 to update the divviup-ts container used in local tests, following the change in the `ci-build` workflow.


tgeoghegan (approved):
Thanks, I should have spotted this in #936.

Issue #998, opened 2023-02-10
Add tests (and fix related bugs) for idempotence of resource creation


tgeoghegan:
Review of #936 uncovered some subtleties about error handling when trying to mutate resources with PUT requests ([comment thread](https://github.com/divviup/janus/pull/936#discussion_r1099171509)). We need to add tests that show we handle various scenarios as expected:

- [x] add test proving that PUT report is idempotent (#1013)
- [x] add test proving that mutating a report with PUT fails with expected error (#1013)
- [x] add test proving that PUT collection job is idempotent (#1036)
- [x] add test proving that mutating a collection job with PUT fails with expected error (#1036)
- [x] add test proving that PUT aggregation job is idempotent (#1033)
- [x] add test proving that mutating an aggregation job with PUT fails with expected error (#1033)
- [x] add test proving that POST aggregation share is idempotent ([see note below](https://github.com/divviup/janus/issues/998#issuecomment-1440659531))

...and of course to fix whatever bugs are uncovered by the new tests.

divergentdave:
We should be sure to test that collection job PUT is idempotent for `current_batch` queries in particular, per https://github.com/divviup/janus/pull/936/files#r1100491397.

tgeoghegan:
It turns out that we already have a test showing that POST to aggregate shares is idempotent. [This portion of `aggregator::tests::aggregate_share_request`](https://github.com/divviup/janus/blob/368ceb85e4b3df68511f436ef98e3b49dd0ff171/aggregator/src/aggregator.rs#L9332) sends valid aggregate share requests multiple times and verifies that the requests succeed each time.

tgeoghegan:
I think we have all the tests we need in place now, so I'm closing this.

PR #1013, opened 2023-02-14
Idempotent report uploads

tgeoghegan:
Uploading reports to the leader should be idempotent, meaning:
    
 - repeated requests to upload the same report (same ID and contents) should succeed
 - PUT with a repeated report ID by varying contents should fail
    
Part of #998

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1971,7 +1971,7 @@ impl VdafOps {
         );
 
         report_writer
-            .write_report(WritableReport::<L, Q, A>::new(report))
+            .write_report(WritableReport::<L, Q, A>::new(vdaf, report))
             .await
     }
 
@@ -4606,12 +4606,12 @@ mod tests {
         );
     }
 
-    fn create_report(task: &Task, report_timestamp: Time) -> Report {
+    fn create_report_with_id(task: &Task, report_timestamp: Time, id: ReportId) -> Report {
         assert_eq!(task.vdaf(), &VdafInstance::Prio3Aes128Count);
 
         let vdaf = Prio3Aes128Count::new_aes128_count(2).unwrap();
         let hpke_key = current_hpke_key(task.hpke_keys());
-        let report_metadata = ReportMetadata::new(random(), report_timestamp);
+        let report_metadata = ReportMetadata::new(id, report_timestamp);
 
         let (public_share, measurements) = vdaf.shard(&1).unwrap();
 
@@ -4643,6 +4643,10 @@ mod tests {
         )
     }
 
+    fn create_report(task: &Task, report_timestamp: Time) -> Report {
+        create_report_with_id(task, report_timestamp, random())
+    }
+
     /// Convenience method to handle interaction with `warp::test` for typical DAP requests.
     async fn drive_filter(
         method: Method,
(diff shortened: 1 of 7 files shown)


divergentdave (approved):

aggregator/src/aggregator.rs
```diff
@@ -4662,6 +4666,45 @@ mod tests {
...
+                    "status": desired_status.as_u16(),
+                    "type": format!("urn:ietf:params:ppm:dap:error:{desired_type}"),
+                    "title": desired_title,
+                    "taskid": format!("{desired_task_id}"),
+                }),
+                "at {file}:{line}"
+            )
+        }
+
+        macro_rules! check_response {
```
branlwyd, line 4694: Rather than inventing a macro to use in one test, IMO we should just call `check_response_fn` directly, but remove the file/line arguments. Even without the file/line information included manually we can run with `RUST_BACKTRACE=1` to get accurate caller information when needed.

aggregator/src/datastore.rs
```diff
@@ -1241,19 +1244,63 @@ impl<C: Clock> Transaction<'_, C> {
...
                     /* public_share */ &encoded_public_share,
                     /* leader_input_share */ &encoded_leader_share,
                     /* helper_encrypted_input_share */ &encoded_helper_share,
                 ],
             )
             .await?;
-        if rows_affected == 0 {
+
+        if rows_affected > 1 {
+            return Err(Error::DbState(format!(
```
branlwyd, line 1260: I guess this should be a panic, since it's impossible/indicates a bug in Postgres. (or just drop this check entirely)

aggregator/src/datastore.rs
```diff
@@ -3622,6 +3705,29 @@ pub mod models {
             }
         }
 
+        /// Decode the fields of a client report using the supplied VDAF and construct a
+        /// [`LeaderStoredReport`].
+        pub(super) fn from_encoded_client_report(
```
branlwyd, line 3710: If this sticks around, it should be an `impl From<EncodedClientReport> for LeaderStoredReport`.

aggregator/src/datastore.rs
```diff
@@ -879,20 +879,14 @@ impl<C: Clock> Transaction<'_, C> {
         )?)
     }
 
-    /// get_client_report retrieves a client report by ID.
+    /// Retrieve a client report by ID, but without decoding the row's columns into VDAF-specific
+    /// types
     #[tracing::instrument(skip(self), err)]
-    pub async fn get_client_report<const L: usize, A>(
+    async fn get_encoded_client_report(
```
branlwyd, line 885: Why do we need `EncodedClientReport`? If it's just to check equality, we can define equality on `LeaderStoredReport` & continue using that. If it's to avoid re-parsing on the slow read path, I'm not sure this is worth the code complexity vs just parsing the existing report into a `LeaderStoredReport`.

In any case, the pattern used throughout `datastore.rs` is for places that need to "share" parsing logic to use `foo_from_row` helper methods that take a `Row` (and other values, if already parsed everywhere they are needed). Let's follow that here.

branlwyd (changes requested):

aggregator/src/aggregator.rs
divergentdave: We may want to revisit `profile.dev.debug = 0` in `Cargo.toml`, I've found myself commenting it out often. We could move it to a separate CI-only profile, perhaps.

branlwyd (approved):

PR #1014, opened 2023-02-14
Check for retryable errors at tokio_postgres boundary.

branlwyd:
Previously, we checked for retryable errors at the point they were returned from the user-supplied callback. This sorta works, but expects every callback to be smart enough to propagate serialization errors appropriately -- which has proven to be error-prone, and required some clever (and complex) code to get right.

Now, we check each tokio_postgres::Transaction operation for a retryable failure. No matter what the return value from the user-supplied callback is, if we ever observed a retryable failure, we retry the transaction. The user-supplied callback no longer needs to be involved in the retry process.


divergentdave (approved):
Very nice! LGTM.

PR #1019, opened 2023-02-16
report IDs should be unique per aggregation job

tgeoghegan:
Adds a uniqueness constraint on the `report_aggregations` table ensuring that any report ID occurs exactly once in any aggregation job.


tgeoghegan:
Per offline discussion, we're not going to do this because the safety/data integrity gain is not worth the performance hit of the new uniqueness constraint. The really important data integrity check -- that aggregation jobs can't be mutated except by stepping them -- will be implemented in the course of #998.

PR #1020, opened 2023-02-16
Update CI build script to keep Build & Test steps consecutive.

branlwyd:
Also, tweak flags so that the test step uses all the same flags as the Build step, except for the `--no-run` flag that keeps the Build step from actually running tests.

(I noticed in CI that our test step was building a few crates: I'm not sure if this is due to the `janus_messages` build between Build & Test, or if it's due to the difference in flags, but ultimately there should be no building required to start running tests.)

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -74,16 +74,17 @@ jobs:
       uses: Swatinem/rust-cache@v2
       with:
         key: ${{ steps.os-version.outputs.release }}
-    - name: Build
-      run: cargo test --no-run --locked --all-targets
     - name: Build minimal janus_messages
       run: cargo build --package janus_messages --no-default-features
+    # Note: keep Build & Test steps consecutive, and match flags other than `--no-run`.
+    - name: Build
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
It looks like this didn't change which crates get rebuilt, but it's nevertheless a worthy change. I noticed that the three crates which get rebuilt are exactly those with build scripts. Looking at https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection, I think the aggregator build script I recently added needs to declare it only depends on the environment variable providing `rustc`.

tgeoghegan (approved):

branlwyd:
> It looks like this didn't change which crates get rebuilt, but it's nevertheless a worthy change.

Indeed -- I will still merge this as these steps should be consecutive. I suspect the build scripts are responsible for the repeated build steps. (anecdotally, these steps are very quick; I'd guess the build scripts are running for some reason & then cargo is realizing there is nothing to do over-and-above the previous build. but they don't run in this way on my local workstation, so I'm currently at a loss as to why they are being triggered.)

PR #1021, opened 2023-02-16
Add rerun-if command to janus_aggregator build.rs

divergentdave:
This adds a `rerun-if-env-changed` command to the new build script, so it doesn't have to be rerun as often. See https://doc.rust-lang.org/cargo/reference/build-scripts.html#change-detection


tgeoghegan (approved):

branlwyd (approved):

divergentdave:
All three crates got rebuilt again in CI, ah well. (I wonder if there's something weird with mtimes in the VMs)

PR #1022, opened 2023-02-17
Fix report replay check in aggregation job init

tgeoghegan:
`handle_aggregate_init_generic` was checking if any of the report shares in an `AggregationJobInitReq` already occurred in the `client_reports` table, and would refuse to begin preparing an input share if so, producing a `report_replayed` error.

However, this is not what DAP requires ([1]):

> Check if the report has already been aggregated *with this aggregation
> parameter*. If this check fails, the input share MUST be marked as invalid
> with the error `report_replayed`. This is the case if the report was used in
> a previous aggregate request and is therefore a replay.

We now check whether there's a `report_aggregations` row for the `ReportId` and whether its corresponding aggregation job uses the same aggregation parameter as the incoming aggregation job. Additionally, we allow `put_report_share` to write the same report twice, because that's perfectly OK for a VDAF like Poplar1.

[1]: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/89e5dc32be05ebd4eb032c7f6d0aaa7074c69b08/draft-ietf-ppm-dap.md?plain=1#L1331

### DIFF
--- a/aggregator/src/datastore.rs
+++ b/aggregator/src/datastore.rs
@@ -919,6 +919,30 @@ impl<C: Clock> Transaction<'_, C> {
     }
 
     #[cfg(feature = "test-util")]
+    pub async fn get_report_metadatas_for_task(
+        &self,
+        task_id: &TaskId,
+    ) -> Result<Vec<ReportMetadata>, Error> {
+        let stmt = self
+            .prepare_cached(
+                "SELECT client_reports.report_id, client_reports.client_timestamp
+                FROM client_reports
+                JOIN tasks ON tasks.id = client_reports.task_id WHERE tasks.task_id = $1",
+            )
+            .await?;
+        self.query(&stmt, &[&task_id.as_ref()])
+            .await?
+            .into_iter()
+            .map(|row| {
+                Ok(ReportMetadata::new(
+                    row.get_bytea_and_convert::<ReportId>("report_id")?,
+                    Time::from_naive_date_time(&row.get("client_timestamp")),
+                ))
+            })
+            .collect()
+    }
+
+    #[cfg(feature = "test-util")]
     pub async fn get_client_reports_for_task<const L: usize, A: vdaf::Aggregator<L>>(
         &self,
         vdaf: &A,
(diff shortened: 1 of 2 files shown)


tgeoghegan:
I encountered this problem while working on aggregation job idempotence as part of #998, and decided to break this out into a smaller PR.

aggregator/src/datastore.rs
```diff
@@ -1344,7 +1315,8 @@ impl<C: Clock> Transaction<'_, C> {
         let stmt = self
             .prepare_cached(
                 "INSERT INTO client_reports (task_id, report_id, client_timestamp)
-                VALUES ((SELECT id FROM tasks WHERE task_id = $1), $2, $3)",
+                VALUES ((SELECT id FROM tasks WHERE task_id = $1), $2, $3)
+                ON CONFLICT DO NOTHING",
```
divergentdave, line 1319: FWIW: in the presence of a misbehaving leader on a task with a VDAF allowing multiple aggregation parameters, `ON CONFLICT DO NOTHING` could lead to a surprise. If the leader incorrectly sends two different reports sharing the same report ID, in aggregation jobs with different aggregation parameters, then the helper will save the first report's timestamp, and eventually reuse that timestamp when preparation of the second report finishes. This timestamp could end up pointing to the "wrong" batch aggregation, compared to the actual timestamp on the second report, and result in two batch aggregations being out of sync between the leader and helper.

The above describes a robustness failure in the presence of a misbehaving aggregator, which is acceptable in the security model, but we may want to catch the changing timestamp in service of debugging the other aggregator. I think the following would let us catch such issues by throwing an error if no rows were modified.

```suggestion
                ON CONFLICT (task_id, report_id) DO UPDATE SET client_timestamp = $3 WHERE excluded.client_timestamp = $3",
```

(or, we could use a PL/pgSQL function to do a similar check in a single DB round-trip, but avoid unnecessary tuple writes)

divergentdave (approved):

aggregator/src/datastore.rs
tgeoghegan: What you describe is a regression introduced by this change (we used to reject aggregation jobs if any report ID therein already existed) so I want to fix it here. I think the behavior we want is to raise an error if we detect an attempt to mutate an existing `client_reports` row. I figured out a way to do this using the `xmax` system column (actually I'm cribbing from StackOverflow, naturally). I'd appreciate some feedback on whether this is too complicated.

divergentdave: I'm wary about using MVCC metadata, particularly given the caveats about `xmax` being nonzero due to rollbacks or locking.

I fleshed out my suggestion snippet above into 7f4167b, this passes the new tests, and has a downside of performing unneccessary no-op UPDATEs on the Poplar1 happy path. If we want to avoid extra UPDATEs, I'd prefer we do so via a PL/pgSQL function, so that we can SELECT, conditionally INSERT, and return a flag all in one call.

tgeoghegan: Fair enough. I suspect we have a bunch more work to do before VDAFs with aggregation parameters actually work in Janus, so I'm happy to take a Poplar1 performance hit for now. Thank you for looking into this.

divergentdave (approved):

aggregator/src/datastore.rs
```diff
@@ -918,6 +918,30 @@ impl<C: Clock> Transaction<'_, C> {
         .transpose()
     }
 
+    #[cfg(feature = "test-util")]
+    pub async fn get_helper_client_reports_for_task(
```
branlwyd, line 922: Q: why `helper`? I think this would work just as well for a leader task. This method also doesn't return reports, maybe `get_report_metadatas_for_task`?

aggregator/src/datastore.rs
```diff
@@ -1615,6 +1625,36 @@ impl<C: Clock> Transaction<'_, C> {
         )
     }
 
+    /// Check whether the report has ever been aggregated with the given parameter.
+    #[tracing::instrument(skip(self), err)]
+    pub async fn check_report_aggregation_exists(
```
branlwyd, line 1630: This method should be typesafe for `aggregation_param` (e.g. it should have an `A: Aggregator` generic parameter, and the `aggregation_param` parameter should be an `&A::AggregationParam`).

aggregator/src/datastore.rs
branlwyd: I like the current `UPDATE` trick to detect if anything is actually changing. This seems more understandable than a trick based on system columns, and we definitely don't need to write a pgSQL function to avoid in-DB no-op updates (yet).

branlwyd (approved):
Just a couple of nits.

aggregator/src/datastore.rs
tgeoghegan: Oops, that was a leftover from when I was going to have this return something else and was dealing with the fact that the helper's `client_reports` rows don't have every column set, but `get_report_metadatas_for_task` is much better.

PR #1033, opened 2023-02-22
Aggregation job PUT idempotence

tgeoghegan:
We now allow a leader to PUT the same aggregation job multiple times into a Janus helper, provided that we can verify that the aggregation job and the report shares it aggregates have not changed.

Part of https://github.com/divviup/janus/issues/998

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -13,7 +13,7 @@ use crate::{
             AggregateShareJob, AggregationJob, AggregationJobState, CollectionJob,
             CollectionJobState, LeaderStoredReport, ReportAggregation, ReportAggregationState,
         },
-        Datastore,
+        Datastore, Transaction,
     },
     messages::{DurationExt, IntervalExt, TimeExt},
     task::{self, Task, VerifyKey, PRIO3_AES128_VERIFY_KEY_LENGTH},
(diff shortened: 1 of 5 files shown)


aggregator/src/aggregator.rs
```diff
@@ -851,6 +854,41 @@ impl<C: Clock> TaskAggregator<C> {
     }
 }
 
+/// Used by the aggregation job initialization handler to represent initialization of a report
+/// share.
+#[derive(Clone, Debug)]
+struct ReportShareData<const L: usize, A>
```
tgeoghegan, line 860: I really wish this struct could be defined closer to the two methods on `VdafOps` that use it, but Rust won't allow you to define a struct inside another `impl`. You can only do that inside a method.

I also considered taking those parts of `VdafOps` that pertain to handling `AggregationJobInitReq` and moving them into a new module `aggregator/src/aggregator/aggregation_job_init_handler.rs` that would define `struct AggregationJobInitHandler`. I think this would help with navigating Janus (`aggregator.rs` is already nearly 10,000 lines), but it would make a mess of this diff, so I'm not doing it now.

divergentdave: FWIW, we could split the `impl` block in two, to group related code together now with a minimally invasive change.

tgeoghegan: That's a good idea, thank you!

divergentdave (approved):

PR #1037, opened 2023-02-23
Shard batch aggregations.

branlwyd:
Batch aggregations (which collect the aggregate share for each given batch) are now sharded: writes during the aggregation process are sent to a (randomly-chosen) shard; reads during the collection process read & merge all shards to retrieve the final aggregate share.

The purpose of this change is to reduce write contention during the aggregation process. The tradeoff is in increased read volume & computation during the collection process.

The number of shards is controlled by a pair of new `batch_aggregation_shard_count` configuration values. The aggregator configuration value controls the helper sharding value, and the aggregation job driver configuration value controls the leader sharding value. (As this sharding behavior is private to the aggregator, there is no reason that these values need to match up.)


db/schema.sql
```diff
@@ -148,18 +148,19 @@ CREATE INDEX report_aggregations_aggregation_job_id_index ON report_aggregations
 CREATE INDEX report_aggregations_client_report_id_index ON report_aggregations(client_report_id);
 
 -- Information on aggregation for a single batch. This information may be incremental if the VDAF
--- supports incremental aggregation.
+-- supports incremental aggregation. Each batch's aggregation is sharded via
```
tgeoghegan, line 151: via...?

tgeoghegan (approved):

tgeoghegan:
Despite having to plumb the max shard count parameter in a ton of places, I'm pleasantly surprised at how unintrusive this change is, given its impact.

branlwyd:
(This appears to have legitimately broken the `janus_divviup_ts` tests; need to investigate why.)

PR #1038, opened 2023-02-23
Database: switch from Serializable to Repeatable Read.

branlwyd:
Several different processes in the system (uploads & aggregation, aggregation & collection) were interfering with one another due to false sharing, which didn't affect small tests much but which caused an unacceptable number of retries.

It may be possible to improve our queries, but for now switching to Repeatable Read allows us to meet our performance needs.


branlwyd:
With this change (and the previous batch-aggregation-sharding change) I was able to handle 1000 QPS of sustained uploads in a local Kind test, with no errors or lost batches, over a ~12-hour period. Without this change, we are not able to generate aggregation jobs fast enough to keep up with incoming reports.

tgeoghegan (approved):

PR #1042, opened 2023-02-25
Fix panic in ReportWriter if upload request is cancelled(?).

branlwyd:
The response-receiver could be dropped if the entire upload future was dropped, which I think corresponds to a cancelled request handling. This would cause the attempt to send the response to panic, which would cause all other waiting response-receivers to panic themselves when the respense-sender was dropped. I saw a few instances of this in the logs from the latest load test:

```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Ok(())', /src/aggregator/src/aggregator/report_writer.rs:139:26
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: RecvError(())', /src/aggregator/src/aggregator/report_writer.rs:56:23
```

With this change, any responses whose receivers have been dropped will simply be dropped.

Also, lower the level of a few spammy-but-less-useful log messages.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_creator.rs
+++ b/aggregator/src/aggregator/aggregation_job_creator.rs
@@ -134,7 +134,7 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
 
         loop {
             tasks_update_ticker.tick().await;
-            info!("Updating tasks");
+            debug!("Updating tasks");
             let start = Instant::now();
             let tasks = self
                 .datastore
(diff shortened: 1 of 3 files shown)


tgeoghegan (approved):
Seems reasonable. My only thought is that we might want to `debug!` something if the sender finds that the receiver has been dropped, or increment a metric since this case seems somewhat unusual. However I'd be OK with filing a follow-up issue since plumbing metrics collectors is kind of a chore, and I think we already know that virtually everywhere we log, we should also/instead record a metric.

divergentdave (approved):

branlwyd:
> Seems reasonable. My only thought is that we might want to `debug!` something if the sender finds that the receiver has been dropped, or increment a metric since this case seems somewhat unusual. However I'd be OK with filing a follow-up issue since plumbing metrics collectors is kind of a chore, and I think we already know that virtually everywhere we log, we should also/instead record a metric.

Seems sensible. I went with a `debug!` message: other than the challenge of plumbing the metric, I think metrics would be more useful around e.g. failed requests/cancellations -- this error is sort of a further symptom/technical detail of that. Note that the logged message doesn't include the upload result, since that result is basically unrelated to the reason that the channel send failed.

PR #1052, opened 2023-02-28
Add `interval` field to collections

### LINKED ISSUE #990: Add `interval` to struct `Collection`
As specified in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/403

tgeoghegan:
Add `interval` field to collections

In DAP-04, the `Collection` message which conveys aggregate shares to the collector now includes an `Interval` representing the exact interval of time spanned by the reports included in the collection.

When servicing a GET request on a finished collection job, Janus now queries the constituent aggregation jobs and merges their `client_report_interval`s together (we were already recording the exact interval spanned by aggregation jobs). `janus_collector::Collection` adds a field `interval: (chrono::DateTime<Utc>, chrono::Duration)` to represent this interval.

Resolves #990

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -22,8 +22,9 @@ use janus_aggregator_core::{
     datastore::{
         self,
         models::{
-            AggregateShareJob, AggregationJob, AggregationJobState, CollectionJob,
-            CollectionJobState, LeaderStoredReport, ReportAggregation, ReportAggregationState,
+            AggregateShareJob, AggregationJob, AggregationJobState, BatchAggregation,
+            CollectionJob, CollectionJobState, LeaderStoredReport, ReportAggregation,
+            ReportAggregationState,
         },
         Datastore, Transaction,
     },
@@ -3211,7 +3212,7 @@ impl VdafOps {
         Vec<u8>: for<'a> From<&'a A::AggregateShare>,
         for<'a> <A::AggregateShare as TryFrom<&'a [u8]>>::Error: Debug,
     {
-        let collection_job = datastore
+        let (collection_job, spanned_interval) = datastore
             .run_tx_with_name("get_collection_job", |tx| {
                 let (task, collection_job_id) = (Arc::clone(&task), *collection_job_id);
                 Box::pin(async move {
@@ -3223,9 +3224,31 @@ impl VdafOps {
                                 Error::UnrecognizedCollectionJob(collection_job_id).into(),
                             )
                         })?;
-                    Q::acknowledge_collection(tx, task.id(), collection_job.batch_identifier())
-                        .await?;
-                    Ok(collection_job)
+
+                    let (batch_aggregations, _) = try_join!(
+                        Q::get_batch_aggregations_for_collect_identifier(
+                            tx,
+                            &task,
+                            collection_job.batch_identifier(),
+                            collection_job.aggregation_parameter()
+                        ),
+                        Q::acknowledge_collection(tx, task.id(), collection_job.batch_identifier()),
+                    )?;
+
+                    // Merge the intervals spanned by the constituent batch aggregations into the
+                    // interval spanned by the collection.
+                    let mut spanned_interval: Option<Interval> = None;
+                    for interval in batch_aggregations
+                        .iter()
+                        .map(BatchAggregation::<L, Q, A>::client_timestamp_interval)
+                    {
+                        match spanned_interval {
+                            Some(m) => spanned_interval = Some(m.merge(interval)?),
+                            None => spanned_interval = Some(*interval),
+                        }
+                    }
+
+                    Ok((collection_job, spanned_interval))
                 })
             })
             .await?;
@@ -3241,6 +3264,14 @@ impl VdafOps {
                 encrypted_helper_aggregate_share,
                 leader_aggregate_share,
             } => {
+                let spanned_interval = spanned_interval
+                        .ok_or_else(|| {
+                            datastore::Error::User(
+                                Error::Internal(format!("collection job {collection_job_id} is finished but spans no time interval")).into(),
+                            )
+                        })?
+                        .align_to_time_precision(task.time_precision())?;
+
                 // §4.4.4.3: HPKE encrypt aggregate share to the collector. We store the leader
                 // aggregate share *unencrypted* in the datastore so that we can encrypt cached
                 // results to the collector HPKE config valid when the current collection job request
@@ -3277,6 +3308,7 @@ impl VdafOps {
                             Q::partial_batch_identifier(collection_job.batch_identifier()).clone(),
                         ),
                         *report_count,
+                        spanned_interval,
                         Vec::<HpkeCiphertext>::from([
                             encrypted_leader_aggregate_share,
                             encrypted_helper_aggregate_share.clone(),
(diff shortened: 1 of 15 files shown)


aggregator/src/aggregator.rs
```diff
@@ -3226,7 +3226,10 @@ impl VdafOps {
                         })?;
                     Q::acknowledge_collection(tx, task.id(), collection_job.batch_identifier())
                         .await?;
-                    Ok(collection_job)
+                    let spanned_interval = tx
+                        .get_interval_spanned_by_collection(&collection_job_id)
```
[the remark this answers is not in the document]
tgeoghegan: I don't think we can do `acknowledge_collection` in parallel with the other two queries because we need the collection job to get the batch identifier, but I put `get_collection_job` and `get_interval_spanned_by_collection` into a `try_join`.

I considered adding a column for the spanned interval to `collection_jobs`, but I decided that a change that doesn't require changing the schema is simpler, and also we can avoid duplicating information in the database. I also figure that GET on collection jobs will happen far, far less often than hot paths like uploads or aggregation jobs, so my intuition is that it shouldn't be too performance critical. However we could record the spanned interval at the time of creating a collection job, if we find that this query is a problem.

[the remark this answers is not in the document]
branlwyd: Ah, I didn't mean to suggest introducing a new field in the schema -- I meant to add an optional interval to the `Collection` model, filling it in "artificially" using logic like that currently written in `get_interval_by_collection`. Still -- this is fine, I strongly suspect pipelining the reads will lead to this being the ~same performance-wise.

branlwyd (dismissed):

aggregator/src/aggregator.rs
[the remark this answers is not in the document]
tgeoghegan: I see what you mean now. I did consider adding that field to `datastore::models::Collection`, but then you end up in this awkward spot: either you make it a required part of the struct and have `Datastore::get_collection_job` work out the interval, but then you force a `JOIN` across aggregation jobs on all of `get_collection_job`'s many callsites, when only a single one (the collection job GET handler) actually needs the spanned interval. Or you could make it an `Option`, but again, that's awkward since only a single code path cares about that option.

aggregator_core/src/datastore.rs
```diff
@@ -2537,6 +2537,44 @@ ORDER BY id DESC
...
+        collection_job_id: &CollectionJobId,
+    ) -> Result<Option<Interval>, Error> {
+        let stmt = self
+            .prepare_cached(
+                "SELECT aggregation_jobs.client_timestamp_interval FROM collection_jobs
+                JOIN aggregation_jobs
+                ON aggregation_jobs.task_id = collection_jobs.task_id
+                AND (aggregation_jobs.batch_id = collection_jobs.batch_identifier
+                    OR aggregation_jobs.client_timestamp_interval && collection_jobs.batch_interval)
+                WHERE collection_jobs.collection_job_id = $1",
```
divergentdave, line 2554: We should probably have separate queries/methods to find this for time interval and fixed size tasks. This logic works for fixed size tasks, because each aggregation job corresponds to a single batch, and we can directly use `aggregation_jobs.client_timestamp_interval`. However, with time interval queries, there's no correspondence between aggregation jobs and batches. We'd have to instead look up the minimum and maximum timestamp of aggregated reports within the query's batch interval.

divergentdave: Bonus: we could aggregate intervals for the fixed-size query with `SELECT tsrange(min(lower(aggregation_jobs.client_timestamp_interval)), max(upper(aggregation_jobs.client_timestamp_interval)), '[)')`. This would save us having to transfer intermediate  values, and may allow a better query plan by pushing down the aggregate.

tgeoghegan: I don't think I follow. This is the same logic used by `delete_expired_aggregation_artifacts` to determine if an aggregation job belongs to a collection, and the `collection_interval_spanned` I added shows that we can correctly determine which aggregation jobs belong to which collections in a time interval task. What case is not covered by this?

divergentdave: Assume we have four reports, two at $T_1$, and two at $T_2$, two collections, one covering only $T_1$, and one covering $T_2$, and two aggregation jobs. If both aggregation jobs contain one report at $T_1$ and one report at $T_2$, then the `client_timestamp_interval` for both aggregation jobs will span both times, and this query will report that each collection's covered interval is longer than the batch interval, when it should only ever be equal to the batch interval or shorter.

tgeoghegan: I think I see now, thank you. I have to think about this more (after lunch), but I think it's already incidentally true that every time interval aggregation job we schedule belongs to a single collection. However I need to take a harder look at how we create aggregation jobs, batch aggregations and collections.

divergentdave: I believe that changed in #967, as we no longer build a group map by rounded report timestamp, and instead just chunk up unordered reports.

tgeoghegan: Ah hah, that'll do it. OK, so we're back to the drawing board on this one. I suspect we have a similar problem with the garbage collection query.

tgeoghegan:
After studying the code a bit, I think the way forward is to determine the spanned time interval from batch aggregations instead of aggregation jobs. For both fixed size and time interval tasks, each batch aggregation belongs to exactly one batch, meaning we can use a single codepath to service both query types. We will have to add a column to `batch_aggregations` for the spanned time interval (I'll call it `client_timestamp_interval` for consistency with the `aggregation_jobs` table), and `Accumulator::update,flush_to_datastore` will have to keep it up to date based on the timestamps of reports that get accumulated. The good news is that this won't introduce new database writes, it's just that we'll update one extra column in `Datastore::update_batch_aggregation`.

core/src/time.rs
```diff
@@ -221,14 +232,117 @@ impl TimeExt for Time {
...
+
+    fn merge(&self, other: &Self) -> Result<Self, Error> {
+        let max_time = std::cmp::max(self.end(), other.end());
+        let min_time = std::cmp::min(self.start(), other.start());
+
+        // This can't actually fail for any valid Intervals
+        Self::new(*min_time, max_time.difference(min_time)?)
+    }
+
+    fn from_time(time: &Time) -> Result<Self, Error> {
```
tgeoghegan, line 260: Arguably this should be `impl From<Time> for Interval`, but that can't be defined in `janus_core` due to the orphan rule.

aggregator/src/aggregator.rs
```diff
@@ -3223,9 +3224,31 @@ impl VdafOps {
...
+                        .iter()
+                        .map(BatchAggregation::<L, Q, A>::client_timestamp_interval)
+                    {
+                        match spanned_interval {
+                            Some(m) => spanned_interval = Some(m.merge(interval)?),
+                            None => spanned_interval = Some(*interval),
+                        }
+                    }
+
+                    Ok((collection_job, spanned_interval))
```
divergentdave, line 3251: We should round up the interval, either before returning it here or during the aggregation flow. The text says "... the interval's start and duration are both multiples of the task's time_precision parameter." Batch aggregations are initialized with the result of `Interval::from_time(timestamp)`, which has duration 1. (currently, the aggregation tests expect batch aggregations with duration 1 intervals, while the collection tests set up batch aggregations with `time_precision`-length intervals and expect the same in collections)

divergentdave (changes requested):

divergentdave (approved):

Issue #878, opened 2022-12-19
Consolidate command line utilities


divergentdave:
We currently build three different command line tools, (#801 contemplates adding a fourth) and each has a different way of managing its dependency on `clap`.

- `target/debug/janus_cli` is in a binary crate that already uses clap for its other targets.
- `target/debug/dap-decode` uses Cargo's `required-features` binary target setting with a feature that enables an optional clap dependency.
- `target/debug/examples/collect` relies on a dev-dependency on clap.

I suggest we create a new binary crate, say `janus_tools`, to house the `dap-decode` and `collect` binaries (and leave `janus_cli` where it is). This would have the following benefits:

- `dap-decode` could drop the required feature.
- `collect` could be tested with a `trycmd` integration test after being promoted from an example to a binary target.
- So  long as the tools crate doesn't depend on `janus_aggregator`, we could publish it to crates.io and make it `cargo install`-able.

PR #1055, opened 2023-02-28
Move dap_decode and collect to new crate

divergentdave:
This implements the idea in #878, moving two of our CLI utilities into their own crate. It also adds a command line help reference test for `collect`.


branlwyd (approved):
👍🏻 hopefully this helps with compile times, even if just a bit.

tgeoghegan (approved):

PR #1059, opened 2023-03-01
Fix building `janus_core` alone

divergentdave:
This enables the `clock` feature on chrono when building `janus_core`, which is necessary for `RealClock::now()`. Our usual builds are unaffected because dependencies from other workspace crates turn on this feature, and those get unified together.


tgeoghegan (approved):
I'm good with this change, but what did this break? I guess releasing the `janus_core` crate would fail? Anyway, maybe we should also add a step to `ci-build.yml` that builds `janus_core` on its own so we can make sure this doesn't regress.

divergentdave:
I was manually building crates one-at-a-time on the command line. Yes, I think this would have broken releasing the crate.

PR #1062, opened 2023-03-03
Dispatch to VDAF instances using macros

### LINKED ISSUE #1040: Macros to help dispatch on VdafInstance
We have a significant number of large match blocks that match on enums representing VDAFs. (Each arm constructs a struct that implements `Vdaf` from the enum variant's values, and then calls a generic ) With the generalization of CountVec into SumVec, and the possibility of maintaining PrgAes128-based VDAFs for special-case uses alongside the standardized PrgSha3-based VDAFs, these could all get bigger. I think we should probably write a small handful of declarative macros to generate these match blocks and get rid of some repetition.

I forget where I first saw this pattern, but I'm thinking of something like `vdaf_dispatch!(vdaf_instance, vdaf => {handle_request_generic(vdaf)?;})`. This could then expand into a match block, much like those we've written by hand. Each arm's block would first construct the VDAF, assign it to a let block with the identifier given in the pseudo-pattern, and then include the token stream for the block on the right of the `=>`.

Once the VDAF macro is in place, we may want to introduce another macro wrapping it to handle query types as well, but that may not be worth it, depending on how much code is query type-dependent.

divergentdave:
This closes #1040. I've defined macros for `VdafInstance` and `VdafOps`. I had to get a little creative with syntax, because the different generated match arm bodies need to use different types when calling generic functions. Thus, the macro sets up a type alias and a const, using passed-in identifiers as names, which can then be used in the rest of the block to explicitly specify types in turbofishes. (Once we define a macro for query types, it will have to do the same thing, creating different type aliases with the same macro parameter name in each match arm)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1768,6 +1768,7 @@ dependencies = [
  "base64 0.21.0",
  "chrono",
  "derivative",
+ "fixed",
  "futures",
  "hex",
  "hpke-dispatch",
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -879,7 +879,6 @@ impl<C: Clock> TaskAggregator<C> {
 /// VdafOps stores VDAF-specific operations for a TaskAggregator in a non-generic way.
 #[allow(clippy::enum_variant_names)]
 enum VdafOps {
-    // For the Prio3 VdafOps, the second parameter is the verify_key.
     Prio3Aes128Count(
         Arc<Prio3Aes128Count>,
         VerifyKey<PRIO3_AES128_VERIFY_KEY_LENGTH>,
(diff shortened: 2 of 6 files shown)


divergentdave:
Important fact: if `#[cfg()]` conditional compilation attributes show up in a macro's body, they always get included literally in the resulting token stream, and the conditional compilation expressions are only evaluated after macro expansion. Thus, they'll refer to crate features, etc., in the consuming crate, not the crate where it's defined. I'll see if I can rework this by moving conditional compilation up to the `macro_rules!` item, and thus keeping the match block in sync with the enum's definition by referring to the same crate's features.

divergentdave:
This is coming together, nice to see a net change of -1k lines. The third case I need to address is match blocks that match on VdafInstance, but do not then construct the VDAF (only using the type in a generic method call). For what it's worth, many or most of these cases will turn into one of the other cases with the forthcoming prio 0.11 upgrade, since more functions will need a &A parameter in order to decode output shares or aggregate shares.

divergentdave:
This is complete for now, here's an inventory of the other VDAF match blocks I found that are going untouched:

- aggregator/src/aggregator.rs, `TaskAggregator::new()`: Adapts VdafInstance enum to VdafOps enum.
- integration_tests/src/client.rs, `json_encode_vdaf()`: Adapts VdafInstance enum to JSON.
- integration_tests/src/daphne.rs, `daphne_vdaf_config_from_janus_vdaf()`: Adapts VdafInstance to Daphne configuration.
- integration_tests/tests/common/mod.rs, `submit_measurements_and_verify_aggregate()`: Generates a different set of measurements for each VDAF.
- tools/src/bin/collect.rs, `run_with_query()`: handles VDAF-related command line arguments.
- interop_binaries/src/lib.rs, `<VdafObject as From<VdafInstance>>::from(): Adapts VdafInstance enum to JSON model.
- interop_binaries/src/lib.rs, `<VdafInstance as From<VdafObject>>::from(): Adapts JSON model to VdafInstance enum.
- interop_binaries/src/bin/janus_interop_client.rs, handle_upload(): Parses measurement from JSON request body.
- interop_binaries/src/bin/janus_interop_collector.rs, handle_collect_start(): Converts aggregation result into JSON.

Each match that converts from one enum to another should stay. Some of the match blocks in test code could potentially be turned into macro invocations in follow-up work if the per-VDAF behavior was encapsulated in traits, like the existing `InteropClientEncoding`.

tgeoghegan (approved):
You had me at `-2,143`.

PR #1064, opened 2023-03-07
Upgrade prio to 0.11.0

### LINKED ISSUE #932: Client should provide nonce to VDAF input sharding algorithm
See https://github.com/divviup/libprio-rs/issues/422, https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/394, https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/117.

divergentdave:
This upgrades to prio 0.11.0, and thus VDAF-04. This required ancillary changes to thread VDAF objects through more code, to make them available for decoding messages. Serialization used for the database and task parameters changed, because of the new VDAF names. Some traits and structs have an additional const generic now, for the nonce size, which we can fix at 16 as DAP does. I was able to get rid of infectious trait bounds on `Vec<u8>` and the like. Closes #932.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -239,6 +239,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bitvec"
+version = "1.0.1"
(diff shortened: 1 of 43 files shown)


tgeoghegan (approved):
Nice to see those trait bounds get simplified.

branlwyd (approved):

PR #1086, opened 2023-03-08
Deal with DAP-04 aggregation job round

### LINKED ISSUE #994: Deal with DAP-04 round skew recovery
Catch up with the text introduced in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/400

tgeoghegan:
`VdafOps::handle_aggregate_continue_generic` now checks whether the incoming `AggregationJobContinueReq` is a replay of a previously seen message. If it is, then we fetch previously-computed prepare shares and serve them back to the leader (`VdafOps::replay_aggregation_job_round`). If not, then we advance to the next round based on the leader's prepare messages (`VdafOps::step_aggregation_job`).

These new methods (and tests for this behavior) are defined in `mod aggregation_job_continue` to avoid growing the already too large `aggregator/src/aggregator.rs`.

Resolves #994

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_continue.rs
+++ b/aggregator/src/aggregator/aggregation_job_continue.rs
@@ -230,13 +230,9 @@ impl VdafOps {
         A: vdaf::Aggregator<L, 16> + 'static + Send + Sync,
         for<'a> A::PrepareState: Send + Sync + Encode + ParameterizedDecode<(&'a A, usize)>,
     {
-        // DAP suggests that when detecting a replayed aggregation job continuation request, helper
-        // implementations should verify that the current leader request matches whatever request we
-        // previously serviced (i.e., that they contain the same report IDs and the same prepare
-        // messages). The idea is to stop the leader from rewinding VDAF preparation and re-running
-        // it with different parameters. However, we have already committed to the previous request
-        // by updating the relevant report_aggregations rows. If the leader did vary the request
-        // contents, the response we generate here will still be based on the previous request.
+        // TODO(issue #994): verify that the replayed request matches the previous one and error out
+        // if not
+
         let response_prep_steps = report_aggregations
             .iter()
             .map(|report_aggregation| {
@@ -518,6 +514,8 @@ mod tests {
 
     #[tokio::test]
     #[allow(clippy::unit_arg)]
+    // TODO(issue #994): enable this test
+    #[ignore]
     async fn aggregation_job_continue_round_recovery_mutate_continue_request() {
         let test_case = setup_aggregation_job_continue_round_recovery_test().await;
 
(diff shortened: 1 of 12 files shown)


aggregator/src/aggregator/aggregation_job_continue.rs
```diff
@@ -1,6 +1,274 @@
...
+    pub(super) fn replay_aggregation_job_round<C, const L: usize, Q, A>(
+        report_aggregations: Vec<ReportAggregation<L, A>>,
+    ) -> Result<AggregationJobResp, datastore::Error>
+    where
+        C: Clock,
+        Q: AccumulableQueryType,
+        A: vdaf::Aggregator<L, 16> + 'static + Send + Sync,
+        for<'a> A::PrepareState: Send + Sync + Encode + ParameterizedDecode<(&'a A, usize)>,
+    {
+        // TODO(issue #994): verify that the current leader request matches the previously seen one
```
tgeoghegan, line 230: This PR is already quite large, despite my efforts to break it into more digestible changes (#911, #1089), so I am opting to defer implementing this last bit to a subsequent change.

messages/src/lib.rs
```diff
@@ -2077,9 +2078,63 @@ impl<Q: QueryType> Decode for AggregationJobInitializeReq<Q> {
     }
 }
 
+/// DAP protocol message representing the round of an aggregation job.
```
branlwyd, line 2081:
```suggestion
/// Type representing the round of an aggregation job.
```

DAP doesn't define an `AggregationJobRound` message or structure, instead using a raw `u16` field. (We could also drop this type entirely, though I'm OK keeping it as it's nice to have type-safe numeric types.)

branlwyd (changes requested):

messages/src/lib.rs
tgeoghegan: That's a fair hair to split. I prefer keeping `AggregationJobRound` for clarity.

aggregator/src/aggregator/aggregation_job_continue.rs
```diff
@@ -1,5 +1,275 @@
...
+    pub(super) fn replay_aggregation_job_round<C, const L: usize, Q, A>(
+        report_aggregations: Vec<ReportAggregation<L, A>>,
+    ) -> Result<AggregationJobResp, datastore::Error>
+    where
+        C: Clock,
+        Q: AccumulableQueryType,
+        A: vdaf::Aggregator<L, 16> + 'static + Send + Sync,
+        for<'a> A::PrepareState: Send + Sync + Encode + ParameterizedDecode<(&'a A, usize)>,
+    {
+        // DAP suggests that when detecting a replayed aggregation job continuation request, helper
```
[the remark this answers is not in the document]
tgeoghegan: Storing a request hash is the approach I had planned to take from the get-go (which is why that's the approach suggested by the DAP text). I agree about the coherence problem, though what I came around to yesterday is that maybe it's not so bad to provide surprising responses to a leader that's already misbehaving, possibly with malicious intent.

I think what I'll do is take a swing at the request-hash based solution and see how complicated it gets (I think it may not be that bad), and then we can better weigh that against the coherence argument. In the interest of not further growing this PR, I'll go back to what I had a couple commits ago, which is a TODO and `#[ignore]` on the test that exercises this case.

divergentdave (approved):

aggregator/src/aggregator/aggregation_job_continue.rs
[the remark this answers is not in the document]
tgeoghegan: Turns out it was pretty easy to do: #1091

branlwyd (approved):

PR #1091, opened 2023-03-09
Replayed `AggregationJobContinueReq`s can't change

### LINKED ISSUE #994: Deal with DAP-04 round skew recovery

tgeoghegan:
When we successfully handle an `AggregationJobContinueReq`, we now record the SHA-256 hash of the request body. Then, if we later receive another request trying to advance an aggregation job to the round it's already on, we check the hash of the incoming request against the recorded one. If they match, then we replay the previously-computed prepare shares. If they don't, the request fails with HTTP 409 Conflict.

Resolves #994

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -65,6 +65,7 @@ use prio::{
 };
 use regex::Regex;
 use reqwest::Client;
+use ring::digest::{digest, SHA256};
 use std::{
     borrow::Borrow,
     collections::{HashMap, HashSet},
@@ -503,6 +504,8 @@ impl<C: Clock> Aggregator<C> {
         }
 
         let req = AggregationJobContinueReq::get_decoded(req_bytes)?;
+        // unwrap safety: SHA-256 computed by ring should always be 32 bytes
+        let request_hash = digest(&SHA256, req_bytes).as_ref().try_into().unwrap();
 
         Ok(task_aggregator
             .handle_aggregate_continue(
@@ -511,6 +514,7 @@ impl<C: Clock> Aggregator<C> {
                 self.cfg.batch_aggregation_shard_count,
                 aggregation_job_id,
                 req,
+                request_hash,
             )
             .await?
             .get_encoded())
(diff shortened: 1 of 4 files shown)


db/schema.sql
```diff
@@ -106,6 +106,8 @@ CREATE TABLE aggregation_jobs(
     client_timestamp_interval TSRANGE NOT NULL,                -- the minimal interval containing all of client timestamps included in this aggregation job
     state                     AGGREGATION_JOB_STATE NOT NULL,  -- current state of the aggregation job
     round                     INTEGER NOT NULL,                -- current round of the VDAF preparation protocol
+    last_continue_request_hash BYTEA,                          -- SHA-256 hash of the most recently received AggregationJobContinueReq
```
branlwyd, line 109: nit: fixup spacing of columns to match

branlwyd (approved):

aggregator/src/aggregator.rs
```diff
@@ -1825,6 +1835,20 @@ impl VdafOps {
                     // but the leader never got our response and so retried stepping the job.
                     // TODO(issue #1087): measure how often this happens with a Prometheus metric
                     if helper_aggregation_job.round() == leader_aggregation_job.round() {
+                        match helper_aggregation_job.last_continue_request_hash() {
+                            None => {
+                                return Err(datastore::Error::User(Error::Internal(format!(
+                                    "aggregation job {aggregation_job_id} is in round {} but has no last request hash",
```
divergentdave, line 1841: Tiny improvment suggestion: we should check that the continue request does not claim to be for round zero before this, and return some sort of bad request error. This would close off a convoluted route through which an external request could trigger this "internal" error.

divergentdave (approved):

aggregator/src/aggregator.rs
tgeoghegan: Good catch. I think we also want some DAP text specifying that an `AggregationJobContinueReq` with `round = 0` is illegal, so I will try to sneak that into DAP-04.

PR #1113, opened 2023-03-16
Janus Aggregator API implementation.

branlwyd:
Based on Trillium; I was pretty happy with how the library worked out.

Note that I currently return the entire task -- including secrets --
wherever a task is returned (e.g. post task, get task). We had
previously discussed that the secrets would be returned by a different
endpoint. I think this can be changed later, and more than that, I think
the most common case will be adding a new task -- where the control
plane will need the secrets in order to relay them to the co-aggregator.

### DIFF
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -1,26 +1,23 @@
-//! This crate implements the Janus aggregator API.
+//! This crate implements the Janus Aggregator API.
 
 use crate::models::{GetTaskIdsResp, PostTaskReq};
+use async_trait::async_trait;
 use janus_aggregator_core::{
-    datastore::{Datastore, Error},
+    datastore::{self, Datastore},
     task::Task,
     SecretBytes,
 };
 use janus_core::{hpke::generate_hpke_config_and_private_key, time::Clock};
-use janus_messages::{Duration, HpkeAeadId, HpkeKdfId, HpkeKemId, TaskId, Time};
+use janus_messages::{Duration, HpkeAeadId, HpkeKdfId, HpkeKemId, Role, TaskId, Time};
 use models::{GetTaskMetricsResp, TaskResp};
-use opentelemetry::{
-    global::meter,
-    metrics::{Histogram, Unit},
-    Context, KeyValue,
-};
 use querystring::querify;
 use rand::{distributions::Standard, random, thread_rng, Rng};
 use ring::constant_time;
-use std::{future::Future, str::FromStr, sync::Arc, time::Instant};
-use tracing::{info_span, warn, Instrument};
-use trillium::{conn_try, conn_unwrap, Body, Conn, Handler, State, Status};
-use trillium_api::ApiConnExt;
+use std::{str::FromStr, sync::Arc};
+use tracing::{error, info_span, warn, Instrument};
+use trillium::{Conn, Handler, Status};
+use trillium_api::{api, Halt, Json, State};
+use trillium_opentelemetry::metrics;
 use trillium_router::{Router, RouterConnExt};
 
 /// Represents the configuration for an instance of the Aggregator API.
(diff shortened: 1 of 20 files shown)


aggregator_api/src/lib.rs
divergentdave, line 1: We'll also need a binary target for this, (plus container builds) either here or in a follow-up.

divergentdave (approved):

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,897 @@
...
+    )
+}
+
+async fn endpoint_handler<C: Clock, Fut>(
+    mut conn: Conn,
+    endpoint_name: impl Into<String>,
+    handler: impl Fn(Arc<AggregatorApi<C>>, Conn) -> Fut,
+) -> Conn
+where
+    Fut: Future<Output = Conn>,
```
tgeoghegan, line 139: nit: I like to put all the trait bounds either in `<>` or in the `where` clause
```suggestion
async fn endpoint_handler<C: Clock, Fut: Future<Output = Conn>>(
    mut conn: Conn,
    endpoint_name: impl Into<String>,
    handler: impl Fn(Arc<AggregatorApi<C>>, Conn) -> Fut,
) -> Conn
```

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,897 @@
...
+struct AggregatorApi<C: Clock> {
+    ds: Datastore<C>,
+    cfg: Config,
+    response_time_histogram: Histogram<f64>,
+}
+
+impl<C: Clock> AggregatorApi<C> {
+    fn check_auth(&self, conn: Conn) -> Conn {
+        if let Some(authorization_value) = conn.headers().get("authorization") {
+            if let Some(received_token) = authorization_value.as_ref().strip_prefix(b"Basic ") {
```
tgeoghegan, line 178: It doesn't really matter because we control both ends and don't expect this API to generally be interoperable, but I think `Authorization: Bearer <token>` is what we want. `Authorization: Basic` is meant to be used with username/password schemes ([RFC 7617](https://www.rfc-editor.org/rfc/rfc7617)).

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,897 @@
...
+            Ok(req) => req,
+            Err(_) => return conn.with_status(Status::BadRequest).halt(),
+        };
+
+        let vdaf_verify_keys = Vec::from([SecretBytes::new(
+            thread_rng()
+                .sample_iter(Standard)
+                .take(req.vdaf.verify_key_length())
+                .collect(),
+        )]);
```
tgeoghegan, line 231: This implementation assumes that it's our responsibility to generate a task ID and a verify key. I think we need to support being handed those values in the request, or we won't be able to test this flow with two Januses.

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,897 @@
...
+    async fn handle_get_task_ids(&self, conn: Conn) -> Conn {
+        const PAGINATION_TOKEN_KEY: &str = "pagination_token";
+        let lower_bound = querify(conn.querystring())
+            .into_iter()
+            .find(|&(k, _)| k == PAGINATION_TOKEN_KEY)
+            .map(|(_, v)| TaskId::from_str(v))
+            .transpose();
+        let lower_bound = match lower_bound {
+            Ok(task_id_param) => task_id_param,
+            Err(_) => return conn.with_status(Status::BadRequest).halt(),
```
tgeoghegan, line 201: We should log the error, and the same goes for several other places in this file. It'd also be nice to put the error's description in the response body, perhaps as an HTTP problem document, but that can be a future enhancement.

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,897 @@
...
+        };
+
+        let rslt = self
+            .ds
+            .run_tx_with_name("delete_task", |tx| {
+                Box::pin(async move { tx.delete_task(&task_id).await })
+            })
+            .await;
+        match rslt {
+            Ok(()) => conn.ok(Body::default()),
```
tgeoghegan, line 316: I think HTTP 204 No Content is more conventional, especially if we don't want to provide a response body: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE

tgeoghegan (changes requested):

db/schema.sql
```diff
@@ -146,8 +147,8 @@ CREATE TABLE report_aggregations(
     error_code          SMALLINT,                           -- error code corresponding to a DAP ReportShareError value; null if in a state other than FAILED
 
     CONSTRAINT report_aggregations_unique_ord UNIQUE(aggregation_job_id, ord),
-    CONSTRAINT fk_aggregation_job_id FOREIGN KEY(aggregation_job_id) REFERENCES aggregation_jobs(id),
-    CONSTRAINT fk_client_report_id FOREIGN KEY(client_report_id) REFERENCES client_reports(id)
+    CONSTRAINT fk_aggregation_job_id FOREIGN KEY(aggregation_job_id) REFERENCES aggregation_jobs(id) ON DELETE CASCADE,
+    CONSTRAINT fk_client_report_id FOREIGN KEY(client_report_id) REFERENCES client_reports(id) ON DELETE CASCADE
```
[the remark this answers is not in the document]
tgeoghegan: Sounds good! #1145

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,897 @@
...
+                })
+                .await,
+            conn
+        );
+
+        conn.with_json(&TaskResp::from(task.as_ref()))
+    }
+
+    async fn handle_get_task(&self, conn: Conn) -> Conn {
+        let task_id = TaskId::from_str(conn_unwrap!(conn.param("task_id"), conn));
```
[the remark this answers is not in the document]
branlwyd: I think we want to return a `500` if the parameter does not exist: this route being taken _should_ require a path with the `task_id` parameter set; if we arrive at this code while handling a path that does not contain a `task_id` parameter, this indicates programmer error, and is thus an Error, Internal to the Server.

`conn_try!` does the right thing for `Result`s where the error path indicates an ISE. I thought `conn_unwrap!` did the same thing, but unfortunately not -- it only returns the conn, unmodified. I switched to code that uses `conn_try!` instead.

(note that we correctly return a `400 Bad Request` if the parameter is provided, but not handled)

aggregator_api/src/lib.rs
branlwyd: Indeed, but I'm planning to wire this into the aggregator. Let's discuss on that eventual PR.

aggregator_core/src/datastore.rs
```diff
@@ -859,6 +819,58 @@ impl<C: Clock> Transaction<'_, C> {
...
+    /// `lower_bound` parameter to be the last task ID retrieved from the previous call.
+    #[tracing::instrument(skip(self), err)]
+    pub async fn get_task_ids(&self, lower_bound: Option<TaskId>) -> Result<Vec<TaskId>, Error> {
+        let lower_bound = lower_bound.map(|task_id| task_id.as_ref().to_vec());
+        let stmt = self
+            .prepare_cached(
+                "SELECT task_id FROM tasks
+                WHERE task_id > $1 OR $1 IS NULL
+                ORDER BY task_id
+                LIMIT 5000",
```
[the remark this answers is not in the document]
branlwyd: A few hundred KiB is fine for server-to-server communication, especially given that this is going to be between two processes running on GCP's network.

aggregator_core/src/datastore.rs
```diff
@@ -859,6 +819,58 @@ impl<C: Clock> Transaction<'_, C> {
...
+        Ok(Some((
+            row.get_bigint_and_convert("report_count")?,
+            row.get_bigint_and_convert("report_aggregation_count")?,
+        )))
+    }
+
+    /// Retrieves task IDs, optionally after some specified lower bound. This method returns tasks
+    /// IDs in lexicographic order, but may not retrieve the IDs of all tasks in a single call. To
+    /// retrieve additional task IDs, make additional calls to this method while specifying the
+    /// `lower_bound` parameter to be the last task ID retrieved from the previous call.
```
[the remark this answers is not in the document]
branlwyd: We don't need to ensure callers see newly-created tasks -- this code will be used as part of a repeated job (e.g. morally equivalent to a cron job), so if we miss a task on one run of the job, we will just grab it on the next run.

The reason I didn't paginate by primary key is that we decided that primary keys are internal to the datastore layer & should not be exposed; paginating by primary key would effectively require providing an encoding of the primary key to callers that the aggregator API could eventually decode back into the appropriate value to plug into its next SQL query.

[this is somewhat beside the main argument, but even paginating in the way you suggest won't provide callers any guarantee that their list of tasks matches the server's list: a task could be deleted from the portion of the list that has already been returned, leaving the caller's list out-of-sync with the server's even before the last `/GET /task-ids` request.]

aggregator_core/src/datastore.rs
```diff
@@ -488,57 +488,17 @@ impl<C: Clock> Transaction<'_, C> {
         Ok(())
     }
 
-    /// Deletes a task from the datastore. Fails if there is any data related to the task, such as
-    /// client reports, aggregations, etc.
+    /// Deletes a task from the datastore, along with all related data (client reports,
+    /// aggregations, etc).
```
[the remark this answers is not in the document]
branlwyd: I think this comment was based on an old snapshot? In any case, the comment in `datastore.rs` was updated.

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,897 @@
+//! This crate implements the Janus aggregator API.
```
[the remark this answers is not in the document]
branlwyd: This matches the use of `aggregator` for Janus' aggregator implementation: `aggregator_api` is the API attached to `aggregator`.

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,897 @@
...
+use janus_messages::{Duration, HpkeAeadId, HpkeKdfId, HpkeKemId, TaskId, Time};
+use models::{GetTaskMetricsResp, TaskResp};
+use opentelemetry::{
+    global::meter,
+    metrics::{Histogram, Unit},
+    Context, KeyValue,
+};
+use querystring::querify;
+use rand::{distributions::Standard, random, thread_rng, Rng};
+use ring::constant_time;
```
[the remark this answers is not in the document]
branlwyd: We use `ring::constant_time` elsewhere in `janus`, and don't yet have a direct dependency on `subtle` for any of the Janus crates. I also don't like `subtle`'s API: ultimately the `ct_eq` operation gives back a `Choice`, which is really a `u8`-masquerading-as-a-boolean. This is handy if you want to combine these values in a constant-time way, but not as great if you just want the outcome of checking equality as a boolean.

IIRC, `libprio-rs` doesn't depend on `ring` because we have folks that want to use `libprio-rs` who can't depend on `ring`. Since Janus doesn't have that restriction, and switching wouldn't let us drop `ring`, and `ring`'s API is (IMO) better, I think we should keep using `ring` for now.

aggregator_api/src/lib.rs
branlwyd: You're right, though this API as-written is "just" for the initial "create and sync tasks between Janus Control & Data Plane" problem, rather than the harder "sync task from ourselves to a co-aggregator" or "have task sync'ed to us from our co-aggregator" problems. These cross-organization syncs will likely need to allow (almost) every parameter to be specified directly, and I think we'd want to define the appropriate API in the Control Plane first (since that's what the co-aggregators would be talking to).

WDYT about waiting to implement this?

tgeoghegan: Sounds good to me!

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,897 @@
...
+
+async fn endpoint_handler<C: Clock, Fut>(
+    mut conn: Conn,
+    endpoint_name: impl Into<String>,
+    handler: impl Fn(Arc<AggregatorApi<C>>, Conn) -> Fut,
+) -> Conn
+where
+    Fut: Future<Output = Conn>,
+{
+    let start_instant = Instant::now();
```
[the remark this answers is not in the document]
tgeoghegan: Yeah, let's leave this as it is since I don't have any ideas for what else to do.

aggregator_api/src/lib.rs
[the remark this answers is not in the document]
tgeoghegan: Well, I don't have a better name in hand, so let's move forward with this.

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+        let task = Arc::clone(&task);
+        Box::pin(async move { tx.put_task(&task).await })
+    })
+    .await
+    .map_err(|err| {
+        error!(err = %err, "Database transaction error");
+        Status::InternalServerError
+    })?;
+
+    Ok((Json(TaskResp::from(task.as_ref())), Halt))
```
jbr, line 161: It's not necessary to halt in a terminal handler, so this could just be `Ok(Json(TaskResp::from(task.as_ref())))`

aggregator_api/src/lib.rs
```diff
@@ -0,0 +1,814 @@
...
+
+fn instrumented<H: Handler>(handler: H) -> InstrumentedHandler<H> {
+    InstrumentedHandler(handler)
+}
+
+struct InstrumentedHandler<H>(H);
+
+#[async_trait]
+impl<H: Handler> Handler for InstrumentedHandler<H> {
+    async fn run(&self, conn: Conn) -> Conn {
```
jbr, line 356: This might be a footgun in trillium that I should document, but wrapping a handler and only calling run may break contained handlers that expect to have before_send called. The api handler does some stuff in before_send — specifically setting a 200 status if the response body was set and there is not another status set. That behavior might not be necessary, but it is probably safest to delegate (and optionally instrument) `Handler::before_send` as well. You may want to delegate `Handler::name` to the inner type as well

branlwyd: Hm! Given how useful wrapping handlers can be, I think figuring out & documenting the correct semantics for wrapping a handler is pretty important. (or, even further, removing the footguns to make writing a wrapping handler as easy as any other handler)

Is it necessary/useful to override the remaining methods? There are a couple of methods around upgrades, and an init function -- I went ahead and overrode those too.

tgeoghegan (approved):
LGTM, just a couple more things to ponder.

messages/src/lib.rs
```diff
@@ -258,11 +258,11 @@ impl From<[u8; Self::LEN]> for ReportId {
 }
 
 impl<'a> TryFrom<&'a [u8]> for ReportId {
-    type Error = &'static str;
+    type Error = anyhow::Error;
```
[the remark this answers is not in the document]
branlwyd: The breaking change argument is compelling, I reverted this portion of the PR as it is not required for the overall change.

Mostly I wanted to settle on a common error type & avoid unnecessary dynamism; using a "proper" error type also allowed the code to be simplified a bit. (We can use `datastore::Error`, but in that case I strongly advocate for introducing an `Other(anyhow::Error)` and using it for every case where we don't expect callers to make programmatic decisions based on the particular error variant.)

PR #1139, opened 2023-03-21
Reformat long string literals

divergentdave:
This PR was produced with the following commands.

```bash
echo "format_strings = true" > rustfmt.toml
cargo +nightly fmt
```

This breaks up various long string literals with escaped newlines, and thus gets some un-format-able blocks unstuck.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -377,9 +377,11 @@ impl<C: Clock> Transaction<'_, C> {
             aggregator_auth_token_ords.push(ord);
             aggregator_auth_tokens.push(encrypted_aggregator_auth_token);
         }
-        let stmt = self.prepare_cached(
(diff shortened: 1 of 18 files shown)


aggregator_core/src/datastore.rs
tgeoghegan, line 1: I'm not sure if I find the SQL queries more readable with `\` in them and with the funky indentation. Our SQL formatting has been inconsistent, but most of these were laid out by hand for some degree of readability. Maybe we'd be better off with `r#""#r` strings for these? Or we could just put the queries back the way they were.

Ideally we'd somehow run a SQL linter on the queries for consistent style but that's far beyond the scope of this change.

tgeoghegan (changes requested):

tgeoghegan (approved):

PR #1141, opened 2023-03-22
Bump Janus crate versions to 0.4.0

tgeoghegan:
Since we want to deploy a DAP-04 environment imminently, we might as well release 0.4.0, even though our DAP-04 implementation technically isn't totally complete.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1661,7 +1661,7 @@ checksum = "fad582f4b9e86b6caa621cabeb0963332d92eea04729ab12892c2533951e6440"
 
 [[package]]
 name = "janus_aggregator"
-version = "0.3.0"
+version = "0.3.1"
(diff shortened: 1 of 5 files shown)


divergentdave (changes requested):
I think we should either make this `0.4.0-alpha.1` or hold off for the VDAF-05 changes. We'll probably be working with the 0.4 versions for some time to come, and I wouldn't want to have knowingly wire-incompatible 0.4.0 versions of janus_core, janus_client, and janus_collector hanging out in the registry.

tgeoghegan:
On second thought, I suspect we can get the last VDAF-05 change in soon (like today or tomorrow if we want to), so let's not bother with 0.4.0-alpha.1.

divergentdave (approved):

PR #1143, opened 2023-03-23
Update lockfile

divergentdave:
This is a follow-up to #1141 to update the lockfile.


tgeoghegan (approved):
Funny, I was sure I'd done this. Thanks!

PR #1178, opened 2023-03-30
Migrate health check to Trillium

divergentdave:
This rewrites the simple health check HTTP server using Trillium.

cc @jbr

### DIFF
--- a/aggregator/src/binary_utils.rs
+++ b/aggregator/src/binary_utils.rs
@@ -15,7 +15,6 @@ use deadpool::managed::TimeoutType;
 use deadpool_postgres::{Manager, Pool, PoolError, Runtime, Timeouts};
 use futures::StreamExt;
 use git_version::git_version;
-use http::StatusCode;
 use janus_aggregator_core::datastore::{Crypter, Datastore};
 use janus_core::time::Clock;
 use opentelemetry::{Context, KeyValue};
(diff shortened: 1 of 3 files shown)


aggregator/src/binary_utils.rs
```diff
@@ -332,11 +332,17 @@ where
...
 async fn health_endpoint_server(address: SocketAddr) {
-    let filter = warp::path("healthz")
-        .and(warp::get().or(warp::head()).unify())
-        .map(|| warp::reply::with_status(warp::reply::reply(), StatusCode::OK));
-    let server = warp::serve(filter);
-    server.bind(address).await;
+    let router = Router::new().get(
+        "/healthz",
+        |conn: trillium::Conn| async move { conn.ok("") },
+    );
```
jbr, line 338:
```suggestion
    let router = Router::new().get("/healthz", trillium::Status::Ok);
```

jbr: A slightly more concise option with the same behavior. `""` instead of `trillium::Status::Ok` would also work, but be less clear.

aggregator/src/binary_utils.rs
```diff
@@ -332,11 +332,14 @@ where
...
 async fn health_endpoint_server(address: SocketAddr) {
-    let filter = warp::path("healthz")
-        .and(warp::get().or(warp::head()).unify())
-        .map(|| warp::reply::with_status(warp::reply::reply(), StatusCode::OK));
-    let server = warp::serve(filter);
-    server.bind(address).await;
+    let router = Router::new().get("/healthz", trillium::Status::Ok);
+    let handler = (Head::new(), router);
+    trillium_tokio::config()
+        .with_port(address.port())
```
jbr, line 338: I should probably add a `with_socketaddr` that handles this port/host splitting, this seems like common behavior

jbr (approved):

tgeoghegan (approved):

PR #1179, opened 2023-03-31
Convert Prometheus scrape endpoint to use Trillium

divergentdave:
This converts the HTTP server for Prometheus scrapes from Warp to Trillium.

### DIFF
--- a/aggregator/src/metrics.rs
+++ b/aggregator/src/metrics.rs
@@ -285,7 +285,7 @@ mod tests {
 
         match handle {
             MetricsExporterHandle::Prometheus(handle) => handle.abort(),
-            MetricsExporterHandle::Noop => {}
+            _ => unreachable!(),
         }
     }
 }
--- a/aggregator/src/metrics.rs
+++ b/aggregator/src/metrics.rs
@@ -12,14 +12,12 @@ use std::{collections::HashMap, net::AddrParseError, sync::Arc};
 
 #[cfg(feature = "prometheus")]
 use {
-    http::StatusCode,
-    hyper::Response,
     opentelemetry::sdk::metrics::{controllers, processors},
     prometheus::{Encoder, TextEncoder},
-    std::net::SocketAddr,
     std::net::{IpAddr, Ipv4Addr},
     tokio::task::JoinHandle,
-    warp::{Filter, Reply},
+    trillium::{KnownHeaderName, Status},
+    trillium_router::Router,
 };
 
 #[cfg(feature = "otlp")]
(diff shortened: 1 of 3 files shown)


aggregator/src/metrics.rs
```diff
@@ -225,3 +230,67 @@ pub fn install_metrics_exporter(
...
+    use std::net::Ipv4Addr;
+    #[cfg(feature = "prometheus")]
+    use tokio::net::TcpListener;
+
+    #[cfg(feature = "prometheus")]
+    #[tokio::test]
+    async fn prometheus_metrics_pull() {
+        let listener = TcpListener::bind((Ipv4Addr::LOCALHOST, 0)).await.unwrap();
+        let port = listener.local_addr().unwrap().port();
+        drop(listener);
```
tgeoghegan, line 257: Is this an attempt to find an open port for this test's metrics listener to bind? Doing this way is racy. I think we should instead see if we can get Trillium to bind an ephemeral port. The documentation on on its [`Config`](https://docs.rs/trillium-server-common/latest/trillium_server_common/struct.Config.html#method.with_port) doesn't say you can do this, so maybe we need a new feature on Trillium? Or is there another Trillium testing affordance we can use?

tgeoghegan (changes requested):

aggregator/src/metrics.rs
divergentdave: Yeah, though we are getting away with this race elsewhere in tests. I was able to get the port number from the `Info` argument passed to `Handler::init()` instead.

aggregator/src/metrics.rs
```diff
@@ -131,10 +138,61 @@ impl AggregatorSelector for CustomAggregatorSelector {
...
+}
+
+#[cfg(feature = "prometheus")]
+#[trillium::async_trait]
+impl Handler for GrabPort {
+    async fn run(&self, conn: Conn) -> Conn {
+        conn
+    }
+
+    async fn init(&mut self, info: &mut Info) {
```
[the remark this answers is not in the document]
divergentdave: Ah, my mental block was that `Init` had to return another handler, which I didn't need, but I suppose `()` is also a handler. That'd be simpler.

aggregator/src/metrics.rs
jbr: I suppose there is still a race condition with portpicker though, but I seem to remember some issue with binding to 0 on windows

tgeoghegan: That's alright, Janus CI doesn't build or test on anything but Linux x86_64, and so anything else is not supported.

tgeoghegan (approved):

PR #1181, opened 2023-03-31
Convert janus_interop_client and janus_interop_collectors to use Trillium

divergentdave:
This rewrites two of the three interop API binaries to use Trillium. `janus_interop_aggregator` is closely tied to the aggregator (it reuses its Warp filter), so those two should be cut over at the same time, to finish out the migration. (otherwise, we'd have to briefly adapt between reqwest and response types)


interop_binaries/Cargo.toml
```diff
@@ -51,6 +51,10 @@ tokio = { version = "1.26", features = ["full", "tracing"] }
 tracing = "0.1.37"
 tracing-log = "0.1.3"
 tracing-subscriber = { version = "0.3", features = ["std", "env-filter", "fmt"] }
+trillium = "0.2.8"
+trillium-api = { version = "0.2.0-rc.2", default-features = false }
+trillium-router = "0.3.5"
+trillium-tokio = "0.2.1"
```
tgeoghegan, line 57: Sorta orthogonal to this change, but: let's put all the trillium deps in the workspace `Cargo.toml` and reference those from each crate.

tgeoghegan (approved):

PR #1218, opened 2023-04-05
Timg/sqlx schema migrations

tgeoghegan:


tgeoghegan:
I need to think more about whether we this change needs to be done in multiple stages to avoid breaking `janus-ops` (maybe we should continue to distribute `schema.sql` alongside the migration scripts), but it's at least ready for review.

divergentdave (approved):

tgeoghegan:
I concluded that there's no reason to force a lockstep update between this repo and `janus-ops`, so I reverted some part of this change and broke it out into #1219, which we can merge only once we are sure things are migrated and working correctly downstream.

Issue #30, opened 2022-03-22
DB schema migrations


tgeoghegan:
We will need to change the database schema over time and so will need to figure out a way to handle versioning, migrations, etc. There should be some product or middleware we can use for this. We can probably make it through early experimental phases by just throwing away databases instead of migrating them, though.

BranLwyd:
I found a couple of potential solutions. These are both free & open-source with a paid support option, and support Postgres:

* [Flyway](https://flywaydb.org)
* [Liquibase](https://www.liquibase.com)

I haven't evaluated these on technical merits, popularity, or anything else.

BranLwyd:
Both of these expect the user to produce "migration scripts", i.e. they do not take a full DB schema and diff it against an existing schema. Flyway works with SQL scripts directly; Liquibase can do this too, but most of the more advanced features (such as rollback) require use of a custom markup language (XML, YAML, JSON are supported). The use of a custom markup language would make it pretty difficult to migrate to a different solution if Liquibase doesn't work out.

Note Liquibase does have a [diff-changelog](https://docs.liquibase.com/commands/diff/diff-changelog.html) command which can produce a changelog based on the diffs between two databases -- but it works on live DBs, rather than schemas, AFAICT.

There are some tools like pgdiff, pgadmin, ... which can produce diffs on schemas, but they don't support CI/CD integration directly, so we'd need to build something.

BranLwyd:
The free version of Flyway does not support "undo", which IMO is disqualifying. (Liquibase's free offering does support undo.) Assuming we want to start with a free option, IMO Liquibase (or something else entirely) is appropriate.

The apparently-most-popular alternative after Liquibase/Flyway appears to be [Bytebase](https://www.bytebase.com). Bytebase appears to be a web application allowing editing & deploying of SQL migration scripts. It also does not appear to have prebuilt CI/CD support. I suspect it's not a good fit for what we hope for.

[Yuniql](https://yuniql.io) is another option. It is effectively a CLI tool which seems to be easy enough to integrate into a CI/CD pipeline. It does use SQL scripts to define the migrations (as opposed to a tool-specific format). However, it appears to be a much less notable project than Liquibase, and it runs on .NET Core which may be slightly more difficult to deploy to our CI/CD pipeline.

divergentdave:
[...]

PR #1221, opened 2023-04-06
Adopt `sqlx migrate`

tgeoghegan:
We now use `sqlx` ([1]) and its CLI ([2]) to manage our SQL schema and migrate it to new versions. The schema is now represented as a series of SQL scripts in `db/`, along with teardown scripts to enable rollbacks. The objective here is to enable easy migrations, namely:

 - adding new tables
 - adding indices to tables
 - adding columns with default values to tables

More complex migrations which transform or delete data may require more sophisticated tools than what's here, and even these "easy" migrations require some planning and operational tools that aren't present in this repository.

This commit contains these changes:

 - the `publish-sql-schema` workflow now publishes all migration scripts in `db` to a Google Cloud Storage bucket to later be consumed by Divvi Up ops tools
 - the interop test containers use `sqlx::migrate::Migrator` instead of `janus_cli` to write the schema into their database
 - `janus_aggregator_core::datastore::test_util::ephemeral_dataastore` uses `sqlx::migrate::Migrator` to apply migrations to ephemeral databases
 - `db/schema.sql` is copied to a `sqlx` migration script, and a corresponding `down` script is added which deletes tables, indices, etc.

[1]: https://crates.io/crates/sqlx
[2]: https://crates.io/crates/sqlx-cli

Part of #30


tgeoghegan:
This is a backport of #1218. The goal here is for our downstream test and deploy pipelines to be able to handle Janus 0.2 the same way as 0.4.

divergentdave (approved):

PR #1232, opened 2023-04-10
Bump crate versions to 0.4.4

tgeoghegan:


divergentdave (approved):

Cargo.toml
```diff
@@ -20,7 +20,7 @@ homepage = "https://divviup.org"
 license = "MPL-2.0"
 repository = "https://github.com/divviup/janus"
 rust-version = "1.65.0"
-version = "0.4.2"
+version = "0.4.4"
```
branlwyd, line 23: Q: why no `0.4.3`?

branlwyd (approved):

Cargo.toml
tgeoghegan: I did 0.4.3 yesterday (https://github.com/divviup/janus/releases/tag/0.4.3) in order to test shipping the SQL migrations to Google Cloud Storage, but I forgot to bump the crate versions when I did (which didn't matter for the test I was doing).

PR #1235, opened 2023-04-11
Use Trillium in aggregator and janus_interop_aggregator

### LINKED ISSUE #1090: Migrate from `warp` to Trillium
The Janus control plane is being built with [Trillium](https://trillium.rs/). For consistency across our projects and also becuase we have some gripes with `warp`, `janus` should migrate to Trillium.

divergentdave:
This finishes and closes #1090, replacing Warp with Trillium in the aggregator and the corresponding interop binary.

Here's an overview of the `trillium::Handler` implementations in play, in the order they appear in tuple handlers.

- `trillium::Init`: Similar to a previous case, we need to grab the server's socket address upon initialization and send it over a oneshot channel, so the `aggregator_server()` function can return it to tests.
- `trillium::Headers`: This adds headers to all responses. We need this for HSTS, until we update our GKE version, and can do this in the load balancer easily.
- `trillium_opentelemetry::metrics`: This records response times and sizes in histograms, following the OTel semantic conventions.
- `trillium_router::Router`: This is straightforward, we use it to dispatch on method and path. Later handlers pull the `Captures` out of connection state to get various IDs from the path, and a `RouteSpec` for logging and metrics.
  - `trillium_api::ApiHandler`: Most of the routes are implemented with trillium_api, and the `FromConnHandler` functions return a `Result` to switch between success and error behaviors (see below). The handlers are also wrapped with `janus_aggregator_api::instrumented`, to take care of creating a tracing span and entering it on each poll. (the span is annotated with the route)
    - `EncodedBody<T>`: This takes a message object, encodes it, and sends that as the body. It also handles setting a Content-Type header.
    - `Error`: Depending on the error type, this either responds with a status code and no body, or a problem details response. (including Content-Type header, JSON body, and appropriate status code) This also stores a label value in the connection state, to be used later, indicating which type of error was seen.
- `StatusCounter`: This checks for an `ErrorCode` stashed in the state by the `Error` handler implementation, if it ran earlier, and increments a metrics counter. There are labels for the method, route, status code, and error code. This happens in `before_send()`, so it sees outgoing responses, whether or not its no-op `run()` was hit.


aggregator/src/aggregator.rs
```diff
@@ -590,17 +603,17 @@ impl<C: Clock> Aggregator<C> {
             .handle_delete_collection_job(&self.datastore, collection_job_id)
             .await?;
 
-        Ok(StatusCode::NO_CONTENT.into_response())
+        Ok(())
```
tgeoghegan, line 606: Does this still return HTTP 204?

tgeoghegan (commented):
I don't think I'm sufficiently up to speed on Trillium to review this in-depth, but I have some questions and comments.

`aggregator/src/aggregator.rs` is a _very_ big file. Since this change touches a lot of it anyway, I think we should take the opportunity to break it out into modules. Could the types and logic for working with problem documents be put into another file? Same goes for the stuff that sets up the Trillium handler(s). It'd be nice if `aggregator.rs` was more focused on `struct Aggregator`, and we had other files/modules focused on doing HTTP stuff.

aggregator/src/aggregator.rs
divergentdave: Yes, I moved that to the end of the `collection_jobs_delete()` function.

aggregator/src/aggregator.rs
```diff
@@ -2751,375 +2650,481 @@ where
...
+/// API handler for the "/tasks/.../reports" PUT endpoint.
+async fn upload<C: Clock>(
+    conn: &mut Conn,
+    (State(aggregator), State(captures), body): (
+        State<Arc<Aggregator<C>>>,
+        State<Captures<'static, 'static>>,
+        Vec<u8>,
+    ),
+) -> Result<Status, Arc<Error>> {
+    validate_content_type(conn, Report::MEDIA_TYPE).map_err(Arc::new)?;
```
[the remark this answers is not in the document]
branlwyd: We could write a handler-wrapper function, similar to `instrumented`, that could provide this.

aggregator/src/aggregator.rs
```diff
@@ -2751,375 +2650,481 @@ where
...
+        )
+    }
+}
 
-/// CollectionId is 16 bytes, hence 22 bytes of Base64URL
-const COLLECTION_JOB_PATH_REGEX: &str = "collection_jobs/[A-Za-z0-9-_]{22}";
+#[async_trait]
+impl Handler for StatusCounter {
+    async fn run(&self, conn: Conn) -> Conn {
+        conn
```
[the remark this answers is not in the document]
divergentdave: I'm intentionally not following that pattern here, because I want the code to run whether or not a preceding handler set the halt flag. (thus it will unconditionally bump the counter) In fact, I think the halt flag will be set by every handler inside the router, and only unset for 404s that didn't match any route. This handler isn't participating in construction of a response, so we don't need to coordinate with other handlers through the halt flag.

aggregator/src/aggregator.rs
```diff
@@ -2751,375 +2650,481 @@ where
...
     listen_address: SocketAddr,
-    response_headers: HeaderMap,
+    response_headers: Headers,
     shutdown_signal: impl Future<Output = ()> + Send + 'static,
 ) -> Result<(SocketAddr, impl Future<Output = ()> + 'static), Error> {
-    let filter = aggregator_filter(datastore, clock, cfg)?;
-    let wrapped_filter = filter.with(warp::filters::reply::headers(response_headers));
-    let server = warp::serve(wrapped_filter);
-    Ok(server.bind_with_graceful_shutdown(listen_address, shutdown_signal))
+    let stopper = Stopper::new();
```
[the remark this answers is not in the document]
divergentdave: I agree, but I think this might be good material for a follow-up PR.

aggregator/src/aggregator.rs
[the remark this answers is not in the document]
branlwyd: I think the idea is that the handler's post-handling in `before_send` doesn't run if the handler halts _before_ the handler's initial handling in `run` occurs -- simulating a "wrapper" handler, I suppose. We definitely want to run `before_send` if the connection (expectedly) halts "inside" the `StatusCounter`.

aggregator/src/aggregator.rs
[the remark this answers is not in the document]
divergentdave: I think a plain handler would work in this case, it could set the status to 400 and set `halted` if the header is missing or incorrect. Then, we could change the `instrumented(api(...))` handlers inside the router to `(check_header("..."), instrumented(api(...)))`. Whether to do this check in the `trillium` world as a handler, or in the `trillium-api` world with traditional error handling is largely a question of style, I think.

aggregator/src/aggregator.rs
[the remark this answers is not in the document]
jbr: When the stopper is stopped, trillium will cease accepting new tcp connections. Open keepalive connections that are between requests are disconnected (dropped), and any open websockets are closed. The [`trillium_tokio signals behavior`](https://github.com/trillium-rs/trillium/blob/main/tokio/src/server.rs#L33-L50) is to stop accepting new requests on the first SIGINT, SIGTERM, or SIGQUIT and wait *indefinitely* for any mid-request conns to complete. A subsequent SIGINT, SIGTERM, or SIGQUIT will abruptly cancel any misbehaving or long lived connections and non-gracefully (gracelessly?) terminate the process.

aggregator/src/aggregator.rs
```diff
@@ -2751,375 +2647,481 @@ where
...
 
-    Ok(filter)
+#[async_trait]
+impl<T> Handler for EncodedBody<T>
+where
+    T: Encode + Sync + Send + 'static,
+{
+    async fn run(&self, mut conn: Conn) -> Conn {
+        conn.headers_mut()
+            .insert(KnownHeaderName::ContentType, self.media_type);
```
jbr, line 2674: Out of curiosity, is this explicit `insert` (vs `with_header`) because it was appending and that wasn't the desired behavior? Asking because I've been wondering if `with_header` should insert by default

divergentdave: Ah, chalk this one up to path dependence. That behavior is not an issue here, because the different places content-type are set are exclusive.

aggregator/src/aggregator.rs
```diff
@@ -2751,375 +2647,481 @@ where
...
+        metrics("janus_aggregator"),
+        Router::new()
+            .get("hpke_config", instrumented(api(hpke_config::<C>)))
+            .with_route(
+                trillium::Method::Options,
+                "hpke_config",
+                hpke_config_cors_preflight,
+            )
+            .put("tasks/:task_id/reports", instrumented(api(upload::<C>)))
+            .with_route(
```
jbr, line 2756: The router has default options handling that we may want to disable: `Router::new().without_options_handling().get…` [`Router::without_options_handling`](https://docs.rs/trillium-router/latest/trillium_router/struct.Router.html#method.without_options_handling)

aggregator/src/aggregator.rs
jbr: I just double checked and it actually *does* insert — I already addressed that. I think it's identical to `conn.with_header(KnownHeaderName::ContentType, self.media_type)` here

aggregator/src/aggregator.rs
```diff
@@ -2751,375 +2647,481 @@ where
...
+    State(aggregator): State<Arc<Aggregator<C>>>,
+) -> Result<EncodedBody<HpkeConfigList>, Error> {
+    let query = serde_urlencoded::from_str::<HpkeConfigQuery>(conn.querystring())
+        .map_err(|err| Error::BadRequest(format!("couldn't parse query string: {err}")))?;
+    let hpke_config_list = aggregator
+        .handle_hpke_config(query.task_id.as_ref().map(AsRef::as_ref))
+        .await?;
+
+    conn.headers_mut()
+        .insert(KnownHeaderName::CacheControl, "max-age=86400");
```
jbr, line 2809: An option here would be to use [`trillium_caching_headers::CacheControlDirective::MaxAge`](https://docs.rs/trillium-caching-headers/latest/trillium_caching_headers/enum.CacheControlDirective.html)

aggregator/src/aggregator.rs
```diff
@@ -3378,25 +3381,32 @@ mod tests {
...
-            .into_response();
+        let mut test_conn = get(&format!("/hpke_config?task_id={}", task.id()))
+            .run_async(&handler)
+            .await;
 
-        assert_eq!(response.status(), StatusCode::OK);
+        assert_eq!(test_conn.status(), Some(Status::Ok));
         assert_eq!(
-            response.headers().get(CACHE_CONTROL).unwrap(),
+            test_conn
```
jbr, line 3390: a slightly more concise alternative here would be to use [`assert_headers`](https://docs.trillium.rs/trillium_testing/macro.assert_headers.html) like 
```rust
assert_headers!(conn,
    "cache-control" => "max-age=86400",
    "content-type" => HpkeConfigList::MEDIA_TYPE)
```

interop_binaries/src/bin/janus_interop_aggregator.rs
```diff
@@ -126,61 +128,43 @@ fn make_filter(
...
-                        status: SUCCESS.to_string(),
-                        error: None,
-                    },
-                    Err(e) => AddTaskResponse {
-                        status: ERROR.to_string(),
-                        error: Some(format!("{e:?}")),
+        .post(
+            "internal/test/add_task",
+            (
+                State(datastore),
```
jbr, line 144: In this case it may be more readable just to move a clone of the datastore and keyring into the closure. Putting stuff in state is primarily useful when it would be otherwise onerous to wire it throughout the app in a bunch of places. Coincidentally moving a clone would slightly more efficient as well, but the primary consideration here is readability

PR #1238, opened 2023-04-12
Use no secrets in `ci-build`

tgeoghegan:
The `ci-build` workflow is not runnable from forks. This is because it wants to authenticate to our private Google Artifact Repository in order to fetch Daphne container images. However, those tests are disabled, and going forward we'll test Daphne using the generic interop test interface and using containers that team publishes to DockerHub. So we should have no need to authenticate to private repositories and thus use secrets. This commit removes mention of `DAPHNE_INTEROP_CONTAINER` from `ci-build.yml` and deletes the steps that authenticate to GCP and the Docker repository it hosts. This should make it possible for PRs on forks to run CI.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -27,7 +27,6 @@ jobs:
     env:
       CARGO_INCREMENTAL: 0
       CARGO_TERM_COLOR: always
-      DAPHNE_INTEROP_CONTAINER: prebuilt=${{ secrets.DAPHNE_PREBUILT_IMAGE_NAME_AND_TAG }}
       RUSTFLAGS: "-D warnings"
     steps:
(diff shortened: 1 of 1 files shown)


divergentdave (approved):

tgeoghegan:
Same changes, but in my personal fork (hopefully the fact that my user is also privileged on `divviup/janus` doesn't affect secret availability): https://github.com/divviup/janus/pull/1239

branlwyd (approved):

tgeoghegan:
This works, it also worked on my fork, and it didn't appear to regress build times. We should also expunge mention of `DAPHNE_INTEROP_CONTAINER` from the codebase and delete the custom code for managing Daphne, but I want to take this just to unblock #1220.

PR #1244, opened 2023-04-12
fix a test failure on macos

jbr:
Previous to this commit, janus tests failed on macos with

```
---- tests::poll_timing stdout ----
thread 'tests::poll_timing' panicked at 'overflow when adding duration to instant', library/std/src/time.rs:408:33
```

This possibly has to do with [OS-specific behaviors of std::time::Instant](https://doc.rust-lang.org/std/time/struct.Instant.html#os-specific-behaviors).

As a fix, this commit does a checked_add before the compare, which is likely safer anyway on all operating systems, but at the very least resolves the macos test failure.

Would it make sense to add a github action to run tests on macos to avoid future regressions?

### DIFF
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -616,7 +616,11 @@ impl<V: vdaf::Collector> Collector<V> {
                 // Check if sleeping for as long as the Retry-After header recommends would result
                 // in exceeding the maximum elapsed time, and return a timeout error if so.
                 if let Some(deadline) = deadline {
-                    if Instant::now() + retry_after_duration > deadline {
+                    let recommendation_is_past_deadline = Instant::now()
+                        .checked_add(retry_after_duration)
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):
LGTM. I'm not sure about adding macOS tests: those runners are very expensive, and it's not clear to me that we want to pretend to support anything besides Linux with Janus. Then again you're not the only person who will want to develop it on a Mac. The real problem is supporting arm64 Macs, since GitHub doesn't have runners for those (yet).

branlwyd (approved):
Thank you! I'd also prefer not to enable MacOS runners at the moment: MacOS runners are exorbitantly expensive (10x as many runner-minutes as Linux, iirc), and our usage is definitely Linux-centric. (we'll definitely accept OS-specific test fixes, of course)

PR #1254, opened 2023-04-17
Fix race between aggregation & collection.

branlwyd:
The basic strategy is to introduce a new "batch aggregation state"
column, which records the current state of the batch aggregation. The
accumulator notes batch aggregations which have been collected & refuses
to aggregate into them. If an aggregation occurs concurrently with a
collection, the aggregation process' write to the batch_aggregations row
(to update the share/count/checksum) will conflict with the collection
process' write to the batch_aggregations row (to update the state),
incurring a retry -- this gives a DB-level protection against races
here.

On collection, we additionally write any relevant batch aggregation rows
which do not yet exist; this protects against the corner case that some
batch aggregation shard didn't get written at all, which would dodge the
write conflict we depend on to protect against race conditions.

Note that this change does not attempt to address the semantics of "when
is the earliest time we can give up on a report share because it is for
a batch that has been collected" -- we might (wastefully) aggregate such
reports; we are just now guaranteed that they won't be merged into an
aggregate share which has been collected, throwing their contribution
away instead. This also does not guarantee that the Leader won't send an
aggregate share request concurrently with an aggregation request which
finishes an aggregation -- this could cause the Leader & Helper to fall
out-of-sync in terms of which reports are included in the batch. (I
think this concurrency issue can still occur, though it is unlikely
since an aggregation job would need to be created & completed in the
time it takes for a single aggregate share request to be processed.)


aggregator/src/aggregator/aggregate_share.rs
```diff
@@ -54,8 +54,14 @@ pub(crate) async fn compute_aggregate_share<
         total_report_count += batch_aggregation.report_count();
 
         match &mut total_aggregate_share {
-            Some(share) => share.merge(batch_aggregation.aggregate_share())?,
-            None => total_aggregate_share = Some(batch_aggregation.aggregate_share().clone()),
+            //Some(share) => share.merge(batch_aggregation.aggregate_share())?,
```
tgeoghegan, line 57:
```suggestion

```

tgeoghegan (commented):
Posting some initial basic thoughts on the change. I want to spend more time thinking about this and will review again tomorrow.

aggregator/src/aggregator.rs
```diff
@@ -2214,6 +2243,40 @@ impl VdafOps {
                                 )
                             )?;
 
+                            // To ensure that concurrent aggregations don't write into a
+                            // currently-nonexistent batch aggregation, we write (empty) batch
+                            // aggregations for any that have not already been written to storage.
```
tgeoghegan, line 2248: I think I get why this is important for the time interval case. Suppose I have contiguous batch aggregations:
```
    b1         b2       b3
|---------|---------|---------|
t0       t2         t3        t4
```

And suppose there are some reports in `b1` and `b3` but none in `b2`. If we service a collection over `[t0, t4)`, then `batch_aggregations` will be `[b1, b2, b3]`, and we'll construct a collection over the reports in `b1` and `b3`. We write out an empty, collected batch aggregation for `b2` to ensure we later will refuse to collect the interval `[t2, t3)`.

But am I right in thinking that for the fixed size query case, this shouldn't be necessary? In that case `Q::get_batch_aggregations_for_collect_identifier` will always yield a single batch aggregation, right?

I'm not proposing any changes here, just want to make sure I understand what this does.

aggregator/src/aggregator/aggregation_job_continue.rs
```diff
@@ -230,14 +230,30 @@ impl VdafOps {
...
+                    .clone()
+                    .with_state(ReportAggregationState::Failed(
+                        ReportShareError::BatchCollected,
+                    ))
+                    .with_last_prep_step(Some(PrepareStep::new(
+                        *report_aggregation.report_id(),
+                        PrepareStepResult::Failed(ReportShareError::BatchCollected),
+                    )));
+            }
+        }
```
tgeoghegan, line 248: What if _all_ the reports in the aggregation job are now failed with `BatchCollected`. Should the state of the aggregation job go to `AggregationJobState::Finished`?

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -678,41 +678,56 @@ impl AggregationJobDriver {
...
+                                                ReportShareError::BatchCollected,
+                                            ))
+                                            .with_last_prep_step(Some(PrepareStep::new(
+                                                *ra.report_id(),
+                                                PrepareStepResult::Failed(
+                                                    ReportShareError::BatchCollected,
+                                                ),
+                                            ))),
+                                    )
+                                    .await
```
tgeoghegan, line 725: Similar observation as in the helper/`aggregation_job_continue.rs` case: should we update the aggregation job state to finished if all reports are failed with `BatchCollected`?

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -200,9 +214,50 @@ impl CollectionJobDriver {
...
+                        None,
+                        0,
+                        Interval::EMPTY,
+                        ReportIdChecksum::default(),
+                    ))
+                } else {
+                    None
+                }
+            })
+            .collect(),
```
tgeoghegan, line 251: nit: this does basically the exact same thing as a similar codepath on the Helper side in `aggregator.rs` and so maybe could be factored out. I'm not sure what you'd hang the new method on, though.

tgeoghegan (changes requested):
Besides the questions I put inline, I think this needs more test coverage. This stuff is quite subtle, so I worry about the risk of regressions. It should be possible to construct tests that fill a database with the appropriate state and then try to step aggregation jobs, collections etc. I suspect some of the tests I wrote in https://github.com/divviup/janus/pull/1180 would help.

aggregator/src/aggregator.rs
divergentdave: I don't think this is so much for future collections over `[t2, t3)` as it is for concurrent aggregations landing in that interval. By responding to an absence of a row with insertion of a "zero" row, we are making phantom read anomalies irrelevant, by forcing the issue and writing relevant rows. If there's a concurrent aggregation at the same time as this larger collection, then either the collection wins, and the zero batch aggregation gets written, filling in the hole, and the aggregation retries, then gets a batch collected error, or the aggregation wins, writing a non-zero batch aggregation into the hole, so the collection retries, reads the now-finished aggregation in its new snapshot, and succeeds.

Without this write, phantom read anomalies mean that we can't be sure from a collection transaction that we see all report aggregations and batch aggregations, thus it's possible we are seeing a unique view of a batch that is missing certain reports compared to future transactions, breaking our replay protections.

db/20230420210648_batch-aggregation-state.up.sql
```diff
@@ -0,0 +1,10 @@
+-- We create a new state for batch_aggregations.
+-- AggregateShare is now optional (if there are no reports), so we drop the NOT NULL constraint.
+
+CREATE TYPE BATCH_AGGREGATION_STATE AS ENUM(
+    'AGGREGATING',  -- this batch aggregation has not been collected & permits further aggregation
+    'COLLECTED'     -- this batch aggregation has been collected & no longer permits aggregation
+);
+
+ALTER TABLE batch_aggregations ADD COLUMN state BATCH_AGGREGATION_STATE NOT NULL;  -- the current state of this batch aggregation
```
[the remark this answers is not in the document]
branlwyd: Hmm... I don't think it makes a difference.

Without `DEFAULT 'AGGREGATING'` (or some other default value), this migration can't be applied if there are any `batch_aggregations` rows since we would be trying to create a new, non-nullable column with no default value.

But `DEFAULT 'AGGREGATING'` is not correct for a system which already has batch aggregations; for correctness, we would need already-collected aggregations to move to state 'COLLECTED'. But we'd need to implement a data migration to make this happen.

Since we are currently not implementing data migrations for non-backwards-compatible schema changes, I'd prefer not specifying a default value: the main application code doesn't need it; it might hypothetically mask bugs in application code; and the lack of a default can be viewed as providing a safeguard against applying this migration to a system with data, since it will fail.

aggregator/src/aggregator/aggregation_job_driver.rs
branlwyd: Good catch, in both locations.

The helper implementation was fixable by just moving some code around; fixing things here required slightly more effort due to needing to modify things which must be in an `Arc` since they are passed into a transaction handler. The most complicated bit was using a `Cow` to avoid copying report aggregations in the (overwhelmingly common) case that there is no race condition.

aggregator/src/aggregator.rs
branlwyd: You have the right idea as to what batch aggregations end up written; David has the right idea as to why. We need an explicit row to exist in order to trigger a write-conflict failure; writing "empty" batch aggregations provides these rows.

This could still impact fixed size in a few ways:
* Batch aggregation rows for a given batch are sharded; through bad luck or low report volume, some batch aggregation shards might end up with no reports assigned to them, and would therefore not be written by the time this code runs. The empty-batch-aggregations functionality would fill in rows for any such shards.
* A Collector could query for a batch that has no reports, as is done in your initial example (`[t2, t3]` is a batch with no reports). This affects both query types, though I think triggering this behavior for fixed-size queries would require some extra-DAP backchannel to communicate the relevant batch ID to the collector, and even then I suppose we'd need to have `max_batch_size = 0` (...and to fix up our code to support setting `max_batch_size = 0`...) for the collection to actually succeed. I think this one is mostly hypothetical, at the moment, for fixed-size.

tgeoghegan: Ok, thanks for the explanations. At the risk of introducing comments that might get out of sync with the code, should all this be documented, since deleting or modifying this code could cause subtle problems that won't manifest obviously in tests?

docs/samples/collection_job_driver_config.yaml
```diff
@@ -75,3 +75,6 @@ worker_lease_clock_skew_allowance_secs: 60
 # Number of times to attempt processing an collection job before abandoning it.
 # (required)
 maximum_attempts_before_failure: 10
+
+# Number of sharded database records per batch aggregation.
```
divergentdave, line 79: It might be worth adding a comment to the various files saying that this setting needs to match across the configuration of different components.

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -217,16 +247,20 @@ impl CollectionJobDriver {
 
                     match maybe_updated_collection_job.state() {
                         CollectionJobState::Start => {
-                            tx.update_collection_job::<SEED_SIZE, Q, A>(&collection_job).await?;
-                            tx.release_collection_job(&lease).await?;
+                            try_join!(
+                                tx.update_collection_job::<SEED_SIZE, Q, A>(&collection_job),
+                                try_join_all(batch_aggregations.iter().map(|ba| tx.update_batch_aggregation(ba))),
```
divergentdave, line 252: I'm concerned that this update could stomp on concurrent changes from a concurrent aggregation flow, running between the `step_collection_job_1` and `step_collection_job_2` transactions. The intent of this update is to flip the batch aggregation's state enum, but if the aggregate share/report count/checksum/timestamp interval have been updated in the database, we'd lose those changes, when we should instead retry in hopes of getting a consistent view of the batch. Our leasing system only protects changes to the collection job, not batch aggregations. (the below `put_batch_aggregation()` call should be okay, because any concurrently created batch aggregation rows would result in an error, and an eventual retry at the job driver level)

aggregator/src/aggregator.rs
branlwyd: We now have tests that check the expected interaction between aggregation & collect jobs. What documentation would you like to see? The remainder of the behavior is standard Postgres behavior, and we document in code [why we trigger retries](https://github.com/divviup/janus/blob/3d15c87f7051acca0416c7948410b5076c958cc7/aggregator/src/aggregator/accumulator.rs#L118-L140) in some cases. We might also trigger retries if two writers attempt to write into the same already-existing batch aggregation row, but that's a standard serialization failure and will be true for all interactions with our DB.

aggregator/src/aggregator.rs
```diff
@@ -2214,6 +2243,40 @@ impl VdafOps {
...
+                            let existing_batch_aggregations: HashSet<_> = batch_aggregations
+                                .iter()
+                                .map(|ba| (ba.batch_identifier(), ba.ord()))
+                                .collect();
+                            let empty_batch_aggregations: Vec<_> = iproduct!(
```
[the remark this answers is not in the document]
branlwyd: I agree, but I think this is beyond the scope of this PR. This was already a problem: a too-small time precision would cause a collect request of a few hours to attempt to read many thousands of rows.

aggregator/src/aggregator.rs
tgeoghegan: Well, what I'm worried about is someone in the future deleting or re-ordering this code, or moving it to another transaction because they don't realize the importance of _where_ it happens. So perhaps a paragraph explaining that these empty batch aggregations need to be written out in this particular transaction?

aggregator/src/aggregator/collection_job_driver.rs
branlwyd: I moved the update of the batch aggregation state to the first transaction to avoid this possibility.

db/20230420210648_batch-aggregation-state.up.sql
divergentdave, line 1: We need to reassign the migration a new date, since the `20230424220336_rm-out-share` landed before this, and our migrations ought to be append-only when sorted by timestamp.

branlwyd: I think we don't need to do this, as the `20230424220336_rm-out-share` migration hasn't been released yet, so both migrations will be released as part of the same Janus version. That they will be applied out of merge order will not cause a problem -- our tests show this as they run in this order.

You highlight a real problem, however, if a release was involved. Something like the following can occur:
* PR A introduces a migration, and ends up stuck in review for a while.
* PR B introduces a separate migration (dated after PR A's migration); this PR is merged quickly.
* A Janus release occurs; PR B's migration is deployed.
* PR A is merged.

The next release will be busted, because PR A's migration is "before" PR B's migration, but has not been applied, which breaks `sqlx`'s expectations; I'm not sure how `sqlx` would respond to this situation but I suspect we'd need to manually fix things up.

I think we should abandon `sqlx`'s date-based scheme and replace it with an arbitrary-number scheme (so migrations' "dates" are e.g. `00000000000001`, `00000000000002`, `00000000000003`, etc). This would ensure a strict ordering of schema updates based on release order. I don't think this would be too hard, as long as `sqlx` doesn't break when handling "dates" of this format -- our current `sqlx migrate add -r <name> --source /path/to/janus/db` would just become "manually create a couple of files with the appropriate naming scheme".

I'll experiment with this change in a separate PR.

branlwyd: https://github.com/divviup/janus/pull/1304 for the change from timestamp-based ordering to opaque-counter-based ordering.

tgeoghegan (approved):

db/20230420210648_batch-aggregation-state.up.sql
branlwyd: (merging from `main`, I renumbered the migration in this PR to match the new numbering scheme, which places it at the end.)

PR #1272, opened 2023-04-18
Check DB schema version at startup

### LINKED ISSUE #1241: Janus should refuse to start up if it sees an unsupported schema version
As of #1218, we now have versioned SQL schema migration scripts. Deploying a migration from schema _n_ to _n+1_ will involve deploying a Janus that can support both schema versions and then applying the migration to schema _n+1_. To make this safer, any Janus version should have a list or range of supported schema versions compiled into it so that at startup, it can query the `_sqlx_migrations` table to see what the current version is and refuse to startup if it doesn't recognize it. This reduces the risk of Janus corrupting data.

tgeoghegan:
Creating a `Datastore` now checks the version of the most recently applied migration script against an array of supported versions compiled into Janus. If the current schema is not supported, `Datastore::new` returns an error.

Resolves #1241

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -41,7 +41,7 @@ use ring::aead::{self, LessSafeKey, AES_128_GCM};
 use std::{
     collections::HashMap,
     convert::TryFrom,
-    fmt::Display,
+    fmt::{Debug, Display},
     future::Future,
     io::Cursor,
     mem::size_of,
@@ -61,6 +61,10 @@ pub mod test_util;
 
 // TODO(#196): retry network-related & other transient failures once we know what they look like
 
+/// List of schema versions that this version of Janus can safely run on. If any other schema
+/// version is seen, [`Datastore::new`] fails.
+const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[20230405185602, 20230417204528];
+
 /// Datastore represents a datastore for Janus, with support for transactional reads and writes.
 /// In practice, Datastore instances are currently backed by a PostgreSQL database.
 pub struct Datastore<C: Clock> {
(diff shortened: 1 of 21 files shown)


divergentdave (approved):

interop_binaries/config/aggregation_job_creator.yaml
```diff
@@ -3,6 +3,7 @@
 # tests, and is not intended for production use.
 database:
   url: postgres://postgres@127.0.0.1:5432/postgres
+  check_schema_version: false
```
branlwyd, line 6: Q: why do we need to set `check_schema_version` to `false` in various interop binaries? Is the schema populated in a different way there?

aggregator_core/src/datastore.rs
```diff
@@ -61,6 +61,10 @@ pub mod test_util;
 
 // TODO(#196): retry network-related & other transient failures once we know what they look like
 
+/// List of schema versions that this version of Janus can safely run on. If any other schema
+/// version is seen, [`Datastore::new`] fails.
+const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[20230405185602, 20230417204528];
```
branlwyd, line 66: It's easy to see when we'd add a new version to `SUPPORTED_SCHEMA_VERSIONS` (presumably, when we're creating a new DB migration, and we'd be reminded to do so because tests would begin failing until we do so).

But what's the process for removing a version from this list: how do we evaluate when it's correct to remove an item? more importantly, what reminds us to remove versions?

(somewhat separately from this, I wonder if it would be useful to have an env var or other configuration knob that controls which DB schema version is used for ephemeral datastores. this could be a useful tool, but not a complete solution, in evaluating if a given schema version still works with a given revision of the code.)

tgeoghegan:
Question for reviewers: what do we think about baking the list of supported schema versions into the binary versus making it a config option? I think I favor the former (which is what I've implemented here) because I think whether a given version of Janus can work with a given schema version is a static property of the code, not a configuration decision. And doing it this way means we don't have to wrangle another configuration option in all our deploys. But maybe we'll be glad during some future incident to be able to force Janus to try running on an arbitrary schema version we didn't know about at the time of compiling Janus?

aggregator_core/src/datastore.rs
```diff
@@ -3852,6 +3916,10 @@ pub enum Error {
     /// An error occurred while manipulating timestamps or durations.
     #[error("{0}")]
     TimeOverflow(&'static str),
+    /// The version of the DB schema found in `_sqlx_migrations` is not supported by this version of
+    /// Janus.
+    #[error("unsupported schema version {version} / {description}")]
+    UnsupportedSchemaVersion { version: i64, description: String },
```
branlwyd, line 3922: Does this need to be a separately-discriminated error? It's hard to imagine we'll ever do anything with this error other than report it to a human being.

If we don't think we'll make program-logic decisions based on this error discriminant, let's introduce an `Other` value that wraps an `anyhow::Error`.

aggregator_core/src/datastore.rs
tgeoghegan: > It's easy to see when we'd add a new version to `SUPPORTED_SCHEMA_VERSIONS` (presumably, when we're creating a new DB migration, and we'd be reminded to do so because tests would begin failing until we do so).
> 
> But what's the process for removing a version from this list: how do we evaluate when it's correct to remove an item? more importantly, what reminds us to remove versions?

With things as they stand in this PR, it's up to the person making a change to remove schema versions from the supported list. That does mean that nothing will remind us or force us to drop unsupported schemas. What would be really nice -- and I think this also speaks to your idea about configuring what schema version tests uses -- is for tests that involve the datastore to automatically be run against all the schema versions that are listed in this array. I think what we'd need is a macro for tests that use storage. The macro would generate code to instantiate an ephemeral datastore with the appropriate schema version, and it could also stamp out multiple tests, each one using a different supported version.

Then we'd get real assurance from the tests that Janus functions with all the schema versions we claim are supported, and it'd be completely obvious when an older schema is not supported: you'd see a plain test failure, and then could decide whether to drop that schema version or not.

aggregator_core/src/datastore.rs
tgeoghegan: In that case we may as well use `Error::DbState` and represent the error as a `String`.

interop_binaries/config/aggregation_job_creator.yaml
tgeoghegan: It's because the interop binaries use `janus_aggregator::binary_utils::janus_main`, just like the "real" Janus components, and that function is responsible for instantiating a `Datastore` before any of the specific application's code gets to run. The interop binaries must opt out of schema version checks so that they can get to their own code (the function pointer invoked by `janus_main`), which will apply the migration scripts in the container image.

aggregator_core/src/datastore.rs
tgeoghegan: I'm experimenting with [`rstest`](https://docs.rs/rstest/latest/rstest/) for this, and it looks pretty promising!

aggregator_core/src/datastore.rs
branlwyd: This works, but my point is: I'm not sure we need a separate `DbState` error discriminant, either -- we generate a number of `DbState` errors, but nothing ever consumes a `DbState` error to make a program-logic decision, so `DbState` should ideally also be `Other`.

I don't think we need to clean up existing error discriminants, but there's no reason to introduce additional unnecessary discriminants, either.

[I'm not married to the idea of using `anyhow::Error` as the internal representation--`String` would work too. I guess I see the advantage of `anyhow::Error` as having a nice API & providing error-wrapping.]

branlwyd (approved):

Issue #1252, opened 2023-04-14
Immutability of SQL migration scripts


tgeoghegan:
The Janus sources now include a series of SQL migration scripts in `db`. However, the series of migrations isn't just a property of a single git commit independent of all other context. Specifically, it should only be possible to change the state of migrations by appending a new migration to the end of the list, because we can't rewrite the history of migrations applied to deployed environments.

Currently, Janus' `publish-sql-schema` workflow simply copies all of `janus/db` into a directory like `gs://janus-artifacts-sql-schemas/0.x`, overwriting whatever was there before. That workflow and `ci-build` should verify that:

- any existing migration scripts have not been changed
- no migration scripts are inserted anywhere except the end of the sequence
- no migration scripts are removed

inahga:
I was able to create a CI script that checks each PR for validity of the migration scripts. https://github.com/divviup/janus/pull/1442.

---

As for validation on the deploy-side, here's the strategy I lean towards:
- Compare the new schema to the last released version (i.e. diff'ing between the two git tags). If the schema is OK, upload the new schema UNLESS the GCS bucket already has stuff in it.
- If there is stuff in the GCS bucket, that means that a schema was already uploaded and this build is probably being run as `workflow_dispatch`. In this case, fail the build, UNLESS a `force` option was provided. 
- If `force` was provided, and the schema is valid, go ahead and upload the schema, overwriting what is in there.

I leave the `force` escape hatch open in case there was some error that caused the original release build to fail to upload its schema part-way through, or if we otherwise know what we're doing.

For point 1, the logic from #1442 can be used, possibly via [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action).

inahga:
I've done a final test of all the changes, but it's not possible to fully test the `publish schema` workflow as it will only deploy from tags, and there is no available tag that has the newly updated workflow. I've tested it thoroughly in isolation, but the final word will have to come from on-call next week. I'll close and re-open if necessary then.

PR #1276, opened 2023-04-19
Undo comment change in old schema, document in new

divergentdave:
#1253 made a change to a comment in `db/20230405185602_initial-schema.up.sql`. Generally speaking, we should avoid this, since it affects the checksum, and sqlx would refuse to apply further migrations if it saw such a discrepancy. This PR undoes that change, and documents the new meaning of `report_aggregations.prep_msg` in the newly added migration instead. We could also not take this PR, and make additional manual database modifications to the staging and load test environments. Separately, we may want to start some living documentation of our schema in the docs directory, to codify an explanation of the current state of our schema, instead of relying on comments in many migration scripts.


tgeoghegan (approved):
This sort of problem would be ruled out by addressing https://github.com/divviup/janus/issues/1252.

branlwyd (approved):

PR #1277, opened 2023-04-19
Run `datastore` tests against multiple schemas

tgeoghegan:
We want to prove in tests that some Janus version can run safely on multiple schema versions to make database schema migrations safer. In this commit:

 - `aggregator_core::test_util::EphemeralDatastore` can now be constructed with a `max_schema_version` argument to control which migration scripts are applied during tests
 - We adopt [`rstest`][1] to inject an `EphemeralDatastore` instance into tests in `aggregator_core::datastore::Datatore::tests`. Using [`rstest_reuse`][2], we can automatically stamp out versions of existing tests that run using multiple schema versions.

We only use this dependency injection technique in the `datastore` module, because that should be the only module that's tightly coupled to the database schema. We could run all tests that use a datastore this way, at the cost of increasing overall test runtime.

[1]: https://docs.rs/rstest
[2]: https://docs.rs/rstest_reuse

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -61,9 +61,32 @@ pub mod test_util;
 
 // TODO(#196): retry network-related & other transient failures once we know what they look like
 
-/// List of schema versions that this version of Janus can safely run on. If any other schema
-/// version is seen, [`Datastore::new`] fails.
-const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[20230405185602, 20230417204528];
+/// This macro stamps out an array of schema versions supported by this version of Janus and an
+/// [`rstest_reuse`][1] template that can be applied to tests to have them run against all supported
+/// schema versions.
+///
+/// [1]: https://docs.rs/rstest_reuse/latest/rstest_reuse/
+macro_rules! supported_schema_versions {
+    ($( $i:literal ),*) => {
+        const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[$($i),*];
+
+        #[cfg(test)]
+        #[rstest_reuse::template]
+        #[rstest::rstest]
+        $(#[case(ephemeral_datastore_max_schema_version($i))])*
+        async fn schema_versions_template(
+            #[future(awt)]
+            #[case]
+            ephemeral_datastore: EphemeralDatastore,
+        ) {
+            // This is an rstest template and never gets run.
+        }
+    }
+}
+
+// List of schema versions that this version of Janus can safely run on. If any other schema
+// version is seen, [`Datastore::new`] fails.
+supported_schema_versions!(20230417204528);
 
 /// Datastore represents a datastore for Janus, with support for transactional reads and writes.
 /// In practice, Datastore instances are currently backed by a PostgreSQL database.
@@ -5408,7 +5431,10 @@ mod tests {
                 LeaderStoredReport, Lease, OutstandingBatch, ReportAggregation,
                 ReportAggregationState, SqlInterval,
             },
-            test_util::{ephemeral_datastore, generate_aead_key},
+            schema_versions_template,
+            test_util::{
+                ephemeral_datastore_max_schema_version, generate_aead_key, EphemeralDatastore,
+            },
             Crypter, Datastore, Error, Transaction,
         },
         query_type::CollectableQueryType,
@@ -5447,10 +5473,10 @@ mod tests {
         time::Duration as StdDuration,
     };
 
+    #[rstest_reuse::apply(schema_versions_template)]
     #[tokio::test]
-    async fn reject_unsupported_schema_version() {
+    async fn reject_unsupported_schema_version(ephemeral_datastore: EphemeralDatastore) {
         install_test_trace_subscriber();
-        let ephemeral_datastore = ephemeral_datastore().await;
         let error = Datastore::new_with_supported_versions(
             ephemeral_datastore.pool(),
             ephemeral_datastore.crypter(),
@@ -5463,10 +5489,10 @@ mod tests {
         assert_matches!(error, Error::DbState(_));
     }
 
+    #[rstest_reuse::apply(schema_versions_template)]
     #[tokio::test]
-    async fn roundtrip_task() {
+    async fn roundtrip_task(ephemeral_datastore: EphemeralDatastore) {
         install_test_trace_subscriber();
-        let ephemeral_datastore = ephemeral_datastore().await;
         let ds = ephemeral_datastore.datastore(MockClock::default()).await;
 
         // Insert tasks, check that they can be retrieved by ID.
@@ -5576,10 +5602,10 @@ mod tests {
         assert_eq!(want_tasks, got_tasks);
     }
 
+    #[rstest_reuse::apply(schema_versions_template)]
     #[tokio::test]
-    async fn get_task_metrics() {
+    async fn get_task_metrics(ephemeral_datastore: EphemeralDatastore) {
         install_test_trace_subscriber();
-        let ephemeral_datastore = ephemeral_datastore().await;
         let ds = ephemeral_datastore.datastore(MockClock::default()).await;
 
         ds.run_tx(|tx| {
@@ -5711,10 +5737,10 @@ mod tests {
         .unwrap();
     }
 
+    #[rstest_reuse::apply(schema_versions_template)]
     #[tokio::test]
-    async fn get_task_ids() {
+    async fn get_task_ids(ephemeral_datastore: EphemeralDatastore) {
         install_test_trace_subscriber();
-        let ephemeral_datastore = ephemeral_datastore().await;
         let ds = ephemeral_datastore.datastore(MockClock::default()).await;
 
         ds.run_tx(|tx| {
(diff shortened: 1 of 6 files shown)


tgeoghegan:
Some additional notes here:

- For now I've only applied this to three tests in `datastore.rs`, because I want to illustrate how this works so we can discuss it without all the noise of adding a `#[future] #[case] ephemeral_datastore: EphemeralDatastore` argument to every test in the module
- We (should) observe that the test `datastore::tests::roundtrip_report_aggregation::case_1_version_20230405185602` fails, because indeed the current Janus code doesn't work on a schema that doesn't define column `report_aggregators.last_prep_step`, so I think that shows the tests working as expected: we've learned that we need to drop schema version `20230405185602` from the supported versions list.

aggregator_core/src/datastore.rs
```diff
@@ -5447,10 +5452,19 @@ mod tests {
         time::Duration as StdDuration,
     };
 
+    #[template]
+    #[rstest]
+    // The version numbers in these cases must match SUPPORTED_SCHEMA_VERSIONS
+    #[case::version_20230405185602(ephemeral_datastore_max_schema_version(20230405185602))]
+    #[case::version_20230417204528(ephemeral_datastore_max_schema_version(20230417204528))]
```
tgeoghegan, line 5459: Ideally this list of cases would be automatically stamped out from `SUPPORTED_SCHEMA_VERSIONS`, but I am not sure how to do that gracefully.

tgeoghegan:
Sample test output:

```
running 48 tests
test datastore::tests::aggregation_job_not_found ... ok
test datastore::tests::aggregation_job_acquire_release ... ok
test datastore::tests::check_report_aggregation_exists ... ok
test datastore::tests::collection_job_acquire_job_max ... ok
test datastore::tests::collection_job_acquire_no_aggregation_job_with_agg_param ... ok
test datastore::tests::collection_job_acquire_no_aggregation_job_with_task_id ... ok
test datastore::tests::collection_job_acquire_release_aggregation_job_in_progress ... ok
test datastore::tests::collection_job_acquire_release_job_finished ... ok
test datastore::tests::collection_job_acquire_report_shares_outside_interval ... ok
test datastore::tests::count_client_reports_for_batch_id ... ok
test datastore::tests::count_client_reports_for_interval ... ok
test datastore::tests::crypter ... ok
test datastore::tests::collection_job_acquire_state_filtering ... ok
test datastore::tests::delete_expired_client_reports ... ok
test datastore::tests::delete_expired_aggregation_artifacts ... ok
test datastore::tests::delete_expired_collection_artifacts ... ok
test datastore::tests::get_aggregation_jobs_for_task ... ok
test datastore::tests::get_collection_job ... ok
test datastore::tests::get_report_aggregations_for_aggregation_job ... ok
test datastore::tests::get_task_ids ... ok
test datastore::tests::get_task_metrics ... ok
test datastore::tests::get_unaggregated_client_report_ids_for_task ... ok
test datastore::tests::fixed_size_collection_job_acquire_release_happy_path ... ok
test datastore::tests::get_unaggregated_client_report_ids_with_agg_param_for_task ... ok
test datastore::tests::report_aggregation_not_found ... ok
test datastore::tests::report_not_found ... ok
test datastore::tests::roundtrip_aggregate_share_job ... ok
test datastore::tests::roundtrip_aggregation_job ... ok
test datastore::tests::roundtrip_batch_aggregation_fixed_size ... ok
test datastore::tests::roundtrip_batch_aggregation_time_interval ... ok
test datastore::tests::roundtrip_interval_sql ... ok
test datastore::tests::roundtrip_outstanding_batch ... ok
test datastore::tests::roundtrip_report ... ok
test datastore::tests::roundtrip_report_aggregation::case_1_version_20230405185602 ... FAILED
test datastore::tests::roundtrip_report_aggregation::case_2_version_20230417204528 ... ok
test datastore::tests::roundtrip_report_share ... ok
test datastore::tests::roundtrip_task::case_1_version_20230405185602 ... ok
test datastore::tests::roundtrip_task::case_2_version_20230417204528 ... ok
test datastore::tests::update_collection_jobs ... ok
test query_type::tests::validate_collect_identifier ... ok
test task::tests::aggregator_endpoints_end_in_slash ... ok
test task::tests::aggregator_request_paths ... ok
test task::tests::collector_auth_tokens ... ok
test task::tests::deserialize_docs_sample_tasks ... ok
test task::tests::reject_invalid_auth_tokens ... ok
test task::tests::task_serde ... ok
test task::tests::task_serialization ... ok
test datastore::tests::time_interval_collection_job_acquire_release_happy_path ... ok

error: test failed, to rerun pass `-p janus_aggregator_core --lib`
failures:

---- datastore::tests::roundtrip_report_aggregation::case_1_version_20230405185602 stdout ----
  2023-04-19T18:25:59.505694Z ERROR janus_aggregator_core::datastore: error: DB error: db error: ERROR: column "last_prep_step" of relation "report_aggregations" does not exist
    at aggregator_core/src/datastore.rs:2036 on ThreadId(108)

thread 'datastore::tests::roundtrip_report_aggregation::case_1_version_20230405185602' panicked at 'called `Result::unwrap()` on an `Err` value: Db(Error { kind: Db, cause: Some(DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E42703), message: "column \"last_prep_step\" of relation \"report_aggregations\" does not exist", detail: None, hint: None, position: Some(Original(169)), where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("parse_target.c"), line: Some(1061), routine: Some("checkInsertTargets") }) })', aggregator_core/src/datastore.rs:7366:18


failures:
    datastore::tests::roundtrip_report_aggregation::case_1_version_20230405185602

test result: FAILED. 47 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 16.42s
```

Note we get distinct test cases per schema version:
```
test datastore::tests::roundtrip_report_aggregation::case_1_version_20230405185602 ... FAILED
test datastore::tests::roundtrip_report_aggregation::case_2_version_20230417204528 ... ok
<...>
test datastore::tests::roundtrip_task::case_1_version_20230405185602 ... ok
test datastore::tests::roundtrip_task::case_2_version_20230417204528 ... ok
```

I like this because I believe it means `cargo` can run these in parallel and also a test failure with one schema version doesn't stop tests on other schema versions from running.

aggregator_core/src/datastore.rs
divergentdave: As declarative macros cannot produce attributes, we'd have to write our own procedural macro in order to do that. We can either do so as a progressive enhancement later, or just make coupled changes to the template and `SUPPORTED_SCHEMA_VERSIONS` by hand.

divergentdave (approved):

aggregator_core/src/datastore.rs
branlwyd: I think the main value of `rstest` is that we run each datastore test on each supported schema version. However, I don't like that we have to remember to manually add a case each time we add a new schema version -- we should have one canonical list of supported schemas, and that is `SUPPORTED_SCHEMAS`. Currently, if I add a version to `SUPPORTED_SCHEMAS` but forget to add a case here, the tests would pass even if the new schema versions was broken; I (or a reviewer) have to manually notice that I'm missing a case. IMO, that's a blocker.

A few potential solutions come to mind:

* I assume something like `#[values(SUPPORTED_SCHEMAS)]` doesn't work, but if so, that's easy enough ([docs](https://docs.rs/rstest/0.17.0/rstest/#creating-a-test-for-each-combinations-of-given-values)).
* We could write a macro to generate the necessary `#[case]` attributes -- based on [this StackOverflow question](https://stackoverflow.com/questions/61449167/generate-attribute-value-via-macro) it looks like generating attributes from a macro is possible.
* If the above aren't feasible/easy, another approach would be to avoid using `rstest` at all. We could write a "wrapper function" like `run_with_ephemeral_datastore` which would take a `Fn(EphemeralDatastore)`, then call that function in a loop with an ephemeral datastore at the proper version. Every place that currently uses the template would be changed to call this function instead, wrapping the current test body. This would work, but wouldn't separate different versions of the same test into separate cases or provide parallelization of tests across versions, and less importantly would induce an additional level of indentation in most test bodies.

I would say writing a quick macro is probably the best approach, assuming there's no syntax that would allow `SUPPORTED_VERSIONS` to be used directly.

divergentdave: That inspires one way we could do this with a declarative macro: we could define and then use a macro that takes multiple schema versions as arguments, and then emits two top-level items: the `const` declaration of `SUPPORTED_SCHEMA_VERSIONS`, and the entire test template function, decorated with one `#[case]` attribute macro per macro argument.

tgeoghegan: That's a great idea. I'll give the macro a try this afternoon.

tgeoghegan: The macro was pretty easy to write (I did in the latest pushed commit). The downside is that [because you can't construct identifiers with macro variables](https://github.com/rust-lang/rust/issues/29599), we can't use `rstest`'s `case::label` syntax to interpolate the schema version into test names.

docs/DEPLOYING.md
```diff
@@ -114,6 +114,12 @@ This will generate two new migration scripts. Fill the `*.up.sql` file with the
 migration you want to run and the `*.down.sql` file with a script that reverses
 the first script.
 
+After adding a migration, you must add its version number to
+`datastore::SUPPORTED_SCHEMA_VERSIONS` as Janus will refuse to run if it does
+not recognize the database schema version. You also must add a `#[case]`
```
branlwyd, line 119: nit: the docs suggesting that a `#[case]` attribute needs to be created can now be dropped.

branlwyd (approved):

PR #1282, opened 2023-04-21
`aggregator_api`: b64 decode received auth token

tgeoghegan:
The bearer token in an HTTP `Authorization` header should be represented in the Base64 alphabet, per RFC 6750[1]. Janus doesn't actually use OAuth 2 here, but it still makes sense for us to decode the bearer token presented by the client from Base64, as you can't transmit arbitrary bytes in HTTP headers anyway.

[1]: https://datatracker.ietf.org/doc/html/rfc6750#section-2.1

### DIFF
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -1,6 +1,7 @@
 //! This crate implements the Janus Aggregator API.
 
 use crate::models::{GetTaskIdsResp, PostTaskReq};
+use base64::{engine::general_purpose::STANDARD, Engine};
 use janus_aggregator_core::{
     datastore::{self, Datastore},
     task::Task,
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
LGTM, but we should be sure to do the same decoding when we implement deserialization of `Config`.

branlwyd (approved):
I guess I don't object, but in practice we're already playing Calvinball with Basic authentication: https://datatracker.ietf.org/doc/html/rfc7617#section-2 suggests that Basic authentication should always encode "username:password", whereas our plan is to encode an arbitrary string.

I'd be OK with keeping the current behavior, which allows specifying an authentication token; if it so happens to be a valid base64 string, great.

OTOH I won't object to switching to base64 here -- I think it's just going to make eventual configuration slightly more complicated, and removes flexibility from what we can specify (though in a way that makes us more-RFC-compatible).

tgeoghegan:
I don't think I follow: we're not using the `Basic` authentication scheme, but rather `Bearer`, which AFAIK is only specified in [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1) and requires that the token be in the Base64 alphabet (though it doesn't require that it be valid Base64, I think). We're not really using OAuth 2.0, so we don't have to follow everything in RFC 6750, but I think using Base64 on the wire makes sense regardless: we generate the auth tokens as random byte sequences, which means the vast majority of them aren't printable. [RFC 7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2) tells us that header values must be made up of printable characters:

```
header-field   = field-name ":" OWS field-value OWS

     field-name     = token
     field-value    = *( field-content / obs-fold )
     field-content  = field-vchar [ 1*( SP / HTAB ) field-vchar ]
     field-vchar    = VCHAR / obs-text

     obs-fold       = CRLF 1*( SP / HTAB )
                    ; obsolete line folding
                    ; see [Section 3.2.4](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4)
```

[7230 1.2](https://datatracker.ietf.org/doc/html/rfc7230#section-1.2):
`VCHAR (any
   visible [[USASCII](https://datatracker.ietf.org/doc/html/rfc7230#ref-USASCII)] character).`

Anyway, I suppose we could also solve this by changing the auth token generation to randomly sample printable ASCII instead of completely random bytes but Base64 encoding when going into/out of HTTP messages seems better to me: it separates a transport layer concern from the application.

PR #1299, opened 2023-04-24
Do not store output share in report aggregations.

### LINKED ISSUE #1284: Space optimization: don't store output shares
One possible time/space tradeoff we could make is to stop storing Prio3 output shares on disk. With a very large Prio3 instance, most of the helper's storage usage comes from output shares, as all other per-report messages are very small. The `report_aggregations.out_share` column could be deleted, and the output share could be re-derived from the seed in the helper share as needed, using the PRG. (in the leader's case, output shares can be recomputed by truncating the input share) Then, aggregation share accumulators could be updated with those in-memory output shares. However, this may require reaching through the VDAF abstraction layer to accomplish.

branlwyd:
This was never necessary: when an output share is recovered, its contribution to the relevant batch aggregation is always recorded, and the collection process solely uses the batch aggregation values. The output share was stored but never (meaningfully) read.

Historically, this is an artifact of the initial implementation of the aggregation process, which occurred before the collection process was written. At time of initial implementation, there was no batch aggregation to record aggregated values into.

Closes #1284.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -63,7 +63,7 @@ pub mod test_util;
 
 /// List of schema versions that this version of Janus can safely run on. If any other schema
 /// version is seen, [`Datastore::new`] fails.
-const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[20230405185602, 20230417204528];
+const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[20230405185602, 20230417204528, 20230424220336];
 
 /// Datastore represents a datastore for Janus, with support for transactional reads and writes.
 /// In practice, Datastore instances are currently backed by a PostgreSQL database.
@@ -1802,8 +1802,7 @@ impl<C: Clock> Transaction<'_, C> {
                     client_reports.report_id, client_reports.client_timestamp,
                     report_aggregations.ord, report_aggregations.state,
                     report_aggregations.prep_state, report_aggregations.prep_msg,
-                    report_aggregations.out_share, report_aggregations.error_code,
-                    report_aggregations.last_prep_step, aggregation_jobs.aggregation_param
+                    report_aggregations.error_code, report_aggregations.last_prep_step
                 FROM report_aggregations
                 JOIN client_reports ON client_reports.id = report_aggregations.client_report_id
                 JOIN aggregation_jobs
@@ -1857,8 +1856,7 @@ impl<C: Clock> Transaction<'_, C> {
                     client_reports.report_id, client_reports.client_timestamp,
                     report_aggregations.ord, report_aggregations.state,
                     report_aggregations.prep_state, report_aggregations.prep_msg,
-                    report_aggregations.out_share, report_aggregations.error_code,
-                    report_aggregations.last_prep_step, aggregation_jobs.aggregation_param
+                    report_aggregations.error_code, report_aggregations.last_prep_step
                 FROM report_aggregations
                 JOIN client_reports ON client_reports.id = report_aggregations.client_report_id
                 JOIN aggregation_jobs
@@ -1911,9 +1909,8 @@ impl<C: Clock> Transaction<'_, C> {
                     aggregation_jobs.aggregation_job_id, client_reports.report_id,
                     client_reports.client_timestamp, report_aggregations.ord,
                     report_aggregations.state, report_aggregations.prep_state,
-                    report_aggregations.prep_msg, report_aggregations.out_share,
-                    report_aggregations.error_code, report_aggregations.last_prep_step,
-                    aggregation_jobs.aggregation_param
+                    report_aggregations.prep_msg, report_aggregations.error_code,
+                    report_aggregations.last_prep_step
                 FROM report_aggregations
                 JOIN client_reports ON client_reports.id = report_aggregations.client_report_id
                 JOIN aggregation_jobs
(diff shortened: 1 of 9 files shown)


branlwyd:
I dropped some code which would attempt to accumulate any aggregations that Finished during initialization, because it is impossible for an aggregation to finish during initialization (i.e. this was effectively dead code).

I'll need to fix this in a smarter way in https://github.com/divviup/janus/pull/1234, since the new aggregation scheme does allow aggregation to complete during initialization.

aggregator_core/src/datastore.rs
```diff
@@ -63,7 +63,7 @@ pub mod test_util;
 
 /// List of schema versions that this version of Janus can safely run on. If any other schema
 /// version is seen, [`Datastore::new`] fails.
-const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[20230405185602, 20230417204528];
+const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[20230405185602, 20230417204528, 20230424220336];
```
tgeoghegan, line 66: You'll have to rebase and shorten this list based on the test results.

db/20230424220336_rm-out-share.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE report_aggregations DROP COLUMN out_share;
```
tgeoghegan, line 1: What's the migration plan here? I think there's no data in `staging-dap-04`'s `report_aggregations` table, but will Janus start throwing errors if we deploy this migration ahead of deploying a new Janus? Or maybe it won't since `staging-dap-04` is idle?

tgeoghegan (approved):

db/20230424220336_rm-out-share.up.sql
branlwyd: Well, the current plan is to accept the errors, since as you say `staging-dap-04` is effectively unused for the time being. Our current migrations effectively exercise the migration infra, but may not fulfill all the requirements of a production migration -- for example, 20230417204528 would have required a data migration/backfill if there were any data to be migrated/backfilled. (one might argue we should eventually yank releases that can't be migrated to higher release versions, except that I'm not sure anyone currently cares.)

That said, maybe that's not the right choice? What we really want is for this migration to run _after_ the code in this PR is fully deployed. There are a couple of ways we could engineer this:
* Implement & document a way to split migrations into "pre-release" and "post-release" migrations, and phrase this as a "post-release" migration.
* Just split the data migration part of the change to a separate PR, and manually wrangle it to be released with the next release after the code changes in this PR.

WDYT? I'd like the ability to engineer data migrations to be pre- or post-release, but this might be too much work immediately. Manually wrangling each data change that needs to happen post-release so that it is included in a later release is easy to choose as a policy, but finicky & somewhat error-prone.

For now, my vote would be "accept whatever errors occur during the release process" since there is no practical downside to doing this, but I'm interested in your thoughts -- especially as to whether pre-/post-migrations are a good idea/feasible to implement.

branlwyd: Well, the current plan is to accept the errors, since as you say `staging-dap-04` is effectively unused for the time being. Our current migrations effectively exercise the migration infra, but may not fulfill all the requirements of a production migration -- for example, 20230417204528 would have required a data migration/backfill if there were any data to be migrated/backfilled. (one might argue we should eventually yank releases that can't be migrated to higher release versions, except that I'm not sure anyone currently cares.)

That said, maybe that's not the right choice? What we really want is for this migration to run _after_ the code in this PR is fully deployed. There are a couple of ways we could engineer this:
* Implement & document a way to split migrations into "pre-release" and "post-release" migrations, and phrase this as a "post-release" migration.
* Just split the data migration part of the change to a separate PR, and manually wrangle it to be released with the next release after the code changes in this PR.

WDYT? I'd like the ability to engineer data migrations to be pre- or post-release, but this might be too much work immediately. Manually wrangling each data change that needs to happen post-release so that it is included in a later release is easy to choose as a policy, but finicky & somewhat error-prone.

For now, my vote would be "accept whatever errors occur during the release process" since there is no practical downside to doing this, but I'm interested in your thoughts -- especially as to whether pre-/post-migrations are a good idea/feasible to implement.

tgeoghegan: I'm comfortable with temporarily breaking `staging-dap-04` while we deploy this migration, because nobody is currently relying on that deployment and thus we shouldn't take on extra work to maintain its availability.

But suppose we were doing this in a live production environment, as a thought exercise: I think what we'd do is this:

- Make a change to Janus that adds the migration script and makes code changes so that Janus can work with or without the migration applied (we can have confidence of this because our tests run against multiple schema versions)
- Deploy the new Janus
- Apply the migration
- Make another change to Janus that removes support for the old schema version

divergentdave: I don't think we need any additional features to differentiate such migrations, rather, more complicated changes should be broken up into multiple deployable changes, and the sequencing of database migrations and code deploys will be determined on a case-by-case basis, then enforced by `SUPPORTED_SCHEMA_VERSIONS` checks at runtime and in tests. This change is uniquely simple, since we're deleting a write-only column. If we were deleting a non-null column that we were both reading and writing, the sequence could look something like:

* Existing state: schema version 0 is deployed, and the deployed code only supports schema version 0
* Code deploy: stop reading from column, supports schema versions 0 and 1
* Schema migration, 0 => 1: make column nullable
* Code deploy: stop writing to column, supports schema versions 0, 1, and 2
* Schema migration, 1 => 2: delete column

tgeoghegan (approved):
Change LGTM, assuming tests pass.

Issue #1281, opened 2023-04-21
Aggregation job size should be dynamic based on message sizes


divergentdave:
Currently, aggregation job size limits (in number of reports) are determined from configuration settings. We typically set the minimum size to 10 reports and the maximum to 500 reports, with the understanding that jobs smaller than the minimum may be created after a long enough time.

We may want to change this dynamically, depending on how large the VDAF messages are. There are five relevant message sizes: the leader share, the helper share, the public share, prepare shares, and prepare messages. Even though the leader share doesn't get sent over the wire, loading many leader shares into memory can be hard on the resources of our deployments.

branlwyd:
At the least, we should consider changing this to be a per-task parameter rather than a per-job parameter.

If we implement this, it should probably be computed based on the VDAF (& any parameters to the VDAF), i.e. derived from the Task's included `VdafInstance`.

It would still be nice to have a tuning knob that controls "relative size" of aggregation jobs -- i.e. if a VDAF has aggregation jobs of size _X_ when the knob is at 1.0, turning the knob to 2.0 will cause the VDAF to have aggregation jobs of size _2*X_.

PR #1301, opened 2023-04-26
Add methods to VdafInstance for message sizes

divergentdave:
This adds closed-form formulas for various VDAF message sizes, accessible from methods on the `VdafInstance` enum. This will likely be needed for #1281, plus these formulas could be useful in estimating bandwidth/storage usage. Their correctness is checked with quickcheck tests, which pick a random VDAF instance, run it, and check against the actual message sizes. Poplar1 support is left incomplete, as the size of any messages from the aggregation flow onwards depends on the aggregation parameter, so it'll need a different API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -257,7 +257,7 @@ dependencies = [
  "log",
  "parking",
  "polling",
- "rustix 0.37.3",
+ "rustix",
  "slab",
  "socket2 0.4.9",
  "waker-fn",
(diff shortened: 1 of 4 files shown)


branlwyd:
I'm not sure we need this -- as I mentioned on #1281, I think the sizes would be driven based on the VDAF type & configuration parameters, rather than specifically on the sizes of the messages inside. While using code like this to estimate the right sizes for the `VdafInstance` is a good idea, keeping this code in-tree is likely to be a maintenance burden. (If nothing else, this functionality should really be exposed from `libprio-rs`, since `libprio-rs` defines the messages themselves.)

PR #1304, opened 2023-04-26
Use opaque counter, rather than timestamp, to order SQL migrations.

branlwyd:
Using a timestamp to order SQL migrations can allow migrations to be merged out-of-order with the order than SQLx expects to be able to apply them. If a "later" migration was released before an "earlier" migration was merged, our DB would be un-migratable and we would have a production incident on our hands.


branlwyd:
See https://github.com/divviup/janus/pull/1254#discussion_r1178362866 for a more detailed explanation of what can go wrong with a timestamp-based ordering.

If we decide to take this change, we'll need to decide what to do with `staging-dap-04`. Probably wipe+redeploy the database.

Note that git merging still won't see conflicts due to the name, but tests will begin failing if we merge two same-versioned (supported) migrations, as SQLx will become confused when it attempts to run the migrations.

tgeoghegan (approved):

Issue #1283, opened 2023-04-24
Revisit acquire_incomplete_time_interval_collection_jobs() query


divergentdave:
Looking at Query Insights after last week's load tests, the collect job acquisition query took an inordinate amount of time to run on the leader. (See #519 previously) Though it ran only ~500 times during the Prio3Histogram load test with 10 reports per second, it nevertheless was #4 on the list of top queries by database load. The database load and latency both rose over time, and the charts suggest they may scale poorly with larger table sizes.

![image](https://user-images.githubusercontent.com/181772/234072399-b9d41da5-bb4f-4325-9b70-9f33c68e630c.png)

The query was long enough that it got truncated inside Query Insights, but I re-ran it manually to get a query plan.

```
postgres=> EXPLAIN ANALYZE WITH updated AS (UPDATE collection_jobs SET lease_expiry = '2023-04-24 00:00:00', lease_token = gen_random_bytes(16), lease_attempts = lease_attempts + 1 FROM tasks WHERE collection_jobs.id IN (SELECT collection_jobs.id FROM collection_jobs INNER JOIN aggregation_jobs ON collection_jobs.aggregation_param = aggregation_jobs.aggregation_param AND collection_jobs.task_id = aggregation_jobs.task_id AND collection_jobs.batch_interval && aggregation_jobs.client_timestamp_interval WHERE tasks.id = collection_jobs.task_id AND tasks.query_type ? 'TimeInterval' AND collection_jobs.lease_expiry <= '2023-04-24 00:00:00' GROUP BY collection_jobs.id HAVING bool_and(aggregation_jobs.state != 'IN_PROGRESS') LIMIT 10) RETURNING tasks.task_id, tasks.query_type, tasks.vdaf, collection_jobs.collection_job_id, collection_jobs.id, collection_jobs.lease_token, collection_jobs.lease_attempts) SELECT task_id, query_type, vdaf, collection_job_id, lease_token, lease_attempts FROM updated ORDER BY id DESC;

                                                                                                         QUERY PLAN                                                                                                          
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=470774493.09..470775221.09 rows=291200 width=176) (actual time=3000.681..3000.690 rows=20 loops=1)
   Sort Key: updated.id DESC
   Sort Method: quicksort  Memory: 30kB
   CTE updated
[...]

PR #1305, opened 2023-04-27
Drop ordering requirement in acquire_incomplete_aggregation_jobs.

branlwyd:
I can't remember why I specified an ordering, but it's not necessary, and ordering can be expensive. While I'm at it, I specify FOR UPDATE SKIP LOCKED in the inner SELECT query; I think this will effectively change concurrent acquiry attempts from always triggering a serialization failure & retry to never triggering a serialization failure & retry.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -1599,12 +1599,14 @@ impl<C: Clock> Transaction<'_, C> {
                     lease_attempts = lease_attempts + 1
                 FROM tasks
                 WHERE tasks.id = aggregation_jobs.task_id
-                AND aggregation_jobs.id IN (SELECT aggregation_jobs.id FROM aggregation_jobs
+                AND aggregation_jobs.id IN (
+                    SELECT aggregation_jobs.id FROM aggregation_jobs
                     JOIN tasks on tasks.id = aggregation_jobs.task_id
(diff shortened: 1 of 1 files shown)


branlwyd:
(trivial PR, I noticed this while poking at approaches for #1283; I'll make a similar change to the collect-job acquiry code as part of the fix to that issue.)

aggregator_core/src/datastore.rs
```diff
@@ -6847,23 +6849,30 @@ mod tests {
...
-                        PRIO3_VERIFY_KEY_LENGTH,
-                        TimeInterval,
-                        Prio3Count,
-                    >::new(
-                        *task.id(),
-                        aggregation_job_id,
-                        (),
-                        (),
-                        Interval::new(Time::from_seconds_since_epoch(0), Duration::from_seconds(1))
+                try_join_all(aggregation_job_ids.into_iter().map(|aggregation_job_id| {
```
tgeoghegan, line 6852: It occurs to me that there are likely a number of places where we iterate over a long list of reports or report aggregations or aggregation jobs or whatever like this, and could get performance wins either by spawning a bunch of tokio tasks or using [`rayon::iter`](https://docs.rs/rayon/latest/rayon/iter/index.html). No need to change that here: #1307

tgeoghegan (approved):

PR #1313, opened 2023-05-01
Fix high cardinality, HTTP client histogram metric

### LINKED ISSUE #1275: Possible memory leak in aggregation_job_driver
I noticed that the aggregation_job_driver memory usage in our current load test is increasing slowly over time. (approximately linear, (edit: or quadratic?) with some jumps up and some other noise) It has increased by 150 MiB over three and a half days. We are seeing approximately 0.45 aggregation job steps per second. This comes out to 1156 bytes per job step.

We could probably track this down by re-enabling debug symbols and running this component under Valgrind's DHAT tool, though it would require writing a lot of configuration files.

divergentdave:
This limits the cardinality of `janus_http_request_duration_seconds_count` by introducing a separate argument for the `endpoint` label, taking a description of the route, rather than extracting the last component of the URL. Fixes #1275.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -2341,17 +2341,13 @@ async fn send_request_to_helper<T: Encode>(
     http_client: &Client,
     method: Method,
     url: Url,
+    route_label: &'static str,
     content_type: &str,
     request: T,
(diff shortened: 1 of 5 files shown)


tgeoghegan (approved):
Nice find!

PR #1337, opened 2023-05-03
Fix typo

jbr:


tgeoghegan (approved):

tgeoghegan:
This change just changes a single letter in a README, so it's odd that it seems to be encountering a flaky integration test that we haven't encountered elsewhere.

PR #1352, opened 2023-05-09
Aggregator API: add three optional task parameters

divergentdave:
This adds three optional fields to POST request bodies for the /tasks endpoint. This allows task IDs, VDAF verify keys, and aggregator authentication tokens to be provided, instead of randomly generated.


branlwyd (approved):

aggregator_api/src/lib.rs
```diff
@@ -109,15 +112,55 @@ async fn get_task_ids<C: Clock>(
...
-            .collect(),
-    )]);
+) -> Result<impl Handler, impl Handler> {
+    let task_id = req.task_id.unwrap_or_else(random);
+    let vdaf_verify_keys = if let Some(encoded) = req.vdaf_verify_key.as_ref() {
+        let bytes = URL_SAFE_NO_PAD.decode(encoded).map_err(|err| {
+            (
+                format!("Invalid base64 value for vdaf_verify_key: {err}"),
+                Status::BadRequest,
+            )
```
divergentdave, line 122: This doesn't work because `<String as Handler>` uses `conn.ok(...)`, which halts the connection. Then, the tuple handler does not run `<Status as Handler>` because the connection is halted. I think I'll set the body with a mutating method instead, and go back to returning `Status` out of the `ApiHandler`.

divergentdave:
The response from divviup-api to a `POST /account/.../tasks` request does not include the collector authentication token, and we have not yet implemented the "fetch secrets" API endpoint either. I'm thinking of adding an optional collector authentication token parameter to this same aggregator API method, both for parallelism with the aggregator authentication token, and to help unblock integration testing.

aggregator_api/src/lib.rs
jbr: This is a known issue that I'd like to change in the next breaking trillium release

PR #1356, opened 2023-05-11
add a base64 Display implementation for HpkePublicKey

jbr:


divergentdave (approved):
LGTM. I notice the Debug implementation uses hex encoding, and we haven't used that  for public keys anywhere else since writing it. (plus, public keys are skipped in many other Debug implementations for length reasons) Shall we change that to use base64url as well?

jbr:
I wasn't sure if that would be considered a breaking change, whereas adding a new impl should be non-breaking

branlwyd (approved):

jbr:
Is the test failure related?

divergentdave:
Suspicious, it could be an issue in new versions of the Bitnami Postgres container we use, or maybe just something transient. It's very interesting that three years are affected, and not all DB tests.

tgeoghegan (approved):

PR #1360, opened 2023-05-12
Pin Postgres container to 14.7-alpine.

branlwyd:
This is a temporary workaround to an issue in the latest version of the Postgres containers, and will be reverted once a fix is available.


divergentdave (approved):

divergentdave:
Actually, we should also change the tag in `aggregator_core/src/datastore/test_util.rs`

PR #1364, opened 2023-05-12
Remove INVALID report aggregation state.

branlwyd:
This state was only used by the Leader, and only if the Helper sent an unexpected preparation message (a Continue when a Finished was expected, or vice-versa). It was treated by all program logic as equivalent to the Failed state. This change replaces all uses of Invalid with Failed(VdafPrepError), and eliminates the Invalid state entirely.


db/00000000000005_rm_invalid.up.sql
```diff
@@ -0,0 +1,14 @@
...
+-- See: https://stackoverflow.com/questions/25811017/how-to-delete-an-enum-type-value-in-postgres
+
+CREATE TYPE REPORT_AGGREGATION_STATE_NEW AS ENUM(
+    'START',     -- the aggregator is waiting to decrypt its input share & compute initial preparation state
+    'WAITING',   -- the aggregator is waiting for a message from its peer before proceeding
+    'FINISHED',  -- the aggregator has completed the preparation process and recovered an output share
+    'FAILED'     -- an error has occurred and an output share cannot be recovered
+);
+ALTER TABLE report_aggregations ALTER COLUMN state TYPE REPORT_AGGREGATION_STATE_NEW
+    USING (state::text::REPORT_AGGREGATION_STATE_NEW);
```
tgeoghegan, line 12: What happens to rows that had `state == INVALID`?

tgeoghegan (approved):

db/00000000000005_rm_invalid.up.sql
branlwyd: The practical answer is that there aren't any -- we made the decisions that migrations currently do not need to be backwards-compatible, as our organization does not yet have a commitment to any particular data model.

The hypothetical answer is that the migration would fail, as the attempt to cast the string `'INVALID'` to the type `REPORT_AGGREGATION_STATE_NEW` would fail since there is no matching value.

tgeoghegan: OK, works for me, just wanted to make sure that this is intentionally a migration that can't be applied to non-empty DBs.

PR #1381, opened 2023-05-17
Make task expiration optional.

branlwyd:
Previously, we were representing a task that did not expire by using a
very large task expiry value. Using an Option is cleaner, and avoids
tricky boundary-condition issues around the various formats a time might
be represented in.


tgeoghegan (approved):

db/00000000000006_optional_task_expiration.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE tasks ALTER COLUMN task_expiration DROP NOT NULL;
```
jbr, line 1: Is there any data that needs to be migrated, replacing MAX with null?

jbr (approved):

db/00000000000006_optional_task_expiration.up.sql
branlwyd: Nope. Our current stance is that we don't need to provide data migrations as there is no data to migrate -- we will need to change this stance once we have a deployed environment with data we care to keep.

PR #1382, opened 2023-05-18
Make extract_bearer_token() just return a Vec<u8>

divergentdave:
This fixes an issue introduced with the `AuthenticationToken` refactor in #1352. Returning an `AuthenticationToken` from this function requires enforcing unnecessary validity conditions on the base64-decoded value. In this context, the returned value was decoded from base64 in a header value, so the DAP-Auth-Token related conditions do not apply, and any byte values should be allowed. For the DAP endpoint, the `AuthenticationToken` conversion and checks are pushed out to `parse_auth_token()` instead, while for the aggregator API endpoint, `AuthenticationToken` is no longer involved.

### DIFF
--- a/aggregator/src/aggregator/http_handlers.rs
+++ b/aggregator/src/aggregator/http_handlers.rs
@@ -532,7 +532,16 @@ fn parse_auth_token(task_id: &TaskId, conn: &Conn) -> Result<Option<Authenticati
     let bearer_token =
         extract_bearer_token(conn).map_err(|_| Error::UnauthorizedRequest(*task_id))?;
     if bearer_token.is_some() {
-        return Ok(bearer_token);
+        return bearer_token
+            .map(AuthenticationToken::try_from)
+            .transpose()
+            .map_err(|_| {
+                Error::BadRequest(
+                    "Authorization: Bearer value decodes to an authentication token containing \
+                     unsafe bytes"
+                        .to_string(),
+                )
+            });
     }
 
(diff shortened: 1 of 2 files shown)


tgeoghegan (commented):
I think `extract_bearer_token` is easier and safer to use if it returns `AuthenticationToken`. If `AuthenticationToken::try_from(&[u8])` is inappropriate for a bearer token, then maybe we need another, infallible constructor on `AuthenticationToken` to use in the case where we are handling a bearer token?

Additionally, it seems awkward to me that we try to guarantee that `DAP-Auth-Token` values are themselves valid ASCII instead of just opaque bytes. It forces a number of awkward validations like this on us, and also it's inefficient. We generate authentication tokens thusly:

```
impl Distribution<AuthenticationToken> for Standard {
    fn sample<R: rand::Rng + ?Sized>(&self, rng: &mut R) -> AuthenticationToken {
        AuthenticationToken(Vec::from(hex::encode(rng.gen::<[u8; 16]>())))
    }
}
```

So there's 16 actual bytes of entropy, but the token stored in DBs and sent on the wire is twice as big after being hex encoded.

But let's keep this change focused on resolving the impedance mismatch you hit. I filed https://github.com/divviup/janus/issues/1384 for the other question.

branlwyd (approved):

branlwyd:
(see discussion on #1384, I think this is a good change.)

PR #1383, opened 2023-05-18
Rewrite in-cluster integration test with divviup-api

### LINKED ISSUE #451: `monolithic_integration_test::janus::Janus` should not conflate container & Kubernetes functionality
These two variants for this type share no functionality whatsoever, and conflating the functionality has made maintenace (of the Container variant, so far) more painful.

Instead, we could use two separate types; we may not even need a Kubernetes type, instead inlining the Kubernetes-relevant code from `Janus` (& `JanusPair`) into the relevant test, since the Kubernetes version is needed only for the "smoke test" we run for deployment config changes. Another approach to the smoke test would be to build a small binary and run it rather than co-opting an existing integration test for this purpose, though this may well be more expensive in terms of build time.

### LINKED ISSUE #1343: Change kind integration test to create tasks via divviup-api
Now that we've got divviup-api deployed in Kind clusters, we'd like to exercise it as part of an integration test. For any Janus deployment, we need to choose between Janus or divviup-api having the authoritative list of tasks. (or else, tasks created via janus_cli or direct database access would get deleted by divviup-api's periodic reconciliation) In the interest of keeping testing simple, I think we should get rid of the Postgres port forward and `put_task()` call that `Janus::new_with_kubernetes_cluster()` uses, and only create DAP tasks in the `kind-integration-test` CI job by way of divviup-api.

This may also be a good time to address #451. We could split the `Janus` enum into two separate structs, for starters. Cluster-based integration tests should have their own integration test binary, which is not compiled and run by default.

divergentdave:
This rewrites the tests used by `kind-integration-test` in the janus-ops repo so that they create tasks using a divviup-api control plane, instead of direct database access. The tests are additionally broken out from the Janus-Janus container-based integration tests to simplify things, and guarded by a new Cargo feature, to gate other dependencies. Note that this will require some lockstep changes in janus-ops when it upgrades, to pass additional environment variables and invoke the correct integration test.

Closes #1343. Closes #451.


integration_tests/src/janus.rs
```diff
@@ -174,29 +58,27 @@ impl<'a> Drop for Janus<'a> {
         // (log export is a no-op for non-containers: when running tests against a cluster, we
```
branlwyd, line 58: nit: this comment should be updated, probably to drop the parenthetical sentence -- the non-container variant is now gone.

branlwyd (approved):

integration_tests/tests/in_cluster.rs
```diff
@@ -0,0 +1,420 @@
...
+}
+
+#[derive(Serialize)]
+struct Sum {
+    bits: usize,
+}
+
+#[derive(Serialize)]
+#[serde(rename_all = "snake_case", tag = "type")]
+enum ApiVdaf {
```
tgeoghegan, line 194: Seems odd to me that this integration test package should have to define API request types for `divviup-api`. Is there a plan to export these somehow from `divviup-api`? Or would that create a dependency cycle?

divergentdave: I hadn't considered that, but it would have to be via a git dependency, unless we want to publish divviup-api (or part of it) to crates.io. Since we can only release all Janus crates at the same time, if there were to be future breaking changes to `HpkeConfig`, we might have to disable the integration test for a release, but otherwise dependency cycles wouldn't be a concern.

tgeoghegan: Let's move forward with this change as-is, and we can consider publishing a Rust client for `divviup-api` separately.

PR #1385, opened 2023-05-18
reexport prio::codec so Decode and Encode implementations are usable

jbr:


branlwyd:
Is this to avoid a direct dependency on `libprio-rs`, presumably in `divviup-api`? If there's already an indirect dependency, could we add a direct dependency on `libprio-rs`?

jbr:
In general, I think that any crate should reexport any traits that would be needed to interact with the exported types

divergentdave (approved):
I'm in favor of this, it'll be better for API ergonomics, and it will also simplify upgrading to the next minor version of this crate in consumer projects.

branlwyd (approved):
That's fair. 👍🏻 I haven't seen good guidelines for when publicly exporting another library's types is appropriate (seems like it would be easy to create undesirably-tight coupling), but this situation seems like a good use case for a `pub use`.

jbr:
If anything, it feels like the codec stuff should be a crate of its own (`tls-encoding`?), since it's not really about prio

jbr:
Because external traits are implemented, the coupling is already 1:1 — if a dependent crate upgrades to an incompatible version of prio, the trait implementations wouldn't be available.

PR #1386, opened 2023-05-18
use STANDARD_NO_PAD for public key

jbr:


branlwyd:
Could you describe why switching from websafe base64 to standard base64 is desirable?

PR #1387, opened 2023-05-18
Docker improvements

divergentdave:
This makes some changes to speed up Docker builds, particularly in CI. Cache mounts are abandoned in favor of using `cargo-chef` to split dependency compilation artifacts and first-party compilation artifacts into separate layers. `ARG` directives are pushed down as far as possible to allow layers to be reused between different image builds. Repetitive `docker build` command lines are replaced with `docker buildx bake` and a new `docker-bake.hcl` file which represents each existing `docker build` invocation declaratively. Using `bake` also allows for better pipelined scheduling of multiple container image builds, which is relevant for our CI jobs. The GitHub Actions cache backend is enabled, with three unique builds writing to the cache, and all builds reading from whichever cache scope is the best fit. The janus_interop_binaries container images are built with the "small" profile in our build-and-test CI job before `cargo build`, so that layer caching can be used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,4 @@
 FROM rust:1.69.0-alpine as builder
-ARG BINARY=aggregator
-ARG GIT_REVISION=unknown
 RUN apk add libc-dev
 WORKDIR /src
 COPY Cargo.toml /src/Cargo.toml
@@ -17,6 +15,8 @@ COPY integration_tests /src/integration_tests
 COPY interop_binaries /src/interop_binaries
 COPY messages /src/messages
 COPY tools /src/tools
+ARG BINARY=aggregator
+ARG GIT_REVISION=unknown
 ENV GIT_REVISION ${GIT_REVISION}
 RUN --mount=type=cache,target=/usr/local/cargo/registry --mount=type=cache,target=/src/target cargo build --release -p janus_aggregator --bin $BINARY --features=prometheus && cp /src/target/release/$BINARY /$BINARY
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,7 @@
 FROM rust:1.69.0-alpine as builder
 RUN apk add libc-dev
 WORKDIR /src
-COPY Cargo.toml /src/Cargo.toml
-COPY Cargo.lock /src/Cargo.lock
+COPY Cargo.toml Cargo.lock /src/
 COPY aggregator /src/aggregator
 COPY aggregator_api /src/aggregator_api
 COPY aggregator_core /src/aggregator_core
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,27 @@
-FROM rust:1.69.0-alpine as builder
+FROM rust:1.69.0-alpine AS chef
 RUN apk add libc-dev
+ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse
+RUN cargo install cargo-chef --version 0.1.60
 WORKDIR /src
+
+FROM chef AS planner
+COPY Cargo.toml Cargo.lock /src/
+COPY aggregator /src/aggregator
+COPY aggregator_api /src/aggregator_api
+COPY aggregator_core /src/aggregator_core
+COPY build_script_utils /src/build_script_utils
+COPY client /src/client
+COPY collector /src/collector
+COPY core /src/core
+COPY integration_tests /src/integration_tests
+COPY interop_binaries /src/interop_binaries
+COPY messages /src/messages
+COPY tools /src/tools
+RUN cargo chef prepare --recipe-path recipe.json
+
+FROM chef AS builder
+COPY --from=planner /src/recipe.json /src/recipe.json
+RUN cargo chef cook --release -p janus_aggregator --features=prometheus
 COPY Cargo.toml Cargo.lock /src/
 COPY aggregator /src/aggregator
 COPY aggregator_api /src/aggregator_api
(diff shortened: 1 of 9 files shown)


docker-bake.hcl
```diff
@@ -0,0 +1,272 @@
...
+    "janus_interop_collector_small",
+  ]
+}
+
+target "janus_aggregator" {
+  args = {
+    GIT_REVISION = "${GIT_REVISION}"
+  }
+  cache-from = [
+    "type=gha,scope=main-janus",
```
tgeoghegan, line 70: What happens when we `docker buildx bake` outside of GitHub Actions? Are the `cache-to/from` directives gracefully ignored?

Could we add a `CACHE_TYPE` variable that defaults to `local` so that we get a speedup when building images locally?

tgeoghegan (approved):
`cargo chef` and `docker buildx bake` are new to me, but the speedups this should yield are compelling.

Should `interop_binaries/build.rs` be updated to use `docker buildx bake`?

Dockerfile
```diff
@@ -1,10 +1,29 @@
-FROM rust:1.69.0-alpine as builder
-ARG BINARY=aggregator
-ARG GIT_REVISION=unknown
-RUN apk add libc-dev protobuf-dev protoc
+FROM rust:1.69.0-alpine AS chef
+RUN apk add libc-dev
```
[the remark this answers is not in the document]
divergentdave: I split installation of those packages out into a later layer (in the builder phase) so that all docker builds can share the same layers up to installation of `cargo-chef`. (We should be able to drop them soon, once we can upgrade opentelemetry, but that requires shaving some other yaks first)

docker-bake.hcl
divergentdave: If the `gha` cache backend is used outside of GitHub Actions, it fails silently, because it doesn't get either the cache API URL or the token from special environment variables. (This was a stumbling block for Jacob and I previously, but it also means these directives won't get in the way of local builds)

The documentation says that the "local BuildKit cache ... is always enabled", and empirically if I run `docker buildx bake` repeatedly, layer caching works as-is. I think the `local` cache backend is distinct, and writes unpacked OCI images to the filesystem at a user-defined location.

divergentdave (commented):
I wasn't originally planning to run `docker buildx bake` from the build script, as either should work well enough, though the pipelining might be nice to reduce the total runtime of the build script. To do so, I think I'd have to either pass `--metadata-file` and parse its output, or override `output` to write to a temporary tar file. Looking at the first CI run, it seems caches aren't being reused as I'd hope between the `docker buildx bake` execution and the later build script commands, based on execution time, so I may need to revisit which Docker drivers and which caches are involved.

tgeoghegan:
> I wasn't originally planning to run `docker buildx bake` from the build script, as either should work well enough, though the pipelining might be nice to reduce the total runtime of the build script. To do so, I think I'd have to either pass `--metadata-file` and parse its output, or override `output` to write to a temporary tar file. Looking at the first CI run, it seems caches aren't being reused as I'd hope between the `docker buildx bake` execution and the later build script commands, based on execution time, so I may need to revisit which Docker drivers and which caches are involved.

I suppose  we could also spawn multiple `docker build` subcommands in parallel in the build script to get parallelism without taking on `docker buildx bake`. In any case, it sounds like we could defer improvements to `build.rs` to a later change, to unblock this change which will achieve its goal of improving CI times.

divergentdave:
Next wrinkle: I have to use `docker buildx build` instead of `docker build` to use the same set of local caches. (and to configure the GHA remote cache source, for good measure) The `--load` flag only causes the final container image, and not the image manifest, to be loaded into the Docker daemon out of the containerized BuildKit worker. When passing `--quiet` and `--iidfile=...`, both stdout and the iidfile get a SHA-256 hash, but they differ. The hash from standard output points to a manifest, (not present on the host) while the hash from the iidfile points to an image. This seems like a bug, as the respective flags for `docker build` provide the same hash, and the documentation for `docker buildx build` says `--quiet` causes it to "Suppress the build output and print image ID on success". It'll be easy enough to work around.

divergentdave:
This is ready to go now. I'm seeing improvements on janus_build, janus_docker, and janus_interop_docker of various magnitudes. (depending on whether dependencies and code have changed) I checked through log inspection that the commands run from inside the build script are reusing locally cached image layers, as intended. I've looked over layers from each build stage, and I think they're looking good in terms of efficiency, there are no large files/directories going unused in cached layers.

interop_binaries/build.rs
```diff
@@ -47,107 +48,129 @@ fn main() {
...
+                            "buildx",
                             "build",
-                            "--quiet",
                             "--file=Dockerfile.interop",
                             "--build-arg=PROFILE=small",
                             "--build-arg=BINARY=janus_interop_client",
-                            ".",
+                            "--load",
                         ])
+                        .arg(format!("--iidfile={image_id_file_name}"))
```
branlwyd, line 62: nit: instead of mixing `args` and `arg` calls, use a single `args` call. (the `format!` parameters can be mixed in by writing `&format!(...)`, since everything will need to be a `&str`.)

README.md
```diff
@@ -43,6 +43,8 @@ DOCKER_BUILDKIT=1 docker build --tag=janus_aggregation_job_driver --build-arg BI
 DOCKER_BUILDKIT=1 docker build --tag=janus_collection_job_driver --build-arg BINARY=collection_job_driver .
 ```
 
+Alternately, run `docker buildx bake` to build images in parallel.
```
branlwyd, line 46: Is there any reason not to suggest only the `docker buildx bake` approach? i.e. is there a use-case for building individual images anymore? If not, I think we can just drop the old instructions entirely.

branlwyd (approved):

Issue #472, opened 2022-09-06
Revisit request authentication


tgeoghegan:
draft-ietf-ppm-dap-02 no longer requires the DAP-Auth-Token:

https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/293
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/305

DAP-Auth-Token isn't forbidden by the new text, so we could leave that support in for backward compatibility with other DAP implementations, but we should look into implementing more secure authentication schemes for Divvi Up, perhaps leaning on managed identity offerings.

tgeoghegan:
I think we will not make changes here as part of our DAP-02 implementation, since this touches on some bigger questions about subscriber management and the Divvi Up control plane. Removing this from the draft-ietf-ppm-dap-02 milestone.

tgeoghegan:
In the future, we might implement more sophisticated authz -- for instance, issue tokens that are only permitted to exercise either the aggregation job or collection job sub-protocols -- but I think we have everything we need in place for the initial production deployment(s) of Divvi Up.

@branlwyd @divergentdave Is there anything else you think we should do w.r.t. request authentication? At most I think we might want to put a version into the bearer tokens we generate, which might make it easier to migrate to more sophisticated schemes in the future, but I'm not sure that's necessary.

branlwyd:
I think we are fine with a bearer token. I wouldn't object to adding a version, but I don't think it's strictly necessary -- presumably any authentication token will have enough unguessable material that it is overwhelmingly unlikely that a valid auth token for one scheme is also valid in another scheme. (or, to think of it another way, a token from one scheme being valid in another scheme had better not be any more likely than a random string of bytes being valid in that scheme) OTOH, including a version would make the hypothetical code to parse multiple versions of tokens much simpler. A nice-to-have but not a requirement.

divergentdave:
We should probably use a different header name when we change to a different authentication scheme, be it `Authorization` or something else. If that's the case, then I think it's fine to continue using `DAP-Auth-Token` with no version number structure in its value.

tgeoghegan:
Following discussion in Slack: we're going to migrate to using proper bearer tokens. The primary motivation here is consistency across our API surfaces. To ease migration (for instance with Daphne), we'll examine both the `DAP-Auth-Token` and `Authorization` headers for some period and then cut over to exclusively honoring bearer tokens. #1300 tracks deprecating `DAP-Auth-Token`.

[...]

PR #1388, opened 2023-05-19
Track authentication mode per-auth-token

tgeoghegan:
Previously, we taught Janus to accept authentication tokens presented in HTTP requests as either `DAP-Auth-Token: <token>` or `Authorization: Bearer <token>`. However, we also want a Janus leader to be able to work with a DAP helper that expects only one or the other authentication mode.

The `task_aggregator_auth_tokens` and `task_collector_auth_tokens` tables now have columns that track the type of auth token, which is either `DAP_AUTH` or `BEARER`. `send_request_to_helper` will now consult this value to decide how it should authenticate requests.

We didn't strictly need this column in `task_collector_auth_tokens` as the leader's collection job handlers will still accept either kind of token, but it's nice to have consistent handling of these types across the two tables.

The distinction between DAP-Auth-Token and bearer tokens is plumbed up from `janus_core` through `janus_aggregator` by combining `janus_core::task::AuthenticationToken` with `janus_collector::Authentication` into an enum that distinguishes between the two authentication modes and can produce suitable HTTP header names and values.

Note that we track authentication mode per-token and not per-task. This means that it's possible for a task to use two different kinds of tokens.

A task defined in YAML and provisioned via `janus_cli` may use Bearer tokens, but this isn't yet possible via `janus_aggregator_api`, which still assumes it is receiving or generating `DAP-Auth-Token`-form tokens. The aggregator API for creating tasks will have to change to accommodate the distinction between token types, and that will arrive in a later change so that it can be coordinated with `divviup-api`.

Part of #472

### DIFF
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -9,8 +9,10 @@ use janus_aggregator_core::{
     SecretBytes,
 };
 use janus_core::{
-    hpke::generate_hpke_config_and_private_key, http::extract_bearer_token,
-    task::AuthenticationToken, time::Clock,
+    hpke::generate_hpke_config_and_private_key,
+    http::extract_bearer_token,
+    task::{AuthenticationToken, DapAuthToken},
+    time::Clock,
 };
 use janus_messages::{Duration, HpkeAeadId, HpkeKdfId, HpkeKemId, Role, TaskId, Time};
 use models::{GetTaskMetricsResp, TaskResp};
@@ -170,17 +172,25 @@ async fn post_task<C: Clock>(
                 Status::BadRequest,
             )
         })?;
-        Vec::from([AuthenticationToken::try_from(token_bytes).map_err(|_| {
-            Error::new(
-                "Invalid HTTP header value in aggregator_auth_token".to_string(),
-                Status::BadRequest,
-            )
-        })?])
+        Vec::from([
+            // TODO(#472): Each token in the PostTaskReq should indicate whether it is a bearer
+            // token or a DAP-Auth-Token. For now, assume the latter.
+            DapAuthToken::try_from(token_bytes)
+                .map(AuthenticationToken::DapAuth)
+                .map_err(|_| {
+                    Error::new(
+                        "Invalid HTTP header value in aggregator_auth_token".to_string(),
+                        Status::BadRequest,
+                    )
+                })?,
+        ])
     } else {
-        Vec::from([random()])
+        // TODO(#472): switch to generating bearer tokens by default
+        Vec::from([AuthenticationToken::DapAuth(random())])
     };
     let collector_auth_tokens = match (req.role, req.collector_auth_token) {
-        (Role::Leader, None) => Vec::from([random()]),
+        // TODO(#472): switch to generating bearer tokens by default
+        (Role::Leader, None) => Vec::from([AuthenticationToken::DapAuth(random())]),
         (Role::Leader, Some(encoded)) => {
             let token_bytes = URL_SAFE_NO_PAD.decode(encoded).map_err(|err| {
                 Error::new(
@@ -188,12 +198,16 @@ async fn post_task<C: Clock>(
                     Status::BadRequest,
                 )
             })?;
-            Vec::from([AuthenticationToken::try_from(token_bytes).map_err(|_| {
-                Error::new(
-                    "Invalid HTTP header value in collector_auth_token".to_string(),
-                    Status::BadRequest,
-                )
-            })?])
+            // TODO(#472): Each token in the PostTaskReq should indicate whether it is a bearer
+            // token or a DAP-Auth-Token. For now, assume the latter.
+            Vec::from([DapAuthToken::try_from(token_bytes)
+                .map(AuthenticationToken::DapAuth)
+                .map_err(|_| {
+                    Error::new(
+                        "Invalid HTTP header value in collector_auth_token".to_string(),
+                        Status::BadRequest,
+                    )
+                })?])
         }
         (Role::Helper, None) => Vec::new(),
         (Role::Helper, Some(_)) => {
@@ -486,7 +500,7 @@ mod tests {
     };
     use janus_core::{
         hpke::{generate_hpke_config_and_private_key, HpkeKeypair, HpkePrivateKey},
-        task::{AuthenticationToken, VdafInstance},
+        task::{AuthenticationToken, DapAuthToken, VdafInstance},
         test_util::{
             dummy_vdaf::{self, AggregationParam},
             install_test_trace_subscriber,
@@ -712,8 +726,8 @@ mod tests {
 
         let vdaf_verify_key =
             SecretBytes::new(thread_rng().sample_iter(Standard).take(16).collect());
-        let aggregator_auth_token = random::<AuthenticationToken>();
-        let collector_auth_token = random::<AuthenticationToken>();
+        let aggregator_auth_token = AuthenticationToken::DapAuth(random());
+        let collector_auth_token = AuthenticationToken::DapAuth(random());
 
         // Verify: posting a task creates a new task which matches the request.
         let req = PostTaskReq {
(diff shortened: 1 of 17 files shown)


aggregator_core/src/task.rs
```diff
@@ -1363,64 +1346,4 @@ mod tests {
             ],
         );
     }
-
-    #[test]
-    fn reject_invalid_auth_tokens() {
```
tgeoghegan, line 1368: The new `janus_core::task::tests::reject_invalid_dap_auth_token` more narrowly tests that invalid `DapAuthToken`s are rejected.

core/src/task.rs
```diff
@@ -533,48 +537,151 @@ macro_rules! vdaf_dispatch {
...
+}
 
 impl AuthenticationToken {
-    /// Constructs a bearer token string suitable for use as the value in an HTTP `Authorization`
-    /// header.
-    pub fn bearer_token(&self) -> String {
-        format!("Bearer {}", STANDARD.encode(self.as_ref()))
+    /// Returns an HTTP header and value that should be used to authenticate an HTTP request with
+    /// this credential.
+    pub fn request_authentication(&self) -> (&'static str, Vec<u8>) {
```
tgeoghegan, line 571: Ideally this would be expressed as something like an extension trait on `reqwest::RequestBuilder` with a method `RequestBuilder::with_authentication(token: &AuthenticationToken)`, but we need to use these values in `mockito` and `trillium` as well, hence this somewhat less ergonomic, lowest-common-denominator approach.

db/00000000000009_auth_token_kind.up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TYPE AUTH_TOKEN_TYPE AS ENUM(
+    'DAP_AUTH', -- DAP-01 style DAP-Auth-Token header
+    'BEARER' -- RFC 6750 bearer token
+);
+
+ALTER TABLE task_aggregator_auth_tokens ADD COLUMN type AUTH_TOKEN_TYPE NOT NULL DEFAULT 'DAP_AUTH';
+ALTER TABLE task_collector_auth_tokens ADD COLUMN type AUTH_TOKEN_TYPE NOT NULL DEFAULT 'DAP_AUTH';
```
tgeoghegan, line 7: Any existing auth tokens are assumed to be DAP auth.

core/src/task.rs
```diff
@@ -533,48 +537,151 @@ macro_rules! vdaf_dispatch {
...
+            (Self::Bearer(own), Self::Bearer(other)) => (own.as_slice(), other.as_slice()),
+            (Self::DapAuth(own), Self::DapAuth(other)) => (own.as_ref(), other.as_ref()),
+            _ => {
+                return false;
+            }
+        };
+        // We attempt constant-time comparisons of the token data. Note that this function still
+        // leaks whether the lengths of the tokens are equal -- this is acceptable because we expect
+        // the content of the tokens to provide enough randomness that needs to be guessed even if
+        // the length is known.
```
inahga, line 597:
```suggestion
        // We attempt constant-time comparisons of the token data to mitigate timing attacks. Note
        // that this function still leaks whether the lengths of the tokens are equal -- this is acceptable
        // because we expect the content of the tokens to provide enough randomness that needs to 
        // be guessed even if the length is known.
```

Optional: Adding "to mitigate timing attacks", prefer to be explicit about what is being accomplished.

docs/samples/tasks.yaml
```diff
@@ -65,20 +65,28 @@
...
+  # header. For example, the first token's value decodes to
   # "aggregator-235242f99406c4fd28b820c32eab0f68".
   aggregator_auth_tokens:
-  - "YWdncmVnYXRvci0yMzUyNDJmOTk0MDZjNGZkMjhiODIwYzMyZWFiMGY2OA"
+  - type: "DapAuth"
+    token: "YWdncmVnYXRvci0yMzUyNDJmOTk0MDZjNGZkMjhiODIwYzMyZWFiMGY2OA"
+  - type: "Bearer"
+    token: "YWdncmVnYXRvci04NDc1NjkwZjJmYzQzMDBmYjE0NmJiMjk1NDIzNDk1NA"
 
   # Authentication tokens shared between the leader and the collector, and used
```
inahga, line 82: Just curious: were there any thoughts about authenticating leader->collector and leader->helper communication using mutual TLS? Just thinking it would simplify authentication logic, insofar as we don't have to compare tokens ourselves. Not saying we should do that, moreso wondering if the conversation ever came up.

aggregator_api/src/lib.rs
```diff
@@ -170,14 +172,21 @@ async fn post_task<C: Clock>(
...
         })?;
-        Vec::from([AuthenticationToken::try_from(token_bytes).map_err(|_| {
-            Error::new(
-                "Invalid HTTP header value in aggregator_auth_token".to_string(),
-                Status::BadRequest,
-            )
-        })?])
+        Vec::from([
+            // TODO(#472): Each token in the PostTaskReq should indicate whether it is a bearer
+            // token or a DAP-Auth-Token. For now, assume the latter.
```
inahga, line 177: Not sure I follow--does this mean that with this PR, all Task POSTs will record themselves as `DapAuth` tokens regardless of them being `Bearer` or `DapAuth` token?

It smells like this would blow up if someone attempts to use a bearer token and has it recorded as the wrong type in the DB.

inahga (commented):
Nicely done. One major question at the end, rest are nits or minor questions.

aggregator_api/src/lib.rs
divergentdave: This is just for the "aggregator API" that the control plane uses to set up new tasks. The control plane does not yet have a concept of different types of authentication tokens, so this won't lead to any surprises.

aggregator_api/src/lib.rs
```diff
@@ -170,14 +172,21 @@ async fn post_task<C: Clock>(
...
+                    Error::new(
+                        "Invalid HTTP header value in aggregator_auth_token".to_string(),
+                        Status::BadRequest,
+                    )
+                })?,
+        ])
     } else {
-        Vec::from([random()])
+        // TODO(#472): switch to generating bearer tokens by default
+        Vec::from([AuthenticationToken::DapAuth(random())])
```
divergentdave, line 189: The `(Role::Leader, None)` arm in the match block below should match this.

core/src/task.rs
```diff
@@ -533,48 +537,151 @@ macro_rules! vdaf_dispatch {
...
+    /// A bearer token. The value is an opaque byte string. Its Base64 encoding is inserted into
+    /// HTTP requests as specified in [RFC 6750 section 2.1][1]. The token is not necessarily an
+    /// OAuth token.
+    ///
+    /// [1]: https://datatracker.ietf.org/doc/html/rfc6750#section-2.1
+    Bearer(
+        #[derivative(Debug = "ignore")]
+        #[serde(
+            serialize_with = "as_url_safe_unpadded_base64",
+            deserialize_with = "from_url_safe_unpadded_base64"
```
divergentdave, line 556: Since this is base64-encoded on the wire in headers, what do you think about changing the YAML/JSON serialization to use base64 as well?

core/src/task.rs
```diff
@@ -533,48 +537,151 @@ macro_rules! vdaf_dispatch {
...
+}
+
+impl Distribution<AuthenticationToken> for Standard {
+    fn sample<R: rand::Rng + ?Sized>(&self, rng: &mut R) -> AuthenticationToken {
+        AuthenticationToken::Bearer(Vec::from(rng.gen::<[u8; 16]>()))
+    }
+}
+
+/// Token presented as the value of the "DAP-Auth-Token" HTTP header. The token is used directly in
+/// the HTTP request without further encoding and so must be a legal HTTP header value. Confirms to
```
divergentdave, line 640: nit: spelling
```suggestion
/// the HTTP request without further encoding and so must be a legal HTTP header value. Conforms to
```

inahga (approved):
Major feedback addressed.

core/src/task.rs
tgeoghegan: Yeah, I think using plain base64 makes more sense in a config file, as URL safety doesn't matter there. It's inconsistent with how we encode DAP-Auth-Tokens, but I can live with that wart.

PR #1401, opened 2023-05-23
Breaking changes to Collector constructors

divergentdave:
This removes the deprecated `CollectorParameters::new()` constructor, and renames `CollectorParameters::new_with_authentication()` to `CollectorParameters::new()`. Since we're already going to be releasing 0.5.0, I'd like to batch up this breaking change as well, to clean out some cruft.


inahga (approved):
What's the implication, if any, for the `draft-ietf-ppm-dap-02` branch?

PR #1402, opened 2023-05-23
docs: don't use podman, use docker

inahga:
The exact reasoning is that there are differences between the output of `docker inspect` and `podman inspect` that [cause serialization panics](https://github.com/testcontainers/testcontainers-rs/blob/b22713aae20f6da1c1a15b007410054014c1ba6b/testcontainers/src/clients/cli.rs#L339) in `testcontainers`. In fact, it's their [intent to discourage podman usage](https://github.com/testcontainers/testcontainers-rs/issues/422).

RHEL derivatives come default with podman, so hopefully this blurb will save the next person a bit of troubleshooting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,9 @@ branch.
 [`docker`](https://www.docker.com) must be installed at build time, and the
 `docker` daemon must be running. To build Janus, execute `cargo build`.
 
+Note that `podman` is not an acceptable substitute for `docker`. There are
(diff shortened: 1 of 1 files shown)


inahga:
Ah, I didn't realize `podman` considered those differences bugs. Good to know, I can report it.

branlwyd (approved):

PR #1403, opened 2023-05-23
Add advice for simple database deployments.

### LINKED ISSUE #1391: Best practice for initializing db for janus with sqlx-cli
Hi!

As described in the [documentation for deployment](https://github.com/dpsa-project/janus/blob/main/docs/DEPLOYING.md), `sqlx-cli` has to be used to apply all migrations in the `db` folder to a fresh postgres database. I have a question with regards to how this could be done in the context of a `docker-compose.yml` file.

Previously, we could start a local janus instance (with databases and 2 aggregation servers) using a [single `docker-compose` file](https://github.com/dpsa-project/dpsa4fl-infrastructure/blob/main/local-docker-setup/docker-compose.yml). This was possible because the standard `postgres` docker container image could be initialized with the single `schema.sql` file that was previously available for janus. Now that the additional `sqlx-cli` executable is required, it is unclear to me how a `docker-compose` setup might look like.

My initial guess is that I want to build an additional container image containing the `sqlx-cli` executable and running the `sqlx-cli migrate run` command. This would be part of my compose file, and by setting explicit dependencies between the containers things should run in the correct order.

Nevertheless, this seems like an unelegant solution, in particular since building an image for the sole purpose of initializing a database seems like overkill. Are there easier ways to achieve a setup where running a single command like `docker-compose up` would setup both databases and also janus? Is the best solution to split database setup and janus setup into two compose files?

I realize that this question might not be directly relevant to janus development, though I'm asking in the hope of someone more experienced having some intuition for what would be the best way to proceed for me. Thanks!

branlwyd:
Closes #1391.

### DIFF
--- a/docs/DEPLOYING.md
+++ b/docs/DEPLOYING.md
@@ -103,6 +103,13 @@ using the `--source` argument to point to `janus/db` and providing database
 connection information in any of the ways supported by `sqlx` (see its
 documentation).
 
+For simple or experimental deployments where the complexity of `sqlx` is not
+warranted, it is possible to create a single schema file by concatenating the
+`.up.sql` scripts, in order, and applying this schema to the database. Note that
+such deployments will not easily be able to migrate to later versions of the
+schema, so this technique is likely not appropriate for deployments which need
+to retain data across deployments.
+
 [sqlx-cli]: https://crates.io/crates/sqlx-cli
 
 ## `janus_cli provision-tasks`
--- a/docs/DEPLOYING.md
+++ b/docs/DEPLOYING.md
@@ -105,10 +105,11 @@ documentation).
 
 For simple or experimental deployments where the complexity of `sqlx` is not
 warranted, it is possible to create a single schema file by concatenating the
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):

docs/DEPLOYING.md
```diff
@@ -103,6 +103,13 @@ using the `--source` argument to point to `janus/db` and providing database
 connection information in any of the ways supported by `sqlx` (see its
 documentation).
 
+For simple or experimental deployments where the complexity of `sqlx` is not
+warranted, it is possible to create a single schema file by concatenating the
+`.up.sql` scripts, in order, and applying this schema to the database. Note that
```
divergentdave, line 108: Since my last comment, I realized that `check_schema_version: false` will have to be set when using this method. I'm working on a separate change to add `check_schema_version: true` in the sample files, after which the below sentence will be self-explanatory.

```suggestion
`.up.sql` scripts, in order, and applying this schema to the database.
When using this technique, `check_schema_version: false` must be set in each configuration file.
Note that
```

divergentdave (approved):

PR #1405, opened 2023-05-23
Documentation and sample configuration updates

divergentdave:
This makes some updates to our documentation, based on recent feedback. I created a separate set of "basic" configuration files, without the optional tracing/metrics sections, because those were getting in the way of first-time users. I also made janus_cli follow the `check_schema_version` setting if present, and made other miscellaneous docs documentation corrections.


branlwyd (commented):
I like the idea of pulling out "minimal" basic config files to reduce confusion for first-time users. My only concern is that placing them all together in a single directory, named in a way that will cause them to be interleaved by a name-based sort, might be confusing. WDYT about a `config_samples/basic` directory and a `config_samples/advanced` directory? or, alternatively, a naming scheme for these files that sorts the basic configs together under lexicographic sort?

tgeoghegan (approved):

inahga (approved):

branlwyd (approved):

PR #1406, opened 2023-05-23
Integration test: give aggregators a chance to settle.

branlwyd:


branlwyd:
N.B. I couldn't reproduce the test failure (https://github.com/divviup/janus/actions/runs/5050514958/jobs/9063568437) on my workstation, so this is somewhat a guess at the correct solution.

tgeoghegan:
I can't ever repro this locally either. I think it has to do with the VMs that GitHub Actions runs tests in, which are presumably slower than our workstations.

tgeoghegan (approved):

PR #1409, opened 2023-05-23
Bump crate versions to 0.5.0

tgeoghegan:
While we still implement `draft-ietf-ppm-dap-04`, breaking API changes have landed in `janus_collector`, so we bump the minor version for SemVer compatibility.


inahga (approved):
Aye.

branlwyd (approved):

PR #1413, opened 2023-05-24
Aggregation job writer improvements.

### LINKED ISSUE #1380: Integration test: improve method to wait for all reports to begin aggregation
Currently, the end-to-end integration test implementation in https://github.com/divviup/janus/blob/main/interop_binaries/tests/end_to_end.rs is racy: it expects that it can upload reports, then with no further synchronization or waiting submit a collection request, and have that collection request include all uploaded reports.

This is technically racy at time of writing: the query used to find collection jobs to process would allow a collection job to be processed even if there are reports that have not yet been added to an aggregation job at all -- I believe this test "worked" because the test's minimum batch size required all the reports to be aggregated for collection to succeed, and prior to #1354, there was nothing stopping an aggregation job from being created & processed concurrently with a collection job over the same batch.

This raciness will soon be more consequential: with #1354, once a collection request is received, no further aggregation jobs can be created if there is ever a point where there are no in-progress aggregation jobs. So unless the aggregation job creator "wins" the race against the collection job request, at least one report will never be aggregated, which will cause the test to fail.

This is worked around in #1354 by waiting (arbitrarily) 5 seconds after report uploads complete, before issuing a collection request. We should investigate a more robust fix, either by modifying our test's implementation or by modifying the semantics of the interop test API.

branlwyd:
Two changes:
* Semantics change: for time-interval tasks, we now require all reports included in the batch to be associated with some aggregation job before the batch can be closed. Effectively, this means that any reports which are received before the collection request arrives are now guaranteed to go through the aggregation process before the collection request is complete, which should remove flakiness from integration test scenarios without requiring arbitrary pauses. It is also useful in a "production" scenario: a collector can now issue a collection request after the end of an interval and be guaranteed that all reports received in the interval will be included in the collection result, even if the aggregation process is running behind.
* Bugfix: initial batch writes, if contended (e.g. if the first two aggregation jobs for a given batch complete concurrently), would fail the DB transaction rather than triggering a retry. This would lead to a retry at the aggregation job level, but if the peer aggregator did not implement idempotent aggregation job initialization, these retries would fail since the DB transaction failure occurred after the peer aggregator received & processed the initial request.

See commit messages for more detail. Closes #1380.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1779,7 +1779,7 @@ impl VdafOps {
                     }
 
                     debug!(collect_request = ?req, "Cache miss, creating new collection job");
-                    let (_, report_count, batches) = try_join!(
+                    let (_, report_count, batches, batches_with_reports) = try_join!(
                         Q::validate_query_count::<SEED_SIZE, C, A>(
                             tx,
                             &vdaf,
@@ -1806,8 +1806,38 @@ impl VdafOps {
                                     Ok::<_, datastore::Error>((batch_identifier, batch))
                                 }
                             }),
-                        )
+                        ),
+                        try_join_all(
+                            Q::batch_identifiers_for_collection_identifier(
+                                &task,
+                                &collection_identifier
+                            )
+                            .map(|batch_identifier| {
+                                let task_id = *task.id();
+                                async move {
+                                    if let Some(batch_interval) =
+                                        Q::to_batch_interval(&batch_identifier)
+                                    {
+                                        if tx
+                                            .interval_has_unaggregated_reports(
+                                                &task_id,
+                                                batch_interval,
+                                            )
+                                            .await?
+                                        {
+                                            return Ok::<_, datastore::Error>(Some(
+                                                batch_identifier.clone(),
+                                            ));
+                                        }
+                                    }
+                                    Ok(None)
+                                }
+                            })
+                        ),
                     )?;
+
+                    let batches_with_reports: HashSet<_> =
+                        batches_with_reports.into_iter().flatten().collect();
 
                     // Batch size must be validated while handling CollectReq and hence before
                     // creating a collection job.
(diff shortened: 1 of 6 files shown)


branlwyd:
(pulling back to draft, I noticed the e2e tests are flaky)

branlwyd:
I saw one more `janus_daphne` flake. I've had this test running in a tight loop on my workstation all morning without observing a single failure. At this point, I don't think this flakiness was induced by the recent batch-state-tracking changes, but it likely merits investigation of its own.

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -197,22 +197,41 @@ impl<const SEED_SIZE: usize, Q: CollectableQueryType, A: vdaf::Aggregator<SEED_S
...
+                self.by_batch_identifier_index
+                    .keys()
+                    .map(|batch_identifier| {
+                        tx.get_batch::<SEED_SIZE, Q, A>(
+                            self.task.id(),
+                            batch_identifier,
+                            aggregation_parameter,
+                        )
+                    })
+            ),
```
tgeoghegan, line 212: These two `try_join_all!` calls will create one DB query for each batch identifier. Would it be more efficient to write a new SQL query that gets all batches matching a list of `batch_identifiers` in one query? If so, we should file an issue to change that separately from this change.

tgeoghegan (approved):
Seems sound, just a couple questions.

aggregator/src/aggregator.rs
```diff
@@ -1806,9 +1806,39 @@ impl VdafOps {
                                     Ok::<_, datastore::Error>((batch_identifier, batch))
                                 }
                             }),
-                        )
+                        ),
+                        try_join_all(
+                            Q::batch_identifiers_for_collection_identifier(
+                                &task,
+                                &collection_identifier
```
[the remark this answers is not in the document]
branlwyd: One of these calls is retrieving `batches`, the other is checking a property of the `client_reports` table.

`Q::batch_identifiers_for_collection_identifier` is an iterator, and doesn't perform network IO or any other blocking operation -- concretely, time-interval tasks just generate the sequence of minimal intervals included in the collection interval, while fixed-size tasks just give back the batch identifier directly. I don't think it's doing enough work to make reusing it valuable, especially since I think we'd either need to (a) collect the batch identifiers into a data structure [using a bit of memory] or (b) map the batch identifiers into a single future that retrieves both the `batches` & `client_reports` information [and have to write some awkward code to split them back out from the result of the `try_join_all`].

(also, this is pedantic, but when evaluating the effective number of network round trips, my model so far has been that concurrent requests -- e.g. in the same `try_join!`/`try_join_all` -- are a single logical round trip. this is because, due to pipelining, we can send all the requests to Postgres at once then receive the responses -- so these requests are already in one "effective" DB round trip. while this model isn't perfect, e.g. I'd bet there is a limit to the number of outstanding pipelined requests that may come into play, we haven't identified performance issues due to it yet.)

aggregator/src/aggregator/aggregation_job_writer.rs
branlwyd: This is https://github.com/divviup/janus/issues/224#issuecomment-1511897659 -- from prior discussions, it might be more performant, but we believe the performance difference to be marginal due to the use of request pipelining.

aggregator/src/aggregator.rs
```diff
@@ -1883,7 +1919,41 @@ impl VdafOps {
...
+                                        // MutationTargetAlreadyExists error.
+                                        //
+                                        // The failing writer, in this case, can't do anything about
+                                        // this problem while in its current transaction: further
+                                        // attempts to read the batch will continue to return `None`
+                                        // (since all reads in the same transaction are from the
+                                        // same snapshot), so it can't update the now-written batch.
+                                        // All it can do is give up on this transaction and try
+                                        // again, by calling `retry` and returning an error.
+                                        tx.retry();
```
inahga, line 1953: Naive question: is it not possible to immediately rollback the transaction and return the error?

I otherwise understand how this all comes together.

branlwyd: Effectively, that's what will happen: we still return `rslt` from this block, so if we followed this codepath to call `tx.retry()` we will also return an error, which will cause the transaction to be rolled back as quickly as possible (and then retried, since we called `tx.retry()`).

Rollback-and-retry is triggered in two ways:
* Automatically by the datastore transaction if any given datastore operation returns a serialization error (which is how Postgres suggests doing a retry).
* Manually if `tx.retry()` is called. This is necessary because, under Repeatable Read, not every error mode that indicates a serialization failure actually results in a serialization failure. For example, here we would "normally" end up with an error around a conflicting index write; but because we wrote the underlying SQL write query `ON CONFLICT DO NOTHING` and return `MutationTargetAlreadyExists` if the query did not affect any rows, we can get around that -- but we need to retry manually.

(We don't automatically retry in the datastore on seeing the `MutationTargetAlreadyExists` because from the datastore's perspective, it doesn't know if this indicates a concurrent-write conflict or a bug, and if it's the latter automatically retrying might lead to an infinite loop of retries. I'm thinking through if/how we might move things into the datastore, though, as peppering our application code with `retry()` calls does not feel like a fantastic solution.)

PR #1418, opened 2023-05-26
Datastore: implement barrier if we see "in failed tx" error.

### LINKED ISSUE #1417: Future cancellation impacts decision to retry transactions
In places where we pipeline many database query futures with `futures::future::try_join_all` or `tokio::try_join!`, the short-circuiting behavior of these combinators may keep us from retrying database transactions when we otherwise would. If one query gets a `T_R_SERIALIZATION_FAILURE`/`T_R_DEADLOCK_DETECTED` error, all queries submitted after it will get `IN_FAILED_SQL_TRANSACTION`. The connection task will submit these error results over per-query future channels in this order, but the query futures may wake up and receive their results in a different order. Thus, a `IN_FAILED_SQL_TRANSACTION` error may be passed out of our datastore method to `try_join_all/try_join!` first, which cancels all outstanding futures, and in turn may cause a `T_R_SERIALIZATION_FAILURE` to be dropped while in a `tokio-postgres` channel, before it gets to our `check_error()` function.

This means that expected serialization errors may inadvertently cause transactions to fail while handling a background processing job or an HTTP request. Normally, job retry logic or whichever HTTP client's retry logic will compensate for this with a larger-scope retry, at the cost of some repeated work. However, the Janus-Daphne integration test is not as forgiving, because Daphne does not yet support idempotent handling of aggregation job requests. There's a critical window from when Daphne receives an aggregation job request from Janus, until Janus successfully store the results in its database, where any error will result in the aggregators becoming out of sync, stranding that job's reports, and causing the integration test's collection to fail. Recent changes in this window increased the degree to which we pipeline requests, and we are now seeing CI failures due to this test. I examined logs for two of them, and saw that an aggregation job PUT was getting retried, failing the first time due to "current transaction is aborted", and then succeeding the second time, but with "ReportReplayed" errors for each report share. The database log showed that the reason the transaction was aborted was a serialization failure. I put together a set of changes to not cancel futures within this window, and confirmed that this fixes the test's flakiness, see d3ac65ed3a722d09f3024acb82662439eeb20637.

I think that addressing this issue would either require intrusive modifications to `tokio-postgres`, to push our transaction retry decision inside the connection task or expose copies of errors through a separate future/stream, or significant changes to our coding discipline, to treat all futures interacting with the database as not cancellation safe, due to their retry flag side effects. I looked at `sqlx`, the other major Postgres client library in Rust, and it doesn't support query pipelining, thus making it both immune to this issue, and not a good fit for our codebase, since we rely on pipelining currently to keep the number of network round trips down. We could work around the impact of this issue in our test suite by adding a retry loop to the Daphne integration test. We should keep an eye on the error rate of aggregation jobs, collection jobs, and HTTP server requests during load tests for possible impact from this issue.

branlwyd:
For some error modes, Postgres will return an error to the caller & then fail all future statements within the same transaction with an "in failed SQL transaction" error. This effectively means one statement will receive a "root cause" error and then all later statements will receive an "in failed SQL transaction" error. In a pipelined scenario, if our code is processing the results of these statements concurrently--e.g. because they are part of a `try_join!`/`try_join_all` group--we might receive & handle one of the "in failed SQL transaction" errors before we handle the "root cause" error, which might cause the "root cause" error's future to be cancelled before we evaluate it. If the "root cause" error would trigger a retry, this would mean we would skip a DB-based retry when one was warranted.

To fix this problem, we (internally) wrap all direct DB operations in `run_op`. This function groups concurrent database operations into "operation groups", which allow us to wait for all operations in the group to complete (this waiting operation is called "draining"). If we ever observe an "in failed SQL transaction" error, we drain the operation group before returning.  Under the assumption that the "root cause" error is concurrent with the "in failed SQL transactions" errors, this guarantees we will evaluate the "root cause" error for retry before any errors make their way out of the transaction code.

Closes #1417.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -48,10 +48,13 @@ use std::{
     mem::size_of,
     ops::RangeInclusive,
     pin::Pin,
-    sync::atomic::{AtomicBool, Ordering},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc, Mutex,
+    },
     time::{Duration as StdDuration, Instant},
 };
-use tokio::try_join;
+use tokio::{sync::Barrier, try_join};
 use tokio_postgres::{error::SqlState, row::RowIndex, IsolationLevel, Row, Statement, ToStatement};
 use tracing::error;
 use url::Url;
@@ -246,9 +249,10 @@ impl<C: Clock> Datastore<C> {
         };
         let tx = Transaction {
             raw_tx,
-            retry: AtomicBool::new(false),
             crypter: &self.crypter,
             clock: &self.clock,
+            retry: AtomicBool::new(false),
+            op_group: Mutex::new(Arc::new(Mutex::new(OperationGroup::Running(0)))),
         };
 
         // Run user-provided function with the transaction, then commit/rollback based on result.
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
Very nice!

tgeoghegan (approved):

branlwyd:
(I'm running my `test_until_failure.sh` script against the `janus_daphne` test -- it took a few hours to observe a failure on my workstation last time, I'm going to leave it running for awhile before merging.)

inahga (approved):
Appreciate the detailed explanations!

Issue #1311, opened 2023-04-28
Error when reverting schema migration


divergentdave:
Our 'initial schema' down migration cannot be run inside a transaction.

```
error: while executing migrations: error returned from database: DROP INDEX CONCURRENTLY cannot run inside a transaction block
```

branlwyd:
I noticed our initial schema migrations includes a bunch of hedging clauses like `...IF EXISTS...`. I think this is probably orthogonal to this issue, but if we're touching the initial down-migration script anyway, I think it should be written without those clauses.

In general, migration scripts should assume they are working with exactly the schema implied by their version; given that, we shouldn't hedge on a DB entity not existing -- if it doesn't, we aren't working with the schema we think we're working with, and we _want_ to error.

tgeoghegan:
Relates to #1279: this would have been caught by an automated test that runs down scripts.

Issue #1279, opened 2023-04-20
Test SQL migration down scripts


tgeoghegan:
Test setups will apply SQL migration up scripts against ephemeral databases, and in #1277, we run tests against multiple schema versions. However we don't currently have coverage of down scripts. We should add tests that run downgrade scripts and as much as possible, run database tests against downgraded databases.

tgeoghegan:
There's a few different things we can check here:

- that down scripts can run at all
- that down scripts return the schema to the state it was in before running the corresponding up script (perhaps by dumping the schema before and after the up/down migration pair and diffing?)
- that Janus can run against a database that has been downgraded (this is tricky because the Janus version under test doesn't necessary work against all the historical database versions, but we can at least do this for schema versions in `datastore::SUPPORTED_SCHEMA_VERSIONS`

inahga:
I was able to write a test that verifies whether the scripts can run :+1:.

As for schema validation, there's a few options:
1. Dump tables/indices/types/etc. by hand and compare.
2. Find a dedicated tool for diff'ing pg schemas.

The latter sounds like the better idea--but I'm still researching what the best options are. Open to input if anyone knows a good tool off-hand.

> that Janus can run against a database that has been downgraded (this is tricky because the Janus version under test doesn't necessary work against all the historical database versions, but we can at least do this for schema versions in datastore::SUPPORTED_SCHEMA_VERSIONS

I lean towards making this an integration test for every `datastore::SUPPORTED_SCHEMA_VERSIONS`, and extend tests to allow running them on a particular schema version. I don't think there's a clean way to re-invoke the whole test suite from within the test-suite. It might also call for extending tests to indicate their own minimum schema version, to accommodate for tests that exercise newer versions of the schema. Might get very tricky...

divergentdave:
Currently, we can re-run datastore tests against multiple schema versions (only using `.up.sql` scripts) using an rstest_reuse template. We have a macro that takes in a variadic list of schema versions, and stamps out a static variable with an array containing those versions, plus an rstest_reuse template with multiple attributes, one for each schema version we want to test against.

I was thinking that we could modify this macro to add another set of case attributes on the template test, one for each schema version except the last one, that runs up scripts until version i+1, then runs the corresponding down script to revert to version i. We'd need to write a new function to pass to the case attribute, change the macro arguments to something like `($( $i:literal ),* , $i_last:literal)`, and change the attribute repetitions like so.

```rust
[...]

PR #1434, opened 2023-05-30
Add test for down migrations.

inahga:
Supports https://github.com/divviup/janus/issues/1279.

Opening in draft for early feedback on strategy/style, and because this
test indeed found problems with our down migrations!


inahga:
With the latest changes, this PR will also close https://github.com/divviup/janus/issues/1311.

aggregator_core/src/datastore/test_util.rs
```diff
@@ -187,6 +224,19 @@ pub async fn ephemeral_datastore() -> EphemeralDatastore {
     ephemeral_datastore_max_schema_version(i64::MAX).await
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum EphemeralDatastoreError {
```
branlwyd, line 228: In Janus, test-only code (i.e. literally test code, or code conditionally compiled under the `test-util` feature) is written infallibly, and panics on error (typically via a call to `unwrap` or `expect`).

I think we should get rid of this error type and panic-on-error instead.

aggregator_core/src/datastore.rs
```diff
@@ -5920,6 +5920,12 @@ mod tests {
         assert_matches!(error, Error::DbState(_));
     }
 
+    #[rstest_reuse::apply(schema_versions_template)]
```
branlwyd, line 5923: This will run an instance of this test for every supported schema version. As a nitpick, this test could instead be a one-off (i.e. not parameterized by the `schema_versions_template`) which starts at the maximum known/available schema version then downgrades to 0 -- this would exercise all schema downgrades just once.

But even better than that might be a different approach: instead, modify the `schema_versions_template` to produce an `EphemeralDatastore` by applying all migrations, then downgrading to the desired version. This would effectively test that downgrade scripts actually produce working schemas. But it wouldn't detect if upgrades were broken -- we could test both the "upgrade to version" and "upgrade past version, then downgrade" approaches work, at the cost of running each parameterized test twice per version.

WDYT? (we could also take both testing approaches, to get "syntax" checking for unsupported schema versions, and "functional" testing for supported schema versions.)

branlwyd (changes requested):

aggregator_core/src/datastore.rs
inahga: > This will run an instance of this test for every supported schema version. As a nitpick, this test could instead be a one-off (i.e. not parameterized by the schema_versions_template) which starts at the maximum known/available schema version then downgrades to 0 -- this would exercise all schema downgrades just once.

:+1:

> But even better than that might be a different approach: instead, modify the schema_versions_template to produce an EphemeralDatastore by applying all migrations, then downgrading to the desired version. This would effectively test that downgrade scripts actually produce working schemas. But it wouldn't detect if upgrades were broken -- we could test both the "upgrade to version" and "upgrade past version, then downgrade" approaches work, at the cost of running each parameterized test twice per version.

Yep! That's incoming in the next PR. (I can't take credit for the idea, @divergentdave had the same idea as you :D https://github.com/divviup/janus/issues/1279#issuecomment-1569374474).

> we could also take both testing approaches, to get "syntax" checking for unsupported schema versions, and "functional" testing for supported schema versions.

That was my intent--one test for basic syntax+functionality checking, then another set of cases that actually runs the downgraded schema against all tests.

aggregator_core/src/datastore/test_util.rs
inahga: Indeed. I originally had wanted to ensure the version that failed was captured in the error, but `expect()` with a good message should be just as good.

branlwyd (approved):

aggregator_core/src/datastore/test_util.rs
```diff
@@ -130,6 +134,35 @@ impl EphemeralDatastore {
...
+                "target version ({}) must be less than the current database version ({})",
+                target, current_version,
+            );
+        }
+
+        // Run down migrations one at a time to provide better context when
+        // one fails.
+        for v in (target..current_version).rev() {
+            if let Err(e) = self.migrator.undo(&mut connection, v).await {
+                panic!("failed to downgrade to version {}: {}", v, e);
```
inahga, line 162: Unsure if there's any preference for `.expect(format!("").as_str())` or just a direct `panic!()` here.

divergentdave: Using the panic macro is fine by me

aggregator_core/src/datastore/test_util.rs
```diff
@@ -130,6 +134,35 @@ impl EphemeralDatastore {
             LessSafeKey::new(UnboundKey::new(&AES_128_GCM, &self.datastore_key_bytes).unwrap());
         Crypter::new(Vec::from([datastore_key]))
     }
+
+    pub async fn downgrade(&self, target: i64) {
```
divergentdave, line 138: We should stick a `#[cfg(feature = "test-util")]` on this, since we don't intend to use it at runtime.

divergentdave (approved):

aggregator_core/src/datastore/test_util.rs
branlwyd: This module declaration in `datastore.rs` is `test-util`, so this isn't necessary.

divergentdave: Ah, sorry, nevermind

aggregator_core/src/datastore/test_util.rs
inahga: Ah, of the options `unwrap_or_else` looks cleanest to me, avoids the `if let` construct (pardon the Rust noobness).

PR #1439, opened 2023-06-01
Test down migrations against all existing tests

inahga:
This uses the technique suggested by @divergentdave in https://github.com/divviup/janus/issues/1279#issuecomment-1569374474 to test down migrations when there are multiple supported versions of the schema.

We can test the validity of this approach by creating a set of benign migrations and adding a new version.

```rust
// in datastore.rs
supported_schema_versions!(10, 9);
```
```sql
-- db/00000000000010_inahga_test.up.sql
CREATE TABLE inahga_test(id BIGINT GENERATED);
```
```sql
-- db/00000000000010_inahga_test.down.sql
DROP TABLE inahga_test;
```

This should result in a passing set of tests, each test should have 3 cases (1 latest, 1 old via normal upgrade, 1 old via downgrade).

We can simulate a regression caused by a downgraded migration by editing the down script to break a feature of Janus.
```sql
-- db/00000000000010_inahga_test.down.sql
DROP TABLE inahga_test;
ALTER TABLE task_collector_auth_tokens DROP COLUMN type;
```

We then see several tests fail on test case 3, because the downgrade script has broken auth token support.

For maximum paranoia, we can add a quick test that exercises the example table, to ensure that the migrations are actually being applied when they should.
```rust
// in datastore.rs
    #[rstest_reuse::apply(schema_versions_template)]
    #[tokio::test]
    async fn inahga_test(ephemeral_datastore: EphemeralDatastore) {
        let datastore = ephemeral_datastore.datastore(MockClock::default()).await;

        datastore
            .run_tx(|tx| {
                Box::pin(async move {
                    tx.query("SELECT * FROM inahga_test", &[]).await.unwrap();
                    Ok(())
                })
            })
            .await
            .unwrap();
    }
```

We then see this test fails on cases 2 and 3, because it exercises functionality that's only present in the newest schema, as expected.

(I'm sure given enough time this manual test of the tests can be automated, but I think that crosses the line of diminishing returns).

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -5863,7 +5863,7 @@ mod tests {
             },
             schema_versions_template,
             test_util::{
-                ephemeral_datastore_max_schema_version, generate_aead_key, EphemeralDatastore,
+                ephemeral_datastore_schema_version, generate_aead_key, EphemeralDatastore,
             },
             Crypter, Datastore, Error, Transaction,
         },
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -5877,7 +5877,7 @@ mod tests {
             test_util::{
                 ephemeral_datastore_schema_version, generate_aead_key, EphemeralDatastore,
             },
-            Crypter, Datastore, Error, Transaction,
+            Crypter, Datastore, Error, Transaction, SUPPORTED_SCHEMA_VERSIONS,
         },
         query_type::CollectableQueryType,
         task::{self, test_util::TaskBuilder, Task},
@@ -5916,6 +5916,13 @@ mod tests {
         time::Duration as StdDuration,
     };
     use tokio::time::timeout;
+
+    #[test]
+    fn check_supported_versions() {
+        if SUPPORTED_SCHEMA_VERSIONS[0] != *SUPPORTED_SCHEMA_VERSIONS.iter().max().unwrap() {
+            panic!("the latest supported schema version must be first in the list");
+        }
+    }
 
     #[rstest_reuse::apply(schema_versions_template)]
     #[tokio::test]
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -5861,6 +5861,10 @@ pub mod models {
 
 #[cfg(test)]
 mod tests {
+    // This function is only used when there are multiple supported versions.
+    #[allow(unused_imports)]
+    use crate::datastore::test_util::ephemeral_datastore_schema_version_by_downgrade;
+
     use crate::{
         datastore::{
             models::{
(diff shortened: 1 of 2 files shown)


aggregator_core/src/datastore.rs
```diff
@@ -71,13 +71,19 @@ pub mod test_util;
...
-        const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[$($i),*];
+    ( $i_latest:literal $(,)? $( $i:literal ),* ) => {
+        const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[$i_latest, $($i),*];
 
         #[cfg(test)]
         #[rstest_reuse::template]
         #[rstest::rstest]
-        $(#[case(ephemeral_datastore_max_schema_version($i))])*
+        // Test the latest supported schema version.
+        #[case(ephemeral_datastore_schema_version($i_latest))]
```
inahga, line 81: I would have liked the cases to have neater titles than `case_1`, `case_2`, etc.

In rstest, that's done using this technique:
```
#[case::foobar(args)]
```
Where `foobar` is a valid Rust `ident`(ifier).

Sadly there doesn't seem to be a _stable_ way to construct idents out of macro arguments, something like the `##` operator in the C preprocessor. In Rust-land it's the experimental [`concat_idents!()`](https://doc.rust-lang.org/std/macro.concat_idents.html) macro.

aggregator_core/src/datastore.rs
```diff
@@ -90,6 +96,8 @@ macro_rules! supported_schema_versions {
 
 // List of schema versions that this version of Janus can safely run on. If any other schema
 // version is seen, [`Datastore::new`] fails.
+//
+// Note that the latest supported version must be first in the list.
 supported_schema_versions!(9);
```
inahga, line 101: Experimenting to see if this can be asserted statically. The logic would be `assert!(max(SUPPORTED_SCHEMA_VERSIONS) == SUPPORTED_SCHEMA_VERSIONS[0])`, but I don't see a const way of getting the max of a slice.

If it's too complicated I'll just add a standard test, just would be nice to have errors like that caught at compile-time.

aggregator_core/src/datastore.rs
```diff
@@ -71,13 +71,19 @@ pub mod test_util;
 ///
 /// [1]: https://docs.rs/rstest_reuse/latest/rstest_reuse/
 macro_rules! supported_schema_versions {
-    ($( $i:literal ),*) => {
-        const SUPPORTED_SCHEMA_VERSIONS: &[i64] = &[$($i),*];
+    ( $i_latest:literal $(,)? $( $i:literal ),* ) => {
```
inahga, line 74: $i_latest has to be first because `$( $i:literal ),* , $i_latest:literal` complains at the call site that the last argument is ambiguous. Fair enough.

aggregator_core/src/datastore.rs
inahga: Ah wait, https://doc.rust-lang.org/std/cmp/fn.max.html `std::cmp::max` takes a `T: Ord`, which arrays do implement https://doc.rust-lang.org/std/primitive.array.html.

inahga: Eh, but it's unstable, and unstable in a somewhat scary way https://github.com/rust-lang/rust/issues/92391#issuecomment-1492469098

divergentdave: Sorry, I hadn't thought my suggestion all the way through. I think it'll be another few years before enough trait implementations are available in const contexts to do a static assertion -- a plain test should be fine here.

branlwyd (approved):

PR #1440, opened 2023-06-02
Enable differential privacy for the fixedvec type.

MxmUrw:
This comprises the following changes:
 - Add a `noise_param` parameter to the fixedvec type(s) in the various vdaf enums.
   This describes the amount of noise to be added.
 - Wire up the new `postprocess()` function of prio to be called in both leader
   and helper. The noising itself happens in the implementation of the fixedvec
   type in prio, see the relevant pr (https://github.com/divviup/libprio-rs/pull/578).

**Notes**:
 1. We currently call the `postprocess()` function of prio in two seperate locations for the leader and for the helper, as it seems that their respective aggregation mechanism utilizes different codepaths. I can imagine that there could be a better way to hook up our noising function. Is there?
 2. We have a new `noise_param` argument for our vdaf. Should we add such an argument to `tools/collect`? Currently, a default value (no noise) is used when instantiating our vdaf.
 3. We added an `experimental` feature for `janus/aggregator` and `janus/aggregator_core` which depends on the `experimental` feature for prio.

**Tasks**:
 - [ ] Switch prio dependency back to crates.io, after our other pr is merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2652,6 +2652,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "num-bigint"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f93ab6289c7b344a8a9f60f88d80aa20032336fe78da341afc91c8a2341fc75f"
+dependencies = [
+ "autocfg",
+ "num-integer",
+ "num-traits",
+]
+
+[[package]]
+name = "num-integer"
+version = "0.1.45"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "225d3389fb3509a24c93f5c29eb6bde2586b98d9f016636dff58d7c6f7569cd9"
+dependencies = [
+ "autocfg",
+ "num-traits",
+]
+
+[[package]]
 name = "num-traits"
 version = "0.2.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


tgeoghegan (commented):
I think we can't really review this until the corresponding libprio-rs change is done, but we can have some useful discussion in the meantime.

> We currently call the postprocess() function of prio in two seperate locations for the leader and for the helper, as it seems that their respective aggregation mechanism utilizes different codepaths. I can imagine that there could be a better way to hook up our noising function. Is there?

Both helper and leader paths ultimately call `aggregator::aggregate_share::compute_aggregate_share` to get an `<A as vdaf::Aggregator>::AggregateShare`. I think that would be the place to apply noise. Both because you'd call it once for either role, but also because in that scope, you have awareness of how many contributions went into the aggregate share, and my intuition is that some DP schemes will want that information to tune the noise they apply.

> We have a new noise_param argument for our vdaf. Should we add such an argument to tools/collect? Currently, a default value (no noise) is used when instantiating our vdaf.

IIUC the noise parameter is meaningless for the collector, which is why over in https://github.com/divviup/libprio-rs/pull/578, I argued for changing the interface so that only the `vdaf::Aggregator` trait deals with DP parameters. But I think we'll have to revisit the question of what the DAP collector needs to do (or not) once we have agreed on the libprio-rs level change.

> We added an experimental feature for janus/aggregator and janus/aggregator_core which depends on the experimental feature for prio.

While it's a bit icky semantically, I'd prefer to put this behind the existing `fpvec_bounded_l2` to minimize feature combinations.

> We added a $FEATURES argument to the dockerfile which allows additional janus features to be enabled, for example from our external docker-compose.yml.

Yes, this is a good idea!

branlwyd (commented):
> We have a new noise_param argument for our vdaf. Should we add such an argument to tools/collect? Currently, a default value (no noise) is used when instantiating our vdaf.

IMO, feel free to extend tooling to be aware of the new VDAF parameter, gated behind an appropriate feature flag.

MxmUrw:
Closing in favor of the up-to-date PR: #1892.

PR #1442, opened 2023-06-02
Add workflow to check the consistency of migrations.

inahga:
Supports https://github.com/divviup/janus/issues/1252.

Adds various checks of migrations that aren't already covered by [sqlx itself](https://docs.rs/sqlx-core/0.6.3/src/sqlx_core/migrate/source.rs.html#21).

### DIFF
--- a/.github/workflows/ci-migrations.yml
+++ b/.github/workflows/ci-migrations.yml
@@ -19,6 +19,7 @@ jobs:
         fetch-depth: 0
 
     - name: Check that existing migrations have not changed
+      if: github.event_name == "pull_request"
       run: |
         DIFF="$(git diff "$BASE_SHA" --name-status db/)"
         printf "%s\n" "$DIFF"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-migrations.yml
```diff
@@ -0,0 +1,58 @@
...
+  pull_request:
+    paths:
+      - "db/**"
+  workflow_dispatch:
+
+jobs:
+  janus_migrations:
+    runs-on: ubuntu-latest
+    env:
+      BASE_SHA: "${{ github.event.pull_request.base.sha }}"
```
divergentdave, line 15: I presume this will be undefined for push and workflow_dispatch events. We should either remove those triggers or skip the `git diff` based check if the pull request base is not available.

inahga: Good point, yes the `git diff` step should be gated to just pull request.

tgeoghegan (approved):

PR #1445, opened 2023-06-05
Merge DB migrations.

branlwyd:
Determining the final schema is becoming more challenging as more migrations are added. I merged this while working through the design of more-incremental GC, I figure it's worthy of consideration whether we want to commit this.

(If this is going to cause a huge operational headache, we can drop it; it's just very useful for me to be able to see the entire schema at once. Maybe there's another way to achieve the same effect?)


inahga:
If we are to merge it, we should definitely merge it before https://github.com/divviup/janus/pull/1442 :smile:.

I in spirit have no problem with this, just some notes:

> Maybe there's another way to achieve the same effect?)

`pg_dump -s` will dump the database schema. It may be useful to have a script or subcommand that is has the UX of something like `janus_cli dump-schema --version 9` which brings up a test pg container, upgrades it to the target version, then gives you the `pg_dump` output.

I might suggest doing this instead of merging migrations, because this is not going to be the last set of migrations we will have to reason about in our heads (similarly for divviup-api).

inahga:
I went ahead and merged https://github.com/divviup/janus/pull/1442, but we can simply ignore it if moving forward with this, since it's not a required check.

tgeoghegan:
IMO we should do this now that we have branched for DAP-05 (which is to say, rebase on `main`). I don't think there will be a graceful transition path from DAP-04 deployments into the ping-pong world, so we have an opportunity to reset the database migrations there.

branlwyd:
Updated against latest main -- we can merge this, though I'm not sure offhand what (if any) hoops we'll need to jump through to make our migration infra happy.

tgeoghegan:
I think that just like with the PR into `release/0.subscriber-01`, we'll just have to ignore the `ci-migrations` failure and merge anyway.

branlwyd:
Yep, if there are no existing 0.6 deployments, I don't think we need to worry.

divergentdave (approved):

PR #1459, opened 2023-06-06
Check validity of migrations before pushing to GCS 

inahga:
Supports #1252

### DIFF
--- a/.github/workflows/publish-sql-schema.yml
+++ b/.github/workflows/publish-sql-schema.yml
@@ -13,19 +13,33 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
     - id: "gcp-auth"
       name: "Authenticate to GCP"
       uses: "google-github-actions/auth@v1"
       with:
         workload_identity_provider: ${{ vars.GCP_ARTIFACT_PUBLISHER_WORKLOAD_IDENTITY_PROVIDER }}
         service_account: ${{ vars.GCP_ARTIFACT_PUBLISHER_DEPLOY_SERVICE_ACCOUNT }}
-    - name: Get the version
+    - name: Determine latest and previous version
       id: get_version
       run: |
         VERSION=${GITHUB_REF/refs\/tags\//}
-        echo "MINOR_VERSION=$(echo ${VERSION} | awk -F'.' '{if (NF != 3) {exit 1}; printf "%s.%s", $1, $2}')" >> $GITHUB_OUTPUT
+        echo "minor_version=$(echo ${VERSION} | awk -F'.' '{if (NF != 3) {exit 1}; printf "%s.%s", $1, $2}')" >> $GITHUB_OUTPUT
+
+        PREVIOUS=$(
+          git tag |
+            grep -P '^[0-9]+\.[0-9]+\.[0-9]+$' |
+            sort -t. -k1,1rn -k2,2rn -k3,3rn |
+            grep -PA1 "^$VERSION\$" | grep -Pv "^$VERSION\$"
+        )
+        echo "previous_version=$PREVIOUS" >> $GITHUB_OUTPUT
+    - name: Check validity of migrations
+      uses: "./.github/actions/validate-migrations"
+      with:
+        from: "${{ steps.get_version.outputs.previous_version }}"
     - name: "Upload schema file(s)"
       run: |-
         gcloud alpha storage cp --recursive \
           db \
-          gs://janus-artifacts-sql-schemas/${{ steps.get_version.outputs.MINOR_VERSION }}/
+          gs://janus-artifacts-sql-schemas/${{ steps.get_version.outputs.minor_version }}/
--- a/.github/actions/validate-migrations/action.yml
+++ b/.github/actions/validate-migrations/action.yml
@@ -8,6 +8,7 @@ runs:
   using: "composite"
   steps:
     - name: Check that existing migrations have not changed
+      shell: bash
       run: |
         DIFF="$(git diff "${{ inputs.from }}" --name-status db/)"
         printf "%s\n" "$DIFF"
@@ -18,6 +19,7 @@ runs:
         fi
 
     - name: Ensure all migrations have the proper filename
+      shell: bash
       run: |
         REGEX='^(README.md)|([0-9]{14}_[a-z_]+\.(down|up)\.sql$)'
         INVALID_FILES=$(find db/ -mindepth 1 -printf "%f\n" | grep -Pv "$REGEX" || true)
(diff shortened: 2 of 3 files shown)


.github/workflows/publish-sql-schema.yml
```diff
@@ -24,8 +30,40 @@ jobs:
...
+        # The meaning of added and removed files are reversed when comparing this
+        # way. If a file comes up deleted, that means that the local git ref has
+        # introduced a new file that is not present in the remote. This is permissible.
+        # If a file comes up added, that means that the local ref has removed an
+        # existing migration which was readded by the overwrite step. This is not
+        # permitted.
+        if [ -n "$DIFF" ] && grep -qPv '^D\b' <<<"$DIFF"; then
+          echo "fatal: migrations cannot be modified or removed, only added."
+          exit 1
+        fi
```
inahga, line 56: I don't really like this approach, it's quite hacky. I'm open to other ideas.

The goal is as so:
- Differences between any existing files on either side should be rejected.
- Files missing on the remote but present in the local should be permitted. (the local is adding a new migration)
- Files present on the remote but absent from the local should be rejected. (the local has deleted a migration)

`diff -r` is not sufficient, because it doesn't differentiate between local/remote differences in a machine-parseable way. I looked at `rclone` to see if it could instead entirely replace `gcloud alpha storage`, but I don't see the right set of flags to make it work as described. We also could just leave out this step and let `gcloud alpha storage cp --no-clobber` do the work, but this has poor UX as it will just bail out with an ambiguous HTTP error (and does not cover for the 3rd case).

.github/workflows/publish-sql-schema.yml
```diff
@@ -24,8 +30,40 @@ jobs:
...
         VERSION=${GITHUB_REF/refs\/tags\//}
         echo "MINOR_VERSION=$(echo ${VERSION} | awk -F'.' '{if (NF != 3) {exit 1}; printf "%s.%s", $1, $2}')" >> $GITHUB_OUTPUT
+    - name: Check validity of migrations
+      uses: "./.github/actions/validate-migrations"
+    - name: Check that existing migrations have not changed
+      if: '!inputs.force'
+      run: |
+        # Overwite the migrations on the local ref so that we can use `git diff` to compare.
+        rm -f db/*.sql
+        trap "git reset --hard" EXIT
```
tgeoghegan, line 40: Perhaps we should run the entire script under `set -eo pipefail` so that we don't have to write traps?

.github/workflows/publish-sql-schema.yml
tgeoghegan: I'm inclined to agree that parsing the output of `git diff --name-status` is dicey:

- I find regexes hard to read (though this one is simple enough that even I was able the understand it, so a short comment explaining what it does would suffice)
- This would break if the format of `diff --name-status` ever changes (I'm not sure if `git` makes guarantees about the stability of its output format)

In any case, while poking around the `git-diff` man page, I think I found something that I think takes the risk out of this and is more self documenting: [`--diff-filter`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---diff-filterACDMRTUXB82308203). If we did `git diff --diff-filter=DM`, then only files that are `D`eleted or `M`odified are included in the diff, and all you have to do is check if the output is empty. I suppose we could do `--diff-filter=CDMRTUXB` to catch everything _except_ added files.

All that said, you might mean that doing `rm db/*` and replacing it with the contents of the GCS bucket is hacky. Perhaps slightly better is [`git diff --no-index`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt-emgitdiffemltoptionsgt--no-index--ltpathgtltpathgt). So you could:

```
git diff --no-index \
    --diff-filter=CDMRTUXB \ # Allow [A]dditions but nothing else
    $BUCKET/db \ # contents of the GCS bucket
    $CHECKOUT/db # contents of the git checkout
```

Then if anything has changed between `$BUCKET/db` and `$CHECKOUT/db` that isn't an addition, the output will be non-empty and you can print the diff and fail the action.

tgeoghegan (changes requested):

.github/workflows/publish-sql-schema.yml
inahga: `run` implicitly runs with `set -eo pipefail`--this trap is more making sure that no matter what happens, i.e. script exiting on another command, the repo is restored. It's additional protection against copying bad state to GCS.

.github/workflows/publish-sql-schema.yml
inahga: That's excellent! The result looks great!
```yaml
    - name: Check that existing migrations have not changed
      if: '!inputs.force'
      run: |
        TEMP=$(mktemp -d)
        gcloud alpha storage cp --recursive \
          "gs://janus-artifacts-sql-schemas/${{ steps.get_version.outputs.minor_version }}/db"\
          "$TEMP"
        if ! git diff --no-index --name-status --diff-filter=CDMRTUXB "$TEMP/db" "./db"; then
          echo "fatal: migrations cannot be modified or removed, only added"
          exit 1
        fi
```
Plus a bonus fix for missing string interpolation (was using `$MINOR_VERSION` to test the script in isolation).

I'll carry it back to https://github.com/divviup/janus/blob/main/.github/workflows/ci-migrations.yml#L19 as well.

.github/workflows/publish-sql-schema.yml
inahga: Moot point now, as the new approach doesn't contaminate the working tree.

.github/workflows/publish-sql-schema.yml
```diff
@@ -23,9 +29,30 @@ jobs:
...
+    - name: Check validity of migrations
+      uses: "./.github/actions/validate-migrations"
+    - name: Check that existing migrations have not changed
+      if: '!inputs.force'
+      run: |
+        TEMP=$(mktemp -d)
+        gcloud alpha storage cp --recursive \
+          "gs://janus-artifacts-sql-schemas/${{ steps.get_version.outputs.minor_version }}/db" \
+          "$TEMP"
+        if ! git diff --no-index --name-status --diff-filter=CDMRTUXB "$TEMP/db" "./db"; then
```
tgeoghegan, line 42: Last thing to consider is whether or not we want `--name-status`. IIUC, the output of `git diff` will go to stdout and show up in the action log, and I think it'd be more helpful to show the entire diff instead of just which files changed. Either way is fine with me, though.

.github/workflows/publish-sql-schema.yml
```diff
@@ -23,9 +29,30 @@ jobs:
...
+          echo "fatal: migrations cannot be modified or removed, only added"
+          exit 1
+        fi
     - name: "Upload schema file(s)"
+      if: '!inputs.force'
```
tgeoghegan, line 47: Just curious: are the single quotes necessary here?

tgeoghegan (approved):

.github/workflows/publish-sql-schema.yml
inahga: My thought was as so: if I get dinged by this check and it tells me that "files cannot be modified", my action is simply to `git restore/revert` whatever was changed. Therefore the most salient information is _what_ files were modified rather than the actual differences. It's a minor optimization though, I can see the full diff still being useful.

PR #1485, opened 2023-06-12
janus_cli observability fixes

divergentdave:
This trues up `main()` in janus_cli with `janus_main()`, fixing a few observability-related issues.

- Hang on to the metrics exporter handle. Dropping this early would stop the exporter if pushing metrics over OTLP.
- Add the `janus_build_info` gauge metric.
- Make the "Starting up" log line more consistent with that in other binaries.
- Flush the trace exporter before exiting by shutting down the current tracer provider. This avoids losing traces when using OTLP for them.

### DIFF
--- a/aggregator/src/bin/janus_cli.rs
+++ b/aggregator/src/bin/janus_cli.rs
@@ -4,7 +4,7 @@ use clap::Parser;
 use janus_aggregator::{
     binary_utils::{database_pool, datastore, read_config, CommonBinaryOptions},
     config::{BinaryConfig, CommonConfig},
-    metrics::install_metrics_exporter,
+    metrics::{install_metrics_exporter, MetricsExporterHandle},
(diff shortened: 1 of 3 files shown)


tgeoghegan (approved):
I think all the changes here are good and useful as-is, but I also don't think there's much  point to tidying up metrics on `janus_cli` unless we teach it to export metrics to `prometheus-pushgateway`, because it's unlikely that any `janus_cli` process will live long enough to service a metrics scrape. I'm comfortable with doing that in a separate PR or filing and issue for it, though.

divergentdave:
I'll file an issue, it looks like it could take a fair bit of boilerplate. The metrics tweaks would be usable as-is with the OTLP exporter, which always works on a push basis (if we wind up using a metrics product that uses the OpenTelemetry Collector).

PR #1489, opened 2023-06-13
Enable divviup-ts integration tests

divergentdave:
This re-enables the integration tests against divviup-ts.


divergentdave:
Each test pins the version of the other DAP implementation, so I think we should still have both tests to maintain coverage of changes in either.

tgeoghegan (approved):
This change accomplishes what it intends to, restores functionality we already had, and does not in and of itself significantly affect CI time, so I think we should take it.

branlwyd (approved):

Issue #1473, opened 2023-06-09
Tracing span improvements


divergentdave:
Our tracing spans do not yet provide a full picture of what's going on inside a process, so we should make the following improvements.

- [x] Remove indefinitely long root spans, see #1463.
- [x] Introduce new root spans, i.e. a span for each iteration of a job acquisition loop, and a span for each job step. Instrument `ReportWriteBatcher::write_batch()`, so that it is a root span.
- [x] Instrument more `Datastore` public methods.
- [x] Instrument `AggregationJobWriter::write()`.
- [x] Adjust the filter in `aggregator/src/trace.rs` to exclude everything in `h2`.
- [x] Instrument `Datastore::run_tx_with_name()` and `Datastore::run_tx_once()`, maybe at the "trace" level.
- [x] Check whether we need to call `tracing::Instrument::in_current_span()` on futures when passing them to each `try_join!` or `try_join_all` primitive from the `futures` or `tokio` crates.
- [x] Evaluate each place where we spawn a new async task, and determine whether the spawned future should be instrumented with the current span, a new span, or not at all.
- [x] janus_cli is missing a call to `opentelemetry::global::shutdown_tracing_provider()`. (the `janus_main()` helper does properly shut down push exporters, but janus_cli does not use this)

PR #1491, opened 2023-06-13
Instrument more methods with tracing spans

divergentdave:
This adds tracing instrumentation to methods throughout the aggregator and datastore modules. Part of #1473.


tgeoghegan (approved):
I wonder if changing the contents of tracing spans should be thought of as a breaking API change, since there is something of a contract between what spans exist and how they're annotated and the monitoring or alerting stacks of anyone running Janus.

Anyway, this change is good and should get merged!

PR #1492, opened 2023-06-13
Don't trace any spans from h2

divergentdave:
This broadens the filter used to exclude h2 spans from traces. Part of #1473.


tgeoghegan (approved):
We have the ability to turn these back on at runtime with `RUST_LOG`, right?

divergentdave:
`RUST_LOG` only controls the `stdout_filter` variable used further up in the function, which is attached as a filter to whichever subscriber is writing to standard output. This filter, for the tracing-opentelemetry layer, is hardcoded for now.

Issue #1467, opened 2023-06-08
Incremental garbage collection


branlwyd:
The current garbage collection implementation requires deleting all related expired reports, aggregation jobs, and collection jobs for a given task in a single transaction. This has a few problems:

* For large-scale tasks, deleting this much data in a single transaction will likely be too expensive.
* For tasks of any scale, if the GC process falls behind, we don't have an effective way to "catch up" and may fall into the same transaction size issue as with large-scale tasks.
* Strange dependencies could exist between entities which could delay GC in confusing ways. For example, if a very old & a recent client report happened to be grouped together into an aggregation job, neither of the reports could be GC'ed until the recent client report was ready for GC. This would lead the very old report to be kept around much longer than it otherwise would be.
* The SQL queries used to implement GC are very complex and likely expensive.

For this reason, an "incremental" GC is desirable. This GC should be able to delete arbitrary expired client reports, aggregation jobs, and collection jobs (& related artifacts) without dependencies on other types of entities being deleted in the same transaction.

Design sketch:

# Incremental GC in Janus

## Client reports

Client reports are stored in the `client_reports` table.

This design will make the following changes:
* Any `client_reports` older than `report_expiry_age` will now be considered eligible for GC, independent of GC-eligibility of any other entities in the system.
* The foreign-key relation of `report_aggregations.client_report_id` on `client_reports.id` will be dropped, to allow client report deletion to occur independently of aggregation job deletion.
* All client-report read & write methods will be updated to act as if `client_reports` rows which are eligible for GC do not exist. (e.g.: read methods will return nothing; write methods will fail with an appropriate row-does-not-exist error)

## Aggregation jobs

Aggregation jobs are stored in the `aggregation_jobs` and `report_aggregations` tables.

This design will make the following changes:
* Any aggregation jobs whose maximum of `client_timestamp_interval` is older than `report_expiry_age` will now be considered eligible for GC, independent of GC-eligibility of any other entities in the system. (Report aggregations are considered to be a part of a single related aggregation job; they will be considered eligible for GC if their related aggregation job is.)
* All aggregation-job read & write methods will be updated to act as if `aggregation_jobs`/`report_aggregations` rows eligible for GC do not exist.
* The start of the aggregation process reads client reports to retrieve report share information. A missing report will now fail that report's aggregation (with `ReportDropped`) rather than failing the entire aggregation job.
[...]

PR #1503, opened 2023-06-15
Break dependency of report_aggregations on client_reports.

branlwyd:
This is required to allow client reports to be GC'ed independently of aggregation jobs; notably, this is the only foreign-key relationship crossing the reports/aggregations/collections boundary.

The changes are, roughly:
  * The report_aggregations table now stores the client report ID, rather than a reference to a client_reports row, as well as the client_timestamp of the report.
  * At the start of the aggregation process, if a client report can't be found, it fails aggregation of that report (with a ReportDropped error) rather than failing the entire aggregtion.

Rather than copying input share information at time of the first aggregation step, we could instead copy it at time of aggregation job creation. The upside is that this would remove the possibility of the client report being GC'ed between aggregation job creation & the first step. However, I think aggregating reports that are actively being GC'ed is a marginal enough case that it does not warrant the extra storage complexity to make this happen.

Part of #1467.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_driver.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver.rs
@@ -21,7 +21,8 @@ use janus_core::{time::Clock, vdaf_dispatch};
 use janus_messages::{
     query_type::{FixedSize, TimeInterval},
     AggregationJobContinueReq, AggregationJobInitializeReq, AggregationJobResp,
-    PartialBatchSelector, PrepareStep, PrepareStepResult, ReportShare, ReportShareError, Role,
+    PartialBatchSelector, PrepareStep, PrepareStepResult, ReportId, ReportShare, ReportShareError,
+    Role,
 };
 use opentelemetry::{
     metrics::{Counter, Histogram, Meter, Unit},
@@ -32,7 +33,12 @@ use prio::{
     vdaf::{self, PrepareTransition},
 };
 use reqwest::Method;
-use std::{collections::HashSet, hash::Hash, sync::Arc, time::Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    hash::Hash,
+    sync::Arc,
+    time::Duration,
+};
 use tokio::try_join;
 use tracing::{info, warn};
 
@@ -169,7 +175,7 @@ impl AggregationJobDriver {
                     // Read client reports, but only for report aggregations in state START.
                     // TODO(#224): create "get_client_reports_for_aggregation_job" datastore
                     // operation to avoid needing to join many futures?
-                    let client_reports =
+                    let client_reports: HashMap<_, _> =
                         try_join_all(report_aggregations.iter().filter_map(|report_aggregation| {
                             if report_aggregation.state() == &ReportAggregationState::Start {
                                 Some(
@@ -184,13 +190,9 @@ impl AggregationJobDriver {
                                             *report_aggregation.report_id(),
                                             lease.leased().task_id(),
                                         ))
-                                        .and_then(|maybe_report| {
-                                            maybe_report.ok_or_else(|| {
-                                                anyhow!(
-                                                    "couldn't find report {} for task {}",
-                                                    report_aggregation.report_id(),
-                                                    lease.leased().task_id(),
-                                                )
+                                        .map(|report| {
+                                            report.map(|report| {
+                                                (*report_aggregation.report_id(), report)
                                             })
                                         })
                                         .map_err(|err| datastore::Error::User(err.into()))
@@ -200,7 +202,10 @@ impl AggregationJobDriver {
                                 None
                             }
                         }))
-                        .await?;
+                        .await?
+                        .into_iter()
+                        .flatten()
+                        .collect();
 
                     Ok((
                         Arc::new(task),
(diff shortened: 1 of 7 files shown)


db/00000000000010_rm_client_report_fk.down.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE report_aggregations DROP COLUMN client_timestamp;
+ALTER TABLE report_aggregations DROP COLUMN client_report_id;
+ALTER TABLE report_aggregations ADD COLUMN client_report_id BIGINT NOT NULL;
+ALTER TABLE report_aggregations ADD CONSTRAINT fk_client_report_id FOREIGN KEY(client_report_id) REFERENCES client_reports(id) ON DELETE CASCADE;
+CREATE INDEX report_aggregations_client_report_id_index ON report_aggregations(client_report_id);
```
inahga, line 5: Need to delete index rather than create.

(I think this should have failed the CI though... isn't creation of a duplicate index an error?)

aggregator_core/src/datastore.rs
```diff
@@ -1332,7 +1332,7 @@ impl<C: Clock> Transaction<'_, C> {
                         ON aggregation_jobs.id = report_aggregations.aggregation_job_id
                         WHERE aggregation_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
                     ) AS report_aggs
-                    ON report_aggs.client_report_id = client_reports.id
+                    ON report_aggs.client_report_id = client_reports.report_id  -- this join is very inefficient, fix before deploying in non-test scenario
```
inahga, line 1335: I see there is no index on client_reports.report_id, is this all that would be necessary?

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -290,34 +295,35 @@ impl AggregationJobDriver {
...
         let mut report_shares = Vec::new();
         let mut stepped_aggregations = Vec::new();
-        for (report_aggregation, report) in reports {
+        for report_aggregation in report_aggregations {
+            // Look up report.
+            let report = if let Some(report) = client_reports.get(report_aggregation.report_id()) {
+                report
+            } else {
+                info!(report_id = %report_aggregation.report_id(), "Attempted to aggregate missing report (most likely garbage collected)");
+                self.aggregate_step_failure_counter.add(
```
inahga, line 316: :+1: for the instrumentation here.

db/00000000000010_rm_client_report_fk.down.sql
inahga: It is an error...
```
postgres@127:postgres> create table foo(id int primary key, bar int);
CREATE TABLE
Time: 0.014s
postgres@127:postgres> create index baz on foo(bar);
CREATE INDEX
Time: 0.013s
postgres@127:postgres> create index baz on foo(bar);
relation "baz" already exists
Time: 0.002s
```
That suggests the smoke test from https://github.com/divviup/janus/pull/1434 is not working. I will investigate.

inahga: Nevermind! I see the column is dropped and the index has to be recreated.

branlwyd: Deleting the `client_report_id` column [as part of a delete/recreate operation] also deletes any indexes on the column, such as `report_aggregations_client_report_id_index`. Surprisingly, since both the down & up migrations do a delete/recreate column operation, both the down & up migration need to recreate the index.

aggregator_core/src/datastore.rs
```diff
@@ -3717,7 +3713,7 @@ impl<C: Clock> Transaction<'_, C> {
                 WHERE task_id = (SELECT id FROM tasks WHERE task_id = $1) AND client_timestamp < $2
                 AND NOT EXISTS(
                     SELECT FROM report_aggregations
-                    WHERE report_aggregations.client_report_id = client_reports.id)",
+                    WHERE report_aggregations.client_report_id = client_reports.report_id)  -- this join is very slow, fix before deploying",
```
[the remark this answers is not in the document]
branlwyd: The first instance is test-only code that (TBH) should not have been committed yet for YAGNI reasons, so I don't mind losing performance on it. Of course, instead of maintaining this code poorly, I should instead just remove it -- I sent #1508 for this.

The second one is in GC code that will be replaced by further work in #1467. I added a reference to the issue in the TODO.

aggregator_core/src/datastore.rs
branlwyd: Possibly. As noted, this is test code; I'll just remove it instead in #1508.

divergentdave (approved):

inahga (approved):

tgeoghegan (approved):

db/00000000000010_rm_client_report_fk.up.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE report_aggregations DROP CONSTRAINT fk_client_report_id;
+ALTER TABLE report_aggregations DROP COLUMN client_report_id;
+ALTER TABLE report_aggregations ADD COLUMN client_report_id BYTEA NOT NULL;
```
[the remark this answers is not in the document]
branlwyd: (I missed this comment before merging, I resolved it in #1508.)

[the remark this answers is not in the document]
branlwyd: Well, maybe not. I can't edit a migration, even just to add a comment. We'll need to fix this up when we next merge migrations, likely at the beginning of the next major release.

PR #1508, opened 2023-06-16
Remove aggregation job creation logic for VDAFs w/ agg param.

branlwyd:
This is test-only code, as such VDAFs (e.g. Poplar1) are not supported for a variety of reasons. This will remain test-only code for the foreseeable future: we don't plan to implement support for Poplar1 until we have gone to production with a Prio3-based deployment. For YAGNI reasons, we should not commit test-only code or code that otherwise we do not plan to use "immediately"; this code constitutes a maintenance burden -- it is more efficient to wait until we need the functionality, and implement it then, to avoid the need to maintain the code in the interim time period. (More concretely, it is plausible that we will be removing time-interval query type support entirely, at which point this code is truly dead.)

(this fell out of discussion on #1503)


divergentdave (approved):

inahga (approved):

tgeoghegan (approved):

tgeoghegan:
Should we put a note in the top-level README explaining that only VDAFs with the unit aggregation parameter are currently supported?

tgeoghegan (approved):

Issue #1509, opened 2023-06-16
Store token hashes instead of tokens


tgeoghegan:
@jbr's comment in https://github.com/divviup/divviup-api/issues/195 made me think that we should change how Janus stores aggregator and collector authentication tokens so that we only store a cryptographic hash of the token instead of the actual token. This is a bit tricky to do, because currently we use the `aggregator_auth_tokens` table to store both tokens that our leader would present to a helper (in which case we need to store the actual token) as well as tokens that our helper would use to authenticate requests from another leader (in which case we only need the hash). So we would have to enrich the representation of tokens in Janus to distinguish between tokens we present and tokens we accept.

tgeoghegan:
@jbr points out that if Janus helpers no longer store the actual aggregator and collector auth tokens, then Janus/Divvi Up need a story for allowing subscribers to recover from losing a collector auth token or allowing a peer aggregator to recover from losing an aggregator auth token, without forcing the rotation of a task, which could be very expensive if it has been deployed to many clients.

One solution strategy would be for Janus' aggregator API to expose some endpoints for rotating the auth tokens associated with a task. Then, if a subscriber loses access to a token, they can revoke the old one and mint a new one.

tgeoghegan:
I am pretty well convinced by @jbr's point about the need to enable recovery from losing a collector authentication token. To that end, I think the Janus aggregator API should support rotation of collector authentication tokens. If we go do that, I think we may as well do it for aggregator auth tokens, too, because they'll mostly work the same.

The big open question here is how much disruption do we want to take on here. Storing collector auth token hashes instead of the unhashed token forces changes on the Divvi Up subscriber, because the existing `GET /tasks/:task_id/collector_auth_tokens` route on divviup-api can't work anymore (Janus will no longer have the token in the clear and so can't service that request). If we're going that far, should we take on the notion of "collector credentials" we discussed previously (https://github.com/divviup/divviup-api/issues/487), and make it the subscriber's responsibility to generate collector auth tokens and then provide divvivup-api the HPKE config + collector auth token hash?

I'm going to proceed on the assumption that we do _not_ want to introduce collector credentials, as I think it can be added after the changes I'm discussing here.

## Collector auth tokens

These are used by a Janus leader to authenticate incoming collection protocol requests. There may be many of these for each task. Any non-revoked token is valid.

### API routes

#### Janus aggregator API

[...]

Issue #1486, opened 2023-06-12
Self-service task provisioning and BYOHelper


tgeoghegan:
## Goals

We want Divvi Up (DU) subscribers to be able to use the DU website or API to provision tasks in which DU acts as the DAP leader. This should require a single API operation, such that DU can guarantee that when the subcriber gets a successful API response and is provided task parameters, both aggregators are ready to process reports in the task.

We want subscribers to be able to use a Janus helper that they are running themselves, which is not previously known to Divvi Up (this last part is what we call BYOHelper; Bring Your Own Helper).

As much as possible, this solution should absolve subscribers of generating secrets. Further, secrets associated with tasks should be exposed to the minimum possible number of protocol participants.

We want all of this to be doable by the subscriber without manual intervention by the Divvi Up team (this is the "self-service" part).

This memo describes a plan for implementing this set of features in our control (`divviup-api`) and data (`janus`) planes. It builds upon but is not necessarily consistent with ideas and APIs described in the [control plane design](https://docs.google.com/document/d/1RgKGb9cnzSxY58fnCZI5qFjFMPn44s0yjt648Jywcno/edit#) and [inter-aggregator task provisioning](https://docs.google.com/document/d/1lrNwfJNWAbf20WyMuhBIYvMIcl9RpscfWYOKKpNmcZw/edit#) documents.

## Non-goals

- This proposal does not address the case where Divvi Up runs a DAP helper.
- This proposal assumes that helpers brought by subscribers are running Janus, which means we can assume they have the Janus aggregator API. In the future, we could write a specification for the aggregator API that any helper implementation can conform to.

## Software components

[Janus](https://github.com/divviup/janus) is Divvi Up's implementation of the [Distributed Aggregation Protocol](https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/) and constitutes the _data plane_ for Divvi Up's deployment. Janus is open source and container images are published to public container registries. Divvi Up encourages partners to use Janus in their deployments, even if they're not using Divvi Up.

The [Janus aggregator API](https://github.com/divviup/janus/tree/main/aggregator_api) is a set of endpoints implemented in Janus that allow managing DAP tasks. It is the interface between Divvi Up's control plane and data plane.

[...]

PR #1525, opened 2023-06-22
Update aggregator API for BYOHelper

### LINKED ISSUE #1506: Revisit aggregator API's `PostTaskReq` for self-service task provisioning
In support of the plot laid out in #1486, the aggregator API's task creation endpoint needs at least the following changes:

- Remove task ID (will be derived from VDAF verify key)
- VDAF verify key is required
- remove collector auth token (will be generated by leader Janus)

tgeoghegan:
Resolves #1506

### DIFF
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -23,7 +23,7 @@ use ring::{
     constant_time,
     digest::{digest, SHA256},
 };
-use std::{str::FromStr, sync::Arc};
+use std::{str::FromStr, sync::Arc, unreachable};
 use tracing::{error, warn};
 use trillium::{Conn, Handler, Status};
 use trillium_api::{api, Halt, Json, State};
@@ -145,6 +145,13 @@ async fn post_task<C: Clock>(
     // representations will be harmonized.
     // https://github.com/divviup/janus/issues/1524
 
+    if !matches!(req.role, Role::Leader | Role::Helper) {
+        return Err(Error::new(
+            format!("invalid role {}", req.role),
+            Status::BadRequest,
+        ));
+    }
+
     // struct `aggregator_core::task::Task` expects to get two aggregator endpoint URLs, but only
     // the one for the peer aggregator is in the incoming request (or for that matter, is ever used
     // by Janus), so we insert a fake URL for "self".
@@ -153,12 +160,7 @@ async fn post_task<C: Clock>(
     let aggregator_endpoints = match req.role {
         Role::Leader => Vec::from([fake_aggregator_url, req.peer_aggregator_endpoint]),
         Role::Helper => Vec::from([req.peer_aggregator_endpoint, fake_aggregator_url]),
-        role => {
-            return Err(Error::new(
-                format!("invalid role {role}"),
-                Status::BadRequest,
-            ))
-        }
+        _ => unreachable!(),
     };
 
     let vdaf_verify_key_bytes = URL_SAFE_NO_PAD
@@ -184,7 +186,7 @@ async fn post_task<C: Clock>(
     // 32 byte task ID by taking SHA-256(VDAF verify key).
     // https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-04#name-verification-key-requiremen
     let task_id = TaskId::try_from(digest(&SHA256, &vdaf_verify_key_bytes).as_ref())
-        .map_err(|err| Error::new(err.to_string(), Status::BadRequest))?;
+        .map_err(|err| Error::new(err.to_string(), Status::InternalServerError))?;
 
     let vdaf_verify_keys = Vec::from([SecretBytes::new(vdaf_verify_key_bytes)]);
 
@@ -229,24 +231,14 @@ async fn post_task<C: Clock>(
             Vec::from([AuthenticationToken::DapAuth(random())])
         }
 
-        role => {
-            return Err(Error::new(
-                format!("invalid role {role}"),
-                Status::BadRequest,
-            ))
-        }
+        _ => unreachable!(),
     };
 
     let collector_auth_tokens = match req.role {
         // TODO(#472): switch to generating bearer tokens by default
         Role::Leader => Vec::from([AuthenticationToken::DapAuth(random())]),
         Role::Helper => Vec::new(),
-        role => {
-            return Err(Error::new(
-                format!("invalid role {role}"),
-                Status::BadRequest,
-            ))
-        }
+        _ => unreachable!(),
     };
     let hpke_keys = Vec::from([generate_hpke_config_and_private_key(
         random(),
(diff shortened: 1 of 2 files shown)


tgeoghegan:
This is working as far as the unit tests on `janus_aggregator_api` go, but next we need to test it against a `divviup-api` that's been updated to use the new `PostTaskReq`.

tgeoghegan:
The test support code in `integration_tests/tests/in_cluster.rs` uses `divviup-api` for task provisioning. At the moment, it hits the `divviup-api` instance associated with either aggregator to separately provision tasks. However, `divviup-api` will soon assume responsibility for provisioning the task in both aggregators (https://github.com/divviup/divviup-api/issues/199). Once that lands (which will come with changes to the messages for task creation), we'll need a further Janus change to adopt those new messages. #1528

aggregator_api/src/lib.rs
```diff
@@ -351,86 +384,147 @@ mod models {
...
+        /// DAP query type for this task.
         pub(crate) query_type: QueryType,
+        /// The VDAF being run by this task.
         pub(crate) vdaf: VdafInstance,
+        /// The role that this aggregator will play in this task.
         pub(crate) role: Role,
-        pub(crate) vdaf_verify_keys: Vec<String>,
+        /// The VDAF verification key used for this DAP task, as Base64 encoded bytes. Task ID is
+        /// derived from the verify key.
+        pub(crate) vdaf_verify_key: String,
```
tgeoghegan, line 431: Ideally this structure would not contain a secret value like VDAF verify key, to reduce the risk of it being exposed. However I'm not sure if that breaks task sync.

tgeoghegan:
Another thing we discussed is moving the aggregator API message definitions (`PostTaskReq`, `TaskResp`) into `janus_messages` so that `divviup-api` and others can avoid re-declaring them. That's a good idea, but it's going to require moving a whole bunch of type declarations between `janus_aggregator_core`, `janus_core` and `janus_messages`, so I'm not going to do that in this PR, both to avoid making this change any bigger, but also to avoid obscuring the changes to `PostTaskReq` and `TaskResp` that this change makes.

I think this change is ready for review so we can unblock further `divviup-api` work. We'll have to make sure to bump the versions used in `janus-ops` in lockstep once both sides are ready.

aggregator_api/src/lib.rs
```diff
@@ -136,87 +140,111 @@ async fn post_task<C: Clock>(
...
+            ),
+            Status::BadRequest,
+        ));
+    }
+
+    // DAP recommends deriving the task ID from the VDAF verify key. We deterministically obtain a
+    // 32 byte task ID by taking SHA-256(VDAF verify key).
+    // https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-04#name-verification-key-requiremen
+    let task_id = TaskId::try_from(digest(&SHA256, &vdaf_verify_key_bytes).as_ref())
+        .map_err(|err| Error::new(err.to_string(), Status::BadRequest))?;
```
tgeoghegan, line 185: From discussion in Slack: we should incorporate a salt into the task ID derivation, since task ID is public(ish). One option is to have the client of the aggregator API (i.e. divviup-api) pass in the same salt to both aggregators when it provisions the task.

tgeoghegan: After discussing further: the VDAF verify keys we currently deal with for Prio3 and Poplar1 are 16 bytes. The search space of 2^128 possible values is big enough that reversing the SHA-256 hash to get the VDAF verify key from a task ID is already impractical, so adding a salt isn't worth the complexity.

aggregator_api/src/lib.rs
```diff
@@ -136,87 +140,113 @@ async fn post_task<C: Clock>(
...
```
inahga, line 187: This strikes me as possibly InternalServerError? We've already validated that the vdaf_verify_key is of the correct length. Thus if we can't `try_from` then likely something went wrong with the `digest`, i.e. a bug.

aggregator_api/src/lib.rs
```diff
@@ -136,87 +140,113 @@ async fn post_task<C: Clock>(
...
+
+    let vdaf_verify_keys = Vec::from([SecretBytes::new(vdaf_verify_key_bytes)]);
```
inahga, line 189: Rust newbie question: what's the preference for `Vec::from([foo])` instead of `vec![foo]`? I would have leaned towards `vec!` but curious if there's something more subtle here.

aggregator_api/src/lib.rs
```diff
@@ -136,87 +140,113 @@ async fn post_task<C: Clock>(
...
+            // TODO(#472): switch to generating bearer tokens by default
+            Vec::from([AuthenticationToken::DapAuth(random())])
+        }
+
+        role => {
+            return Err(Error::new(
+                format!("invalid role {role}"),
+                Status::BadRequest,
+            ))
         }
```
inahga, line 237: Nit: In the spirit of DRY we could have a role check at the top of the function
```rust
if !matches(role, Role::Leader | Role::Helper) {
    /* same error logic */
}
```
Then subsequent matches over role can have the last case be `_ => unreachable!()`.

aggregator_api/src/lib.rs
inahga: Refresh my memory on what task sync accomplishes? I thought it was to sync up task definitions from janus to control plane. If that's the case then presumably the leader and helper already have the vdaf_verify_key and handing this over is not necessary?

aggregator_api/src/lib.rs
```diff
@@ -351,86 +384,147 @@ mod models {
...
+        /// requests from the Leader.
+        // TODO(#1509): This field will have to change as Janus helpers will only store a salted
+        // hash of aggregator auth tokens.
+        pub(crate) aggregator_auth_token: String,
+        /// The authentication token used by the task's Collector to authenticate to the Leader, as
+        /// Base64 encoded bytes.
+        /// `Some` if `role` is Leader, `None` otherwise.
+        // TODO(#1509) This field will have to change as Janus leaders will only store a salted hash
+        // of collector auth tokens.
+        pub(crate) collector_auth_token: Option<String>,
```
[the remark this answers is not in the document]
inahga: Where is it intended for the subscriber to get the collector auth token to run a collector?

inahga (commented):
Mostly questions to check my understanding, a few improvements available.

aggregator_api/src/lib.rs
tgeoghegan: I'm not sure if there's a Rust-wide convention for this, but we've used `Vec::from([T])` in Janus. There's nothing explicit about this in our [code style](https://github.com/divviup/janus/blob/main/docs/DEVELOPMENT.md#code-style), but I remember discussing this in the past -- @branlwyd do you remember if we wrote down anything about this style preference somewhere?

aggregator_api/src/lib.rs
tgeoghegan: I think I agree. But I'm not certain whether `divviup-api` ever needs to see this value beyond provisioning it into the aggregators, so I'm just making a note that we have to think about this before we risk breaking task sync.

aggregator_api/src/lib.rs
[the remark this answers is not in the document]
tgeoghegan: See [the sequence diagram](https://private-user-images.githubusercontent.com/19212995/244223772-4e0db56a-f55c-48de-9b09-dd15c0bed3b4.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJrZXkiOiJrZXkxIiwiZXhwIjoxNjg3MzcxNjc5LCJuYmYiOjE2ODczNzEzNzksInBhdGgiOiIvMTkyMTI5OTUvMjQ0MjIzNzcyLTRlMGRiNTZhLWY1NWMtNDhkZS05YjA5LWRkMTVjMGJlZDNiNC5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBSVdOSllBWDRDU1ZFSDUzQSUyRjIwMjMwNjIxJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDIzMDYyMVQxODE2MTlaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1lYjQwNDUyMTVkMWUzY2Q3NmIyMzRiZjU4NjJhY2I2ZTYyOTdlNWI4NDE0MzZkYWQwOTI4MDgwYzRhYjA1ZWIzJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZhY3Rvcl9pZD0wJmtleV9pZD0wJnJlcG9faWQ9MCJ9.ngM2dTLOSF0ZuUDdLI9r6GWn34Qp2qTRWUYy_9OXC8c) in #1486. Short version:

- When the Leader is provisioning the task into itself, it generates the collector auth token and stores it in its own datastore.
- Leader delivers the collector auth token to divviup-api in this `TaskResp` message
- divviup-api then provides the collector auth token to the subscriber in the response to the `NewTask` message
- Collector is expected to store the auth token and use it later to authenticate collection job requests to the leader

The nuance here is that we need to give the subscriber the collector auth token "in the clear" when the task is created. But when we do #1509, Janus will start storing a token hash instead of the token. That means that a `GET /tasks/:id` request cannot contain the actual auth token, because the leader doesn't have it anymore at that point. I guess at that point this field can just be `None`, so maybe no message field changes will be needed.

aggregator_api/src/lib.rs
tgeoghegan: Thanks, that is tidier

inahga (approved):

PR #1542, opened 2023-06-26
Incremental aggregation-artifact GC.

branlwyd:
I removed part of the GC binary tests, as the collection artifact portion of these tests no longer work but it doesn't make sense to fix them up as the model will soon be changing. I will fix them up with the collection artifact GC update.

Part of #1467.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -1063,7 +1063,7 @@ impl<C: Clock> Transaction<'_, C> {
     /// Retrieves report & report aggregation metrics for a given task: either a tuple
     /// `Some((report_count, report_aggregation_count))`, or None if the task does not exist.
     #[tracing::instrument(skip(self), err)]
-    pub async fn get_task_metrics(&self, task_id: TaskId) -> Result<Option<(u64, u64)>, Error> {
+    pub async fn get_task_metrics(&self, task_id: &TaskId) -> Result<Option<(u64, u64)>, Error> {
         let stmt = self
             .prepare_cached(
                 "SELECT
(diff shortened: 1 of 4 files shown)


aggregator_core/src/datastore.rs
```diff
@@ -1574,7 +1578,7 @@ impl<C: Clock> Transaction<'_, C> {
                     .prepare_cached(
                         "DELETE FROM client_reports
                         USING tasks
-                        WHERE client_reports.id = tasks.task_id
+                        WHERE client_reports.task_id = tasks.id
```
tgeoghegan, line 1581: Was this just broken and we didn't notice because nothing was exercising garbage collection?

aggregator_core/src/datastore.rs
```diff
@@ -1932,10 +1935,38 @@ impl<C: Clock> Transaction<'_, C> {
...
             )
-            .await?,
-        )
+            .await?;
+        let row_count = rows.len().try_into()?;
+
+        if let Some(row) = rows.into_iter().next() {
+            // We got a row back, meaning we wrote an aggregation job. We check that it wasn't
+            // expired per the task's report_expiry_age, but otherwise we are done.
+            if row.get("is_expired") {
```
tgeoghegan, line 1947: IIUC this checks whether the interval of time spanned by the reports in this aggregation job is past the report expiry age for the task. Why do we check this _after_ writing an aggregation job instead of not writing the row in the first place?

tgeoghegan (approved):

aggregator_core/src/datastore.rs
branlwyd: This strategy is used because we need to differentiate between (a) a conflict with an existing row, which due to `ON CONFLICT DO NOTHING` causes nothing to be written and (b) the attempted write being already GC'ed. Case (a) leads to an error (in `check_insert`), while case (b) should silently not write anything (so that writing an aggregation job 1 second before expiry behaves the same as writing an aggregation job 1 second after expiry).

Another option would be to evaluate the expiry before writing anything, but then we'd need to read the task's report expiry age (or have this method take the task/report expiry age) before attempting a write -- this would be additional round-trips for the fast path of "write is not expired".

aggregator_core/src/datastore.rs
branlwyd: It was broken, but not because GC wasn't exercised -- this query was silently not deleting anything. But this wasn't noticed because this codepath is only taken if the attempted write is already expired; and in that case the read methods would treat the row as nonexistent. So this problem was undetectable without doing something like a custom SQL query to read the row disregarding expiration status.

The practical upshot of this was small -- a row would be written (and correctly ignored as garbage) and deleted once the garbage collector got to it, instead of no row being written.

If we want to do better, I think there are a few options:
(a) write some custom SQL / a `test-util`-only method that reads client report rows regardless of expiration status.
(b) drop this codepath entirely -- writing an expired entity will actually write a row to the DB, and we lean on the reader's expiration check to act as if the row is not there.

Thoughts?

tgeoghegan: Oh, I got mixed up: I thought the problem was that the query referred to columns on `client_reports` and `tasks` that don't exist, and was surprised that it wasn't straightforwardly rejected by Postgres. I understand the problem better now.

I'm not sure how much we should invest in detailed testing on this. It sounds like despite this bug, all the external behavior was correct: the row in question would get ignored by most queries and eventually would get deleted by GC. I'm not sure if the extra test code and the extra coupling to the code under test would be worth the assurance that the row actually gets deleted.

I think you understand this code much better than I do now so I'll defer to your judgment on the practicality of testing this.

aggregator_core/src/datastore.rs
tgeoghegan: OK, thanks. I don't find this code path particularly objectionable, I just wanted to make sure I understood why we're doing the less-than-screamingly-obvious thing.

PR #1543, opened 2023-06-28
Version the aggregator API with Content-Type

### LINKED ISSUE #1475: Aggregator API should be versioned
We should version the aggregator API, using a custom media type as we did in divviup-api.

tgeoghegan:
The aggregator API now requires that requests have a `Content-Type` and `Accept` header specifying the appropriate API version, matching `divviup-api`'s behavior.

Resolves #1475

### DIFF
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -25,7 +25,12 @@ use ring::{
 };
 use std::{str::FromStr, sync::Arc, unreachable};
 use tracing::{error, warn};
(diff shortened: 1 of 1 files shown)


tgeoghegan:
Moving to draft while I test this against a corresponding divviup-api change.

tgeoghegan:
Successfully tested against the corresponding divviup-api change in https://github.com/divviup/divviup-api/pull/244. We have to be careful about deploying these two changes in lockstep.

divergentdave (approved):

Issue #1470, opened 2023-06-09
acquire_incomplete_aggregation_jobs() fails to find jobs too often


divergentdave:
During the most recent cloud-based load test, `acquire_incomplete_aggregation_jobs()` frequently returned zero jobs, though there were thousands of jobs meeting `state = 'IN_PROGRESS' AND lease_token IS NULL`. I had raised the log filter level on the job_driver module, so I can get a longitudinal view of job acquisition with this log explorer query:

```
resource.type="k8s_container"
jsonPayload.message="No jobs available" OR jsonPayload.message="Acquired jobs"
jsonPayload.target="janus_aggregator::binary_utils::job_driver"
```

Over two hours and 40 minutes, I see 32,951 "No jobs available" messages, and 2,791 "Acquired jobs" messages. There was a backlog of unprocessed jobs until the end of this timeframe, and the aggregation job drivers were overall falling behind, so this query is not performing as we wish.

divergentdave:
This may or may not be related: I saw one instance of the following panic.

```
thread 'main' panicked at 'Acquired 24 jobs exceeding maximum of 4'
```

This never recurred, so it's a very rare problem, but it may help point to underlying problems in this query.

divergentdave:
I have put together a single-process test harness to reproduce this issue, with all the same parameters. The code is on the [david/experiment-agg-job-test-harness branch](https://github.com/divviup/janus/compare/david/experiment-agg-job-test-harness), (see commit e14c86bbf53e5b40febc91cea1adb8bb9eb82cf2) and the command to run is `RUST_LOG=job_perf=info JANUS_INTEROP_CONTAINER=skip cargo test -p janus_integration_tests --test job_perf --release`. (Release builds are necessary to shard reports fast enough) This runs a leader aggregator, aggregation job creator, and aggregation job driver, a helper aggregator, and some client tasks for five minutes. The goal is that almost all aggregation jobs should be finished, and there should not be more than a couple incomplete aggregation jobs eligible for acquisition. I've found after running this several times that I typically get 40-60 in progress aggregation jobs, most of which are eligible for acquisition, though on occasion I see that all aggregation jobs are finished.

I tried moving the aggregation job subquery to a CTE, and that didn't work. Next up, I plan to [inspect the lock manager](https://www.postgresql.org/docs/15/view-pg-locks.html) periodically while the test is running, to see what's locking the aggregation_jobs table and its rows.

For reference here's a representative query plan.

```
[...]

PR #1544, opened 2023-06-28
Move subqueries using LIMIT to WITH clauses

divergentdave:
This addresses an issue discovered while testing for #1470. Postgres warns that subqueries may be executed more than once, and in particular advises against performing any modifications within subqueries. It appears that read-only subqueries using `LIMIT` can also be executed multiple times, thus defeating our intended purpose of writing `column IN (SELECT ... LIMIT n)`. This PR moves all such subqueries to `WITH` clauses, which are guaranteed to either run only once, or appear to run only once (if they're used in a single place, which ours are). See https://www.postgresql.org/docs/current/sql-select.html#SQL-WITH.


tgeoghegan (approved):
Good find, that's a rather scary bug!

PR #1547, opened 2023-06-29
Optionally serve aggregator API alongside DAP API

### LINKED ISSUE #1505: Optionally serve aggregator API on the same TCP port as the DAP API
The automated task provisioning scheme in #1486 urges Divvi Up subscribers to expose the aggregator API to the internet so that Divvi Up can access it. This would be easier to do if both the DAP API and the aggregator API could be served from the same TCP port. Then the subscriber would only need to configure a single internet ingress.

However, in Divvi Up, we still want to restrict the aggregator API to a distinct port exposed only on a private virtual network, so we want this behavior to be configurable.

In the case where the aggregator API is on the same port as the DAP API, we should also allow configuring path prefixes at which either API is served to mitigate the risk of future API path collisions.

tgeoghegan:
Extends aggregator configuration so that the aggregator API can either:

 - be served on its own socket address independently of the DAP API
 - be served on the same address as the DAP API but at some arbitrary path prefix
 - not be served at all

Resolves #1505

### DIFF
--- a/aggregator/src/bin/aggregator.rs
+++ b/aggregator/src/bin/aggregator.rs
@@ -17,6 +17,7 @@ use std::{iter::Iterator, net::SocketAddr, sync::Arc, time::Duration};
 use tokio::join;
 use tracing::info;
 use trillium::Headers;
+use trillium_router::router;
 use trillium_tokio::Stopper;
 
 #[tokio::main]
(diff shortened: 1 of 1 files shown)


aggregator/src/bin/aggregator.rs
divergentdave, line 1: docs/samples/advanced_config/aggregator.yaml and any other files with the old settings should be updated as well.

divergentdave (approved):

PR #1548, opened 2023-06-29
aggregator API: default to bearer tokens

tgeoghegan:
Previously, the aggregator API assumed that any aggregator or collector auth tokens it was handling were `AuthenticationToken::DapAuth`. Janus must support those for the interop testing API and to work with Daphne, but generally, we prefer to use more boring `Authorization: Bearer` tokens. Certainly any tasks provisioned via the aggregator API and `divviup-api` should use bearer tokens.

With this PR, we augment the `PostTaskReq` and `PostTaskResp` messages so that they use `AuthenticationToken` to represent aggregator and collector tokens, meaning that on the wire, a token now looks like:

```
{
  "type": "Bearer",
  "token": "AAAAAAAAA<etc.>"
}
```

Also, when the aggregator API generates either kind of token, it now generates a bearer token. I'm not aware of any scenarios where we need to generate `DapAuth` tokens in the aggregator API so there is no affordance for requesting that kind of token.

See #472

### DIFF
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -10,10 +10,8 @@ use janus_aggregator_core::{
     SecretBytes,
 };
 use janus_core::{
-    hpke::generate_hpke_config_and_private_key,
-    http::extract_bearer_token,
-    task::{AuthenticationToken, DapAuthToken},
-    time::Clock,
+    hpke::generate_hpke_config_and_private_key, http::extract_bearer_token,
+    task::AuthenticationToken, time::Clock,
 };
 use janus_messages::{
     query_type::Code as SupportedQueryType, Duration, HpkeAeadId, HpkeKdfId, HpkeKemId, Role,
@@ -42,7 +40,7 @@ use url::Url;
 /// Represents the configuration for an instance of the Aggregator API.
 #[derive(Clone)]
 pub struct Config {
-    pub auth_tokens: Vec<SecretBytes>,
+    pub auth_tokens: Vec<AuthenticationToken>,
     pub public_dap_url: Url,
 }
 
@@ -217,6 +215,7 @@ async fn post_task<C: Clock>(
     // struct `aggregator_core::task::Task` expects to get two aggregator endpoint URLs, but only
     // the one for the peer aggregator is in the incoming request (or for that matter, is ever used
     // by Janus), so we insert a fake URL for "self".
+    // TODO(#1524): clean this up with `aggregator_core::task::Task` changes
     // unwrap safety: this fake URL is valid
     let fake_aggregator_url = Url::parse("http://never-used.example.com").unwrap();
     let aggregator_endpoints = match req.role {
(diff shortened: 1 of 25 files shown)


tgeoghegan:
Corresponding `divviup-api` change: https://github.com/divviup/divviup-api/pull/247

core/src/task.rs
divergentdave, line 1: I think we should avoid representing these tokens as base64 in some places and base64url in others.

A more flexible approach would be to not round-trip bearer tokens through either base64 or base64url, but instead validate that they match the syntax in RFC6750 when deserializing, store them as-is, and compare them as-is. This should be addressed separately, at any rate.

aggregator_api/src/lib.rs
```diff
@@ -441,10 +416,10 @@ mod models {
         pub(crate) time_precision: Duration,
         /// HPKE configuration for the collector.
         pub(crate) collector_hpke_config: HpkeConfig,
-        /// If this aggregator is the leader, this is the bearer token to use to authenticate
-        /// requests to the helper, as Base64 encoded bytes. If this aggregator is the helper, the
-        /// value is `None`.
-        pub(crate) aggregator_auth_token: Option<String>,
+        /// If this aggregator is the leader, this is the token to use to authenticate requests to
+        /// the helper, as Base64 encoded bytes. If this aggregator is the helper, the value is
+        /// `None`.
```
divergentdave, line 421: nit: the token is now a composite value

```suggestion
        /// If this aggregator is the leader, this is the token to use to authenticate requests to
        /// the helper. If this aggregator is the helper, the value is `None`.
```

aggregator_api/src/lib.rs
```diff
@@ -484,13 +459,13 @@ mod models {
         /// requests from the Leader.
         // TODO(#1509): This field will have to change as Janus helpers will only store a salted
         // hash of aggregator auth tokens.
-        pub(crate) aggregator_auth_token: String,
+        pub(crate) aggregator_auth_token: AuthenticationToken,
```
divergentdave, line 462: This doc comment should be updated as well, and same below.

core/src/task.rs
tgeoghegan: I agree with you that we need to be consistent about encodings. We previously discussed how we prefer unpadded Base64URL where possible, because:

- we are forced to use it to interpolate values into URLs
- it's more copy-pasteable than Base64 (whose alphabet contains `/`)

I had thought that the `Authorization: Bearer` header was the one place where tokens _must_ be represented as plain Base64. For that reason, I have been insisting that we need to consider the canonical form of an authentication token to be the bytes, and then Base64 or unpadded Base64URL encodings are details of a particular serialization context.

However, I just noticed that I have misread [RFC 6750 2.1](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1): its grammar for the `Authorization: Bearer` scheme allows both the Base64 and Base64URL alphabets (and some extra symbols like `.` and `~`).

I think that means we can use unpadded Base64URL _everywhere_: config files, JSON messages sent to APIs, and HTTP headers. Doing this would break existing clients/users of:

1. `divviup-api`
2. the Janus aggregator API
3. `janus_cli`'s task provisioning subcommand
4. the Janus DAP API, if authenticated by bearer tokens instead of `DAP-Auth-Token`

But (1), (2) and (3) are broken by the recent changes anyway. And I'm pretty sure (4) doesn't exist, because the only external users of Janus that have existed so far have used `DAP-Auth-Token`. Meaning I think we can still change the representation of bearer tokens in HTTP headers without (further) breaking anyone.

Now, if unpadded Base64URL is permissible everywhere we need to transmit an authentication token, then that means that we can, as you suggest, treat them opaquely. In other words, we can start treating the unpadded Base64URL encoding of some bytes as the *canonical* form of the token.

Long story short: I want to implement your suggestion. I think this PR is the right place to do it, as well.

tgeoghegan:
Downgrading to a draft; I'll tinker with this while on trains next week.

tgeoghegan:
We now use Base64URL unpadded everywhere, and this works in conjunction with the changes in https://github.com/divviup/divviup-api/pull/247 to do an end to end test where we use the BYOHelper flow to pair aggregators, provision a task and then run an aggregation. This is ready for review.

divergentdave (approved):
While this will be a disruptive change, I think it will be a nice improvement to have all auth tokens have the same representation in memory, on the wire in DAP requests, on the wire in aggregator API requests, in YAML files, and on the command line. We should mention this in the next release's notes, and also communicate it to our partners that have currently deployed Janus for testing purposes. (for example, old task YAML files may no longer parse, bearer-type authentication tokens in a database would be interpreted differently when loading and will likely throw an error)

PR #1549, opened 2023-06-30
aggregator API: idempotent task creation

### LINKED ISSUE #1507: Aggregator API idempotence
The aggregator API's endpoints should be idempotent. The only two currently existing endpoints that modify tasks are `POST /tasks` and `DELETE /tasks/:task_id`. The VDAF verify key and task ID (respectively) are natural idempotence keys.

tgeoghegan:
Makes the task creation endpoint in the aggregator API idempotent, by checking whether a task corresponding to the incoming `PostTaskReq` already exists. If it does, and all the other task's parameters match the `PostTaskReq`, then we return 201 Created. If the task exists but some other parameter (e.g., min batch size) differs, we return 409 Conflict.

Deletion was already correctly handled so no new changes or test were needed.

Resolves #1507

### DIFF
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -176,6 +176,8 @@ async fn get_task_ids<C: Clock>(
 }
 
 /// A simple error type that holds a message and an HTTP status. This can be used as a [`Handler`].
+#[derive(Debug, thiserror::Error)]
+#[error("{message}")]
 struct Error {
     message: String,
     status: Status,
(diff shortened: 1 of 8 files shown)


aggregator_api/src/lib.rs
```diff
@@ -293,7 +295,32 @@ async fn post_task<C: Clock>(
...
+                && existing_task.time_precision() == task.time_precision()
+                && existing_task.collector_hpke_config() == task.collector_hpke_config() {
+                    return Ok(())
+                }
+
+                let err = Error::new(
+                    "task with same VDAF verify key and task ID already exists with different parameters".to_string(),
+                    Status::Conflict,
+                );
+                return Err(datastore::Error::User(err.into()));
```
divergentdave, line 319: We need to match on the error received by map_err below for `datastore::Error::User`, try downcasting to this module's error type, and return that unwrapped error instead if it succeeds. Otherwise, duplicate tasks would just get the "Error storing task" message.

tgeoghegan: Good catch, I forgot to write a test for the task mutation case.

divergentdave (approved):

aggregator_api/src/lib.rs
tgeoghegan: Fixed: I did this to avoid writing identical `match` arms for the `_` case and the case where `datastore::Error:User` wraps something besides `aggregator_api::Error`. Too clever?

inahga (approved):
:+1:

branlwyd (approved):

PR #1551, opened 2023-06-30
Don't take explicit row locks in the tasks table

### LINKED ISSUE #1470: acquire_incomplete_aggregation_jobs() fails to find jobs too often

divergentdave:
This fixes our three `FOR UPDATE SKIP LOCKED` queries to only take row locks in the table of interest, and not in the tasks table as well. By removing contention on the task row, this removes a cause of getting zero rows out of these queries when we shouldn't.

This fixes #1470. The rewritten query broke the correlation between aggregation job queue liveness issues and table scan types used in query plans I had seen previously, and I haven't been able to trigger the problem again since.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -1306,7 +1306,7 @@ impl<C: Clock> Transaction<'_, C> {
                     WHERE tasks.task_id = $1
                       AND client_reports.aggregation_started = FALSE
                       AND client_reports.client_timestamp >= COALESCE($2::TIMESTAMP - tasks.report_expiry_age * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)
-                    FOR UPDATE SKIP LOCKED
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):
Great detective work here!

PR #1555, opened 2023-07-03
Job driver: wait for in-flight jobs upon SIGTERM

### LINKED ISSUE #1502: Graceful shutdown of job drivers
It would be nice to wait for in-flight jobs to complete when shutting down the aggregation job driver or collection job driver. Currently, the process will shut down almost immediately after a SIGTERM, so in-flight jobs would be abandoned, and then retried once their leases expire. This could be done with [CloneCounter](https://docs.rs/trillium-server-common/0.4.3/trillium_server_common/struct.CloneCounter.html).

_Originally posted by @jbr in https://github.com/divviup/janus/pull/1501#discussion_r1231293310_

divergentdave:
Fixes #1502.

### DIFF
--- a/aggregator/src/binary_utils/job_driver.rs
+++ b/aggregator/src/binary_utils/job_driver.rs
@@ -1,5 +1,6 @@
 //! Discovery and driving of jobs scheduled elsewhere.
 
+use anyhow::Context as _;
 use chrono::NaiveDateTime;
 use janus_aggregator_core::datastore::{self, models::Lease};
 use janus_core::{time::Clock, Runtime};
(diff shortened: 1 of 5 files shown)


aggregator/src/binary_utils/job_driver.rs
```diff
@@ -72,28 +77,33 @@ where
...
+            .context("max_concurrent_job_workers was too large")?;
+        Ok(Self {
             clock,
             runtime,
             meter,
+            stopper,
             min_job_discovery_delay,
             max_job_discovery_delay,
             max_concurrent_job_workers,
+            max_concurrent_job_workers_u32,
```
branlwyd, line 98: IMO, rather than storing the same value twice, do the conversion at point-of-use. It looks like the `_u32` version is only used in one place, during shutdown, so this should be OK. We can keep an attempted conversion in `new` to validate that the given value can be represented as a `u32`, then unwrap when we do the conversion "for real".

branlwyd (approved):

Issue #1568, opened 2023-07-03
Implement `taskprov` extension


branlwyd:
The taskprov extension is defined here: https://datatracker.ietf.org/doc/draft-wang-ppm-dap-taskprov/

We need to implement this for both the latest version of DAP (DAP-04, taskprov-03, at time of writing) as well as DAP-02 (taskprov-02). We can likely implement taskprov-03 on our `main` branch, then adapt to taskprov-02 when we fork `main` to implement DAP-02.

[The largest difference between taskprov-02 and taskprov-03 appears to be how the task configuration is conveyed: taskprov-02 conveys the task config in a client report extension, taskprov-03 conveys the task config in a header value.]

cjpatton:
Just a heads up that there is at least one change on the main branch that we want to include before you implement the latest draft. cc/ @wangshan

inahga:
Helper support for draft-wang-ppm-dap-taskprov-04 is implemented. At this time we don't have interest in implementing the other roles.

PR #1576, opened 2023-07-07
Add feature flag for taskprov to configuration

inahga:
Supports #1568.


divergentdave (approved):
LGTM. One suggestion, take it or leave it: For other configuration options that are shared by some but not all components, we repeat them in each binary's config struct, rather than putting it in CommonConfig. We may want to do the same with taskprov configuration (i.e. only include it in config structs for the aggregator and aggregation_job_driver) but I'm not 100% sure.

PR #1577, opened 2023-07-07
Explicitly pass OpenTelemetry Meters

### LINKED ISSUE #1531: Unify use of opentelemetry::metrics::Meter
Currently, we set up two `Meter`s with the global `MeterProvider` in each binary. For three binaries, we create one named "janus_aggregator" and one named after the executable. For the aggregator, we create one named "janus_aggregator" and one named "janus_aggregator_api". Since this is all first-party code, I think we should unify on "janus_aggregator". These names are visible on the `otel_scope_name` attribute attached to all metrics (including the info metric `otel_scope_info`).

Relatedly, I think we should explicitly pass `Meter` clones or references to components that need to set up metrics instruments, rather than using globals. Currently, we use a mix of these tactics. If we only use the global `MeterProvider` in top-level code, and then explicitly pass `Meter` or instrument dependencies down to components, this makes metrics easier to use in testing, by creating isolated `BasicController` instances for each test. (see #1470 for an example)

divergentdave:
This refactors our OpenTelemetry metrics handling by passing a `Meter` instance around explicitly when setting up components. We now only emit metrics under a single library name, and most tests use a `Meter` from `NoopMeterProvider`. By explicitly passing this dependency, we can now replace it by using a non-global `BasicController` in tests, in order to either test metrics themselves, or use metrics as an introspection hook to test some other behavior.

Closes #1531.

### DIFF
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -83,7 +83,7 @@ pub fn aggregator_api_handler<C: Clock>(ds: Arc<Datastore<C>>, cfg: Config) -> i
         State(ds),
         State(Arc::new(cfg)),
         // Metrics.
-        metrics("janus_aggregator_api").with_route(|conn| conn.route().map(ToString::to_string)),
+        metrics("janus_aggregator").with_route(|conn| conn.route().map(ToString::to_string)),
         // Authorization check.
         api(auth_check),
         // Check content type and accept headers
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -606,8 +606,10 @@ mod tests {
                 ReportAggregationState,
             },
             test_util::{ephemeral_datastore, EphemeralDatastore},
+            Datastore,
         },
         task::{test_util::TaskBuilder, QueryType, Task},
+        test_util::noop_meter,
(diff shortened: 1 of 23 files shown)


aggregator_api/src/lib.rs
```diff
@@ -83,7 +83,7 @@ pub fn aggregator_api_handler<C: Clock>(ds: Arc<Datastore<C>>, cfg: Config) -> i
         State(ds),
         State(Arc::new(cfg)),
         // Metrics.
-        metrics("janus_aggregator_api").with_route(|conn| conn.route().map(ToString::to_string)),
+        metrics("janus_aggregator").with_route(|conn| conn.route().map(ToString::to_string)),
```
inahga, line 86: Clarification: is it still possible to differentiate metrics coming from the aggregator api vs aggregator?

inahga (approved):

aggregator_api/src/lib.rs
divergentdave: The only metric names that the two crates both use are "http.server.duration" et. al., but those can be differentiated via the `http_route` label. Otherwise, our custom metrics were partitioned between different scopes.

PR #1580, opened 2023-07-10
Improve aggregation job creator graceful shutdown

divergentdave:
This PR replaces oneshot channels and `select!` with `Stopper` in per-DAP task job creation tasks, checks for shutdown signals during each loop's sleep, and waits for all tasks to complete before exiting instead of cancelling the main task and letting per-DAP task tasks run detached.

Using `Stopper` should be more lightweight than `select!`, as the combinator future is better optimized for this simple case, and random number generators don't need to get involved. Avoiding cancellation of futures that interact with the database means we get a chance to finish work-in-progress in open transactions, and we will cause fewer transaction rollbacks and connection errors during normal operation.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_creator.rs
+++ b/aggregator/src/aggregator/aggregation_job_creator.rs
@@ -48,7 +48,7 @@ use tokio::{
     try_join,
 };
 use tracing::{debug, error, info};
-use trillium_tokio::Stopper;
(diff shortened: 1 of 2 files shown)


inahga (approved):
Nicely written.

jbr (approved):
🎉

branlwyd (approved):

Issue #1574, opened 2023-07-06
Implement time-bucketing for fixed-size batches.


branlwyd:
Idea: the fixed-size query type allows the Leader to select the assignment of client reports to batches. The Leader could choose to generate batches such that all reports happen to fall into the same time window/bucket, for example, all reports are in the same day.

The upsides of this are:
* The collector could aggregate the collection results into a time-based bucket. These time buckets could be updated over time as additional batches arrive, unlike time-interval query types, where the first collection "fixes" the aggregate value of the time bucket. This would be especially important for use-cases where there is a long tail of late-arriving reports.
* This might provide an alternative time-bucketing strategy to the time-interval query type, allowing the time-interval query type to be dropped entirely.

Downsides/challenges:
* Batches must be at least `min_batch_size` in size; thus, up to `min_batch_size` reports per time window can't be placed into a batch. The most likely solution is to simply drop these reports, but this will only work if the overall report volume is large enough & `min_batch_size` small enough that these reports are statistically insignificant.

The implementation strategy should likely be to add a new (non-standard/implementation-specific) task parameter to the fixed-size query type, an interval of time similar to `time_precision`, which determines the batch window bucket size. Then augment the `aggregation_job_creator` to generate aggregation jobs such that all reports for a given batch fall into the same window.

PR #1600, opened 2023-07-10
Use sharded report counts in get_filled_outstanding_batch()

divergentdave:
This PR simplifies the `get_filled_outstanding_batch()` query, so that it only joins against `batch_aggregations`, not `aggregation_jobs` and `report_aggregations`. Instead of counting finished report aggregations directly, we add up `report_count` columns from sharded batch aggregations, and perform the same filtering.

While looking at #1574, I was concerned that this query would fare poorly with many never-to-be-filled outstanding batches. This doesn't address the underlying asymptotic issue, but it does knock down the cardinality by using a much cheaper join and aggregate function. This rewrite may also be of use to #1467, since the query no longer touches report_aggregations.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -3608,17 +3608,14 @@ impl<C: Clock> Transaction<'_, C> {
                 "WITH batches AS (
                     SELECT
                         outstanding_batches.batch_id AS batch_id,
-                        COUNT(DISTINCT report_aggregations.client_report_id) AS count
+                        SUM(batch_aggregations.report_count) AS count
                     FROM outstanding_batches
-                    JOIN aggregation_jobs
-                      ON aggregation_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
-                     AND aggregation_jobs.batch_id = outstanding_batches.batch_id
-                    JOIN report_aggregations
-                      ON report_aggregations.aggregation_job_id = aggregation_jobs.id
-                     AND report_aggregations.state = 'FINISHED'
-                    GROUP BY outstanding_batches.batch_id
+                    JOIN batch_aggregations
+                      ON batch_aggregations.batch_identifier = outstanding_batches.batch_id
+                    WHERE outstanding_batches.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                    GROUP BY outstanding_batches.batch_id, batch_aggregations.aggregation_param
                 )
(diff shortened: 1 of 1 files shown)


aggregator_core/src/datastore.rs
```diff
@@ -3608,17 +3608,14 @@ impl<C: Clock> Transaction<'_, C> {
...
-                      ON aggregation_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
-                     AND aggregation_jobs.batch_id = outstanding_batches.batch_id
-                    JOIN report_aggregations
-                      ON report_aggregations.aggregation_job_id = aggregation_jobs.id
-                     AND report_aggregations.state = 'FINISHED'
-                    GROUP BY outstanding_batches.batch_id
+                    JOIN batch_aggregations
+                      ON batch_aggregations.batch_identifier = outstanding_batches.batch_id
+                    WHERE outstanding_batches.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                    GROUP BY outstanding_batches.batch_id, batch_aggregations.aggregation_param
```
branlwyd, line 3616: I'm not sure grouping by aggregation param is helpful; `outstanding_batches` is scoped to the batch (i.e. `(batch_id)` only) rather than batch & aggregation param (i.e. `(batch_id, aggregation_param)`). We'll need to fix this up once we support aggregating the same batch multiple times, but for now I don't think we should look to the aggregation param at all.

aggregator_core/src/datastore.rs
```diff
@@ -3608,17 +3608,14 @@ impl<C: Clock> Transaction<'_, C> {
...
                 )
-                SELECT batch_id FROM batches WHERE count >= $2 LIMIT 1",
+                SELECT DISTINCT batch_id FROM batches WHERE count >= $2::bigint LIMIT 1",
```
branlwyd, line 3618: Q: why distinct (given `LIMIT 1`)? why explicitly cast to `bigint`?

divergentdave: The bigint cast is necessary because otherwise the prepared query expects a `numeric` type, and there isn't a canned `ToSql` implementation for that.

aggregator_core/src/datastore.rs
divergentdave: Both the aggregation_param GROUP BY and the SELECT DISTINCT were intended to be parallel to the `COUNT(DISTINCT ...)` in the old query, for this purpose. I can take them out.

PR #1603, opened 2023-07-13
Incremental collection artifact GC.

branlwyd:
The implemented GC policy:
* batches, batch_aggregations, and outstanding_batches will be considered part of the same entity for purposes of GC, and will be considered eligible for GC once the maximum of the batch interval (for time-interval) or client_timestamp_interval (for fixed-size) of the batches row is older than report_expiry_age.
* collection_jobs and aggregate_share_jobs use the same rule to determine GC-eligiblity, but this rule is query type-specific.
  * For time-interval tasks, collection_jobs and aggregate_share_jobs are considered eligible for GC if the minimum of the collection interval is older than report_expiry_age. (The minimum is used instead of the maximum to ensure that collection jobs are not GC'ed after their underlying aggregation information from batch_aggregations.)
  * For fixed-size tasks, collection_jobs and aggregate_share_jobs are considered eligible for GC if the related batch is eligible for GC or has been GC'ed. [This choice induces a dependency between GC of different kinds of entities, but this is likely survivable -- thoughts on alternate approaches welcome.]

Note that I also had to make a few data-model changes to implement this. Most notably, `batches` are now written by the Helper when an aggregation job is created; this is used to track the client timestamp interval for each given batch, to allow Helper `aggregate_share_jobs` and `batch_aggregations` to be GC'ed appropriately.

Part of #1457.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -3468,6 +3468,7 @@ impl<C: Clock> Transaction<'_, C> {
         task_id: &TaskId,
         batch_id: &BatchId,
     ) -> Result<(), Error> {
+        // XXX
         let stmt = self
             .prepare_cached(
                 "INSERT INTO outstanding_batches (task_id, batch_id)
@@ -3493,22 +3494,32 @@ impl<C: Clock> Transaction<'_, C> {
         &self,
         task_id: &TaskId,
     ) -> Result<Vec<OutstandingBatch>, Error> {
+        // XXX: update tests
         let stmt = self
             .prepare_cached(
                 "SELECT batch_id FROM outstanding_batches
-                WHERE task_id = (SELECT id FROM tasks WHERE task_id = $1)",
+                JOIN tasks ON tasks.id = outstanding_batches.task_id
+                JOIN batches ON batches.batch_identifier = outstanding_batches.batch_id
+                WHERE tasks.task_id = $1
+                  AND UPPER(batches.client_timestamp_interval) >= COALESCE($2::TIMESTAMP - tasks.report_expiry_age * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)",
             )
             .await?;
 
         try_join_all(
-            self.query(&stmt, &[/* task_id */ task_id.as_ref()])
-                .await?
-                .into_iter()
-                .map(|row| async move {
-                    let batch_id = BatchId::get_decoded(row.get("batch_id"))?;
-                    let size = self.read_batch_size(task_id, &batch_id).await?;
-                    Ok(OutstandingBatch::new(*task_id, batch_id, size))
-                }),
+            self.query(
+                &stmt,
+                &[
+                    /* task_id */ task_id.as_ref(),
+                    /* now */ &self.clock.now().as_naive_date_time()?,
+                ],
+            )
+            .await?
+            .into_iter()
+            .map(|row| async move {
+                let batch_id = BatchId::get_decoded(row.get("batch_id"))?;
+                let size = self.read_batch_size(task_id, &batch_id).await?;
+                Ok(OutstandingBatch::new(*task_id, batch_id, size))
+            }),
         )
         .await
     }
@@ -3595,6 +3606,7 @@ impl<C: Clock> Transaction<'_, C> {
         task_id: &TaskId,
         min_report_count: u64,
     ) -> Result<Option<BatchId>, Error> {
+        // XXX
         // TODO(#1467): fix this to work in presence of GC.
         let stmt = self
             .prepare_cached(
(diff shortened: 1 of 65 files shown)


branlwyd:
(I tried to make the individual commits in this branch as useful as possible, as this PR turned out quite large.)

aggregator_core/src/datastore.rs
```diff
@@ -3658,10 +3660,40 @@ impl<C: Clock> Transaction<'_, C> {
...
             )
-            .await?,
-        )
+            .await?;
+        let row_count = rows.len().try_into()?;
+
+        if let Some(row) = rows.into_iter().next() {
+            // We got a row back, meaning we wrote a batch. We check that it wasn't expired per the
+            // task's report_expiry_age, but otherwise we are done.
+            if row.get("is_expired") {
```
inahga, line 3890: Since this overall insertion function is relatively simple ([as compared to this](https://github.com/divviup/janus/pull/1526/files#diff-ad4962043a4d91715b26c52cd1ac0b490a8ac647f82f7cf34dd61f56ff134972R1565)), can't we check whether the batch is expired outside of SQL and just no-op?

aggregator_core/src/datastore.rs
```diff
@@ -3758,20 +3799,27 @@ impl<C: Clock> Transaction<'_, C> {
...
-                let aggregation_parameter =
-                    A::AggregationParam::get_decoded(row.get("aggregation_param"))?;
-                Self::batch_from_row(*task_id, batch_identifier, aggregation_parameter, row)
-            })
-            .collect()
+        self.query(
+            &stmt,
+            &[
+                /* task_id */ task_id.as_ref(),
+                /* now */ &self.clock.now().as_naive_date_time()?,
```
inahga: Remark: I do wish there was some way to dedupe this pattern, but I haven't thought of anything that would be more complexity than its worth.

aggregator_core/src/datastore.rs
```diff
@@ -3468,24 +3468,53 @@ impl<C: Clock> Transaction<'_, C> {
...
-            &[
-                /* task_id */ task_id.as_ref(),
-                /* batch_id */ batch_id.as_ref(),
-            ],
-        )
-        .await?;
+        if let Some(row) = rows.into_iter().next() {
+            // We got a row back, meaning we wrote an outstanding batch. We check that it wasn't
+            // expired per the task's report_expiry_age, but otherwise we are done.
+            if row.get("is_expired") {
```
inahga, line 3668: Similar to my point above, can we check for expiry outside of sql? Well... we can't outside of sql because we have to select the batch's client_timestamp_interval.

So in this case I don't think the same feedback applies. It's less expensive to unconditionally insert and rarely delete versus unconditionally select then insert. (sorry, just thinking out loud).

inahga (commented):
So far so good. Checkpoint of my feedback, I'll have to continue Monday.

branlwyd:
(I fixed up the GC-specific tests; technically this could have gone in a different PR, but these are tests pertaining to the application-logic changes in this PR, so they should really go with this PR.)

aggregator_core/src/datastore.rs
branlwyd: We could do that, but to do so, we'd need to know the task's report expiry age (since expiry is determined by this parameter).

Since this method doesn't take a `Task`, we'd need to read the report expiry age, inducing an extra round-trip for each insertion. With the strategy I take here, a non-expired insertion is one round-trip, and an expired insert is two round-trips. In the (hopeful) case that the large majority of inserts are non-expired, this is a savings.

There are a few alternate strategies I could have taken:
* I could take the `Task` (or just the expiry) as an argument. I _think_ this would work out in the sense that I think application uses of these methods either have or could plumb the relevant `Task` data through to the insertion method calls; but this would complicate the usage of every insertion method, and we'd need to deal with (unlikely, natch) issues of passing the wrong task.
* I could skip expiry-checking on insert at all, always writing a row to the DB. We would then rely on the expiry-checking on read to avoid reading expired data, and lean on the GC to lazily delete these rows eventually. (TBH, maybe I should have gone this way...)

aggregator_core/src/datastore.rs
branlwyd: Agreed. 👍🏻

aggregator_core/src/datastore.rs
branlwyd: Yup, that's the right reasoning. See my response to your other comment along these lines for how I think about this approach.

aggregator_core/src/datastore.rs
inahga: Hmm taking another look with a fresh brain I'm not sure why I thought this was doable without an extra round-trip, it's right there `(SELECT report_expiry_age FROM tasks WHERE task_id = $1)`.

> I could skip expiry-checking on insert at all, always writing a row to the DB. We would then rely on the expiry-checking on read to avoid reading expired data, and lean on the GC to lazily delete these rows eventually. (TBH, maybe I should have gone this way...)

I like that but I don't find enough rationale to suggest executing on it.

inahga (approved):
Nice work!

PR #1604, opened 2023-07-13
Abandon release/0.2 branch.

branlwyd:
(We'll soon be forking `main` back to a DAP-02 release to pick up all of the work that has gone in since release/0.2 was under active development.)


inahga (approved):
:headstone:

divergentdave (approved):

Issue #239, opened 2022-06-13
HPKE keys: explicitly mark one key "primary"


branlwyd:
At time of writing, the "primary" HPKE config for a task (used for encryption, passed to clients when they request the current HPKE config, prioritized for decryption) is determined as the HPKE config with maximal config ID.

This could be problematic since the HPKE config ID is a `u8`, and so may plausibly wrap around. We should instead explicitly mark one of the HPKE configs as "primary".

inahga:
Do we *need* to determine a primary HPKE key?

We can present a list to the client, and the client can make that decision for themselves, i.e. choose index 0 or choose the key with ciphers that it supports/prefers. The client has to index the response anyways, because we return a list of configs.

We will know which key was used for decryption because its ID is supplied in plaintext along with the encrypted report.

inahga:
This is more or less moot with global HPKE keys support and the key rotator https://github.com/divviup/janus/issues/1641.

PR #1610, opened 2023-07-14
Expose global HPKE configuration.

inahga:

As called for by https://github.com/wangshan/draft-wang-ppm-dap-taskprov/blob/main/draft-wang-ppm-dap-taskprov.md#supporting-hpke-configurations-independent-of-tasks-hpke-config-no-task-id.

~This provides two options for storing a global HPKE configuration. One is the config file method, which is mostly there incidentally since the Config struct is used to propagate the configuration. The preferred method is storing it in the database. A `task_hpke_key` with no task ID is denoted as the global HPKE key, and only one is allowed in the database. A future PR will extend `janus_cli` to provision this key on a fresh database.~

~We will maintain the 1:1 mapping of HPKE config to task ID. Each newly provisioned task will simply copy the global HPKE config (this also entails re-encrypting the private key with the correct AAD). This also aids in key rotation--we only have to change the key and all previous tasks will retain linkage to the old key.~

Introduces a new feature for advertising a global HPKE key configuration to clients. Configuration is specified in and periodically refreshed from the database.

It also provides some handling for key rotation. A global key can be in two states:
- Expired: Should no longer be used by clients. No longer advertised to clients on `/hpke_config`. Still available to Janus for report share decryption. Represented by a `global_hpke_key` having a `expired_at` timestamp in the past.
- Active: Advertised on `/hpke_config` and valid for clients to use.

I would suggest this strategy also be carried over for task_hpke_keys to address https://github.com/divviup/janus/issues/239.

What's missing from this PR, to come in future PR to keep this one small:
- Report share decryption for shares that have used a global HPKE key.
- `janus_cli` extension for inserting keys into DB (expecting operator to insert by DB console is not reasonable IMO).


aggregator_core/src/datastore.rs
```diff
@@ -3952,6 +3952,71 @@ impl<C: Clock> Transaction<'_, C> {
...
+        if hpke_key_rows.is_empty() {
+            Ok(None)
+        } else {
+            let config = HpkeConfig::get_decoded(hpke_key_rows[0].get("config"))?;
+
+            let config_id = u8::try_from(hpke_key_rows[0].get::<_, i16>("config_id"))?;
+            let encrypted_private_key: Vec<u8> = hpke_key_rows[0].get("private_key");
+            let private_key = HpkePrivateKey::new(self.crypter.decrypt(
+                "task_hpke_keys",
+                &config_id.to_be_bytes(),
```
inahga, line 3974: This AAD subfield is normally the concatenation of `task_id | config_id`. My original iteration had this be `"janus_global_hpke_config" | config_id`, but I wasn't sure if that really had any benefit so I just went with `config_id` for simplicity. Open to feedback on this.

aggregator/src/bin/aggregator.rs
```diff
@@ -32,13 +32,28 @@ async fn main() -> Result<()> {
...
             .context("failed to parse response headers")?;
 
+        let mut cfg = ctx.config.aggregator_config();
+
+        if cfg.taskprov_config.enabled && cfg.taskprov_config.global_hpke_keypair.is_none() {
+            match datastore
+                .run_tx_with_name("get_global_hpke_config", |tx| {
+                    Box::pin(async move { tx.get_global_hpke_keypair().await })
+                })
+                .await?
```
inahga, line 42: I opted to just grab the config once at start-up time. Thus we have to restart the process if the key is changed. I figure this is a once-a-year kind of thing so we can be spared cache + cache invalidation logic.

aggregator_core/src/datastore.rs
divergentdave: This seems fine, as they are domain separated from per-task keys by the length of the AAD.

branlwyd (changes requested):

divergentdave:
I have a few high-level comments that may apply more to the whole taskprov feature, relevant to how we handle keypairs.

First off, if we advertise global HPKE configs in the `GET /hpke_config` response, then any client may request keys without specifying a task_id, see them, and then use them on non-taskprov tasks. Thus, rather than copy global keys into per-task key rows for every task, I think we should instead always consider both per-task and global keys when receiving reports or aggregation requests, filter both lists by HpkeConfigId, and then do trial decryption. Moreover, not copying keys will make rotation of the global keys easier. (and I think we wouldn't want to have previously created taskprov tasks use copies old keys -- those clients ought to follow HPKE config caching behavior just the same as non-taskprov clients)

I'd prefer to drop providing global HPKE keypairs via the configuration file, as we will likely only use the database route, to keep our config files free of secrets. Plus, related to the above, the keys will have impact on non-taskprov tasks, so nesting this within the taskprov configuration would be confusing.

db/00000000000012_global_hpke_config.up.sql
```diff
@@ -0,0 +1,6 @@
+-- A task_hpke_key with no task_id will be the global HPKE key.
```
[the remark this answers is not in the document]
inahga: My thought was just to keep the number of tables low and keep duplication low (new table would be the same minus task_id column). But as you said, tables are cheap. I can split it out.

db/00000000000012_global_hpke_config.up.sql
```diff
@@ -0,0 +1,6 @@
...
+ALTER TABLE task_hpke_keys ALTER COLUMN task_id DROP NOT NULL;
+
+-- Ensure only one task_hpke_key can have a null task_id, ensuring that there
+-- is only one global HPKE key.
+CREATE UNIQUE INDEX task_hpke_keys_task_id_only_one_null ON task_hpke_keys((task_id IS NULL)) WHERE task_id IS NULL;
```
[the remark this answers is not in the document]
inahga: I had made the incorrect conclusion that taskprov doesn't provide any indication of which HPKE is used for an initial report (so we must assume use of only one global config), but that's not correct as the report payload is encrypted and we can use trial decryption (as @divergentdave pointed out) to determine which key is in use. With the note above, we should be able to comfortably have multiple global keypairs.

aggregator/src/config.rs
```diff
@@ -96,6 +97,12 @@ pub struct TaskprovConfig {
     ///
     /// [spec]: https://datatracker.ietf.org/doc/draft-wang-ppm-dap-taskprov/
     pub enabled: bool,
+
+    /// The global HPKE config to advertise when no task ID is provided in the HPKE config request.
+    /// It's suggested to keep this configuration in the database instead of providing it in the
+    /// plaintext config.
+    #[serde(default)]
+    pub global_hpke_keypair: Option<HpkeKeypair>,
```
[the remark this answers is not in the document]
inahga: Hah, my initial iteration actually had this be a orthogonal feature, but I rolled it into taskprov thinking it was unlikely to be used elsewhere. I can get that back out of the reflog.

db/00000000000012_global_hpke_config.down.sql
```diff
@@ -0,0 +1,3 @@
+DELETE FROM task_hpke_keys WHERE task_id IS NULL;
+ALTER TABLE task_hpke_keys ALTER COLUMN task_id SET NOT NULL;
+DROP INDEX task_hpke_keys_task_id_only_one_null;
```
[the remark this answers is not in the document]
inahga: Oof, editor misconfiguration.

inahga:
> Thus, rather than copy global keys into per-task key rows for every task, I think we should instead always consider both per-task and global keys when receiving reports or aggregation requests, filter both lists by HpkeConfigId, and then do trial decryption.

My kneejerk is that this could get computationally expensive for the leader if multiple trial decryptions have to happen on each report. But I suppose simple heuristics could mitigate: if taskprov header is present, use global keys first, if not use task keys first.

> (and I think we wouldn't want to have previously created taskprov tasks use copies old keys -- those clients ought to follow HPKE config caching behavior just the same as non-taskprov clients)

I was under the faulty conclusion that clients will always use the same HPKE config for the task lifetime, but on a reread of the spec I see I'm mistaken. Your approach makes sense :+1:.

> I'd prefer to drop providing global HPKE keypairs via the configuration file, as we will likely only use the database route, to keep our config files free of secrets. Plus, related to the above, the keys will have impact on non-taskprov tasks, so nesting this within the taskprov configuration would be confusing.

:+1:

divergentdave:
We haven't built out key rotation for per-task HPKE keypairs yet, but it is definitely more important for global HPKE keypairs. Currently, our `GET /hpke_config` endpoint just returns the HPKE config with the largest ID. There's an open issue, #239, to track which keypair is the primary keypair that should be advertised. I think it would also make sense to record in the database whether non-primary keypairs are on the way in (recently generated, but not yet primary) or on the way out (used to be primary, now only used for decryption until they are destroyed).

inahga:
My thought was to have a `tombstoned_at` column which records the date the keypair was marked for deletion. My imagination is a janus operator comes by and does `janus_cli global-hpke-config expire <id>`, or this could even be done on a cronjob.

 Then `/hpke_config` sans task_id would return all non-tombstoned keys. That way, DAP clients refreshing their cache would pick up a new key. Decryption operations would ignore the tombstone column when querying for the HPKE config. We can have an async GC process that comes by and fully deletes tombstoned keys when they're well past the cache expiration time.

(I think this more or less follows what you suggested, with just less explicit semantics).

inahga (commented):
Feedback addressed. I've went with the approach I described above, and updated the PR description.

Note that I will address decryption of reports using global keys in a separate PR, to keep this one relatively small.

divergentdave (approved):
Not having a "not yet primary" state for keys is fine so long as our decryption code paths can hit the database for new keys. (which will be covered in a future PR)

aggregator_core/src/datastore.rs
```diff
@@ -3952,6 +3952,154 @@ impl<C: Clock> Transaction<'_, C> {
...
+        &self,
+        config_id: &HpkeConfigId,
+    ) -> Result<Option<HpkeKeypair>, Error> {
+        let stmt = self
+            .prepare_cached(
+                "SELECT config_id, config, private_key FROM global_hpke_keys WHERE config_id = $1;",
+            )
+            .await?;
+        let hpke_key_rows = self
+            .query(&stmt, &[&(Into::<u8>::into(*config_id) as i16)])
```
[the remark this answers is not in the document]
inahga: Much better, thanks! (also noticed this pattern elsewhere uses `transpose()`
```rust
        self.query_opt(&stmt, &[&(u8::from(*config_id) as i16)])
            .await?
            .map(|row| self.global_hpke_keypair_from_row(&row))
            .transpose()
```

inahga:
From discussion in https://isrg.slack.com/archives/C0167LT4C73/p1689810185430009 we did identify a need for a third not-yet-active state. I've updated the PR with the changes necessary for that.

The problem is that we want to cache global_hpke_keys so that we don't have to roundtrip on every upload. Maintaining coherency of this cache across aggregator replicas is challenging without the third state.

Now, global key rotation will go like so:
1. Keypairs for decryption are stored in a cache of `Hashmap<HpkeConfigId, HpkeKeypair>`. This cache is never invalidated during runtime.
1. Operator adds a new key with `janus_cli global_keys add`. The key is added in an inactive state. It is not yet advertised by /hpke_configs.
1. Operator restarts all aggregator replicas. This flushes any cached keypairs in the hashmap.
1. Operator activates the keypair with `janus_cli global_keys activate` and expires+demotes the old keypair with `janus_cli global_keys expire`.
1. `/hpke_config` caches will refresh and over time start to advertise the new key and retire the old one.

aggregator/src/aggregator.rs
```diff
@@ -221,24 +245,79 @@ impl<C: Clock> Aggregator<C> {
...
+        // Slow path: the cache has expired and needs to be refreshed.
+        let now = self.clock.now();
+        if now
+            > global_hpke_configs
+                .last_updated
+                .add(&GlobalHpkeConfigCache::REFRESH_INTERVAL)?
+        {
+            let configs = self
+                .datastore
+                .run_tx_with_name("refresh_global_hpke_configs", |tx| {
```
branlwyd, line 298: We should not run a DB operation while holding the lock; this will induce unnecessary/artificial latency between different readers of the HPKE configs.

Instead, I suggest:
* On creation of the `GlobalHpkeConfigCache`, start a "refresh" task which periodically refreshes the global HPKE configs (with the first refresh happening immediately).
* On dropping the `GlobalHpkeConfigCache`, stop the task.
* To retrieve the global HPKE configs, use one of the following strategies (see https://docs.rs/tokio/latest/tokio/sync/index.html):
  * an `mpsc::Sender<oneshot::Sender<Arc<Vec<HpkeConfig>>>>` to allow the reader to send a message to the refresh task, which it will respond to by sending a (clone of the `Arc` of) the current `HpkeConfig`s.
  * use a `Watch<Option<Arc<Vec<HpkeConfig>>>>`[1] to store the value -- this works great except that you must have an initial value, so we'd need to somehow block readers from trying to read before the initial HPKE config refresh is finished. One way to do this would be to use an `RwLock` which is initially write-locked; have the refresher task release the write-lock once a value is available in the `Watch`, then have the readers acquire & release a read lock before attempting to read the `Watch` value.

With this strategy, no reader will ever wait after the first time the refresh is completed.

[1] lol @ this type name

aggregator_core/src/datastore.rs
```diff
@@ -5905,6 +6041,42 @@ pub mod models {
 
         to_sql_checked!();
     }
+
+    #[derive(Debug, Clone, PartialEq, Eq)]
+    pub struct GlobalHpkeKeypair {
+        hpke_keypair: HpkeKeypair,
+        is_active: bool,
+        expired_at: Option<Time>,
```
branlwyd, line 6049: Q: why time-based expiry, instead of an enum with values like `Active`, `Primary`, `Expired`?

aggregator_core/src/datastore.rs
```diff
@@ -3952,6 +3952,141 @@ impl<C: Clock> Transaction<'_, C> {
...
+            row.get::<_, Option<NaiveDateTime>>("expired_at")
+                .as_ref()
+                .map(Time::from_naive_date_time),
+        ))
+    }
+
+    /// Unconditionally and fully drop a keypair. This is a dangerous operation,
+    /// since report shares encrypted with this key will no longer be decryptable.
+    #[tracing::instrument(skip(self), err)]
+    pub async fn delete_global_hpke_keypair(&self, config_id: &HpkeConfigId) -> Result<(), Error> {
```
branlwyd, line 4014: nit: it's totally fine to keep this since it's written & looks good, but in general it's a good idea to not implement functionality until you are ready to use it in application code.

aggregator_core/src/datastore.rs
inahga: We have to keep track of the expiry time either way to determine when the key can be fully dropped from the database. It's a bit of a tradeoff--the state is implicit instead of explicit, but in exchange we don't have to maintain coherency between an `expired_at` column and state enum.

It's fair enough to record `created_at` and an explicit state though. That actually might work a lot nicer when it comes to determining advertise-able keypairs.

aggregator/src/aggregator.rs
inahga: Would it be sufficient to instead drop the lock before querying the database and reacquire the lock before updating the cache? That way only one `/hpke_config` call will be blocked to update the cache.

This pattern we use already in our task cache https://github.com/divviup/janus/blob/main/aggregator/src/aggregator.rs#L428

inahga: Well wait, that won't quite work. An updating reader could be in the middle of a tx while another reader queries the cache again. That other reader will move into its tx too, thus duplicating database work.

Isn't this a problem too for the task cache, linked above?

aggregator_core/src/datastore.rs
inahga: Explicit state works out a lot nicer, simplifies the datastore API and tests. Great suggestion!

aggregator/src/aggregator.rs
branlwyd: The task cache has the same problem, but it's "lesser" since the task is only retrieved once (per process). The global HPKE config retrieval happens once an hour (per process), so the extra DB hits will recur once an hour.

Still, IMO your suggestion is likely an acceptable middle ground/first pass -- I really just want to avoid doing IO while holding a lock.

If you go this route, would you mind filing an issue to fix up both of these? I think the fix would be pretty small but concurrency code is always tricky.

inahga: > The task cache has the same problem, but it's "lesser" since the task is only retrieved once (per process). The global HPKE config retrieval happens once an hour (per process), so the extra DB hits will recur once an hour.

:+1:

> acceptable middle ground/first pass

I've put together a small isolated experiment to try various caching methods, and I think the two lock method I just suggested is the least effective. If we assume 1000 QPS and 100ms database roundtrip, then that's an additional 100 hits to the database. Those tasks also have wait for the full 100ms roundtrip each, compared to the single lock where blocked requests have less time to wait as the one tx approaches completion.

Here's that harness https://gist.github.com/inahga/3531c7341346b3224d99059cf80928b0, if interested or want to experiment.

I'll use this harness as well to try the two sync approaches you suggested. I _suspect_ that the `Watch` method won't be any better because it internally uses a RwLock on its value, which IIUC will block readers when a write lock is queued. Either way this is a good exercise since I've not delved into tokio's sync primitives very much yet.

---

The radical suggestion would also be to drop the cache and just load all advertiseable keys at runtime. I don't like this though because it means key rotation involves two application restarts since a key state change requires restart.

inahga: I'll also note that regardless of the caching method used, I believe now a background task for refreshing is strictly necessary. With the refresh on every `get`, the first request after the refresh interval will *always* be blocked to query the DB, so performance is suffering even in the absence of high QPS.

branlwyd: N.B. a `Watch`-based implementation (w/ a refresher task) would not be holding the lock during IO, so lock contention would be very cheap -- the lock would only be held long enough to clone the `Arc` inside (for readers), or to replace the `Arc` (for the writer task).

inahga: Going through the `Watch` docs again, this simpler approach probably should have occurred to me in the first place :laughing:. We could just have a background task that only locks the cache after it's queried the DB.

Refresh task:
```rust
        let refresh_task_handle = spawn(async move {
            loop {
                sleep(REFRESH_INTERVAL_DURATION).await;
                let new_values = datastore.get_values().await;

                {
                    let mut values = values_mutex.lock().await;
                    if new_values != *values {
                        *values = new_values;
                    }
                }
            }
        });
```

The cache itself is still `Mutex<Vec<HpkeConfig>>` (I'll also spend some time coaxing it into a `Mutex<Arc<Vec` to avoid extraneous allocations).

In my isolated experiment I was able to coax this up to 30k QPS before seeing even 1ms of latency due to lock contention, and cache updates don't block readers for the duration of the tx. As a bonus we can downgrade (upgrade?) to a `std::sync::Mutex` which tokio claims provides better performance, since we're not holding locks across awaits.

How does this approach look @branlwyd?

branlwyd: LGTM :) (I'm not sure we save much by checking if `new_values != *values`, but I don't object to it -- I suppose we might end up tracking fewer `Arc`s overall.)

inahga: Indeed. That has the regrettable side affect of turning a few constructors `async` (who says Rust doesn't have function coloring?).

This happens even if we want to use the RwLock method to block readers until the cache is filled. We have to create a write-locked mutex, which AFAICT requires the function to be async. We can't send a `std::sync::RwLock`'s guard across futures, and we have to `.await` to acquire a `tokio::sync::RwLock`.

inahga: Opened https://github.com/divviup/janus/issues/1639 to address the concurrency error we saw in `task_aggregator_for()`.

branlwyd (approved):
Nicely done!

PR #1611, opened 2023-07-14
Remove &Meter parameter from EphemeralDatastore::datastore().

branlwyd:
This was categorically set to &noop_meter(), so that value is now hard-wired. If we want to introduce tests for the meter behavior later, we can introduce a separate constructor for datastores, perhaps named something like `datastore_with_meter()`.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -2740,12 +2740,11 @@ mod tests {
         .build();
 
         let ephemeral_datastore = ephemeral_datastore().await;
-        let meter = noop_meter();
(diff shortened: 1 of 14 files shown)


branlwyd:
(In general, it's a good idea to consider whether a default might save a little complexity -- it's definitely true now, and seems overwhelmingly likely to remain true in the future, that most ephemeral datastores will use a no-op meter.)

inahga (approved):
This collides with the recent change https://github.com/divviup/janus/pull/157, so probably ought to get @divergentdave's feedback. I'm fine with it in principle though, as you said we can have a `datastore_with_meter` function to achieve tests involving metrics or metrics introspection.

divergentdave (approved):

PR #1619, opened 2023-07-18
In-aggregator GC.

branlwyd:
This moves GC from a separate binary to an in-aggregator process. The idea is that we'll set up one (or possibly a subset) of aggregator replicas to be responsible for garbage collection.

Stacked on #1603. Part of #1457.


aggregator/src/bin/aggregator.rs
```diff
@@ -63,6 +63,27 @@ async fn main() -> Result<()> {
...
+            let datastore = Arc::clone(&datastore);
+            async move {
+                if let Some(gc_config) = ctx.config.garbage_collection {
+                    let gc = GarbageCollector::new(
+                        datastore,
+                        gc_config.report_limit,
+                        gc_config.aggregation_limit,
+                        gc_config.collection_limit,
+                    );
+                    let mut interval = interval(Duration::from_secs(gc_config.gc_frequency_s));
```
inahga, line 76: Are we operating on the assumption that will continue to operate one aggregator replica?

If not, could consider introducing jitter to prevent GC from running at the same time across all replicas. Any kubectl rollout will have all replicas coming up at roughly the same time.

Optional, I can't think of any real problem at the moment with GC running at the same time.

aggregator/src/bin/aggregator.rs
```diff
@@ -250,6 +279,24 @@ struct Config {
     batch_aggregation_shard_count: u64,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+struct GarbageCollectorConfig {
```
inahga, line 283: Consider adding these parameters to docs/samples/advanced_config.

aggregator/src/bin/aggregator.rs
```diff
@@ -250,6 +279,24 @@ struct Config {
...
+    /// of the garbage collector.
+    report_limit: u64,
+
+    /// The limit to the number of aggregation jobs, and related aggregation artifacts, deleted for
+    /// a single task by a single run of the garbage collector.
+    aggregation_limit: u64,
+
+    /// The limit to the number of batches, and related collection artifacts, deleted for a single
+    /// task by a single run of the garbage collector.
+    collection_limit: u64,
```
inahga, line 297: What I find most helpful with config flags (besides _what_ they do), is _why, or what factors should I consider before changing them. If that's open ended for now, that's fine, just wondering if you had an explicit thought in mind for how these should be derived.

aggregator/src/bin/aggregator.rs
```diff
@@ -221,9 +246,13 @@ pub enum AggregatorApi {
 struct Config {
     #[serde(flatten)]
     common_config: CommonConfig,
+
     #[serde(default)]
     taskprov_config: TaskprovConfig,
 
+    #[serde(default)]
```
inahga, line 253: I _think_ these are defaulting to zero, insofar as I don't see where a non-zero default is specified in this PR. Is this intentional? If so it's a clever way of no-oping the GC process unless it's opted-into, but want to make sure that's the intent.

inahga (approved):
Nice simplification, combining the aggregator process with the aggregator is definitely the right call. I have only questions or non-critical suggestions.

aggregator/src/bin/aggregator.rs
branlwyd: It's pretty open-ended, but: "large enough to keep up with the rate of creation, and not much larger than that." If we go below rate of creation we'll fall behind (of course), but at the same time, going too large will create larger-than-necessary transactions.

aggregator/src/bin/aggregator.rs
branlwyd: This actually defaults to `None` (since this is an `Option`), but you're correct about the overall behavior: unless specified, GC is disabled.

aggregator/src/bin/aggregator.rs
branlwyd: We will almost surely run more than one replica; I am hoping to use configuration to run GC on only one replica. I _think_ we can use a `StatefulSet` for this (since that gives access to an "index", which allows us to control behavior on a replica-by-replica basis -- see e.g. https://itnext.io/kubernetes-statefulset-initialization-with-unique-configs-per-pod-7e02c01ada65). If that's too complicated or otherwise infeasible, we can always create a "leader" deployment with one replica, and a "followers" deployment with the remaining replicas.

In the long run, we might want to rethink this strategy (what if the one GC replica goes down?), but I think this is a pretty good starting point.

PR #1625, opened 2023-07-20
[Experimental] Remove time-interval query type.

branlwyd:


tgeoghegan:
The experiment this supported was an idea I pitched for DAP back at IETF 116 or so. We didn't go that route, so the experiment is concluded and we can close this PR.

PR #1640, opened 2023-07-24
Allow decrypting with global key on report upload

inahga:
Follow up to https://github.com/divviup/janus/pull/1610, supports https://github.com/divviup/janus/issues/1568.

All code paths involving report decryption should now be able to use a global HPKE keypair to attempt decryption of a report. If  a config ID is given that matches both a task and global key, trial decryption is used.

The keypair is cached using the same mechanism in https://github.com/divviup/janus/pull/1610. Importantly, this has cache invalidation. Because we have the `Pending` state, we can safely refresh keys without worrying about them being advertised before all aggregators have picked up on the new key. This means that the key rotation strategy is:
1. Add a new key to DB
2. Wait ~30 minutes (plus some safety factor) for all aggregators to pick up on the new key OR restart the application. Because the key isn't being advertised, we don't have to worry about reports encrypted with a key that an aggregator hasn't become aware of.
3. Set new key to the active state.
4. Wait ~30 minutes for all aggregators to have this new key be active before retiring any old keys.

I've renamed the cache to `GlobalHpkeKeypairCache` to denote that it does store the private key in memory (`Config` could imply that it doesn't).

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1206,22 +1206,6 @@ impl VdafOps {
             }
         }
 
-        // Verify that the report's HPKE config ID is known.
-        // https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.3.2
-        let global_hpke_keypair =
-            global_hpke_keypairs.keypair(leader_encrypted_input_share.config_id());
-        let task_hpke_keypair = task
-            .hpke_keys()
-            .get(leader_encrypted_input_share.config_id());
-
-        // No task key is present, so see if this report was encrypted with a global key.
-        if task_hpke_keypair.is_none() && global_hpke_keypair.is_none() {
-            return Err(Arc::new(Error::OutdatedHpkeConfig(
-                *task.id(),
-                *leader_encrypted_input_share.config_id(),
-            )));
-        }
-
         // Decode (and in the case of the leader input share, decrypt) the remaining fields of the
         // report before storing them in the datastore. The spec does not require the /upload
         // handler to do this, but it exercises HPKE decryption, saves us the trouble of storing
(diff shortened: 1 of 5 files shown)


aggregator/src/aggregator.rs
```diff
@@ -1400,24 +1487,15 @@ impl VdafOps {
...
+
+            let global_hpke_keypair =
+                global_hpke_keypairs.keypair(report_share.encrypted_input_share().config_id());
 
             // If decryption fails, then the aggregator MUST fail with error `hpke-decrypt-error`. (§4.4.2.2)
-            let plaintext = hpke_keypair.and_then(|hpke_keypair| {
+            let try_hpke_open = |hpke_keypair: &HpkeKeypair| {
                 hpke::open(
                     hpke_keypair.config(),
                     hpke_keypair.private_key(),
```
inahga, line 1501: This isn't quite right, it'll log to info on every decryption trial failure. Let me move to draft and fix that.

aggregator/src/aggregator.rs
```diff
@@ -195,44 +196,61 @@ impl Default for Config {
...
-    /// Cache of HPKE configs. We use a [`std::sync::Mutex`] since we won't hold
-    /// locks across `.await`s and it is lighter weight than [`tokio::sync::Mutex`].
-    configs: Arc<StdMutex<Arc<Vec<HpkeConfig>>>>,
+pub struct GlobalHpkeKeypairCache {
+    // We use a std::sync::Mutex in this cache because we won't hold locks across
+    // `.await` boundaries. StdMutex is lighter weight than `tokio::sync::Mutex`.
+    /// Cache of HPKE configs for advertisement.
+    configs: Arc<StdMutex<HpkeConfigs>>,
+
+    // Cache of HPKE keypairs for report decryption.
```
tgeoghegan, line 215:
```suggestion
    /// Cache of HPKE keypairs for report decryption.
```

aggregator/src/aggregator.rs
```diff
@@ -1179,19 +1226,52 @@ impl VdafOps {
...
+                &InputShareAad::new(
+                    *task.id(),
+                    report.metadata().clone(),
+                    report.public_share().to_vec(),
+                )
+                .get_encoded(),
+            ) {
+                Ok(results) => Ok(results),
+                Err(error) => Err(error.into()),
+            }
```
tgeoghegan, line 1244: I think you can do `Result::map_err` instead of writing out this `match`.

aggregator/src/aggregator.rs
```diff
@@ -1179,19 +1226,52 @@ impl VdafOps {
...
+                    *task.id(),
+                    *leader_encrypted_input_share.config_id(),
+                )));
+            }
+            (None, Some(global_hpke_keypair)) => try_hpke_open(&global_hpke_keypair),
+            (Some(task_hpke_keypair), None) => try_hpke_open(task_hpke_keypair),
+            (Some(task_hpke_keypair), Some(global_hpke_keypair)) => {
+                match try_hpke_open(task_hpke_keypair) {
+                    Ok(plaintext) => Ok(plaintext),
+                    Err(_) => try_hpke_open(&global_hpke_keypair),
```
tgeoghegan, line 1268: I think we should only fall back to the global HPKE keypair if the error from `try_hpke_open` is `hpke::Error::Hpke`, because otherwise we might suppress evidence of some other problem.

aggregator/src/aggregator.rs
```diff
@@ -1430,6 +1502,35 @@ impl VdafOps {
...
+
+            let plaintext = check_keypairs.and_then(|_| {
+                match (task_hpke_keypair, global_hpke_keypair) {
+                    (None, None) => unreachable!("already checked this condition"),
+                    (None, Some(global_hpke_keypair)) => try_hpke_open(&global_hpke_keypair),
+                    (Some(task_hpke_keypair), None) => try_hpke_open(task_hpke_keypair),
+                    (Some(task_hpke_keypair), Some(global_hpke_keypair)) => {
+                        match try_hpke_open(task_hpke_keypair) {
+                            Ok(result) => Ok(result),
+                            Err(_) => try_hpke_open(&global_hpke_keypair),
```
tgeoghegan, line 1530: Similar to above, let's check what the error is before we fall back to the global key.

aggregator/src/aggregator.rs
```diff
@@ -1430,6 +1502,35 @@ impl VdafOps {
...
+                    &[KeyValue::new("type", "unknown_hpke_config_id")],
+                );
+                Err(ReportShareError::HpkeUnknownConfigId)
+            } else {
+                Ok(())
+            };
+
+            let plaintext = check_keypairs.and_then(|_| {
+                match (task_hpke_keypair, global_hpke_keypair) {
+                    (None, None) => unreachable!("already checked this condition"),
```
tgeoghegan, line 1524: This is awkward. Why not move the block on lines 1508-1517 into this match arm?

aggregator/src/aggregator.rs
```diff
@@ -3182,6 +3292,80 @@ mod tests {
...
+                    tx.put_global_hpke_keypair(&global_hpke_keypair_different_id)
+                        .await?;
+                    Ok(())
+                })
+            })
+            .await
+            .unwrap();
+
+        // Let keypair cache refresh.
+        sleep(StdDuration::from_millis(750)).await;
```
tgeoghegan, line 3338: We should avoid this kind of thing in tests, as they tend to be brittle and increase test runtime. Could you instead expose some test method on the keypair cache or on the aggregator struct that lets you force the cache refresh?

tgeoghegan (changes requested):

aggregator/src/aggregator.rs
inahga: The two code paths (`None,None` and the rest) have different errors and behaviors on error. Match arms must return the same type. We _could_ force them to match... but for decryption, we can't directly return `ReportShareError::HpkeDecryptError` because it conceals error context that we need to trace AND we also want to only trace once (not on each failure of trial decryption). I also hoped to avoid copying the tracing calls around, as it's fairly verbose.

We don't have this problem with 1509-1517 because the control flow is simpler, we can just directly return from the function.

I played around with it for a while and this was as concise as I could get, open to other ideas.

aggregator/src/aggregator.rs
inahga: Agreed, it does occur to me that if the build machine is overloaded (which is likely, they're fairly weak), the cache may not refresh in time.

Can I make the change in a separate PR? I'll have to modify existing tests.

aggregator/src/aggregator.rs
inahga: I'm surprised this isn't a clippy lint!

aggregator/src/aggregator.rs
inahga: Here's how that'll look roughly https://github.com/divviup/janus/commit/655cc8a0736b293e706e953c019b78e5c8b0e438 (I should have done this in the first place!)

tgeoghegan: Sure, we can take https://github.com/divviup/janus/commit/655cc8a0736b293e706e953c019b78e5c8b0e438 separately so long as we don't let it linger.

aggregator/src/cache.rs
tgeoghegan, line 1: nit: it'd be nice to have a module-level doccomment, perhaps explaining that this module is intended to contain various kinds of caches used in Janus.

tgeoghegan (approved):

PR #1644, opened 2023-07-25
Fix bug with trailing slashes in URL

inahga:
In https://github.com/divviup/janus/pull/1579#discussion_r1261499802 I identified cases where `https://example.com` would be encoded as `https://example.com/` (notice the trailing slash). At the time I didn't think this was relevant, but it turns out to affect task ID derivation.

Task ID derivation is `sha256sum(encoded_task_config)`. If we are peered with a non-janus aggregator and if we introduce trailing slashes, Janus could derive a different task ID than its peer. (This is actually somewhat unlikely since we'd derive task ID directly from the encoded bytes, but is a subtle inconsistency that could hit us later).

Instead, provide `TryInto<url::Url>` to ease converting DAP Urls to a more useful type.

### DIFF
--- a/messages/src/lib.rs
+++ b/messages/src/lib.rs
@@ -45,7 +45,7 @@ pub enum Error {
 
 /// Wire-representation of an ASCII-encoded URL with minimum length 1 and maximum
 /// length 2^16 - 1.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, PartialEq, Eq)]
 pub struct Url(Vec<u8>);
 
 impl Url {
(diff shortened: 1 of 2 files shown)


messages/src/lib.rs
```diff
@@ -86,7 +70,9 @@ impl Decode for Url {
 
 impl Display for Url {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        (&self.0 as &dyn Display).fmt(f)
+        // This isn't actually lossy, since this type cannot be constructed from
+        // non-ASCII bytes.
+        write!(f, "{}", String::from_utf8_lossy(&self.0))
```
tgeoghegan, line 75: Since this method returns `std::fmt::Result` anyway, could we use `String::from_utf8` and map errors `std::fmt::Error`?

tgeoghegan (approved):

messages/src/lib.rs
tgeoghegan: [`std::fmt::Result`](https://doc.rust-lang.org/std/fmt/type.Result.html) is just `pub type Result = Result<(), Error>;` and `std::fmt::Error` in turn is just `pub struct Error` so I think `.map_err(|_| std::fmt::Error)` should work. If not, then this is fine.

tgeoghegan (approved):

messages/src/lib.rs
inahga: Aha, that works. I was trying to make use of the `|_|`.

PR #1646, opened 2023-07-26
add GET {aggregator_api}/ → AggregatorApiConfig

jbr:
Pairs with https://github.com/divviup/divviup-api/pull/340

Explanation of changes:

* Previously the aggregator-api handler was unconditionally built and discarded, but as of this PR, we require the aggregator-api config in order to build the aggregator-api handler. This seems more generally correct and adaptable to other configuration keys that are mandatory if and only if aggregator-api is enabled
* Previously, the BinaryContext was partially moved in order to move the non-clone datastore into an Arc, but that did not work well with extracting behvior to a function. To address this, we now destructure the BinaryContext and explicitly take or borrow fields from it as needed.
* `Box::pin(async {})` was replaced with `Box::pin(std::future::ready(()))`
* This PR puts Config back into the Conn in auth_check in order to use it in the get_config endpoint

### DIFF
--- a/aggregator/src/bin/aggregator.rs
+++ b/aggregator/src/bin/aggregator.rs
@@ -4,71 +4,63 @@ use clap::Parser;
 use janus_aggregator::{
     aggregator::{self, garbage_collector::GarbageCollector, http_handlers::aggregator_handler},
     binary_utils::{
-        janus_main, setup_server, setup_signal_handler, BinaryOptions, CommonBinaryOptions,
+        janus_main, setup_server, setup_signal_handler, BinaryContext, BinaryOptions,
+        CommonBinaryOptions,
     },
     cache::GlobalHpkeKeypairCache,
     config::{BinaryConfig, CommonConfig, TaskprovConfig},
 };
 use janus_aggregator_api::{self, aggregator_api_handler};
-use janus_aggregator_core::SecretBytes;
+use janus_aggregator_core::{datastore::Datastore, SecretBytes};
 use janus_core::time::RealClock;
 use serde::{Deserialize, Serialize};
-use std::{future::Future, pin::Pin};
+use std::{
+    future::{ready, Future},
+    pin::Pin,
+};
 use std::{iter::Iterator, net::SocketAddr, sync::Arc, time::Duration};
 use tokio::{join, time::interval};
 use tracing::{error, info};
-use trillium::Headers;
+use trillium::{Handler, Headers};
 use trillium_router::router;
 use trillium_tokio::Stopper;
+use url::Url;
 
 #[tokio::main]
 async fn main() -> Result<()> {
     janus_main::<_, Options, Config, _, _>(RealClock::default(), |ctx| async move {
-        let datastore = Arc::new(ctx.datastore);
+        let BinaryContext {
+            clock,
+            options,
+            mut config,
+            datastore,
+            meter,
+        } = ctx;
+
+        let datastore = Arc::new(datastore);
         let stopper = Stopper::new();
         setup_signal_handler(stopper.clone())
             .context("failed to register SIGTERM signal handler")?;
-        let response_headers = ctx
-            .config
+        let response_headers = config
             .response_headers()
             .context("failed to parse response headers")?;
 
         let mut handlers = (
             aggregator_handler(
                 Arc::clone(&datastore),
-                ctx.clock,
-                &ctx.meter,
-                ctx.config.aggregator_config(),
+                clock,
+                &meter,
+                config.aggregator_config(),
             )
             .await?,
             None,
         );
 
-        let aggregator_api_auth_tokens = ctx
-            .options
-            .aggregator_api_auth_tokens
-            .iter()
-            .filter(|token| !token.is_empty())
-            .map(|token| {
-                let token_bytes = STANDARD
-                    .decode(token)
-                    .context("couldn't base64-decode aggregator API auth token")?;
-
-                Ok(SecretBytes::new(token_bytes))
-            })
-            .collect::<Result<Vec<_>>>()?;
-
-        let inner_aggregator_api_handler = aggregator_api_handler(
-            Arc::clone(&datastore),
-            janus_aggregator_api::Config {
-                auth_tokens: aggregator_api_auth_tokens,
-            },
-        );
-
         let garbage_collector_future = {
             let datastore = Arc::clone(&datastore);
+            let gc_config = config.garbage_collection.take();
             async move {
-                if let Some(gc_config) = ctx.config.garbage_collection {
+                if let Some(gc_config) = gc_config {
                     let gc = GarbageCollector::new(
                         datastore,
                         gc_config.report_limit,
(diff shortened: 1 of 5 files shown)


aggregator_core/src/lib.rs
```diff
@@ -69,6 +69,7 @@ impl<H: Handler> InstrumentedHandler<H> {
     }
 
     async fn before_send(&self, conn: Conn) -> Conn {
+        let conn = self.0.before_send(conn).await;
```
jbr, line 72: Drawing attention to this: Without this line, any before_send implementations inside of this handler would not be called. If this handler always wraps simple Fn handlers, the behavior is identical, but at some point in the future omitting any wrapped before_send calls may be confusing.

aggregator_api/src/lib.rs
```diff
@@ -409,6 +421,22 @@ mod models {
...
+    #[allow(dead_code)] // allowed in order to fully describe the interface and for later use
+    #[derive(Serialize, PartialEq, Eq, Debug)]
+    pub(crate) enum AggregatorRole {
+        Either,
+        Leader,
+        Helper,
+    }
+
+    #[derive(Serialize, PartialEq, Eq, Debug)]
+    pub(crate) struct AggregatorApiConfig {
```
inahga, line 433: Pedantic, take it or leave it: This is a slight misnomer, since it's not describing the aggregator API. I'd also argue it's not describing a configuration either--I would suggest `AggregatorCapabilities`.

aggregator_api/src/lib.rs
```diff
@@ -113,13 +115,23 @@ async fn auth_check(conn: &mut Conn, State(cfg): State<Arc<Config>>) -> impl Han
...
         // Authorization succeeds.
+        conn.set_state(cfg);
         return None;
     }
 
     // Authorization fails.
     Some((Status::Unauthorized, Halt))
 }
 
+async fn get_config(_: &mut Conn, State(config): State<Arc<Config>>) -> Json<AggregatorApiConfig> {
```
inahga, line 126: Similarly, `get_capabilities` or `get_aggregator_capabilities`.

aggregator_api/src/lib.rs
```diff
@@ -113,13 +115,23 @@ async fn auth_check(conn: &mut Conn, State(cfg): State<Arc<Config>>) -> impl Han
...
+    Json(AggregatorApiConfig {
+        dap_url: config.public_dap_url.clone(),
```
inahga, line 128: Remind me, I think I missed context on this choice: why do we need the aggregator to supply its own URL? The downstream client can't infer this based on how the aggregator is being reached?

aggregator_api/src/lib.rs
```diff
@@ -113,13 +115,23 @@ async fn auth_check(conn: &mut Conn, State(cfg): State<Arc<Config>>) -> impl Han
...
+        role: AggregatorRole::Either,
+        vdafs: vec![1, 2, 3],
+        query_types: vec![1, 2],
```
inahga, line 131: I believe we can reuse `messages::query_type::Code` in `messages/src/lib.rs` for this (somewhat unfortunately named):
```
    /// DAP protocol message representing the type of a query.
    #[derive(Copy, Clone, Debug, PartialEq, Eq, TryFromPrimitive, Serialize, Deserialize)]
    #[repr(u8)]
    pub enum Code {
        Reserved = 0,
        TimeInterval = 1,
        FixedSize = 2,
    }
```

Or I would define a model in `mod models`.

aggregator_api/src/lib.rs
```diff
@@ -113,13 +115,23 @@ async fn auth_check(conn: &mut Conn, State(cfg): State<Arc<Config>>) -> impl Han
...
 
     // Authorization fails.
     Some((Status::Unauthorized, Halt))
 }
 
+async fn get_config(_: &mut Conn, State(config): State<Arc<Config>>) -> Json<AggregatorApiConfig> {
+    Json(AggregatorApiConfig {
+        dap_url: config.public_dap_url.clone(),
+        role: AggregatorRole::Either,
+        vdafs: vec![1, 2, 3],
```
inahga, line 130: There's `VdafInstance` in `core/src/task.rs` for this, unfortunately you'd have to construct the enum instances to get its discriminant (likewise for `taskprov::VdafType`)

If that's too ugly, I'd still prefer we define a model (similar to AggregatorRole) to assign these numbers some meaning.

inahga (commented):
Nice job overall, some minor comments

aggregator_api/src/lib.rs
jbr: Correct; the aggregator api may be hosted at a different domain from the dap server or the same domain at a different path. There's no way for divviup-api to know how the dap server and aggregator-api are configured in relation to each other

aggregator_api/src/lib.rs
jbr: I started out calling it capabilities but decided its "how the aggregator api is configured" from the perspective of the api consumer (divviup-api). Specifically, the public dap url isn't really a capability, but it is a configuration. Other properties of this struct are both capabilities and also configuration

aggregator_api/src/lib.rs
jbr: addressed in 05026e5e02500027dfc843bc0710789b4c8d595a

aggregator_api/src/lib.rs
jbr: Addressed in 05026e5e02500027dfc843bc0710789b4c8d595a

aggregator_api/src/lib.rs
```diff
@@ -405,10 +419,34 @@ mod models {
...
+    pub(crate) struct AggregatorApiConfig {
+        pub dap_url: Url,
+        pub role: AggregatorRole,
+        pub vdafs: Vec<VdafId>,
+        pub query_types: Vec<QueryTypeId>,
+    }
+
+    #[derive(Debug, Clone, Copy, Eq)]
+    #[repr(u8)]
+    pub enum VdafId {
```
inahga, line 444:
```suggestion
    pub(crate) enum VdafId {
```
Consistent with the other models.

aggregator_api/src/lib.rs
```diff
@@ -15,8 +15,10 @@ use janus_core::{
     task::{AuthenticationToken, DapAuthToken},
     time::Clock,
 };
-use janus_messages::{Duration, HpkeAeadId, HpkeKdfId, HpkeKemId, Role, TaskId};
-use models::{GetTaskMetricsResp, TaskResp};
+use janus_messages::{
+    query_type::Code as QueryTypeId, Duration, HpkeAeadId, HpkeKdfId, HpkeKemId, Role, TaskId,
```
inahga, line 19: :+1:

inahga (approved):

aggregator_api/src/lib.rs
```diff
@@ -113,13 +117,23 @@ async fn auth_check(conn: &mut Conn, State(cfg): State<Arc<Config>>) -> impl Han
         constant_time::verify_slices_are_equal(bearer_token.as_ref(), key.as_ref()).is_ok()
     }) {
         // Authorization succeeds.
+        conn.set_state(cfg);
```
divergentdave, line 120: Since it took me a bit to figure out -- I'd appreciate a comment explaining `FromConn for State<T>` takes the config out of the connection state, and this line is putting it back for use by a later handler.

Alternately, we could just access it via `conn.state:<...>()`, and avoid shuffling it around.

aggregator_api/src/lib.rs
```diff
@@ -405,10 +419,37 @@ mod models {
...
+
+    #[derive(Serialize, PartialEq, Eq, Debug)]
+    pub(crate) struct AggregatorApiConfig {
+        pub dap_url: Url,
+        pub role: AggregatorRole,
+        pub vdafs: Vec<VdafId>,
+        pub query_types: Vec<QueryTypeId>,
+    }
+
+    #[allow(clippy::enum-variant-names)] //allowed because it just happens to be the case that all of the supported vdafs are prio3
```
divergentdave, line 444: nit: this needs to be in identifier form
```suggestion
    #[allow(clippy::enum_variant_names)] //allowed because it just happens to be the case that all of the supported vdafs are prio3
```

divergentdave (approved):

aggregator/src/bin/aggregator.rs
```diff
@@ -145,6 +143,39 @@ async fn main() -> Result<()> {
...
 }
 
+fn build_aggregator_api_handler<'a>(
+    options: &Options,
+    config: &'a Config,
+    datastore: &Arc<Datastore<RealClock>>,
+) -> anyhow::Result<Option<(impl Handler, &'a AggregatorApi)>> {
+    let Some(aggregator_api) = &config.aggregator_api else {
+        return Ok(None);
+    };
```
tgeoghegan, line 153: Does this work on Rust 1.65, currently Janus' MSRV? I don't think bumping MSRV is a dealbreaker, FWIW.

aggregator/src/bin/aggregator.rs
```diff
@@ -185,14 +216,35 @@ pub struct HeaderEntry {
...
+    },
+    PathPrefix {
+        /// The Janus aggregator API will be served on the same address as the DAP API, but relative to
+        /// the provided prefix. e.g., if `path_prefix` is `aggregator-api`, then the DAP API's uploads
+        /// endpoint would be `{listen-address}/tasks/{task-id}/reports`, while task IDs could be
+        /// obtained from the aggregator API at `{listen-address}/aggregator-api/task_ids`.
+        path_prefix: String,
+        /// Resource location at which the DAP service managed by this
+        /// aggregator api can be found on the public
+        /// internet. Required.
```
tgeoghegan, line 236: nit: wrap these comments to the same width as everything else

aggregator_api/src/lib.rs
```diff
@@ -113,13 +117,23 @@ async fn auth_check(conn: &mut Conn, State(cfg): State<Arc<Config>>) -> impl Han
...
 
     // Authorization fails.
     Some((Status::Unauthorized, Halt))
 }
 
+async fn get_config(_: &mut Conn, State(config): State<Arc<Config>>) -> Json<AggregatorApiConfig> {
+    Json(AggregatorApiConfig {
+        dap_url: config.public_dap_url.clone(),
+        role: AggregatorRole::Either,
+        vdafs: vec![VdafId::Prio3Count, VdafId::Prio3Sum, VdafId::Prio3Histogram],
```
tgeoghegan, line 132: I don't think this is right:

- it's missing `Prio3CountVec` and `Prio3SumVec`
- it needs to include `[Prio3FixedPoint16BitBoundedL2VecSum, Prio3FixedPoint32BitBoundedL2VecSum, Prio3FixedPoint64BitBoundedL2VecSum]` when the `fpvec_bounded_l2` feature is enabled

You're going to point out that several of those VDAFs don't have assigned IDs in any protocol draft, which is a good point. The point of this PR is to set up the plumbing for advertising capabilities, so I filed #1654.

aggregator_api/src/lib.rs
tgeoghegan, line 444: Wrap long lines

```suggestion
    // allowed because it just happens to be the case that all of the supported vdafs are prio3
    #[allow(clippy::enum-variant-names)] 
```

tgeoghegan (changes requested):
Needs a rebase and some compilation error fixes but looks sound

aggregator_api/src/lib.rs
jbr: Tbh I was surprised by this as well. I think it's worth changing how FromConn for State works, and require T be Clone

aggregator/src/bin/aggregator.rs
jbr: https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html stable since 1.65

aggregator/src/bin/aggregator.rs
jbr: Previously the comments weren't consistently wrapped to 100 so I wasn't sure what to set my editor to. I rewrapped everything to 100 in rebase.

aggregator_api/src/lib.rs
jbr: several of those VDAFs don't have assigned IDs in any protocol draft 😁

Cargo.lock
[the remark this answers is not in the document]
jbr: I was having build issues that a cargo update resolved, but will stash those changes and see if it was transient

aggregator_core/src/lib.rs
```diff
@@ -70,15 +70,18 @@ impl<H: Handler> InstrumentedHandler<H> {
 
     async fn before_send(&self, conn: Conn) -> Conn {
         if let Some(span) = conn.state::<InstrumentedHandlerSpan>() {
+            let conn = self.0.before_send(conn).instrument(span.0).await;
```
divergentdave, line 73: Apologies, this needs a clone
```suggestion
            let conn = self.0.before_send(conn).instrument(span.0.clone()).await;
```

tgeoghegan (approved):

PR #1652, opened 2023-07-28
Track what process created a task

inahga:

Introduces an enum `TaskCreator` which tracks possible processes that can create a task. Fixes janus_cli and the aggregator API to properly record themselves in this new column.

This is required so that we can correctly join taskprov-specific collector and aggregator auth tokens with a task. The taskprov-specific auth tokens will be in separate tables `taskprov_aggregator_auth_tokens` and `taskprov_collector_auth_tokens`. If a task was created by taskprov, we will query for taskprov auth tokens (using the peer aggregator URL to join), and concatenate the resulting auth tokens with the normal `*_auth_tokens` vector.

Alternative approaches considered:
- Unconditionally join all tasks against `taskprov_*_auth_tokens`. I avoided this because it creates a footgun where a non-taskprov task is provisioned with a taskprov peer aggregator, the task could have the wrong auth tokens joined to it.
- Have just a `created_by_taskprov BOOL` column. We can, but I figured that since we have multiple means of provisioning tasks, we might as well track them all. YAGNI might apply, I'm open to feedback on this.


inahga:
I think this is not required anymore. It turns out we use the auth tokens in relatively few places. In those places, we can identify if we're running in a taskprov context, and fetch the appropriate auth tokens from the `taskprov_*_auth_tokens` table.

e.g.
```rust
        if taskprov_header.is_some() {
            fetch taskprov keys from db
            match auth tokens against taskprov keys
        } else {
            match auth tokens against task keys
        }
```

It may still some day be useful to track how tasks are created, but at the moment it's not required.

PR #1653, opened 2023-07-28
remove apparently-unused async-std dev dep from aggregator_core

jbr:
Async-std is a heavy dependency and `rg async_std` turns up no results from janus root


divergentdave (approved):
The `#[rstest::rstest]` macro may emit `#[async_std::test]` on async tests, but we currently avoid this by using `#[tokio::test]` on the function first. (see [here](https://github.com/la10736/rstest/blob/master/README.md#inject-test-attribute))

tgeoghegan (approved):

PR #1656, opened 2023-07-28
Split datastore.rs into models.rs and tests.rs

inahga:
Prevent this file from attaining critical mass.

To get this to merge and retain git history, we will have to temporarily allow merge or rebase commits on this repository. Squashing the commits will cause the history to be lost.

The drawback to keeping the distinct commits is that this harms use of `git bisect`, these commits would have to be `git bisect skip`'d.


tgeoghegan (approved):

jbr:
External opinion: As long as you're moving stuff around, why not break the models file into a file per model and reexport from the models module? One file per type makes it really easy to jump to exactly what you're looking for and keep type definitions, impls, and small associated types together

inahga:
I'll have to defer to the rest of the team @divviup/committers for an opinion on that, since they have spent far more time in the codebase than I. Then maybe we can carry that forward into the style guide https://github.com/divviup/janus/blob/main/docs/DEVELOPMENT.md.

I myself get along decently with go-to-definition and go-to-reference, but that workflow has reached its limits at this large of a file, hence my immediate urge to break it down.

tgeoghegan:
While `models.rs` is rather large (1,700+ lines), I think mandating one file per struct is an overcorrection in the other direction. I also think we should try to avoid imposing format or layout rules on a project unless they can be enforced by automation, because policing conformance with such rules is a poor use of engineer time. I say we go ahead with `models.rs` as it is.

Issue #1669, opened 2023-08-02
Implement `draft-ietf-ppm-dap-07`


tgeoghegan:
[DAP-05](https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/05/) is out and Janus should implement it. We need:

- [x] update to new libprio-rs that does VDAF-06 (#1673)
- [x] update to new libprio-rs that does VDAF-07 (https://github.com/divviup/libprio-rs/issues/701; #1811)
- [x] ping pong topology https://github.com/divviup/janus/pull/1811
- [x] deal with error type changes https://github.com/divviup/janus/pull/1853
- [x] verify that we use appropriate media types when serving HPKE configs, aggregate share requests
- [x] HPKE domain separation strings
- [x] update versions table in README.md (https://github.com/divviup/janus/pull/1676)
- [x] Branch for DAP-04 (https://github.com/divviup/janus/pull/1676)
- [x] Daphne is skipping ahead to draft-ietf-dap-09 <s>re-enable Daphne test once they have a DAP-06 implementation</s>
- [x] re-enable divviup-ts tests once it has a DAP-07 implementation (https://github.com/divviup/janus/pull/2388)
- [x] re-enable Prio3Histogram in-cluster test once divviup-api can deal with DAP-05 (https://github.com/divviup/divviup-api/issues/410, #1743)

[Draft diff](https://author-tools.ietf.org/iddiff?url1=draft-ietf-ppm-dap-04&url2=draft-ietf-ppm-dap-06&difftype=--html)

tgeoghegan:
[DAP-06](https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/06/) is out with a more clear and implementable ping-pong spec, so we'll jump ahead to implementing that.

tgeoghegan:
The wrong commit got tagged as DAP-06, leaving it with the wrong HPKE domain separation strings and an outdated reference to DAF. So we published `draft-ietf-ppm-dap-07` which fixes those issues, and that's what we're going to implement in Janus.

PR #1673, opened 2023-08-04
Adopt `prio` 0.14.0

tgeoghegan:
The bulk of the changes here deal with the change to the representation of `Prio3Histogram`. Since `prio` 0.14.x implements VDAF-06, taking this change will break compatibility with DAP-04.

Part of #1669


tgeoghegan:
I'm pretty sure this will also break `divviup-api`, which will have to update its `Prio3Histogram` representation (or will have to adapt how it deals with `Prio3Histogram` based on what version of Janus the aggregator runs).

aggregator_core/src/datastore/tests.rs
```diff
@@ -107,18 +107,8 @@ async fn roundtrip_task(ephemeral_datastore: EphemeralDatastore) {
...
-            Role::Leader,
-        ),
-        (
-            VdafInstance::Prio3Histogram {
-                buckets: Vec::from([0, 25, 50, 75, 100]),
-            },
-            Role::Leader,
-        ),
+        (VdafInstance::Prio3Histogram { buckets: 4 }, Role::Leader),
+        (VdafInstance::Prio3Histogram { buckets: 5 }, Role::Leader),
```
divergentdave, line 111: Technically the direct translation of these parameters would be 5 and 6 buckets, respectively, but it doesn't really matter for the purposes of checking serialization/deserialization.

core/src/task.rs
```diff
@@ -30,10 +30,7 @@ pub enum VdafInstance {
     /// A vector of `Prio3` sums.
     Prio3SumVec { bits: usize, length: usize },
     /// A `Prio3` histogram.
-    Prio3Histogram {
-        #[derivative(Debug(format_with = "bucket_count"))]
-        buckets: Vec<u64>,
-    },
+    Prio3Histogram { buckets: usize },
```
divergentdave, line 33: I have a slight preference for naming the parameter `length` instead of `buckets` here, in order to match with what VDAF uses.

integration_tests/src/divviup_api_client.rs
```diff
@@ -21,7 +21,7 @@ pub enum ApiVdaf {
     /// Corresponds to Prio3Count
     Count,
     Histogram {
-        buckets: Vec<u64>,
+        buckets: usize,
```
divergentdave, line 24: If we rename the parameter to `length` in the aggregator API, we'll want to do the same in the control plane API.

interop_binaries/src/lib.rs
```diff
@@ -117,7 +117,7 @@ pub enum VdafObject {
         length: NumberAsString<usize>,
     },
     Prio3Histogram {
-        buckets: Vec<NumberAsString<u64>>,
+        buckets: NumberAsString<usize>,
```
divergentdave, line 120: I used `length` instead of `buckets` in divergentdave/draft-dcook-ppm-dap-interop-test-design#41, again to align with https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-7.4.3-2.

tools/src/bin/collect.rs
```diff
@@ -367,10 +332,9 @@ struct Options {
         required = false,
         num_args = 1,
         action = ArgAction::Set,
-        value_parser = BucketsValueParser::new(),
         help_heading = "VDAF Algorithm and Parameters"
     )]
-    buckets: Option<Buckets>,
+    buckets: Option<usize>,
```
divergentdave, line 337: If we replace `buckets` with `length` elsewhere, we could eliminate this flag and simplify a bit.

divergentdave (commented):
Looking good, I have one parameter naming comment on three different APIs

integration_tests/src/divviup_api_client.rs
tgeoghegan: https://github.com/divviup/divviup-api/issues/410 for the divviup-api side of this.

tgeoghegan:
I've renamed `buckets` to `length` everywhere and simplified the collect tool by collapsing `--buckets` into `--length`.

integration_tests/tests/common/mod.rs
```diff
@@ -322,21 +322,15 @@ pub async fn submit_measurements_and_verify_aggregate(
...
         }
-        VdafInstance::Prio3Histogram { buckets } => {
-            let vdaf = Prio3::new_histogram(2, buckets).unwrap();
+        VdafInstance::Prio3Histogram { length } => {
+            let vdaf = Prio3::new_histogram(2, *length).unwrap();
 
-            let mut aggregate_result = vec![0; buckets.len() + 1];
-            aggregate_result.resize(buckets.len() + 1, 0);
+            let mut aggregate_result = vec![0; *length];
+            aggregate_result.resize(*length, 0);
```
divergentdave, line 329: Not introduced by this PR: This line is a no-op, because the vector is already the correct length. It's likely leftover from past clippy jousting or conflict resolution.
```suggestion
```

divergentdave (approved):
Looks good, just one more arithmetic issue in some transitional code.

PR #1675, opened 2023-08-05
taskprov: Implement helper behavior

inahga:

Please review this on a per-commit basis. Each commit also contains details on the included changes.


aggregator/src/aggregator.rs
```diff
@@ -336,12 +378,25 @@ impl<C: Clock> Aggregator<C> {
         aggregation_job_id: &AggregationJobId,
         req_bytes: &[u8],
         auth_token: Option<AuthenticationToken>,
+        taskprov_header: Option<&[u8]>,
```
tgeoghegan, line 383: I don't think it makes sense for either this handler or the aggregate share handler, below, to deal with taskprov. How could we be handling an aggregation job continue or aggregate share request if we didn't first handle an aggregation job init request that provisioned the task?

The [spec](https://datatracker.ietf.org/doc/html/draft-wang-ppm-dap-taskprov-04#name-the-taskprov-extension) says:

>Otherwise, if the Leader does opt-in, it advertises the task to the Helpers during the aggregate protocol ([Section 4.4](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-04#section-4.4) of [[DAP](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-04)]). In particular, it includes the task configuration in an HTTP header *of its first aggregate request for that task*.

(emphasis mine) 

Then the helper section says:
>Finally, the Helper completes the *aggregate initialize* request as usual, deriving the VDAF verification key for the task as described in [Section 3.2](https://datatracker.ietf.org/doc/html/draft-wang-ppm-dap-taskprov-04#vdaf-verify-key).

IMO the leader text is ambiguous and we should ask the taskprov authors to clarify exactly which requests the leader should augment with the taskprov header, but I think the intent here is what I concluded: only the leader's `AggregationJobInitReq` should include the taskprov header.

aggregator/src/aggregator.rs
```diff
@@ -2474,6 +2724,7 @@ impl VdafOps {
         vdaf: Arc<A>,
         req_bytes: &[u8],
         batch_aggregation_shard_count: u64,
+        peer_aggregator: Option<&PeerAggregator>,
```
tgeoghegan, line 2729: It's a shame that `handle_aggregate_share_generic` has to do gymnastics with the `Option<&PeerAggregator>` just so it can figure out how to get a collector HPKE config. More broadly, I think this layer of Janus shouldn't need to be aware of taskprov at all. What if somewhere higher up in the callstack, some code that is already dealing with the `dap-taskprov` header could figure out what collector HPKE config is relevant and then construct a `Task` with that?

aggregator/src/aggregator.rs
```diff
@@ -478,26 +564,184 @@ impl<C: Clock> Aggregator<C> {
...
+    /// we can opt into the task. This function might return an opt-out error for conditions that
+    /// are relevant for all DAP workflows (e.g. task expiration).
+    #[tracing::instrument(skip(self, aggregator_auth_token), err)]
+    async fn taskprov_authorize_request(
+        &self,
+        peer_role: &Role,
+        task_id: &TaskId,
+        taskprov_header: &[u8],
+        aggregator_auth_token: Option<&AuthenticationToken>,
+    ) -> Result<(TaskConfig, Arc<PeerAggregator>, Vec<Url>), Error> {
```
tgeoghegan, line 692: I'm very concerned about parsing the `TaskConfig` object before we can even authenticate taskprov requests. While using Rust does mean parsing untrusted inputs is safer, it still should be avoided as much as possible (I acknowledge that we have already parsed a whole lot of untrusted HTTP before getting to this point). That was a lot of the reason that we refactored the DAP API to put the task ID and things like aggregation job IDs into request paths.

By this point, we already have the `aggregator_auth_token`. Could we instead scan the `taskprov_aggregator_auth_tokens` table looking for a match? Alternately, we could augment the schema for `taskprov_peer_aggregators` so that it includes some unique peer aggregator ID and then require incoming requests to have a `dap-taskprov-peer-aggregator-id` header in them.

aggregator/src/aggregator.rs
```diff
@@ -307,17 +322,44 @@ impl<C: Clock> Aggregator<C> {
         aggregation_job_id: &AggregationJobId,
         req_bytes: &[u8],
         auth_token: Option<AuthenticationToken>,
+        taskprov_header: Option<&[u8]>,
```
tgeoghegan, line 327: I'd prefer if we decoded the taskprov header as soon as possible, and then all these method signatures (this,  `handle_aggregate_continue`) would have the more expressive `taskprov: Option<&TaskConfig>`. It's also friendlier to tests, because constructing a fake `TaskConfig` value is easier than constructing the encoding thereof.

Many of these methods pass the request down as `&[u8]`, but we are forced to do that because often you can't parse the incoming request until you have the task, because sometimes you need to know the query type to parse the message. That isn't the case for the `TaskConfig`, so I'd like to parse it as soon as possible. Actually, I think it could be done in a trillium handler.

tgeoghegan (commented):
The design this PR takes is that the peer aggregators which support taskprov are stored in a Janus database table, and we expect to have to use `janus_cli` to create entries in that table. A comment in code states that we only ever expect to have or two such aggregators, so did we consider providing their information to Janus as config instead of as persistent database rows?

The advantage of providing them via config is that we wouldn't have to commit to a permanent representation of a peer aggregator and also that changing the list of them becomes a simple matter of changing config and a restart (and the current implementation requires a restart for Janus to notice a new peer aggregator anyway).

I understand that there's an appeal to having taskprov peer aggregators work the same way as aggregators managed through divviup-api, but I think we should be cautious about putting a representation of a peer aggregator into the Janus database, especially if we expect that pairing a taskprov peer will _always_ involve DU operators anyway.

db/00000000000015_taskprov.up.sql
```diff
@@ -0,0 +1,42 @@
...
+    role AGGREGATOR_ROLE NOT NULL,  -- the role of this aggregator relative to the peer
+    verify_key_init BYTEA NOT NULL, -- the preshared key used for VDAF verify key derivation.
+
+    -- Per task parameters.
+    tolerable_clock_skew   BIGINT NOT NULL, -- the maximum acceptable clock skew to allow between client and aggregator, in seconds
+    report_expiry_age      BIGINT,          -- the maximum age of a report before it is considered expired (and acceptable for garbage collection), in seconds. NULL means that GC is disabled.
+    collector_hpke_config BYTEA NOT NULL,   -- the HPKE config of the collector (encoded HpkeConfig message)
+
+    CONSTRAINT taskprov_peer_aggregator_endpoint_and_role_unique UNIQUE(endpoint, role)
+);
```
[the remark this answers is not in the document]
inahga: Under normal conditions we should either query all rows, or query by `(endpoint, role)`

The `CONSTRAINT ... UNIQUE` should take care of making an index on `(endpoint, role)` for us (I believe `CONSTRAINT ... UNIQUE` is just sugar for `CREATE UNIQUE INDEX`?)
```
postgres@127:postgres> EXPLAIN ANALYZE SELECT * FROM taskprov_peer_aggregators WHERE endpoint = 'https://example.com' AND role = 'HELPER';
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| QUERY PLAN                                                                                                                                                                    |
|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Index Scan using taskprov_peer_aggregator_endpoint_and_role_unique on taskprov_peer_aggregators  (cost=0.15..8.17 rows=1 width=124) (actual time=0.012..0.013 rows=0 loops=1) |
|   Index Cond: ((endpoint = 'https://example.com'::text) AND (role = 'HELPER'::aggregator_role))                                                                               |
| Planning Time: 0.126 ms                                                                                                                                                       |
| Execution Time: 0.043 ms                                                                                                                                                      |
+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
EXPLAIN 4
Time: 0.008s
```
```
postgres@127:postgres> SELECT * FROM pg_indexes WHERE tablename = 'taskprov_peer_aggregators';
+------------+---------------------------+---------------------------------------------------+------------+-------------------------------------------------------------------------->
| schemaname | tablename                 | indexname                                         | tablespace | indexdef                                                                 >
|------------+---------------------------+---------------------------------------------------+------------+-------------------------------------------------------------------------->
| public     | taskprov_peer_aggregators | taskprov_peer_aggregators_pkey                    | <null>     | CREATE UNIQUE INDEX taskprov_peer_aggregators_pkey ON public.taskprov_pee>
| public     | taskprov_peer_aggregators | taskprov_peer_aggregator_endpoint_and_role_unique | <null>     | CREATE UNIQUE INDEX taskprov_peer_aggregator_endpoint_and_role_unique ON >
+------------+---------------------------+---------------------------------------------------+------------+-------------------------------------------------------------------------->
```

aggregator_core/src/datastore.rs
```diff
@@ -1046,7 +1056,9 @@ impl<C: Clock> Transaction<'_, C> {
             )?));
         }
 
-        Ok(Task::new(
+        // Don't validate a task on its way out. If it's in the database, we'll assume that it was
+        // constructed in a valid way.
+        Ok(Task::new_without_validation(
```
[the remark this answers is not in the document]
inahga: Indeed. Taskprov tasks don't have any auth tokens or a collector HPKE config, which Task::validate considers invalid.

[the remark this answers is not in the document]
inahga: An alternative could to track in the database whether a task was created by taskprov, then invoke separate validation logic. Curious on your thoughts on this--I was wary of mixing taskprov logic too much with normal logic, hence the approach of building a taskprov task onto the normal task abstraction as much as possible.

aggregator_core/src/taskprov.rs
```diff
@@ -0,0 +1,403 @@
...
+            None,
+            Vec::new(),
+            Vec::new(),
+            Vec::new(),
+        ));
+        task.validate()?;
+        Ok(task)
+    }
+
+    fn validate(&self) -> Result<(), Error> {
```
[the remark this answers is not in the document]
inahga: Clarification: two implementations of `validate()` or the entire struct?

For `validate()`, this is confounded by the fact that taskprov validation is not really a subset of normal task validation. I could possibly factor out the common parts into 3 functions `validate_normal` `validate_common` `validate_taskprov`, but not sure if that's really an improvement.

aggregator/src/aggregator.rs
inahga: My apologies, I should have linked the pending change that this logic follows https://github.com/wangshan/draft-wang-ppm-dap-taskprov/commit/73859696afcf7a1ddfdcb83de3632e7d1152c327. This PR is based off of the main branch of `draft-wang-ppm-dap-taskprov`, as we went with in https://github.com/divviup/janus/issues/1568#issuecomment-1619156088.

This sentence in particular is what drove me to do it for `continue`.

> The Leader MUST advertise the task to the Helper in every request incident to the task as described in {{definition}}.

And for `share`.

> The Leader MUST advertise the task in every aggregate share request issued to the Helper as described in {{definition}}.

That being said, you are correct that it doesn't really make sense for `share` or `continue` to follow this flow since nothing can be done in these functions without a provisioned task**. It does seem worth it to ask the authors, I'll do so tomorrow.

** I guess you could collect nothing to trigger the taskprov flow and provision a task... perhaps that was their intent?

aggregator/src/aggregator.rs
inahga: > Could we instead scan the taskprov_aggregator_auth_tokens table looking for a match?

I think this can be done. It would be costly with lots of peer aggregators and/or auth tokens, but we don't plan on having that many anyways.

inahga:
> so did we consider providing their information to Janus as config instead of as persistent database rows?

Yes, conversation here https://isrg.slack.com/archives/C0167LT4C73/p1688684256768919?thread_ts=1688678359.202149&cid=C0167LT4C73. The main rationale is to keep secrets out of plaintext configuration, like auth tokens, vdaf verify init, and global HPKE config.

This simplifies deploying this--the only modifications to `janus-ops` that are necessary are enabling taskprov in the configuration. We won't have to change how we handle the config file due to it containing secrets.

aggregator/src/aggregator.rs
inahga: Ahah, this is the intent of "The Leader MUST advertise the task to the Helper in every request incident to the task as described in {{definition}}" https://github.com/wangshan/draft-wang-ppm-dap-taskprov/issues/31, it's so that tasks can be generated on the fly rather than stored in the DB.

tgeoghegan:
> Yes, conversation here https://isrg.slack.com/archives/C0167LT4C73/p1688684256768919?thread_ts=1688678359.202149&cid=C0167LT4C73. The main rationale is to keep secrets out of plaintext configuration, like auth tokens, vdaf verify init, and global HPKE config.

OK, I find that argument persuasive, and also if I felt strongly about it then I shoulda gotten involved at the time the design was being discussed! That said I think it'd be nice to record somewhere why we do it this way, besides a link to a private Slack. Maybe we could add a `docs/TASKPROV.md` that discusses various design decisions and alternatives considered. No need to block this change on that writeup, though.

inahga:
> Maybe we could add a docs/TASKPROV.md that discusses various design decisions and alternatives considered. 

Already in the works!

aggregator_core/src/taskprov.rs
```diff
@@ -0,0 +1,403 @@
...
+    /// The URL at which the peer aggregator can be reached. This, along with `role`, is used to
+    /// uniquely represent the peer aggregator.
+    endpoint: Url,
+
+    /// The role that the peer aggregator takes in DAP. Must be [`Role::Leader`] or [`Role::Helper`].
+    /// This, along with `endpoint`, uniquely represents the peer aggregator.
+    role: Role,
+
+    /// The preshared key used to derive the VDAF verify key for each task.
+    #[derivative(Debug = "ignore")]
```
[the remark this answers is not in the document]
inahga: That works!
```
#[derive(Derivative, Clone, Copy, PartialEq, Eq)]
#[derivative(Debug)]
pub struct VerifyKeyInit(#[derivative(Debug = "ignore")] [u8; Self::LEN]);
```

`SecretBytes` doesn't do it this way... but perhaps it should?

aggregator/src/cache.rs
```diff
@@ -137,3 +141,34 @@ impl Drop for GlobalHpkeKeypairCache {
...
+                    Box::pin(async move { tx.get_taskprov_peer_aggregators().await })
+                })
+                .await?
+                .into_iter()
+                .map(Arc::new)
+                .collect(),
+        })
+    }
+
+    pub fn get(&self, endpoint: &Url, role: &Role) -> Option<Arc<PeerAggregator>> {
```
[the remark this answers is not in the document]
inahga: Nice. I think this also applies to `GlobalHpkeKeypairCache`, I can experiment with that in a separate PR.

aggregator/src/aggregator.rs
tgeoghegan: Ah, I can see how this requirement helps in the face of that design choice. But as you point out, Janus doesn't need to generate tasks on the fly, so while we might still expect to see the `dap-taskprov` header in aggregation job continue and aggregate share requests, we can just ignore it and use the task ID in the request path just like any other request. Regardless of the presence of taskprov, an aggregation job continue req for some (task ID, aggregation job ID) pair is going to fail if it was not preceded by an aggregation job init req.

I know that there's a MUST about checking `TaskConfig` against the task ID in the request path, but I suspect that can be checked higher up in the stack. Or frankly I'd be comfortable with just not checking the `TaskConfig` anywhere but aggregation job init.

aggregator_core/src/taskprov.rs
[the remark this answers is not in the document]
tgeoghegan: I buy that validation logic differs across the two kinds of task, but it appears a lot of logic is duplicated across the two. For instance, we will soon change `Task` so that instead of  `aggregator_endpoints: Vec<Url>`, it'll just have two fields `leader_endpoint: Url, helper_endpoint: Url`. When we make that change, I only want to have to update a single `validate()` implementation.

db/00000000000015_taskprov.up.sql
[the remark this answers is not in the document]
tgeoghegan: Oh, neat, you've clearly looked into this more than I have!

messages/src/lib.rs
```diff
@@ -110,7 +110,7 @@ impl TryFrom<&[u8]> for Url {
     }
 }
 
-impl TryInto<url::Url> for Url {
+impl TryInto<url::Url> for &Url {
```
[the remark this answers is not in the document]
tgeoghegan: Doing `TryFrom` does sound more Rustful so I'd try that if you feel like it, but this isn't worth spending too much time on.

aggregator_core/src/datastore.rs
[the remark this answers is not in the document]
tgeoghegan: I think you're right that errors here are unlikely, but validating data here is still nice because it can protect us against a botched database migration or some other funny thing happening.

I wonder if tracking the source of the task will become necessary anyway so that divviup-api can make good decisions. But in the near term, I'm wary of making schema changes unless we're sure we need them. How about a compromise: you've written two flavours of task validation routine, one for taskprov tasks and one for regular ones. We could check if either succeeds and error out otherwise. This way all we're doing is adding code, which is much easier to change than a DB schema. If at some point in the future we begin tracking task provenance anyway, we can revisit this validation logic.

aggregator/src/aggregator.rs
inahga: Ah, looking deeper at implementing this, these requests do need to be concerned with taskprov so that we can use the peer aggregator's auth tokens. The only indication about what auth tokens need to be used comes from the taskprov header.

In essence, this PR using the property "The Leader MUST advertise the task to the Helper in every request incident" to identify the peer aggregator being used per request. This goes away if we agree upon a better authentication or peer aggregator identification mechanism with our peer.

I've noted in slack that some of this complexity comes from supporting multiple peer aggregators. If we accept that there would only be one peer aggregator in one role per janus deployment, then we don't need to parse the taskprov header before authentication.

aggregator/src/aggregator.rs
inahga: Discussed in https://isrg.slack.com/archives/C0167LT4C73/p1691514380751889. This requires some additional discussion with our peers about how authentication should work.

I will leave it as-is with a FIXME comment and an issue for follow up.

aggregator_core/src/datastore.rs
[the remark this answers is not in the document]
inahga: I've implemented this. I think it could probably use some unit tests around validation, but I'd like your input before I get too deep into that.

db/00000000000015_taskprov.up.sql
```diff
@@ -0,0 +1,42 @@
...
+    token BYTEA NOT NULL,                -- bearer token used to authenticate messages to/from the other aggregator (encrypted)
+    type AUTH_TOKEN_TYPE NOT NULL DEFAULT 'BEARER',
+
+    CONSTRAINT task_collector_auth_tokens_unique_peer_aggregator_id_and_ord UNIQUE(peer_aggregator_id, ord),
+    CONSTRAINT fk_peer_aggregator_id FOREIGN KEY(peer_aggregator_id) REFERENCES taskprov_peer_aggregators(id) ON DELETE CASCADE
+);
+
+-- A task can now have no collector HPKE config set, in which case we'll look at the peer aggregator
+-- for the config.
+ALTER TABLE tasks ALTER COLUMN collector_hpke_config DROP NOT NULL;
```
tgeoghegan, line 42: nit: missing newline

messages/src/lib.rs
```diff
@@ -110,13 +110,11 @@ impl TryFrom<&[u8]> for Url {
...
 
-impl TryInto<url::Url> for Url {
+impl TryFrom<&Url> for url::Url {
     type Error = url::ParseError;
 
-    fn try_into(self) -> Result<url::Url, Self::Error> {
-        // Unwrap safety: this type can't be constructed without being validated
-        // as consisting only of ASCII.
-        url::Url::parse(str::from_utf8(&self.0).unwrap())
+    fn try_from(value: &Url) -> Result<Self, Self::Error> {
```
tgeoghegan, line 116: We should keep the unwrap safety comment

aggregator/src/aggregator.rs
```diff
@@ -478,24 +575,169 @@ impl<C: Clock> Aggregator<C> {
...
+        if self.clock.now() > *task_config.task_expiration() {
+            return Err(Error::InvalidTask(*task_id, OptOutReason::TaskExpired));
         }
+
+        info!(
+            ?task_id,
+            ?task_config,
+            ?peer_aggregator,
+            "taskprov: authorized request"
+        );
```
tgeoghegan, line 739: I don't think we should log every time we successfully authorize a request.

tgeoghegan (approved):
Few more nits but this looks good

aggregator/src/aggregator.rs
inahga: I've demoted it to `debug` level.

PR #1676, opened 2023-08-07
Prepare to branch for DAP-04

tgeoghegan:
We are starting our DAP-05 implementation (#1669) and so must branch for DAP-04.


tgeoghegan:
Despite a confusing branch name, the point of this change is to branch `release/0.5` from `main`, which is our maintained DAP-04 implementation.

PR #1683, opened 2023-08-07
Split aggregator_api/src/lib.rs

inahga:
Quick cleanup PR before adding more stuff to this crate. Strictly optional, if others are comfortable with the file length we can skip this PR.


aggregator_api/src/routes.rs
```diff
@@ -0,0 +1,321 @@
...
+use querystring::querify;
+use rand::random;
+use ring::digest::{digest, SHA256};
+use std::{str::FromStr, sync::Arc, unreachable};
+use tracing::{error, warn};
+use trillium::{Conn, Handler, Status};
+use trillium_api::{Halt, Json, State};
+use url::Url;
+
+pub(super) async fn get_config(
```
inahga, line 30: This is the only actual code change--these handlers are `pub(super)` now.

tgeoghegan (approved):

PR #1688, opened 2023-08-09
Fix Content-Type of problem details responses

### LINKED ISSUE #1687: Problem details documents returned with wrong Content-Type
Our aggregator is returning a media type of "application/json" (and a status code of 400) in response to a failing collection job creation request. This results in the response body correctly getting ignored by `janus_core::http::response_to_problem_details()`, since it expects "application/problem+json".

divergentdave:
This fixes #1687 by ~~reordering two calls~~ updating trillium-api. ~~The `with_json()` method from an extension trait also sets the Content-Type header, so it was stomping on the previously-set header.~~ All tests that check problem details response bodies now also check the Content-Type header.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4779,9 +4779,9 @@ dependencies = [
 
 [[package]]
 name = "trillium-api"
-version = "0.2.0-rc.3"
+version = "0.2.0-rc.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b9f41bf4e58d42642f4ba58d0d609d7d06499fea2df21d66f371a63f2849f8a"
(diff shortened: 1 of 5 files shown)


jbr:
this probably should be treated as a flaw in trillium-api. Explicitly set content type shouldn't be stomped by with_json

jbr:
released trillium-api v0.2.0-rc.4 🎉

tgeoghegan:
So it sounds like instead of this change, we should just move to a newer `trillium-api`?

divergentdave:
Done, this now bumps the trillium-api version, and I backed out the reordering of method calls.

tgeoghegan (approved):
Glad to see the test coverage improvement here!

Issue #1685, opened 2023-08-08
Taskprov: determine secrets rotation strategy


inahga:
There are several secrets whose lifecycle needs to be considered:

- Aggregator auth tokens: changes need to be communicated to and kept in sync with the peer.
- Collector auth tokens, if operating as the leader.
- VDAF verify key init: needs to be kept in sync with the peer for new tasks to derive the same VDAF verify key.

Relevant discussion: https://github.com/divviup/janus/pull/1675#discussion_r1286300929

inahga:
Note that in the aggregator API I have left the case where we want to modify an existing peer aggregator unhandled, we should address this in this issue as the implementation will likely depend on how we want to handle key rotation.

PR #1689, opened 2023-08-09
Add peer aggregator endpoints to aggregator API

inahga:
Supports https://github.com/divviup/janus/issues/1568, stacked on https://github.com/divviup/janus/pull/1675.

Adds peer aggregator endpoints to the aggregator API to allow CRUD. Similar to https://github.com/divviup/janus/pull/1686, I will hold off on adding functionality to janus_cli to be conscious of time. The risks and rationale are also similar.

Notice that the format of operations is `OPERATION /taskprov/peer_aggregators?role=foo&endpoint=http://example.com`. I considered several alternatives:
- Having role/endpoint being specified in the request body isn't very RESTful.
- Having them specified like `/taskprov/peer_aggregators/role/endpoint` is unergonomic, as both `querystring` and `serde_urlencoded` expect their I/O to be in the form of key-value pairs. This is the best alternative IMO, if we're willing to give up proper urlencoding.
- The database ID should ideally be completely opaque, so addressing by DB ID is out.

I also punt on allowing modification of peer aggregators, to leave that for https://github.com/divviup/janus/issues/1685.

I replaced `querystring` with `serde_urlencoded`, since `serde_urlencoded` pretty much does everything `querystring` does. There's no sense in keeping around both IMO.

### DIFF
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -161,8 +161,6 @@ enum Error {
     #[error(transparent)]
     Url(#[from] url::ParseError),
     #[error(transparent)]
-    UrlEncoding(#[from] serde_urlencoded::de::Error),
-    #[error(transparent)]
     Role(#[from] RoleParseError),
 }
 
@@ -197,9 +195,6 @@ impl Handler for Error {
                 .with_status(Status::BadRequest)
                 .with_body(message.to_string()),
             Self::Url(err) => conn
-                .with_status(Status::BadRequest)
-                .with_body(err.to_string()),
-            Self::UrlEncoding(err) => conn
                 .with_status(Status::BadRequest)
                 .with_body(err.to_string()),
             Self::Role(err) => conn
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -5,13 +5,10 @@ mod routes;
 mod tests;
 
 use async_trait::async_trait;
-
 use janus_aggregator_core::datastore;
 use janus_aggregator_core::{datastore::Datastore, instrumented};
 use janus_core::{http::extract_bearer_token, task::AuthenticationToken, time::Clock};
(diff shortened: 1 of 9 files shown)


aggregator_api/src/lib.rs
```diff
@@ -141,15 +149,22 @@ enum Error {
...
-    /// Errors that should return HTTP 400.
-    #[error("{0}")]
-    BadRequest(String),
     /// Errors that should return HTTP 404.
     #[error("Target resource was not found")]
     NotFound,
     /// Errors that should return HTTP 409.
     #[error("{0}")]
     Conflict(String),
+
```
tgeoghegan, line 158: nit: delete blank line

tgeoghegan (changes requested):
I disagree with the rationale for using query parameters. We should use JSON request bodies like we do for everything else in the aggregator API.

I do agree that providing a body to a GET request is a bad idea. We should use HTTP QUERY ([trillium lets you specify methods as strings](https://docs.trillium.rs/trillium_router/struct.router#method.with_route)) to work around that.

inahga:
Trillium doesn't appear to support the `QUERY` method.
```
thread 'tests::get_global_hpke_configs' panicked at 'called `Result::unwrap()` on an `Err` value: UnrecognizedMethod("Invalid HTTP method")', /home/inahga/.cargo/registry/src/index.crates.io-6f17d22bba15001f/trillium-router-0.3.5/src/router.rs:306:42
```

Should it? AFAICT it didn't make it out of the draft stage https://datatracker.ietf.org/doc/draft-ietf-httpbis-safe-method-w-body/.

Either way this isn't a showstopper for having role/endpoint in the request body. I think it's actually reasonable for us to omit GET by role and endpoint, because in practice we're very unlikely to have more than 1 peer aggregator configured. Request body is otherwise a nice cleanup.

tgeoghegan (approved):

Issue #1691, opened 2023-08-10
aggregator API: don't return HTTP error when deleting resource that doesn't exist


tgeoghegan:
The aggregator API's endpoints for DELETE-ing resources return HTTP 204 No Content if a resource was successfully deleted, but HTTP 404 Not Found if the resource in question didn't already exist. For idempotence reasons, I think both cases should return HTTP 204 No Content because regardless of whether the resource existed, the client's desired postcondition is achieved. Returning distinct HTTP status codes just makes clients more complicated, because I doubt they'll do anything different if they get 404 instead of 204.

inahga:
Resolved by https://github.com/divviup/janus/pull/1694

PR #1694, opened 2023-08-11
Minor aggregator API cleanup

inahga:
- Deletion should return 204 on not found, not 404.

    Supports https://github.com/divviup/janus/issues/1691.

- Provide explicit return type in routes

    From conversation in https://github.com/divviup/janus/pull/1686#discussion_r1290401987.

- Remove needless Halt

    I'm not sure why this is here--no other routes use Halt in this manner (should they)?

### DIFF
--- a/aggregator_api/src/routes.rs
+++ b/aggregator_api/src/routes.rs
@@ -289,7 +289,7 @@ pub(super) async fn delete_task<C: Clock>(
     })
     .await
     .map_err(|err| match err {
-        datastore::Error::MutationTargetNotFound => Status::NotFound,
+        datastore::Error::MutationTargetNotFound => Status::NoContent,
         _ => {
             error!(err = %err, "Database transaction error");
             Status::InternalServerError
@@ -450,7 +450,7 @@ pub(super) async fn delete_global_hpke_config<C: Clock>(
     })
     .await
     .map_err(|err| match err {
-        datastore::Error::MutationTargetNotFound => Status::NotFound,
+        datastore::Error::MutationTargetNotFound => Status::NoContent,
         _ => {
             error!(err = %err, "Database transaction error");
             Status::InternalServerError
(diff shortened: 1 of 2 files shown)


aggregator_api/src/routes.rs
```diff
@@ -281,15 +278,15 @@ pub(super) async fn get_task<C: Clock>(
...
+) -> Result<Status, Status> {
     let task_id = conn.task_id_param()?;
 
     ds.run_tx_with_name("delete_task", |tx| {
         Box::pin(async move { tx.delete_task(&task_id).await })
     })
     .await
     .map_err(|err| match err {
-        datastore::Error::MutationTargetNotFound => Status::NotFound,
+        datastore::Error::MutationTargetNotFound => Status::NoContent,
```
tgeoghegan, line 289: Were there not tests that had to be updated to account for this? If not, we should take the opportunity to add tests that exercise this.

aggregator_api/src/routes.rs
```diff
@@ -76,19 +76,16 @@ pub(super) async fn get_task_ids<C: Clock>(
         })?;
     let pagination_token = task_ids.last().cloned();
 
-    Ok((
-        Json(GetTaskIdsResp {
-            task_ids,
-            pagination_token,
-        }),
-        Halt,
```
tgeoghegan, line 84: Hmm, I think maybe we had to do this in order to stop other routes from also matching on the request, but I don't think any other handler would have matches on `/task_ids` so I suspect this wasn't needed.

tgeoghegan (changes requested):

aggregator_api/src/routes.rs
inahga: Whoops, that got lost in a `git stash`.

aggregator_api/src/routes.rs
inahga: Looking into it more I think it'll stop any additional handlers from running, but we don't have any at the moment. https://github.com/divviup/janus/blob/8c2761d4c438bf488227a90d66c2682b8bbb2d7b/aggregator_api/src/lib.rs#L71

tgeoghegan (approved):

PR #1697, opened 2023-08-14
use divviup-api-client in in-cluster tests

jbr:


integration_tests/Cargo.toml
```diff
@@ -35,6 +35,11 @@ serde_json = "1.0.103"
 testcontainers = "0.14.0"
 tokio.workspace = true
 url = { version = "2.4.0", features = ["serde"] }
+divviup-client = { git = "https://github.com/divviup/divviup-api", features = [
+    "admin",
+] } # , tag = "0.0.19" }
```
jbr, line 40: eventually we'll want to use versions from crates.io

integration_tests/tests/in_cluster.rs
```diff
@@ -118,68 +125,84 @@ impl InClusterJanusPair {
...
+                },
+                other => panic!("unsupported vdaf {other:?}"),
+            },
             min_batch_size: task.min_batch_size(),
             max_batch_size: match task.query_type() {
                 QueryType::TimeInterval => None,
                 QueryType::FixedSize { max_batch_size, .. } => Some(*max_batch_size),
             },
-            expiration: "3000-01-01T00:00:00Z".to_owned(),
+            expiration: None,
```
jbr, line 183: this is the only known functional change here, and i don't know if it'll break anything. it was easier to type "None" than parse an OffsetDateTime

integration_tests/tests/in_cluster.rs
```diff
@@ -101,15 +100,23 @@ impl InClusterJanusPair {
...
         // service port numbers.
 
-        let divviup_api = DivviupApiClient::new(
-            cluster
-                .forward_port(&divviup_api_namespace, "divviup-api", 80)
-                .await,
+        let port = cluster
+            .forward_port(&divviup_api_namespace, "divviup-api", 80)
+            .await
+            .local_port();
```
divergentdave, line 106: The struct returned by `Cluster::forward_port()` needs to be stored in a separate binding, as it acts as a drop guard for shutting down the port forwarding.

Cargo.lock
```diff
@@ -2174,6 +2214,24 @@ dependencies = [
...
+ "derivative",
+ "hex",
+ "num_enum 0.6.1",
+ "prio",
+ "rand",
+ "serde",
+ "thiserror",
+ "url",
+]
+
```
divergentdave, line 2234: The presence of a second copy of `janus_messgaes` is a bad smell. I tried updating the dependency and adapting the code to pass a `HpkeConfig` to `create_hpke_config()` instead of encoding it first, and I now get an error due to mismatched crate versions.

```
error[E0308]: mismatched types
    --> integration_tests/tests/in_cluster.rs:151:17
     |
149  |             .create_hpke_config(
     |              ------------------ arguments to this method are incorrect
150  |                 account.id,
151  |                 task.collector_hpke_config().unwrap(),
     |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&HpkeConfigContents`, found `&HpkeConfig`
     |
     = note: `HpkeConfig` and `HpkeConfigContents` have similar names, but are actually distinct types
note: `HpkeConfig` is defined in crate `janus_messages`
    --> /home/david/Code/ppm/janus-ops/janus-checkout/messages/src/lib.rs:1076:1
     |
1076 | pub struct HpkeConfig {
     | ^^^^^^^^^^^^^^^^^^^^^
note: `HpkeConfigContents` is defined in crate `janus_messages`
    --> /home/david/.cargo/registry/src/index.crates.io-6f17d22bba15001f/janus_messages-0.5.14/src/lib.rs:1076:1
     |
1076 | pub struct HpkeConfig {
     | ^^^^^^^^^^^^^^^^^^^^^
     = note: perhaps two different versions of crate `janus_messages` are being used?
note: method defined here
    --> /home/david/.cargo/git/checkouts/divviup-api-9ebde8d8443dfb74/97ded06/client/src/lib.rs:322:18
     |
322  |     pub async fn create_hpke_config(
     |                  ^^^^^^^^^^^^^^^^^^
```

I think we can fix this by adding a `patch.crates-io.janus_messages.path` entry in the workspace-level `Cargo.toml`. I confirmed that `cargo publish --dry-run -p janus_messages` doesn't complain with this present.

```toml
[patch.crates-io]
janus_messages = { path = "messages" }
```

divergentdave (changes requested):

Cargo.lock
jbr: I addressed this a different way — I'm not convinced it's a smell to have two different versions of a crate. The janus_messages in the client should be considered part of its interface, and reflects the version of janus_messages the client expects to be interacting with. Instead I just converted between the types at the api boundary. Do you still prefer patching janus_messages?

divergentdave: Sure, converting sounds good to me

integration_tests/tests/in_cluster.rs
```diff
@@ -11,16 +13,11 @@ use janus_core::{
...
-use janus_integration_tests::{
-    client::ClientBackend,
-    divviup_api_client::{
-        DivviupApiClient, NewAggregatorRequest, NewHpkeConfigRequest, NewTaskRequest,
-    },
-    TaskParameters,
-};
+use janus_integration_tests::{client::ClientBackend, TaskParameters};
 use janus_messages::TaskId;
 use prio::codec::Encode;
```
divergentdave, line 18: This is now unused
```suggestion
```

integration_tests/tests/in_cluster.rs
```diff
@@ -118,68 +123,99 @@ impl InClusterJanusPair {
...
-                    contents: STANDARD.encode(task.collector_hpke_config().unwrap().get_encoded()),
-                },
+                account.id,
+                HpkeConfigContents::new(
+                    (hpke_config.id() as u8).into(),
+                    (hpke_config.kem_id() as u16).try_into().unwrap(),
+                    (hpke_config.kdf_id() as u16).try_into().unwrap(),
+                    (hpke_config.aead_id() as u16).try_into().unwrap(),
+                    hpke_config.public_key().as_ref().to_vec().into(),
+                ),
```
divergentdave: These conversions need some fixes:
```suggestion
                &HpkeConfigContents::new(
                    u8::from(*hpke_config.id()).into(),
                    (*hpke_config.kem_id() as u16).try_into().unwrap(),
                    (*hpke_config.kdf_id() as u16).try_into().unwrap(),
                    (*hpke_config.aead_id() as u16).try_into().unwrap(),
                    hpke_config.public_key().as_ref().to_vec().into(),
                ),
```

integration_tests/tests/in_cluster.rs
```diff
@@ -118,68 +123,99 @@ impl InClusterJanusPair {
...
+                VdafInstance::Prio3Count => Vdaf::Count,
+                VdafInstance::Prio3Sum { bits } => Vdaf::Sum {
+                    bits: bits.try_into().unwrap(),
+                },
+                VdafInstance::Prio3SumVec { bits, length } => Vdaf::SumVec {
+                    bits: bits.try_into().unwrap(),
+                    length: length.try_into().unwrap(),
+                },
+                VdafInstance::Prio3Histogram { length } => Vdaf::Histogram {
+                    buckets: (0..*length - 1)
```
divergentdave, line 178: `length` is a `usize` and not a reference in this case
```suggestion
                    buckets: (0..length - 1)
```

divergentdave (approved):
I tested this prior to a recent rebase with some previously-discussed fixes and it worked against the cluster. At this point, the main branch is now using prio 0.14, which implements VDAF-06, and thus the integration test can no longer talk with the container images running in the cluster. I think we should fix the remaining compilation issues and land this, then backport the changes to the release/0.5 branch, which will still be run in the janus-ops CI, at least for the next few weeks.

integration_tests/Cargo.toml
```diff
@@ -15,6 +15,7 @@ in-cluster = ["dep:k8s-openapi", "dep:kube"]
 anyhow.workspace = true
 backoff = { version = "0.4", features = ["tokio"] }
 base64.workspace = true
+divviup-client = { git = "https://github.com/divviup/divviup-api", features = ["admin"], tag = "0.0.24" }
```
tgeoghegan, line 18: This won't work until we have `divviup-api:0.0.25` that incorporates https://github.com/divviup/divviup-api/pull/512

jbr: Could always tag=main ¯\_(ツ)_/¯

tgeoghegan: Yeah, or pin a commit on `main`. But it's nicer to use a tag and releases aren't hard to do.

PR #1698, opened 2023-08-14
Aggregator API error handling cleanup

inahga:
As this crate grows, having to map each error is becoming burdensome.

Mainly turns `Error` into a enum from a struct. We can map errors into variants that handle HTTP status codes for us, and use `#[from]` to let us use `?` on datastore errors.

We do lose a little bit of tracing in this PR. Datastore errors are unaffected, because datastore methods are already traced. But handler-local internal error paths aren't well traced, some context might be missing since we'll only see the final result. There's ways around this, but the internal error cases look to be _very_ unlikely to happen (and likely should be unwraps anyways), so I don't think it's worth spending much time on.


aggregator_api/src/lib.rs
```diff
@@ -133,56 +134,76 @@ async fn auth_check(conn: &mut Conn, (): ()) -> impl Handler {
...
-        .map_err(|err| {
-            warn!(err = ?err, "Couldn't parse task_id parameter");
-            Status::BadRequest
-        })
+    fn task_id_param(&self) -> Result<TaskId, Error> {
+        TaskId::from_str(
+            self.param("task_id")
+                .ok_or_else(|| Error::Internal("Missing task_id parameter".to_string()))?,
+        )
+        .map_err(|err| Error::Err(Status::BadRequest, format!("{:?}", err)))
```
tgeoghegan, line 198: Should this be `Error::BadRequest`?

tgeoghegan (changes requested):

PR #1732, opened 2023-08-15
`release/0.subscriber-01`: Bump crate versions

tgeoghegan:
Cargo versions must be SemVer compatible, which means that a version like `0.subscriber-01.1` is not valid, so instead we set `0.5.12-subscriber-01.1`. This reflects the release this branched from (`0.5.12`), the specific purpose of this series of releases (`subscriber-01`), and allows us to increment versions (`.1`).

What's unfortunate about this is that it doesn't match the git tags we were planning to cut (which look like `0.subscriber-01.1`). We don't publish crates from this release branch, so that might not matter: the Cargo version may only ever show up in log lines. However we could choose to change our tagging scheme to match the Cargo versions, at which point there would be a mismatch with the branch name.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1818,7 +1818,7 @@ checksum = "af150ab688ff2122fcef229be89cb50dd66af9e01a4ff320cc137eecc9bacc38"
 
 [[package]]
 name = "janus_aggregator"
-version = "0.5.12"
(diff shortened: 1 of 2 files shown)


inahga (approved):
That's unfortunate, but I don't see a better way.

PR #1736, opened 2023-08-16
Document taskprov extension and global HPKE keys

inahga:


docs/CONFIGURING_GLOBAL_HPKE_KEYS.md
```diff
@@ -0,0 +1,98 @@
...
+also [considering having this replace per-task keys entirely][1].
+
+If a global key is configured, requests to `/hpke_configs` that don't provide
+the task ID will provide the list of active global HPKE keys.
+
+This document describes the operational overhead of managing these keys.
+
+[1]: https://github.com/divviup/janus/issues/1641
+
+## Lifecycle
```
tgeoghegan, line 13:
```suggestion
## Lifecycle

```

docs/CONFIGURING_GLOBAL_HPKE_KEYS.md
```diff
@@ -0,0 +1,98 @@
...
+A global keypair has three states:
+- `pending`: The key is in the database, but is not being advertised to clients.
+- `active`: The key is used by clients, and clients should use it to encrypt
+  reports.
+- `expired`: The key is not advertised to client and will eventually be deleted.
```
tgeoghegan, line 18:
```suggestion
- `expired`: The key is not advertised to clients and will eventually be deleted.
```

docs/CONFIGURING_GLOBAL_HPKE_KEYS.md
```diff
@@ -0,0 +1,98 @@
...
+
+This document describes the operational overhead of managing these keys.
+
+[1]: https://github.com/divviup/janus/issues/1641
+
+## Lifecycle
+A global keypair has three states:
+- `pending`: The key is in the database, but is not being advertised to clients.
+- `active`: The key is used by clients, and clients should use it to encrypt
+  reports.
```
tgeoghegan, line 17: To align with the other two bullets, plus we can't really say what clients are or are not doing:
```suggestion
- `active`: The key is being advertised to clients, and clients should use it to encrypt
  reports.
```

docs/CONFIGURING_GLOBAL_HPKE_KEYS.md
```diff
@@ -0,0 +1,98 @@
...
+- `expired`: The key is not advertised to client and will eventually be deleted.
+
+These states are to facilitate key caching and rotation. The lifecycle of a key
+is as follows:
+1. The key is created in the `pending` state.
+1. The in-memory caches of all Janus replicas must pick up the new key. This can
+   be done by waiting or restarting the replicas.
```
tgeoghegan, line 24: Back in ENPA, we solved a similar problem by making worker processes exit hourly so they'd restart and pick up fresh configuration values from their environment. I think we've discussed doing something similar for Janus. Can you find a relevant issue if it exists and link it here? If no such issue exists, write one up.

docs/CONFIGURING_TASKPROV.md
```diff
@@ -0,0 +1,144 @@
...
+[1]: https://datatracker.ietf.org/doc/draft-wang-ppm-dap-taskprov/
+
+## Compatibility
+
+See this table for the matrix of branches to Taskprov versions. Taskprov is not
+implemented on any branches missing from this list.
+
+| Git branch | Taskprov draft version | Conforms to protocol? | Status |
+| ---------- | ------------- | --------------------- | ------ |
+| `release/0.subscriber-01` | [`draft-wang-ppm-dap-taskprov-02`][2] | Helper only | Supported |
```
tgeoghegan, line 22: I think we don't currently have a taskprov implementation merged into `release/0.subscriber-01`, so if there's an issue open for that, you could link it here.

docs/CONFIGURING_TASKPROV.md
```diff
@@ -0,0 +1,144 @@
...
+VERIFY_KEY_INIT=$(openssl rand 32 | basenc -w0 --base64url | sed 's/=//g')
+AGGREGATOR_AUTH_TOKEN=$(echo "some_helpful_identifier_$(openssl rand -hex 32)" | 
+    basenc -w0 --base64url | sed 's/=//g')
+
+curl -v \
+    -H "Authorization: Bearer $AGGREGATOR_API_TOKEN" \
+    -H "Accept: application/vnd.janus.aggregator+json;version=0.1" \
+    -H "Content-Type: application/vnd.janus.aggregator+json;version=0.1" \
+    "$AGGREGATOR_URL/taskprov/peer_aggregators" \
+    --data "{
```
tgeoghegan, line 104: nit: I think if you used `'` to delimit the value of `--data`, then you wouldn't need to escape `"`s:
```
--data '{ "endpoint": "thing" }'
```

tgeoghegan (approved):

docs/CONFIGURING_TASKPROV.md
inahga: Indeed, but `'` will also stop variable substitution `$VAR` due to the $ being literal. I could just avoid substitution, but my hope was to make this command as copy-pastable as possible.

docs/CONFIGURING_TASKPROV.md
inahga: It's hanging out in https://github.com/divviup/janus/pull/1742, I got ahead of myself a bit in this doc to spare having to change it for a change that's <1 day from being applied.

Issue #1641, opened 2023-07-24
Should we eliminate task-specific HPKE keys?


inahga:
From discussion in https://isrg.slack.com/archives/C0167LT4C73/p1689871160489639?thread_ts=1689810185.430009&cid=C0167LT4C73.

Since we *must* implement [global HPKE keys for taskprov](https://www.ietf.org/archive/id/draft-wang-ppm-dap-taskprov-02.html#section-3.4) (or more precisely, task-independent HPKE configuration), we can consider doing away with task-specific HPKE keys entirely.

The main possible benefit is that we don't have to worry about per-task HPKE key rotation. It's more difficult to implement than for global HPKE keys, since we have to manage a set of keys for each discrete task. (Note that we could just never rotate a task HPKE and just let them age out).

Possible drawbacks:
- The subscriber can never be in control over a global key. (i.e. they may want to choose different ciphers, but this could be mitigated by having a wide set of global keys using different ciphers to choose from).
- Could have a large blast radius if a global key is compromised (although in the current architecture, if any report encryption key is leaked it's highly likely they're all compromised).
- Spinning up janus from scratch is slightly more complicated, because the operator has to provision a global key after the database schema is applied.

inahga:
Superseded by https://github.com/divviup/janus/issues/2147

PR #1738, opened 2023-08-16
Fix panic when trying to get HPKE configs on a task that doesn't have any configured

inahga:

If a taskprov task is present, the end-user can still query `/hpke_configs?task_id=foo`. If they do this, the handler will panic because we have an `unwrap` in the task-specific `handle_hpke_config()` function. Taskprov tasks don't have any HPKE keys directly associated with them.
```rust
        HpkeConfigList::new(Vec::from([self
            .task
            .hpke_keys()
            .iter()
            .max_by_key(|(&id, _)| id)
            .unwrap()
            .1
            .config()
            .clone()]))
```

Partially fix this by unconditionally returning the global keys when in taskprov mode, so any client `task_id` parameters are ignored.

This doesn't handle the case where a taskprov task is in the database, the system is out of taskprov mode, and a client queries `/hpke_configs?task_id=foo`. I figure this is a niche enough case that we can ignore it, however I wrapped it in a comment and nicer error handling logic.

This awkward logic goes away if we adopt https://github.com/divviup/janus/issues/1641. I'm also open to cleaner ways of handling this case.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -264,33 +264,51 @@ impl<C: Clock> Aggregator<C> {
         &self,
         task_id_base64: Option<&[u8]>,
     ) -> Result<HpkeConfigList, Error> {
-        match task_id_base64 {
-            Some(task_id_base64) => {
-                let task_id_bytes = URL_SAFE_NO_PAD
-                    .decode(task_id_base64)
-                    .map_err(|_| Error::UnrecognizedMessage(None, "task_id"))?;
-                let task_id = TaskId::get_decoded(&task_id_bytes)
-                    .map_err(|_| Error::UnrecognizedMessage(None, "task_id"))?;
-                let task_aggregator = self
-                    .task_aggregator_for(&task_id)
-                    .await?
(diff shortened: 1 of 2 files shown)


aggregator/src/aggregator.rs
```diff
@@ -264,33 +264,51 @@ impl<C: Clock> Aggregator<C> {
...
+                        .task_aggregator_for(&task_id)
+                        .await?
+                        .ok_or(Error::UnrecognizedTask(task_id))?;
+
+                    match task_aggregator.handle_hpke_config() {
+                        Some(hpke_config_list) => Ok(hpke_config_list),
+                        // Assuming something hasn't gone horribly wrong with the database, this
+                        // should only happen in the case where the system has been moved from taskprov
+                        // mode to non-taskprov mode. Thus there's still taskprov tasks in the database.
+                        // This isn't a supported use case, so the operator needs to delete these tasks
```
inahga, line 297: I've updated https://github.com/divviup/janus/pull/1736 to back up this assertion.

tgeoghegan (approved):

PR #1743, opened 2023-08-17
add `protocol` to aggregator api config/capabilities

jbr:


aggregator_api/src/models.rs
```diff
@@ -23,6 +23,7 @@ pub(crate) enum AggregatorRole {
 
 #[derive(Serialize, PartialEq, Eq, Debug)]
 pub(crate) struct AggregatorApiConfig {
+    pub dap_versions: &'static [&'static str],
```
branlwyd, line 26: I think we should use a schema where the capabilities can vary by DAP version. i.e. something like `{"DAP-04": {/*remainder of capabilities*/}, "DAP-05": {/*remainder of capabilities*/}}`. This is because DAP capabilities are likely to vary by version, both in terms of what capabilities need to be represented by a given version, and by which capabilities are supported by any given implementation on a per-version basis.

branlwyd (changes requested):

aggregator_api/src/models.rs
jbr: I continue to disagree on this.

jbr:
Reopened with a singular `dap_version` to solve the immediate problem (histogram representation) instead of the uncertain future problem

branlwyd (approved):

divergentdave (approved):

aggregator_api/src/models.rs
jbr: Marking this as resolved, as the current PR only attempts to address single-protocol-version deployments

aggregator_api/src/routes.rs
```diff
@@ -33,6 +33,7 @@ pub(super) async fn get_config(
     State(config): State<Arc<Config>>,
 ) -> Json<AggregatorApiConfig> {
     Json(AggregatorApiConfig {
+        protocol: "DAP-05",
```
tgeoghegan, line 36: Janus skipped ahead to DAP-07
```suggestion
        protocol: "DAP-07",
```

aggregator_api/src/tests.rs
```diff
@@ -82,7 +82,7 @@ async fn get_config() {
             .await,
         Status::Ok,
         concat!(
-            r#"{"dap_url":"https://dap.url/","role":"Either","vdafs":"#,
+            r#"{"protocol":"DAP-05","dap_url":"https://dap.url/","role":"Either","vdafs":"#,
```
tgeoghegan, line 85: As elsewhere, we should advertise DAP-07
```suggestion
            r#"{"protocol":"DAP-07","dap_url":"https://dap.url/","role":"Either","vdafs":"#,
```

tgeoghegan (changes requested):

tgeoghegan (approved):

tgeoghegan:
This is failing CI because the PR was branched long enough ago that the SQL migrations clash (which is our fault for letting this perfectly good change languish for so long). @jbr, I believe that should go away if you rebase the change and force push. It's time to take this change so we can re-enable `divviup-api` based integration testing against DAP-07.

jbr:
Rebased

divergentdave (approved):

PR #1750, opened 2023-08-18
Add end-to-end test of aggregation using taskprov

divergentdave:
This adds another taskprov test that sends `AggregateInitializeReq`, `AggregateContinueReq`, and `AggregateShareReq` requests in sequence, and checks the results against a VDAF transcript. Each individual request already has test coverage, but this additional test mainly confirms that the fixtures set up in each test correctly model the actual outputs from the prior requests.


inahga (approved):
Excellent, thanks!

inahga:
It may worth forward-porting this to `main` as well.

Issue #1607, opened 2023-07-13
Refactor aggregator http_handler tests with rstest


inahga:
The standup process for tests in https://github.com/divviup/janus/blob/main/aggregator/src/aggregator/http_handlers.rs#L616 can be deduplicated with [rstest](https://docs.rs/rstest/latest/rstest/).

In particular, this pattern is repeated a lot.
```rust
async fn test() {
        install_test_trace_subscriber();

        let task = TaskBuilder::new(
            QueryType::TimeInterval,
            VdafInstance::Prio3Count,
            Role::Leader,
        )
        .build();
        let clock = MockClock::default();
        let ephemeral_datastore = ephemeral_datastore().await;
        let meter = noop_meter();
        let datastore = ephemeral_datastore.datastore(clock.clone(), &meter).await;
        ...
``

PR #1765, opened 2023-08-21
Aggregator test cleanup

inahga:
Closes https://github.com/divviup/janus/issues/1607. Mainly factor out test setup into its own function.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_continue.rs
+++ b/aggregator/src/aggregator/aggregation_job_continue.rs
@@ -278,6 +278,7 @@ impl VdafOps {
 #[cfg(feature = "test-util")]
 #[cfg_attr(docsrs, doc(cfg(feature = "test-util")))]
 pub mod test_util {
+    use crate::aggregator::http_handlers::test_util::{take_problem_details, take_response_body};
     use janus_aggregator_core::task::Task;
     use janus_messages::{AggregationJobContinueReq, AggregationJobId, AggregationJobResp};
     use prio::codec::{Decode, Encode};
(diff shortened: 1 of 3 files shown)


tgeoghegan (approved):
This change is good, but it looks like you didn't wind up using `rstest`. I guess all that would have done is add a messy `(MockClock, EphemeralDatastore, Arc<Datastore<MockClock>>, impl Handler)` argument to every test function, right?

inahga:
Indeed, at the time of issue writing thought it was a good candidate for rstest, but introduces a lot of noise for what is easier: a function returning a tuple.

aggregator/src/aggregator/aggregation_job_continue.rs
```diff
@@ -321,12 +322,7 @@ pub mod test_util {
                 .unwrap(),
             AggregationJobResp::MEDIA_TYPE
         );
-        let body_bytes = test_conn
-            .take_response_body()
-            .unwrap()
-            .into_bytes()
-            .await
-            .unwrap();
+        let body_bytes = take_response_body(&mut test_conn).await;
```
branlwyd, line 325: 100% of the places we use `take_response_body`, in this PR and in pre-existing code, we immediately decode the body bytes into some message via `get_decoded` (or, in a few places, a nearly-equivalent `decode` that should actually be a `get_decoded` to check that we didn't receive too much data).

WDYT about augmenting `take_response_body` to be generic on some `T: Decode` and return the decoded value, rather than the body bytes?

branlwyd (approved):

aggregator/src/aggregator/aggregation_job_continue.rs
inahga: Automerge beat me to it, but I'll explore this in another PR.

PR #1793, opened 2023-08-22
Release Janus 0.subscriber-01.3.

branlwyd:


inahga (approved):

tgeoghegan (changes requested):
You'll also need to update the versions in lines 32-41 of `Cargo.toml`. Then again, judging by the diff on `Cargo.lock`, perhaps those versions are being ignored?

divergentdave:
I don't think the dependency versions in Cargo.toml are being ignored, rather they define a minimum and maximum version range. Cargo.toml then records the actual versions in use. The different version numbers will have limited impact on this branch, since none of the crates will be published, so they'll always be resolved using their relative workspace paths instead.

branlwyd:
To match other release branches (& make our release instructions correct), I removed the final minor-minor version from the internal version specifications in our `Cargo.toml`. Assuming Rust can parse these versions properly, I think this should resolve the issue (though also, see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html -- `cargo update` should happily take our version past what is directly specified).

Issue #1796, opened 2023-08-22
Collector API redesign


divergentdave:
We will need to make the API of `janus_collector` more flexible in order for it to be used in stateful collector services, and recover from errors or downtime without unnecessarily losing data. Since draft-ietf-ppm-dap-04, collectors can no longer "rediscover" collection jobs by sending the same collection request multiple times, which makes the limitations of the current API worse.

The simplest straw man proposal to achieve this would be making the `janus_collector::CollectionJob` struct public, implementing `serde::Serialize` and `serde::Deserialize` on it, and making the methods `Collector::start_collection()` and `Collector::poll_once()`/`Collector::poll_until_complete()` public. This would allow consumers to save the collection job ID (along with other necessary parameters) and query it later beyond the scope of one Rust future.

Exposing a collection job structure would also let us address #762, adding an API to delete a collection job.

Per #1655, it would be natural to return a collection job structure when querying existing collection jobs by query with a Janus-specific API.

This would be good take care of before we cut 0.6.0.

branlwyd:
Exposing the collection job as part of the API is a good idea, and something like it is required to make a collector work out.

One question:

> Since draft-ietf-ppm-dap-04, collectors can no longer "rediscover" collection jobs by sending the same collection request multiple times,

Why not? I quickly scanned the diff between -03 & -04 and didn't spot anything -- I think resending the same collection request multiple times should be supported, and it's likely a spec bug if it is not supported.

divergentdave:
This is because of the problem described in #1655, we switched from the leader choosing collect URIs to the collector choosing collection job IDs. Note that sending the same collection request with the same collection job ID is idempotent, but if you forget the ID and remember the collection request, sending the same request under a fresh collection job ID results in Janus returning a "batch collected too many times" error.

branlwyd:
Ah, fair -- I think that can be worked around by having the collector store the collection job ID to durable storage before sending it the first time. (This would imply `janus_collector` taking a caller-provided `CollectionJobId` when starting a collection, rather than determining one randomly -- this may be what you had in mind already.)

branlwyd:
[...]

Issue #1655, opened 2023-07-28
Allow lookup of collection jobs by query and aggregation parameter


tgeoghegan:
DAP makes the collector responsible for choosing collection job IDs ([1](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-05#name-collection-job-initializati)). To poll collection job status and eventually obtain aggregate shares, the collector must keep track of the IDs it has chosen, and more to the point the mapping of collection job IDs to the query (e.g., an interval of time). Neither DAP nor Janus provide a means of discovering what collection jobs exist for a task, nor checking whether a collection job exists for some query and aggregation parameter.

The DAP API makes it the collector's responsibility to choose collection job IDs for two reasons:

- Consistency with the aggregation job API
- Collection job ID is an obvious idempotency key for the `PUT /tasks/{task-id}/collection_jobs/{collection-job-id}` endpoint

However, the fields of [`CollectionReq`](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-05#section-4.6.1-6) (namely the query and the aggregation parameter) are sufficient to uniquely identify a collection job. The collector has no choice but to keep track of those values, so collection job interfaces that operate on those would reduce the statefulness of collector implementations.

It's possible that DAP should revisit the design of the collection job API, but as a step in that direction, we could add a end API endpoint to Janus:

`QUERY /tasks/{task-id}/collection_jobs`

We use the [HTTP QUERY method](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html), which is just GET but with a request body. The body would be a `CollectionReq`. If some collection job exists in the task that matches the `CollectionReq`, then the response is HTTP 302 Found with a `Location` header containing the full collection job ID (i.e., `https://dap.example.org/tasks/{task-id}/collection_jobs/{collection-job-id}`.

We could also provide an endpoint that enumerates all available collection jobs in a task (something like `GET /tasks/{task-id}/collection_jobs`).

tgeoghegan:
@simon-friedberger this issue is based on our discussion yesterday. Please let me know what you think!

simon-friedberger:
The trigger for this was that it was previously possible to repeatedly query a specific time interval. Since the query is theoretically idempotent, this seems reasonable. But that's not possible with the current API because the Collector has to create a new collection job. This seems to fail because it creates another aggregation job and we currently always limit those with `max_batch_query_count = 1`.

[...]

PR #1799, opened 2023-08-23
Account max_batch_query_count against distinct requests.

branlwyd:
Before, the Leader would account for `max_batch_query_count` over all requests, so multiple identical requests would count as multiple queries.

If this is acceptable for Janus, we should consider a change to DAP to clarify that `max_batch_query_count` is accounted for by distinct aggregation parameters; currently, the text is ambiguous and permits both the preexisting behavior & the behavior in this PR.

Likely obviates the need for #1655. Inspired by & hopefully simplifies #1796.

### DIFF
--- a/aggregator/src/aggregator/collection_job_tests.rs
+++ b/aggregator/src/aggregator/collection_job_tests.rs
@@ -33,7 +33,7 @@ use janus_messages::{
 use prio::codec::{Decode, Encode};
 use rand::random;
 use serde_json::json;
-use std::sync::Arc;
+use std::{collections::HashSet, sync::Arc};
 use trillium::{Handler, KnownHeaderName, Status};
 use trillium_testing::{
     prelude::{post, put},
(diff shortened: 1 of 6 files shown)


tgeoghegan:
I think this is relevant to https://github.com/divviup/janus/issues/1655

aggregator/src/aggregator/collection_job_tests.rs
```diff
@@ -495,6 +495,65 @@ async fn collection_job_put_idempotence_time_interval() {
         .unwrap();
 }
 
+#[tokio::test]
+async fn collection_job_put_idempotence_time_interval_varied_collection_id() {
```
inahga, line 499: I think a test for fixed size is justified as well, since that code path was modified as well.

aggregator/src/aggregator/query_type.rs
```diff
@@ -114,11 +118,16 @@ impl CollectableQueryType for TimeInterval {
...
-        // https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.5.6
-        let intersecting_intervals: Vec<_> = match task.role() {
+        aggregation_param: &A::AggregationParam,
+    ) -> Result<(), datastore::Error>
+    where
+        A::AggregationParam: Send + Sync + Eq + Hash,
+    {
+        // Check how distinct aggregation parameters appear in rows in the relevant table with an
+        // intersecting batch interval. Each distinct aggregation parameter consumes one unit of
+        // query count. https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.5.6
```
tgeoghegan, line 128: Could you update this reference to link to a more recent DAP draft?

tgeoghegan (commented):
I think this change's reading of DAP is correct, but I wonder about the implications for the semantics of collection jobs. In this change, we allow two collection jobs to be created for the same query and aggregation parameter while consuming a single unit of max batch query count.

After that happens, what will the leader do with the collection jobs? Will it run them both? Will the helper agree to run both of those collection jobs, or will it refuse the second aggregate share request it gets?

If we allow the creation of two identical collection jobs with distinct IDs, how would we implement the API for looking up collection jobs by query and agg param, which is what #1655 requested? And incidentally, for that reason, I don't think this change _resolves_ #1655, because it doesn't add a new HTTP route for looking up collection jobs.

branlwyd:
> I think this change's reading of DAP is correct, but I wonder about the implications for the semantics of collection jobs. In this change, we allow two collection jobs to be created for the same query and aggregation parameter while consuming a single unit of max batch query count.
> 
> After that happens, what will the leader do with the collection jobs? Will it run them both?

Yes, it runs both -- though much of the work is deduplicated (as it will reuse existing aggregation jobs). Merging the results of the aggregation jobs into a final aggregate share will be re-done currently; arguably we should augment our logic to cache this data, as the Helper implementation does with aggregate share jobs, but this is optional (the cost is re-reading the aggregation job results & re-combining them into a final aggregate share).


> Will the helper agree to run both of those collection jobs, or will it refuse the second aggregate share request it gets?

The Helper must run both collection jobs: the Helper's interface to this is through the `AggregateShareReq`, which identifies the collected batch via a batch identifier & aggregation param (no collection job ID). These requests are required to be idempotent: https://www.ietf.org/archive/id/draft-ietf-ppm-dap-05.html#section-4.6.2-15


> If we allow the creation of two identical collection jobs with distinct IDs, how would we implement the API for looking up collection jobs by query and agg param, which is what #1655 requested? And incidentally, for that reason, I don't think this change _resolves_ #1655, because it doesn't add a new HTTP route for looking up collection jobs.

I think we shouldn't implement that API at all, instead restoring the functionality that led to the request in the first place -- the initial request for #1655 was actually restoring the ability to re-send identical collection requests. See https://github.com/divviup/janus/issues/1655#issuecomment-1658368789: "The trigger for this was that it was previously possible to repeatedly query a specific time interval. Since the query is theoretically idempotent, this seems reasonable. But that's not possible with the current API because the Collector has to create a new collection job."

I'm also concerned about implementing that API because it would not allow a Collector implementation to skip needing to track collection job IDs -- the Collector is still going to need to track state for a given collection job, e.g. have I read the results & written them to my storage. (Indeed, even without this fix I would recommend that a Collector implementation write collection job ID/batch ID/aggregation param, in a `START` state, to its own storage before sending any requests to the Aggregator -- that would also obviate the need for the API in #1655.)

This is a bugfix that restores functionality that was lost as we upgraded past DAP-02. I've removed the "resolved" comment, I'm going to comment on #1655 to see if we can close.

tgeoghegan (approved):
I understand how this solves Simon's problem now.

branlwyd:
I filed https://github.com/divviup/janus/issues/1802 to track caching of aggregate share information by the Leader, similar to what is already done by the Helper.

Issue #1628, opened 2023-07-21
OpenTelemetry trace propagation


divergentdave:
We should enable OpenTelemetry tracing systems to link spans for collection requests to spans in the collection job driver, and spans in the aggregation job creator and aggregation job driver.

I'm still wrapping my head around the available interfaces. Most distributed tracing use cases are HTTP-centric, and propagate this information through the `tracestate` and `traceparent` HTTP headers. If possible, I'd like to just use aggregation job IDs and collection job IDs as a trace ID directly. (cutting corners by not propagating per-system IDs) If that doesn't work out, worst case we can add a `jsonb` column to each job table to act as a carrier.

divergentdave:
I think we do need `jsonb` columns in the database for trait propagation after all. For two spans to be connected, they need to share a 16-byte trace ID and 8-byte span ID (either on the span itself and its parent span ID, or in a link). Additionally, we need to be able to transfer the span flags field in order to make sampling work correctly. For example, the aggregator process could be configured to sample 1% of its root spans, then the "sampled" flag in any one collection job's trace will determine whether all subsequent related job processing spans are exported or not.

PR #1800, opened 2023-08-24
Add database columns to jobs for W3C trace context

divergentdave:
This adds `jsonb` database columns to aggregation jobs and collection jobs to propagate distributed tracing metadata through our job queue system. This is part of #1628.


tgeoghegan (approved):

branlwyd (changes requested):
As this is on `main` for an as-yet-unreleased version, the schema change should be phrased as a change to the `initial_schema` script rather than a migration.

branlwyd (approved):

branlwyd:
(the label might be too late unless you want to try pushing again, poke me if you'd like me to merge.)

PR #1804, opened 2023-08-24
Use ContainerLogsDropGuard instead of custom Drop impls

inahga:
Minor cleanup: use [`ContainerLogsDropGuard`](https://github.com/divviup/janus/blob/main/interop_binaries/src/lib.rs#L391) instead of these custom Drop impls. It's already used for the [client container](https://github.com/divviup/janus/blob/main/integration_tests/src/client.rs#L189), not sure why it's not used for the other containers.

### DIFF
--- a/integration_tests/src/client.rs
+++ b/integration_tests/src/client.rs
@@ -265,7 +265,7 @@ where
                 .with_network(network)
                 .with_container_name(client_container_name),
         );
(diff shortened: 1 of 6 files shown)


branlwyd (approved):
I'd say induce a test-failure to see that this works if you haven't already, otherwise LGTM.

PR #1741, opened 2023-08-16
Draft: DAP-05 ping-pong

tgeoghegan:
Please do not review yet. I just want to see if CI works.

Issue #1524, opened 2023-06-21
Simplify `aggregator_core::task::Task`


tgeoghegan:
Janus' task representation contains a bunch of unnecessary information:

- an aggregator never needs to know its own DAP API URL
- a helper doesn't need to know the other aggregator's DAP API URL
- a task cannot have more than one VDAF verify key
- a task cannot have multiple collector auth tokens

And it's possible a task will only ever have a single aggregator auth token, too (we have to decide whether to support rotation of that credential). As we revisit the task creation API in the aggregator API (#1506), this leads to some awkward impedance mismatches. We should resolve those by amending the `Task` struct.

This goes hand-in-hand with changing the database schema for tasks, tracked in #1521.

Note that we serialize this structure to/from YAML when provisioning tasks with `janus_cli`, so as long as we intend to support task provisioning with that tool, the `Task` struct is arguably public API, so we should be careful about breaking this.

PR #1805, opened 2023-08-24
DAP-05: Specialize to exactly two aggregators

tgeoghegan:
This commit takes changes to various messages so that we handle exactly one leader and exactly one helper, as specified in [DAP-05][1]. These changes are voluminous but mostly mechanical.

This commit also adds variants for `Poplar1` to a couple of enums and some macros, because later changes will need it.

[1]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/05/

Part of #1669

### DIFF
--- a/aggregator/src/aggregator/collection_job_tests.rs
+++ b/aggregator/src/aggregator/collection_job_tests.rs
@@ -1,10 +1,4 @@
-use crate::aggregator::{
-    http_handlers::{
-        aggregator_handler,
-        test_util::{decode_response_body, take_problem_details},
-    },
-    Config,
-};
+use crate::aggregator::{http_handlers::aggregator_handler, Config};
 use http::StatusCode;
 use janus_aggregator_core::{
     datastore::{
@@ -42,7 +36,6 @@ use serde_json::json;
 use std::{collections::HashSet, sync::Arc};
 use trillium::{Handler, KnownHeaderName, Status};
 use trillium_testing::{
-    assert_headers,
     prelude::{post, put},
     TestConn,
 };
@@ -353,10 +346,23 @@ async fn collection_job_success_fixed_size() {
         }
 
         let mut test_conn = test_case.post_collection_job(&collection_job_id).await;
-        assert_eq!(test_conn.status(), Some(Status::Ok));
-        assert_headers!(&test_conn, "content-type" => (Collection::<FixedSize>::MEDIA_TYPE));
 
-        let collect_resp: Collection<FixedSize> = decode_response_body(&mut test_conn).await;
+        assert_eq!(test_conn.status(), Some(Status::Ok));
+        assert_eq!(
+            test_conn
+                .response_headers()
+                .get(KnownHeaderName::ContentType)
+                .unwrap(),
+            Collection::<FixedSize>::MEDIA_TYPE
+        );
+        let body_bytes = test_conn
+            .take_response_body()
+            .unwrap()
+            .into_bytes()
+            .await
+            .unwrap();
+        let collect_resp = Collection::<FixedSize>::get_decoded(body_bytes.as_ref()).unwrap();
+
         assert_eq!(
             collect_resp.report_count(),
             test_case.task.min_batch_size() + 1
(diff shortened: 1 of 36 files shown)


tgeoghegan:
I carved this change out of #1741 so that the noisy mechanical change of going to exactly two aggregators wouldn't drown out the more substantial changes from moving to the ping-pong topology.

tgeoghegan:
Looks like we need `Poplar1AggregationParam: Hash` for this to work: https://github.com/divviup/libprio-rs/pull/703

- [ ] move prio dep back to a released version once the above PR is merged and released

core/Cargo.toml
```diff
@@ -49,7 +49,7 @@ k8s-openapi = { workspace = true, optional = true }
 lazy_static = { version = "1", optional = true }
 prio.workspace = true
 rand = "0.8"
-reqwest = { version = "0.11.19", default-features = false, features = ["rustls-tls", "json"] }
+reqwest = { version = "0.11.18", default-features = false, features = ["rustls-tls", "json"] }
```
branlwyd, line 52: nit: revert this version change

aggregator/src/aggregator/collection_job_tests.rs
```diff
@@ -417,7 +422,23 @@ async fn collection_job_success_fixed_size() {
...
+        &test_conn
+            .take_response_body()
+            .unwrap()
+            .into_bytes()
+            .await
+            .unwrap(),
+    )
+    .unwrap();
+    assert_eq!(
+        problem_details,
```
inahga, line 441: This might be on an outdated branch? This was cleaned up in https://github.com/divviup/janus/commit/a76f06ea539e1ea8b119c8ce54f479e778cd2677 to use utility functions.

aggregator_api/src/routes.rs
```diff
@@ -93,9 +93,9 @@ pub(super) async fn post_task<C: Clock>(
     // TODO(#1524): clean this up with `aggregator_core::task::Task` changes
     // unwrap safety: this fake URL is valid
     let fake_aggregator_url = Url::parse("http://never-used.example.com").unwrap();
-    let aggregator_endpoints = match req.role {
-        Role::Leader => Vec::from([fake_aggregator_url, req.peer_aggregator_endpoint]),
-        Role::Helper => Vec::from([req.peer_aggregator_endpoint, fake_aggregator_url]),
+    let (leader_aggregator_endpoint, helper_aggregator_endpoint) = match req.role {
+        Role::Leader => (fake_aggregator_url, req.peer_aggregator_endpoint),
+        Role::Helper => (req.peer_aggregator_endpoint, fake_aggregator_url),
```
inahga, line 98: While we're here, should we make both aggregator endpoints Option (or tagged enum, or even expand role into tagged enum `enum Role { Leader { endpoint: Url ...`) so that we don't have to carry around a fake aggregator endpoint? 

It does sound like a large yak-shave, so if it takes too much time or is too complex, we can leave it alone.

messages/Cargo.toml
```diff
@@ -20,7 +20,9 @@ hex = "0.4"
 num_enum = "0.7.0"
 # We can't pull prio in from the workspace because that would enable default features, and we do not
 # want prio/crypto-dependencies
-prio = { version = "0.14.0", default-features = false }
+# TODO(timg): move back to released prio
```
branlwyd, line 23: reminder: need to resolve this TODO, as well (same as other reminder)

inahga:
If you push again, this label ^^ should allow the ci-migrations build to pass.

branlwyd (approved):

aggregator_api/src/routes.rs
branlwyd: I think the eventual cleanup is to make our view of tasks track what is needed by the particular role; specifically, a Leader aggregator needs to know only the Helper's aggregator endpoint, and a Helper doesn't need to know anyone's aggregator endpoint. This is tracked by #1524.

aggregator/src/aggregator/collection_job_tests.rs
tgeoghegan: Yes, this is rebase carnage. I'll clean it up.

aggregator_api/src/routes.rs
tgeoghegan: Yes, #1524 will address this. That change will be pretty significant, so I don't want to do it here.

PR #1809, opened 2023-08-25
Add validation to enforce listen_address and path_prefix mutual exclusion

inahga:
From discussion in https://isrg.slack.com/archives/C0167LT4C73/p1692908321449989, we noticed that providing both `listen_address` and `path_prefix` results in `listen_address` being prioritized, which is misleading. It should be an error to provide both fields.

The quick and easy solution to this is to add `#[serde(deny_unknown_fields)]` to the `AggregatorApi` enum. However, this results in an unfriendly error message: `data did not match any variant of untagged enum AggregatorApi`, which IMO is essentially meaningless to an operator without knowledge of Janus internals. Instead, we implement a custom deserializer that performs additional validation and lets us have a nicer error message.

### DIFF
--- a/aggregator/src/bin/aggregator.rs
+++ b/aggregator/src/bin/aggregator.rs
@@ -208,13 +208,13 @@ pub struct HeaderEntry {
 pub struct AggregatorApi {
     /// Address on which this server should listen for connections to the Janus aggregator API
     /// and serve its API endpoints, independently from the address on which the DAP API is
-    /// served. This is mutually exclusive with [`path_prefix`].
+    /// served. This is mutually exclusive with `path_prefix`.
     listen_address: Option<SocketAddr>,
     /// The Janus aggregator API will be served on the same address as the DAP API, but relative
     /// to the provided prefix. e.g., if `path_prefix` is `aggregator-api`, then the DAP API's
     /// uploads endpoint would be `{listen-address}/tasks/{task-id}/reports`, while task IDs
     /// could be obtained from the aggregator API at `{listen-address}/aggregator-api/task_ids`.
-    /// This is mutually exclusive with [`listen_address`].
+    /// This is mutually exclusive with `listen_address`.
     path_prefix: Option<String>,
     /// Resource location at which the DAP service managed by this aggregator api can be found
     /// on the public internet. Required.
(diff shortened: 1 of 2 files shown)


docs/samples/advanced_config/aggregator.yaml
```diff
@@ -69,9 +69,10 @@ listen_address: "0.0.0.0:8080"
 
 # How to serve the Janus aggregator API. If not set, Janus aggregator API is not served. (optional)
 aggregator_api:
-  # Socket address on which to listen for requests.
+  # Serve the aggregator API on an address and port that is separate from the DAP API.
   listen_address: "0.0.0.0:8081"
-  # Alternately, the aggregator API may be served on `listen_address`, at an arbitrary path prefix.
+  # Alternatively, serve the aggregator API on the same address as the DAP API, but on a separate
```
branlwyd, line 74: super-nit, not caused by this PR: add newline above this line

aggregator/src/bin/aggregator.rs
```diff
@@ -209,36 +204,45 @@ pub struct HeaderEntry {
...
-        /// uploads endpoint would be `{listen-address}/tasks/{task-id}/reports`, while task IDs
-        /// could be obtained from the aggregator API at `{listen-address}/aggregator-api/task_ids`.
-        path_prefix: String,
-        /// Resource location at which the DAP service managed by this aggregator api can be found
-        /// on the public internet. Required.
-        public_dap_url: Url,
-    },
+#[serde(deny_unknown_fields)]
+pub struct AggregatorApi {
+    /// Address on which this server should listen for connections to the Janus aggregator API
```
branlwyd, line 209: IMO, note in the doccomments that these fields are mutually exclusive.

branlwyd (approved):

aggregator/src/bin/aggregator.rs
```diff
@@ -209,36 +204,45 @@ pub struct HeaderEntry {
...
-        match self {
-            AggregatorApi::ListenAddress { public_dap_url, .. } => public_dap_url,
-            AggregatorApi::PathPrefix { public_dap_url, .. } => public_dap_url,
+fn deserialize_aggregator_api<'de, D>(deserializer: D) -> Result<Option<AggregatorApi>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let aggregator_api: Option<AggregatorApi> =
+        serde_yaml::from_value(Deserialize::deserialize(deserializer)?)
+            .map_err(de::Error::custom)?;
```
divergentdave, line 229: This can be simplified by using the deserializer directly. (ideally, we'd like the body of Serialize and Deserialize implementations to be independent of any serializer or deserializer)
```suggestion
    let aggregator_api: Option<AggregatorApi> = Deserialize::deserialize(deserializer)?;
```

divergentdave (approved):

PR #1812, opened 2023-08-27
janus_collector: Store collection job ID, not URL

divergentdave:
This re-applies #1797, which was partially rolled back by #1805, likely due to conflict resolution issues.


branlwyd (approved):
apologies, I missed this during review :-|

PR #1844, opened 2023-08-28
`collect` tool: accept HPKE keypair from `divviup`

### LINKED ISSUE #1810: `collect` binary should accept HPKE configs in the format that `divviup hpke-config generate` outputs
`divviup hpke-config generate` emits an HPKE config object that looks like:

```
{
  "aead": "AesGcm128",
  "id": 202,
  "kdf": "Sha256",
  "kem": "X25519HkdfSha256",
  "private_key": "AAAAAAAAA",
  "public_key": "jlk7lPLnIw4wvoLjWLreF0HVhpK1W0W4ecdPkvEiND8"
}
```

The `collect` binary takes an HPKE config, but expects an [`HpkeConfig` DAP message](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-05#section-4.4.1), encoded in base64url. This is an awkward impedance mismatch. Ahead of a more coherent, end-to-end story for how collection works, we can make this a bit nicer by having `collect` accept HPKE configs in the format that `divviup hpke-config generate` outputs.

tgeoghegan:
Adds a new option to `collect` so that an HPKE keypair encoded in JSON by `divviup hpke-config generate` may be used instead of providing the Base64URL unpadded private key plus the RFC 8446 encoding of the HPKE config.

Resolves #1810

### DIFF
--- a/core/src/hpke.rs
+++ b/core/src/hpke.rs
@@ -2,7 +2,7 @@
 
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use derivative::Derivative;
-use hpke_dispatch::{HpkeError, Kem, Keypair};
+use hpke_dispatch::{Aead, HpkeError, Kdf, Kem, Keypair};
 use janus_messages::{
     HpkeAeadId, HpkeCiphertext, HpkeConfig, HpkeConfigId, HpkeKdfId, HpkeKemId, HpkePublicKey, Role,
 };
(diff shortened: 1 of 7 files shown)


tools/tests/cmd/collect.trycmd
```diff
@@ -26,6 +26,9 @@ DAP Task Parameters:
           
           [env: HPKE_PRIVATE_KEY=]
 
+      --hpke-config-json <HPKE_CONFIG_JSON>
```
divergentdave, line 29: How about `--hpke-key-json`? For those in the know, "HPKE config" should only refer to the public part, while for the uninitiated, the word "key" indicates that the file is crypto-related. On the other hand, the control plane and CLI do talk about HPKE configs.

core/src/hpke.rs
```diff
@@ -253,13 +253,59 @@ impl HpkeKeypair {
...
+pub struct DivviUpHpkeConfig {
+    id: HpkeConfigId,
+    kem: Kem,
+    kdf: Kdf,
+    aead: Aead,
+    public_key: HpkePublicKey,
+    private_key: HpkePrivateKey,
+}
+
+// We use a fallible TryFrom conversion because it's possible that HpkeDispatch, Janus and
```
jbr, line 267: in this case, if the struct stores `String`s for each of the values, the `Kem/Kdf/Aead as TryFrom<str>` implementations in the latest hpke-config (0.5.1) are far more permissive than serde

tools/src/bin/collect.rs
```diff
@@ -295,19 +296,30 @@ struct Options {
...
+    hpke_config: Option<HpkeConfig>,
     /// The collector's HPKE private key, encoded with base64url
     #[clap(
         long,
         value_parser = PrivateKeyValueParser::new(),
         env,
         help_heading = "DAP Task Parameters",
-        display_order = 3
+        display_order = 3,
+        conflicts_with = "hpke_config_json"
```
divergentdave, line 310: We should make the two old arguments require each other
```suggestion
        conflicts_with = "hpke_config_json"
        requires = "hpke_config"
```

tools/src/bin/collect.rs
```diff
@@ -295,19 +296,30 @@ struct Options {
         long,
         value_parser = HpkeConfigValueParser::new(),
         help_heading = "DAP Task Parameters",
-        display_order = 2
+        display_order = 2,
+        conflicts_with = "hpke_config_json"
```
divergentdave, line 300: See below
```suggestion
        conflicts_with = "hpke_config_json"
        requires = "hpke_private_key"
```

divergentdave (approved):

Issue #1842, opened 2023-08-28
Move build information from janus_build_info to target_info


divergentdave:
Since upgrading to opentelemetry-rust v0.20, we once again have the problem that gauges that are only observed once appear in one metrics scrape, and then do not appear subsequently. We'd like `janus_build_info` to always be present in metrics scrapes, so we can tell what versions of components are running from a single instant query. Instead of trying to fix this, we could instead add the version information as `Resource` attributes, so that they appear in the `target_info` metric. (see the [specification](https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#resource-attributes-1)) This metric is output by special-case code that runs every time metrics are gathered, bypassing OpenTelemetry instrument handling, so it won't be affected by the behavior change like `janus_build_info`.

divergentdave:
Done in #1845.

PR #1845, opened 2023-08-28
Move version information to OTel resource

divergentdave:
This implements #1842, moving build-time version information from a custom `janus_build_info` gauge into `target_info`, which is produced by the `opentelemetry-prometheus` exporter. I tried to follow the spirit of the semantic conventions for resource attributes, relating to [`service.version`](https://opentelemetry.io/docs/specs/otel/resource/semantic_conventions/#service-experimental) and [`process.runtime.name` and `process.runtime.version`](https://opentelemetry.io/docs/specs/otel/resource/semantic_conventions/process/#process-runtimes).

Note that there's only one `service.version` attribute defined, while we previously recorded both a semantic version and a git hash separately. The conventions for [version attributes](https://opentelemetry.io/docs/specs/otel/resource/semantic_conventions/#version-attributes) say that the value may be an arbitrary identifier, so I decided to interpolate both the crate's semantic version and the git commit hash, rather than come up with a different attribute name for one or both values.

I ran this with the environment variable `OTEL_SERVICE_NAME=aggregator` set, and I get the following in the response. It sticks around after waiting and re-scraping.
```
# HELP target_info Target metadata
# TYPE target_info gauge
target_info{process_runtime_name="Rust",process_runtime_version="1.72.0",service_name="aggregator",service_version="0.5.14-c529b01f"} 1
```

### DIFF
--- a/aggregator/src/bin/janus_cli.rs
+++ b/aggregator/src/bin/janus_cli.rs
@@ -2,9 +2,7 @@ use anyhow::{anyhow, Context, Result};
 use base64::{engine::general_purpose::STANDARD_NO_PAD, Engine};
 use clap::Parser;
 use janus_aggregator::{
-    binary_utils::{
-        database_pool, datastore, read_config, record_build_info_gauge, CommonBinaryOptions,
-    },
+    binary_utils::{database_pool, datastore, read_config, CommonBinaryOptions},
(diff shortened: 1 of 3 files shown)


tgeoghegan (approved):
LGTM. This will need to be backported to `release/0.5` and `release/0.subscriber-01`, right?

PR #1853, opened 2023-08-29
Adopt DAP-05 error types

tgeoghegan:
DAP-05 changes a few error types.

- `unrecongizedMessage -> invalidMessage`
- `queryMismatch` is gone (now we use `invalidMessage` in that case)
- `roundMismatch -> stepMismatch`

We still use the word "round" in several places where "step" would be more appropriate given DAP-05 text. Renaming those variables, etc., will arrive in a later commit to avoid adding unnecessary noise here.

Part of #1669

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -290,9 +290,9 @@ impl<C: Clock> Aggregator<C> {
                 Some(task_id_base64) => {
                     let task_id_bytes = URL_SAFE_NO_PAD
                         .decode(task_id_base64)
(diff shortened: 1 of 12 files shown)


inahga (approved):
I reviewed only [775724d](https://github.com/divviup/janus/pull/1853/commits/775724dc6ff114265979ce117e586c8af41ae40c), and https://github.com/divviup/janus/pull/1853/commits/f00595d4f29f503fcd72a4237872e5219e615da5, which LGTM.

PR #1859, opened 2023-08-30
`collect` tool: accept HPKE keypair from `divviup`

tgeoghegan:
Adds a new option to `collect` so that an HPKE keypair encoded in JSON by `divviup hpke-config generate` may be used instead of providing the Base64URL unpadded private key plus the RFC 8446 encoding of the HPKE config.

Backport of #1844

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -40,7 +40,7 @@ chrono = { workspace = true, features = ["clock"] }
 derivative = "2.2.0"
 futures = "0.3.28"
 hex = "0.4"
-hpke-dispatch = "0.5.1"
+hpke-dispatch = { version = "0.5.1", features = ["serde"] }
 http = "0.2.9"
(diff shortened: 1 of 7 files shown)


core/Cargo.toml
```diff
@@ -40,7 +40,7 @@ chrono = { workspace = true, features = ["clock"] }
 derivative = "2.2.0"
 futures = "0.3.28"
 hex = "0.4"
-hpke-dispatch = "0.5.1"
+hpke-dispatch = { version = "0.5.0", features = ["serde"] }
```
divergentdave, line 43: We may as well leave the version requirement as-is. (note that the lockfile is still pointing at 0.5.1)
```suggestion
hpke-dispatch = { version = "0.5.1", features = ["serde"] }
```

divergentdave (approved):

PR #1866, opened 2023-08-31
Set up Dependabot groups

divergentdave:
This sets up several groups in our Dependabot configuration for related crates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -54,3 +54,34 @@ updates:
     schedule:
       interval: "weekly"
     target-branch: "release/0.5"
+
+groups:
+  serde:
+    patterns:
+      - serde
+      - serde_*
+  tracing:
+    patterns:
+      - tracing
+      - tracing-*
+  opentelemetry:
+    patterns:
+      - opentelemetry
+      - opentelemetry-*
+      - tonic
+  trillium:
+    patterns:
+      - trillium
+      - trillium-*
+  kube:
+    patterns:
+      - kube
+      - k8s-openapi
+  postgres:
+    patterns:
+      - tokio-postgres
+      - postgres-*
+  error-handling:
+    patterns:
+      - thiserror
+      - anyhow


.github/dependabot.yml
```diff
@@ -54,3 +54,34 @@ updates:
...
+    patterns:
+      - trillium
+      - trillium-*
+  kube:
+    patterns:
+      - kube
+      - k8s-openapi
+  postgres:
+    patterns:
+      - tokio-postgres
```
tgeoghegan, line 82: Should we add a `[tokio, tokio-*]` group? 8 such crates show up in Cargo.lock (7, if we exclude `tokio-postgres` which is covered in the `postgres` group.

tgeoghegan (approved):

jbr:
Does this mean there's a way to have it open/update a single PR with all changes, if the group is *?

.github/dependabot.yml
divergentdave: I don't think so, we have only ever gotten Dependabot PRs for tokio out of that group. IIUC we don't get version updates because they are not direct dependencies, while we would still get security updates on them as transitive dependencies. The grouping feature only applies to version updates. (plus, the churn on tokio-stream and tokio-util is pretty low)

divergentdave:
> Does this mean there's a way to have it open/update a single PR with all changes, if the group is *?

I think so. There are also other filters available, to make groups for all patch-level updates, all updates to development dependencies, etc.

divergentdave:
This is annoying, the file is failing to parse with the following error (when checking https://github.com/divviup/janus/network/updates)

> The property '#/' contains additional properties ["groups"] outside of the schema when none are allowed

The [documentation](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) includes one example with the same structure (the "npm" one). I'll file a follow-up to copy and paste the groups configuration in the right place.

PR #1867, opened 2023-08-31
Fix Dependabot groups configuration

divergentdave:
This is a follow-up to #1866 to fix the structure of the configuration file.


tgeoghegan (approved):

tgeoghegan:
Oof, that _is_ annoying.

Issue #1860, opened 2023-08-30
Better error when collection can't be created due to insufficient prepared reports


tgeoghegan:
If a Collector tries to create a collection _after_ a sufficient number of uploads have happened but _before_ Janus asynchronously schedules and runs the relevant aggregation jobs, then `collect` yields this error:

```
collect --task-id <snip> --leader 'https://dap-04.api.divviup.org/' --vdaf count --current-batch`
Error: HTTP response status 400 Bad Request - The batch implied by the query is invalid.
```

This isn't very helpful. The query is actually perfectly valid, it's just that the reports that it captures aren't aggregated yet. Janus should be able to distinguish the case from the one where not enough reports have been uploaded (we have the `client_reports` and `report_aggregations` rows, they're just in the wrong state), so we should be able to provide a better error and even a suggestion of what to do: try again in a few minutes after the next time `aggregation_job_{creator,driver}` runs. We could even make Janus aware of the scheduling parameters for the `aggregation_job_*` components so that it could provide a corresponding `Retry-After` header.

A related case is where the Collector tries to collect before enough reports have been uploaded. In that case, there's no way to provide a useful `Retry-After` -- Janus has no idea when uploads will happen -- but at least the error could specifically explain that there aren't enough uploaded reports rather than the broad "query is invalid".

divergentdave:
#989 is related, but I vaguely recall we had a previous conversation somewhere else about spec interpretation on this point.

We could provide this additional guidance in the "detail" field of the problem details response, and then have the CLI print that as well.

branlwyd:
This is indeed confusing, but I think that the behavior in Janus could be changed such that this error goes away entirely (as well as the related error regarding attempting to collect before enough reports are uploaded).

Currently, upon receipt of a collection job creation request with a `current-batch` query, Janus will greedily attempt to select a batch which has not yet been collected & which has enough reports associated with it (via aggregation job creation) to allow collection. If no such batch is available, we terminate the request with the error in the OP.

Instead, we could accept this request, noting that it is a `current-batch` request. Then, when we attempt to step the collection job, one of the first steps would be to select a specific batch ID for this collection job to collect. That is, we would _lazily_ select a specific batch to collect for each `current-batch` collection job. This would allow the Collector to issue requests at their leisure, and then poll them until they complete -- delays due to report uploads & aggregation job creation delays would effectively be hidden inside the polling process.

[...]

PR #1878, opened 2023-09-01
Check collection job equivalence by query, not batch ID.

### LINKED ISSUE #1877: Janus collection checks allow fixed-size batches to be lost
Specifically, the data loss scenario looks like:
* Collector generates a collection ID to send, stores it to durable storage (along with an aggregation parameter, if applicable to the VDAF in use).
* Collector issues `current-batch` request with the collection ID / aggregation parameter from the previous request. Janus creates a collection job based on this request.
* Collector crashes, losing state, before it can note that the collection job has been created.

At this point, the Collector has effectively lost this batch: it can't reissue the `current-batch` request because idempotency checking is based on the batch ID that was selected for the collection by the Leader, rather than the query sent by the Collector [1]; the Collector does not know this batch ID.

(another bug around this is that we technically fail to be DAP-compliant: DAP states "Changing a collection job's parameters is illegal, so further requests to `PUT /tasks/{tasks}/collection_jobs/{collection-job-id}` for the same `collection-job-id` but with a different `CollectionReq` in the body `MUST` fail with an HTTP client error status code." Currently, Janus would accept a `PUT` of a `current-batch` query, followed by a `PUT` of a `by-batch-id` query for the selected batch, against the same collection job ID. This is a marginal bug because, without extra-DAP communications channels, there is no way for the Collector to know the correct batch ID.)

One workaround a Collector in this situation could do would be to attempt to blindly poll the collection job, then if that poll returns unrecognized-collection-job, give up or re-attempt creation of the collection job at that point. I think this is unsuitable because it adds complexity to the Collector to work around a Janus implementation choice, and IMO Collector implementors are unlikely to realize this complexity is required until they encounter this issue.

My suggested fix would be for Janus to store the query used by the collection job, as well as the assigned batch ID which is currently stored. Then use the query for request-deduplication purposes. This would solve this issue & would be one step along the path to implementing https://github.com/divviup/janus/issues/1860#issuecomment-1699992694.

[1] https://github.com/divviup/janus/blob/b28b149aa9a1e3e2e3ec6315ccff590585dd963c/aggregator/src/aggregator.rs#L2219-L2220

branlwyd:
This allows current-batch requests to be reissued in fixed-size tasks, simplifying implementation of a Collector & avoiding the potential for loss of a collection job due to an inopportunely-timed task failure.

While I'm at it, I also update `get_collection_job` to take a task ID in addition to a collection job ID. This provides isolation between different tasks, which is especially imporant now that the Collector generates the collection job IDs -- otherwise, it would be trivial for Collectors to generate the same collection job ID for two tasks, which would cause some very strange behavior in Janus.

Resolves #1877. Part of #1860.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -2213,10 +2213,10 @@ impl VdafOps {
 
                     // Check if this collection job already exists, ensuring that all parameters match.
                     if let Some(collection_job) = tx
-                        .get_collection_job::<SEED_SIZE, Q, A>(&vdaf, &collection_job_id)
+                        .get_collection_job::<SEED_SIZE, Q, A>(&vdaf, task.id(), &collection_job_id)
                         .await?
                     {
-                        if collection_job.batch_identifier() == &collection_identifier
+                        if collection_job.query() == req.query()
                             && collection_job.aggregation_parameter() == aggregation_param.as_ref()
                         {
                             debug!(
@@ -2366,8 +2366,9 @@ impl VdafOps {
                     let collection_job = CollectionJob::<SEED_SIZE, Q, A>::new(
                         *task.id(),
                         collection_job_id,
-                        collection_identifier,
+                        req.query().clone(),
                         aggregation_param.as_ref().clone(),
+                        collection_identifier,
                         initial_collection_job_state,
                     );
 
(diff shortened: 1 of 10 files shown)


inahga (approved):
> While I'm at it, I also update get_collection_job to take a task ID in addition to a collection job ID. This provides isolation between different tasks, which is especially imporant now that the Collector generates the collection job IDs -- otherwise, it would be trivial for Collectors to generate the same collection job ID for two tasks, which would cause some very strange behavior in Janus.

I imagine this should be backported to our other active branches?

branlwyd:
> I imagine [the get_collection_job change] should be backported to our other active branches?

Yes, I think this change should be backported.

The batch-equivalence check should probably also be backported; we must backport it if we want to implement https://github.com/divviup/janus/issues/1860#issuecomment-1699992694 (since implementing that requires knowing the query that was used to generate a collection job).

PR #1879, opened 2023-09-02
Do not compare preparation states in non-test code.

branlwyd:
This required a few updates to non-test code:

* The aggregation job creator used equality checks to determine if a given report aggregation was in state `START`. These are easily replaced by a call to `matches!`, which does not require `PartialEq + Eq`.

* Checking if a repeated aggregation job `PUT` matched the preexisting aggregation job checked if the report aggregations were equal. This is replaced by populating the last request hash for initialization requests (as well as continuation requests) and checking if the aggregation job is equal; since the aggregation job carries both the round number as well as the request hash, this should suffice to check that the requests are equal. As a bonus, we no longer need to read the preexisting report aggregations.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1502,10 +1502,8 @@ impl VdafOps {
     /// the sense that no new rows would need to be written to service the job.
     async fn check_aggregation_job_idempotence<'b, const SEED_SIZE: usize, Q, A, C>(
         tx: &Transaction<'b, C>,
-        vdaf: &A,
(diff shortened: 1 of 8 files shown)


inahga (approved):

branlwyd:
(removing resolution of https://github.com/divviup/libprio-rs/issues/722 as that issue is for libprio-rs, and we should do the work to remove comparability of relevant types from libprio-rs before counting that issue done. this is just the application-level change allowing the libprio-rs change to be made.)

PR #1916, opened 2023-09-12
Adopt `prio` 0.15.0

tgeoghegan:


messages/Cargo.toml
```diff
@@ -20,7 +20,7 @@ hex = "0.4"
 num_enum = "0.7.0"
 # We can't pull prio in from the workspace because that would enable default features, and we do not
 # want prio/crypto-dependencies
-prio = { version = "0.14.1", features = ["multithreaded"] }
+prio = { version = "0.15.0", default-features = false, features = ["multithreaded"] }
```
tgeoghegan, line 23: This fixes a bug I introduced in #1805: we should not enable default features on `prio` in `janus_messages`.

inahga (approved):

aggregator/src/aggregator.rs
```diff
@@ -807,13 +812,18 @@ impl<C: Clock> TaskAggregator<C> {
             }
 
             VdafInstance::Prio3SumVec { bits, length } => {
-                let vdaf = Prio3::new_sum_vec_multithreaded(2, *bits, *length)?;
+                let vdaf = Prio3::new_sum_vec_multithreaded(
+                    2,
+                    *bits,
+                    *length,
+                    VdafInstance::chunk_size(*length),
```
divergentdave, line 819:
```suggestion
                    VdafInstance::chunk_size(*bits * *length),
```

divergentdave (commented):
I have one suggestion in several places regarding the chunk length for Prio3SumVec. (This will mostly evaporate once we outsource the chunk length to whoever chooses task parameters)

PR #1919, opened 2023-09-13
Update rustls-webpki

divergentdave:
This will resolve a CPU DoS security alert.


inahga (approved):

inahga (commented):
I'm somewhat surprised dependabot didn't make a PR for this... is it because it's a transient dependency?

divergentdave:
It did previously create #1791. We have two incompatible versions of this crate in our lockfile, and this updates the other one. It appears the advisory has also been updated since then to reflect that subsequent releases had incomplete fixes.

PR #1922, opened 2023-09-13
Replace "round" with "step"

tgeoghegan:
When DAP-05 introduced the ping-pong topology, it also started talking about aggregation protocol "steps" instead of "rounds". This commit corrects the word usage across Janus to line up with the specification.

This change is pretty pedantic, but I think it's worth it for readers trying to resolve Janus against the DAP specification.

Relevant to #1669

### DIFF
--- a/aggregator/src/aggregator/aggregate_init_tests.rs
+++ b/aggregator/src/aggregator/aggregate_init_tests.rs
@@ -448,7 +448,7 @@ async fn aggregation_job_mutation_report_aggregations() {
 }
 
 #[tokio::test]
-async fn aggregation_job_init_two_round_vdaf_idempotence() {
+async fn aggregation_job_init_two_step_vdaf_idempotence() {
     // We must run Poplar1 in this test so that the aggregation job won't finish on the first step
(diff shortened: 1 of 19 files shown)


inahga (approved):
Not pedantic at all, `round` meaning `step` is not obvious, bound to confuse future us or newcomers if not changed.

Uber nit: on messages/src/lib.rs:4565 the comment is `// round`. (Sorry, I don't think github lets you make a suggestion for code that isn't touched in a PR).

aggregator/src/aggregator/aggregate_init_tests.rs
```diff
@@ -448,7 +448,7 @@ async fn aggregation_job_mutation_report_aggregations() {
 }
 
 #[tokio::test]
-async fn aggregation_job_init_two_round_vdaf_idempotence() {
+async fn aggregation_job_init_two_step_vdaf_idempotence() {
```
divergentdave, line 451: We should probably keep this as-is. Poplar1 is two rounds, so this will still take one DAP step, just testing different code paths.
```suggestion
async fn aggregation_job_init_two_round_vdaf_idempotence() {
```

PR #1923, opened 2023-09-14
Minor changes for DAP-07

tgeoghegan:
See individual commits


inahga:
We should probably update the README to indicate `main` is targeting DAP-06, and perhaps that DAP-05 is not planned?

inahga (approved):
Code changes look good.

tgeoghegan:
> We should probably update the README to indicate `main` is targeting DAP-06, and perhaps that DAP-05 is not planned?

Yeah, fixing the readme is in my TODO list in #1669. But there's no reason not to do it here, especially since Chris Patton helpfully pushed DAP-07 today.

PR #1957, opened 2023-09-18
Add missing non_exhaustive to enums

inahga:
Mainly targeting the ExtensionType, but I noticed other enums that could expand in the future in this crate are missing it.

Adding it to the HPKE types comes with unfortunate propagation of an error and unwraps. I'm open to whether this change is worth it. We could also alternatively have a panic here https://github.com/divviup/janus/pull/1957/files#diff-9c7b948f33912f9d1847860795daa701030c35ed114ba5d7e1fcd9674f234ba2R219.

### DIFF
--- a/aggregator_api/src/routes.rs
+++ b/aggregator_api/src/routes.rs
@@ -146,12 +146,14 @@ pub(super) async fn post_task<C: Clock>(
         _ => unreachable!(),
     };
 
+    // Unwrap safety: we always use a supported KEM.
     let hpke_keys = Vec::from([generate_hpke_config_and_private_key(
         random(),
         HpkeKemId::X25519HkdfSha256,
         HpkeKdfId::HkdfSha256,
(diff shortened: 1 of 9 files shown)


aggregator_api/src/routes.rs
```diff
@@ -321,7 +323,8 @@ pub(super) async fn put_global_hpke_config<C: Clock>(
         req.kem_id.unwrap_or(HpkeKemId::X25519HkdfSha256),
         req.kdf_id.unwrap_or(HpkeKdfId::HkdfSha256),
         req.aead_id.unwrap_or(HpkeAeadId::Aes128Gcm),
-    );
+    )
+    .unwrap();
```
divergentdave, line 327: We should propagate this error instead of unwrapping it, since the algorithms could be unsupported if the operator makes a mistake.

divergentdave (approved):

PR #1958, opened 2023-09-18
Release Janus `0.6.0-prerelease-1`

tgeoghegan:
This release is being cut for integration testing, but tagged as a pre-release because we anticipate making more disruptive changes (especially SQL schema changes) before deploying it anywhere.

The Janus maintainers DO NOT recommend deploying this version as it will be incompatible with eventual 0.6.x releases and no migration path will be provided.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1804,7 +1804,7 @@ checksum = "af150ab688ff2122fcef229be89cb50dd66af9e01a4ff320cc137eecc9bacc38"
 
 [[package]]
 name = "janus_aggregator"
-version = "0.6.0"
+version = "0.6.0-prerelease-1"
 dependencies = [
  "anyhow",
  "assert_matches",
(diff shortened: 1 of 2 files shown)


inahga (approved):

divergentdave (approved):
FYI this is going to copy the schema to the 0.6 directory of our GCS bucket. We need this in order to deploy our integration test, but we will also need to manually clobber this prior to releasing `0.6.0`.

tgeoghegan:
> FYI this is going to copy the schema to the 0.6 directory of our GCS bucket. We need this in order to deploy our integration test, but we will also need to manually clobber this prior to releasing `0.6.0`.

Thanks for this reminder. Fortunately @inahga added a flag to the `publish-sql-schema` workflow to force overwrite of migrations, so while this will require operator intervention, it shouldn't require fiddling with `gsutil`.

PR #1962, opened 2023-09-19
Change default auth token type to `BEARER`

tgeoghegan:
I don't think this makes a real difference, because all the code paths that write auth tokens into the database should be setting the token type explicitly anyway, and we switched the default type in the aggregator API to `BEARER` a while ago. But this at least is consistent with the other token tables and will be less confusing to readers.


inahga:
Is it in fact time to do away with DAP tokens anyways?

divergentdave:
I don't think so, they're currently used by both Daphne and draft-dcook-ppm-dap-interop-test-design. Due to inertia, they are still valuable for interoperation; I'm not yet aware of anyone else supporting `Authentication: Bearer`.

divergentdave (approved):
LGTM, note that you'll need to add the `allow-changed-migration` label to the PR and force-push

tgeoghegan:
My opinion is that Daphne's usage of `DAP-Auth-Token` will go away and be replaced by whatever authentication scheme Cloudflare uses on their other production services. However we will keep it in Janus indefinitely to support the interop standard.

inahga (approved):

PR #1963, opened 2023-09-19
Handle chunk_length VDAF parameter

### LINKED ISSUE #1900: Deal with `Prio3SumVec` and `Prio3Histogram` chunk length parameters
In VDAF-07, `Prio3SumVec` and `Prio3Histogram` are parameterized by a chunk length for the underlying `ParallelSum` gadget. There is [some guidance on how to tune that value](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-07#name-selection-of-parallelsum-ch), but the conclusion is that a DAP implementation like Janus can't simply compute a good value.

For now, we are using the square root of the measurement size, which VDAF-07 says will give us non-optimal but not terrible proof lengths. Janus' representation of these VDAFs will have to account for chunk size so that divviup-api or a peer aggregator can instruct us what chunk size to use.

divergentdave:
This adds chunk_length parameters to `VdafInstance` and updates to the latest editor's copy of draft-wang-ppm-dap-taskprov. This will require some additional changes in divviup-api to handle aggregator API changes.

Closes #1900.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -951,7 +951,7 @@ dependencies = [
 [[package]]
 name = "divviup-client"
 version = "0.0.1"
-source = "git+https://github.com/divviup/divviup-api?tag=0.0.25#8eda807aff7a6ed20d7ea8873b4d36b2421cdb95"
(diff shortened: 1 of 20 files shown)


divergentdave:
I updated `divviup-client` to reflect the new parameter there, and rebased on the tip of main. This is ready to go now.

tgeoghegan (approved):

PR #1964, opened 2023-09-19
Setup Github Action to auto-approve/merge dependabot PRs.

branlwyd:


inahga (approved):
I admit I've been ignoring the idea of automerging dependabot because I'm somewhat uncomfortable with the idea. It means that our vendors can more or less push code to our repository without any developer vetting them.

Let me think about it out loud: the threat model in my head is that a vendor releases malicious code. 

I suppose that threat is better mitigated with the likes of cargo audit and cargo vet. We weren't really doing code review anyway and IMO it's unreasonable for us to. We also have lots of time for the community to discover such a thing before such code is actually released to production, assuming we deploy on wednesday or thursday. (note that malicious code can still exfiltrate CI secrets--but we already accept that risk since dependabot PRs run builds automatically).

My threat model also includes a vendor releasing buggy code. I do read the change notes and like to do code review for features that I know impact the code base. I think @jbr's idea of a changelog rollup is nice, because it lets the developer review the changes at release time, but I don't think we have to block this PR for that (I wish dependabot actually copied the release notes into the commit message, that'd pretty much solve it for me).

An entirely different approach we could take is just to freeze dependencies except for  security updates and bug fixes that directly affect us. We would probably also want major version upgrades, to avoid being too far behind.

Overall I'm fine with this approach, just making some notes.

branlwyd:
Your thinking on the threat model is pretty much aligned with mine: since we are not actually reviewing the changes for safety/correctness (nor would it be feasible for us to do so), or doing any other vetting beyond our CI, the toil of manually approving each PR is not useful.

I think we can review changelogs just by looking to the merged PRs in our notifications (or in the git commit log, etc). I'll leave potentially switching tools to later work.

PR #1966, opened 2023-09-20
Fix dependabot automerge.

branlwyd:
The important change is GITHUB_TOKEN -> GH_TOKEN. Despite the gh CLI allowing either, apparently in the context of Github Actions GITHUB_TOKEN is not allowed. (I couldn't find documentation to this effect, but the error I was receiving earlier suggested setting GH_TOKEN, and switching from GITHUB_TOKEN to GH_TOKEN does indeed cause the action to work as expected.)


divergentdave (approved):

divergentdave:
LGTM. The [docs](https://cli.github.com/manual/gh_help_environment) say that `GH_TOKEN` takes precedence over `GITHUB_TOKEN`, so what's probably going on is that we're clobbering a `GH_TOKEN` environment variable set by GHA with the wrong user's token.

Issue #1728, opened 2023-08-15
`release/0.subscriber-01`: trim unneeded features


tgeoghegan:
We drop support for the `l2_fixedpoint` stuff in https://github.com/divviup/janus/pull/1695 because we know that deployment will never need it. In the interest of reducing maintenance costs, we might consider dropping other unneeded features as the exact requirements of that deployment come into focus. For instance, we may be able to drop one of the query types or some more VDAFs, which would reduce the amount of code we carry around in `release/0.subscriber-01`. Then again, it might make it harder to apply patches across release branches.

branlwyd:
Once we know more about the operational configuration we'll be using, we may be able to drop a query type, too, to save quite a bit of code.

branlwyd:
I think this is wise. We're going to be viewing this as effectively a custom deployment, so we can trim anything that isn't actively used by the deployment.

At least for VDAFs/query types, I think merge issues when doing patches won't be too bad -- almost all of the "complex" code is parameterized by VDAF/query-type; it won't be too hard to fixup merge issues on our VDAF/query-type dispatch code.

inahga:
We have abandoned this branch of Janus.

PR #1969, opened 2023-09-20
subscriber-01: Remove taskprov feature flag, always behave as if taskprov is enabled

inahga:
This environment will always operate with taskprov enabled. Make simplifications based on this assumption. We no longer have to consider cases where taskprov isn't enabled.

The main functionality change is that we always advertise a global HPKE key and not bother with the task key. Some tests are deleted based on this, the only operative test is `global_hpke_config()`.

Notice that a lot of our tests still work as-is--many of them insert a task specific HPKE key and use it directly without roundtripping through the advertisement endpoint. This works since we still trial decrypt global keys and task-specific keys. However, `taskprov_tests.rs` still exercises the roundtrip, so there is coverage.

This won't break janus-ops, since the extraneous config will be ignored (i.e. this doesn't block deployment).

In spirit of https://github.com/divviup/janus/issues/1728.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -9,10 +9,8 @@ use crate::{
         report_writer::{ReportWriteBatcher, WritableReport},
     },
     cache::{GlobalHpkeKeypairCache, PeerAggregatorCache},
-    config::TaskprovConfig,
     Operation,
 };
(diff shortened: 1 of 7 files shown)


inahga:
Integration tests need work.

branlwyd (approved):

Issue #1521, opened 2023-06-21
Collapse `task_vdaf_verify_keys`, `task_collector_auth_tokens`, `task_aggregator_auth_tokens` tables into `tasks`


tgeoghegan:
As of DAP-04, task parameters like the VDAF verify key are immutable and cannot be rotated (save by creating a new task). Janus' DB schema has separate tables `task_vdaf_verify_keys`, `task_collector_auth_tokens` and `task_aggregator_auth_tokens` with foreign keys into `tasks` to allow many:1 relationships, but simplify the schema if a task is only ever going to have a single verify key, auth token, etc.

branlwyd:
VDAF verify keys are immutable, but it's not necessarily true that the implementation-specific collector/aggregator auth tokens need to be. OTOH, since rotating VDAF verify keys requires task rotation, maybe we decide it's OK to require task rotation for rotating other kinds of secrets.

tgeoghegan:
See discussion in #1509: if we choose to enable aggregator or collector auth token rotation, then we will still need the many tokens:1 task relationship for those two things.

PR #1973, opened 2023-09-21
Tasks have only one aggregator, collector token

tgeoghegan:
We've decided that each task has exactly one aggregator and collector auth token, and rotating those means rotating the task. This commit updates the representations of auth tokens in the database and in memory (`aggregator_core::task::Task`) accordingly.

Several tests relied upon tasks constructed in tests having two aggregator auth tokens, one of each supported type. Those tests are fixed to explicitly construct tasks using `DAP-Auth-Token` tokens where necessary.

Given that there is now a single aggregator or collector auth token per task, we could remove the `task_aggregator_auth_tokens` and `task_collector_auth_tokens` tables and instead add columns `aggregator_auth_token`, `aggregator_auth_token_type`, `collector_auth_token` and `collector_auth_token_type` to `tasks`.

I chose not to do this because validating the correctness of those columns would be tricky. We couldn't make them all `NOT NULL`, because a helper task doesn't have a collector auth token and a task provisioned via taskprov won't have either token (instead it'll use tokens from the `taskprov_*_auth_tokens` tables). So we would have to write constraints on the `tasks` table to ensure pairs of columns are either `NULL` or `NOT NULL` in tandem, which I think are expressed more clearly in the independent `task_*_auth_tokens` tables. Plus, if we ever _do_ decide to support auth token rotation, it'll be easier to do with these tables in place.

Part of #1524, #1521

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -14,7 +14,7 @@ use crate::{
     SecretBytes,
 };
 use chrono::NaiveDateTime;
-use futures::future::try_join_all;
+use futures::future::{try_join_all, Either};
 use janus_core::{
     hpke::{HpkeKeypair, HpkePrivateKey},
     task::{AuthenticationToken, VdafInstance},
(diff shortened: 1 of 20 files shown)


branlwyd (changes requested):
I would eliminate the separate tables for auth tokens; with this PR, the only reason they exist is because we used to support multiple tokens. If we were to implement the token functionality for the first time, we would not add these tables. I understand the argument that it is challenging to express the necessary constraints, but we could either lean on application-level checks only or implement a [CHECK constraint](https://www.postgresql.org/docs/15/ddl-constraints.html) in the DB.

aggregator_core/src/datastore.rs
```diff
@@ -927,47 +904,33 @@ impl<C: Clock> Transaction<'_, C> {
             )
             .map(SecretBytes::new)?;
 
-        // Aggregator authentication tokens.
-        let mut aggregator_auth_tokens = Vec::new();
-        for row in aggregator_auth_token_rows {
-            let ord: i64 = row.get("ord");
-            let auth_token_type: AuthenticationTokenType = row.get("type");
-            let encrypted_aggregator_auth_token: Vec<u8> = row.get("token");
+        let aggregator_auth_token = if let Some(row) = aggregator_auth_token_row {
```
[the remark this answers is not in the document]
branlwyd: (assuming the error types work out, it's hard to reason about the operation of the `?` operator without an IDE)

tgeoghegan:
> I would eliminate the separate tables for auth tokens; with this PR, the only reason they exist is because we used to support multiple tokens. If we were to implement the token functionality for the first time, we would not add these tables. I understand the argument that it is challenging to express the necessary constraints, but we could either lean on application-level checks only or implement a [CHECK constraint](https://www.postgresql.org/docs/15/ddl-constraints.html) in the DB.

Done, the auth token tables are gone.

aggregator_core/src/datastore.rs
```diff
@@ -584,81 +584,66 @@ impl<C: Clock> Transaction<'_, C> {
...
-        let mut aggregator_auth_tokens = Vec::new();
-        for (ord, token) in task.aggregator_auth_tokens().iter().enumerate() {
-            let ord = i64::try_from(ord)?;
-
-            let mut row_id = [0; TaskId::LEN + size_of::<i64>()];
-            row_id[..TaskId::LEN].copy_from_slice(task.id().as_ref());
-            row_id[TaskId::LEN..].copy_from_slice(&ord.to_be_bytes());
-
+        // Aggregator auth token.
+        let aggregator_auth_token_future = if let Some(token) = task.aggregator_auth_token() {
```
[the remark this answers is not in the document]
tgeoghegan: This is moot because now that the distinct auth token tables are gone, we don't need to do these inserts separately from the one in `tasks` at all.

aggregator_core/src/datastore.rs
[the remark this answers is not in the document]
tgeoghegan: I was a little surprised to see that I could write an `Option::map` closure with two different fallible calls in it by using `Option::transpose`, but it works, along with an `Option::zip` since we no longer have a distinct `Option<Row>` for the auth tokens.

aggregator_core/src/task.rs
```diff
@@ -129,10 +129,12 @@ pub struct Task {
...
     /// HPKE configuration for the collector.
     collector_hpke_config: Option<HpkeConfig>,
-    /// Tokens used to authenticate messages sent to or received from the other aggregator.
-    aggregator_auth_tokens: Vec<AuthenticationToken>,
-    /// Tokens used to authenticate messages sent to or received from the collector.
-    collector_auth_tokens: Vec<AuthenticationToken>,
+    /// Token used to authenticate messages sent to or received from the other aggregator. Only set
+    /// if the task was not created via taskprov.
+    aggregator_auth_token: Option<AuthenticationToken>,
+    /// Token used to authenticate messages sent to received from the collector. Only set if this
```
inahga, line 135:
```suggestion
    /// Token used to authenticate messages sent to or received from the collector. Only set if this
```

db/00000000000001_initial_schema.up.sql
```diff
@@ -89,34 +89,24 @@ CREATE TABLE tasks(
...
     collector_hpke_config       BYTEA,                     -- the HPKE config of the collector (encoded HpkeConfig message)
-    vdaf_verify_key             BYTEA NOT NULL             -- the VDAF verification key (encrypted)
+    vdaf_verify_key             BYTEA NOT NULL,            -- the VDAF verification key (encrypted)
+
+    -- Authentication token used to authenticate messages to/from the other aggregator.
+    -- These columns are NULL if the task was provisioned by taskprov.
+    aggregator_auth_token_type  AUTH_TOKEN_TYPE,    -- the type of the authentication token
+    aggregator_auth_token       BYTEA,              -- encrypted bearer token
+    -- The aggregator_auth_token columns must either both be NULL or both be non-NULL
+    CONSTRAINT aggregator_auth_token_null CHECK ((aggregator_auth_token_type IS NULL) = (aggregator_auth_token IS NULL)),
```
inahga, line 99: WDYT about unit tests that exercise this constraint, i.e. try to insert values that violate it?

This is my main worry with inserting business logic into the database--it's difficult to test for. But I think our test harness can support such a test without too much pain.

inahga (approved):
Nice simplification.

docs/samples/tasks.yaml
```diff
@@ -59,31 +59,29 @@
     aead_id: Aes128Gcm
     public_key: 4qiv6IY5jrjCV3xbaQXULmPIpvoIml1oJmeXm-yOuAo
 
-  # Authentication tokens shared beteween the aggregators, and used to
+  # Authentication token shared beteween the aggregators, and used to
   # authenticate leader-to-helper requests. In the case of a leader-role task,
-  # the leader will include the first token in a header when making requests to
```
divergentdave, line 64: Oof, this was wrong the whole time. (the last token was the primary auth token)

divergentdave (approved):

db/00000000000001_initial_schema.up.sql
tgeoghegan: Good idea. Our definition of `struct Task` doesn't allow constructing a value where the auth token is set but not its type (or vice-versa), but it's easy enough to write SQL `UPDATE` statements that exercise these constraints with illegal mutations to an existing task so I did that.

inahga (approved):
:100:

Issue #1684, opened 2023-08-08
Taskprov: determine better authentication scheme


inahga:
Relevant discussions:
- https://github.com/divviup/janus/pull/1675#discussion_r1286467888
- https://isrg.slack.com/archives/C0167LT4C73/p1691514380751889

We are aiming for multi-tenancy of taskprov peers.

As the helper, for us to appropriately authenticate a request coming from a taskprov peer, we have to first parse the taskprov header, determine the URL of the peer, then look the peer up in the database/cache to retrieve their authentication tokens.

Parsing of the untrusted input of the taskprov header before authentication is undesirable.

We should discuss with our potential taskprov peers as to what authentication scheme works best. Perhaps an additional header identifying the peer and providing the authentication token, so those can be used before parsing the taskprov header.

branlwyd:
Only one point: I think our practical deployment plans only require a single taskprov peer, right? (by which I mean an Author, in taskprov spec terms) So while it would be very nice to support multiple peers, our technical requirements are to support a single peer.

PR #1976, opened 2023-09-22
subscriber-01: Specialise to only one peer aggregator

inahga:
This deployment will have only one peer aggregator. Make simplifications based on that assumption.

The peer aggregator config is no longer stored in the database. Instead, keep non-secret parameters in the plaintext configuration, and have secret values supplied via environment variable (which in turn will be sourced from kubernetes secrets).

This incidentally [simplifies the authentication mechanism](https://github.com/divviup/janus/issues/1684)--we no longer have to parse untrusted input before authenticating (note that we still do this, I'll leave the final optimization for a future PR to keep this one simple).

Supports https://github.com/divviup/janus/issues/1728.


aggregator/src/aggregator.rs
```diff
@@ -636,45 +673,27 @@ impl<C: Clock> Aggregator<C> {
...
-            .get(peer_aggregator_url, peer_role)
-            .ok_or(Error::InvalidTask(
-                *task_id,
-                OptOutReason::NoSuchPeer(*peer_role),
-            ))?;
-
-        if !aggregator_auth_token
-            .map(|t| peer_aggregator.check_aggregator_auth_token(t))
-            .unwrap_or(false)
+            .any(|token| token == request_token)
```
inahga, line 696: This can (and should be) replaced with a trillium handler, along with rejecting non-taskprov requests. However, I'm leaving that for a later PR to keep this one simpler.

interop_binaries/src/bin/janus_interop_aggregator.rs
```diff
@@ -161,7 +165,21 @@ async fn make_handler(
             max_upload_batch_size: 100,
             max_upload_batch_write_delay: std::time::Duration::from_millis(100),
             batch_aggregation_shard_count: 32,
-            ..Default::default()
+            global_hpke_configs_refresh_interval: GlobalHpkeKeypairCache::DEFAULT_REFRESH_INTERVAL,
```
tgeoghegan, line 168: It bothers me that the interop test stuff needs to be aware of taskprov at all. That doesn't really make sense, since taskprov isn't exercised in the interop test setup and shouldn't be used. If the higher level goal here is to move toward specializing the Janus on `release/0.subscriber-01` to exclusively be a taskprov helper, then I think we could just drop support for the interop testing API altogether.

tgeoghegan (approved):
I'm a little wary of doing more work on authentication on this branch because we don't yet understand the requirements, so effort we expend here may be thrown away. That said, this PR doesn't change how authentication works, just the arity of the tokens.

I'd also like to better understand the direction we want to take this release branch. Is the goal to strip out functionality to the point that it only implements a helper that can only have tasks provisioned via taskprov?

inahga:
> Is the goal to strip out functionality to the point that it only implements a helper that can only have tasks provisioned via taskprov?

This is the direction I'm vaguely aiming for, if everyone else thinks that's wise.

I'm not sure that we should entirely remove the interop test harness quite yet, as we haven't determined what a taskprov E2E test will look like--such a decision IMO should wait until we have some engineering collaboration with our partner. We might be able to use the pattern of running their leader in a container.

aggregator/src/bin/aggregator.rs
```diff
@@ -171,6 +165,18 @@ fn build_aggregator_api_handler<'a>(
     )))
 }
 
+fn parse_auth_tokens(auth_tokens: &[String]) -> Result<Vec<AuthenticationToken>> {
```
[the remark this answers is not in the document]
inahga: Neither turns out very straightforward--I don't think you can in-place turn a `[String]` into a `[&str]` at the call site, and IntoIterator requires lifetime annotations. For a library function I would try harder for IntoIterator, but I don't think this is improvement for a small private helper function.

interop_binaries/src/bin/janus_interop_aggregator.rs
inahga: Discussed in slack. We can move forward with this PR. If we have any more intrusive changes to the interop API, we should drop it.

PR #1978, opened 2023-09-25
Relax format restrictions on authentication tokens

divergentdave:
This relaxes the requirements we place on authentication tokens. The last time we changed this, we required that all `DAP-Auth-Token` and `Authorization: Bearer` values were in unpadded base64url. This causes problems when using the interop test API outside of Janus's integration tests, because it's more restrictive than it used to be. In one case, I was using structured values for these, consisting of a word describing the token's role, a hyphen, and then some random hexadecimal bytes. This failed the base64url check in one case because it had an impossible length. (base64 cannot produce an output with length one more than a multiple of four, ignoring padding)

This PR places different, looser, validation requirements on each kind of authentication token. For `DAP-Auth-Token`, we require that the value be a valid HTTP header value. (i.e. we exclude some nonprintable control characters) Unlike in older versions, we also require that the value be valid UTF-8. (this makes implementation simpler, and there's no need to go out of our way supporting invalid UTF-8 inside a HTTP payload) For `Authorization: Bearer`, we check against the language described in RFC 6750. In addition to unpadded base64url, we will now support other base64 variants, compact JWT, and other formats that fit in the expanded alphabet. For both kinds of authentication tokens, we still generate their values internally by base64url-encoding random bytes.

This change ought to go in now, during our pre-release window, because it makes breaking changes to variants of a public `enum`.

### DIFF
--- a/core/src/task.rs
+++ b/core/src/task.rs
@@ -1,6 +1,6 @@
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use derivative::Derivative;
-use http::header::AUTHORIZATION;
+use http::{header::AUTHORIZATION, HeaderValue};
 use janus_messages::taskprov;
 use rand::{distributions::Standard, prelude::Distribution};
 use ring::constant_time;
@@ -622,7 +622,7 @@ pub enum AuthenticationToken {
     /// The token is not necessarily an OAuth token.
     ///
     /// [1]: https://datatracker.ietf.org/doc/html/rfc6750#section-2.1
-    Bearer(TokenInner),
+    Bearer(BearerToken),
 
     /// Token presented as the value of the "DAP-Auth-Token" HTTP header. Conforms to
     /// [draft-dcook-ppm-dap-interop-test-design-03][1], sections [4.3.3][2] and [4.4.2][3], and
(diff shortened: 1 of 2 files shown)


core/src/task.rs
```diff
@@ -632,30 +632,30 @@ pub enum AuthenticationToken {
     /// [2]: https://datatracker.ietf.org/doc/html/draft-dcook-ppm-dap-interop-test-design-03#section-4.3.3
     /// [3]: https://datatracker.ietf.org/doc/html/draft-dcook-ppm-dap-interop-test-design-03#section-4.4.2
     /// [4]: https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-01#name-https-sender-authentication
-    DapAuth(TokenInner),
+    DapAuth(DapAuthTokenInner),
```
inahga, line 635: Consider just having these be `BearerToken` and `DapAuthToken`. The `Inner` suffix suggests that there's some different public opaque outer type, which is not the case here.

inahga (approved):
:+1:. This should also make it a bit easier to generate valid tokens at the shell, i.e. can just do `openssl rand 32 | base64`.

core/src/task.rs
```diff
@@ -692,71 +692,202 @@ impl Distribution<AuthenticationToken> for Standard {
...
-/// A token value used to authenticate HTTP requests.
+/// A token value used to authenticate HTTP requests. This token is used in the "DAP-Auth-Token"
+/// HTTP request header.
 ///
-/// The token is used directly in HTTP request headers without further encoding and so much be a
-/// legal HTTP header value. More specifically, the token is restricted to the unpadded, URL-safe
-/// Base64 alphabet, as specified in [RFC 4648 section 5][1]. The unpadded, URL-safe Base64 string
-/// is the canonical form of the token and is used in configuration files, Janus aggregator API
-/// requests and HTTP authentication headers.
+/// This token is used directly in HTTP request headers without further encoding and so much be a
```
tgeoghegan, line 698:
```suggestion
/// This token is used directly in HTTP request headers without further encoding and so must be a
```

tgeoghegan (approved):

PR #1984, opened 2023-09-25
break up `janus_core::task`

tgeoghegan:
Moves handling of authentication tokens to a new module `janus_core::auth_tokens`. Once that's gone, the only thing of substance left is `enum VdafInstance` and the macros for dispatching on VDAFs, so that module is renamed to `janus_core::vdaf` to better describe its purpose.

### DIFF
--- a/core/src/vdaf.rs
+++ b/core/src/vdaf.rs
@@ -1,15 +1,7 @@
-use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use derivative::Derivative;
-use http::{header::AUTHORIZATION, HeaderValue};
 use janus_messages::taskprov;
-use rand::{distributions::Standard, prelude::Distribution};
-use ring::constant_time;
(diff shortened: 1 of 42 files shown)


core/src/vdaf.rs
```diff
@@ -607,288 +600,6 @@ macro_rules! vdaf_dispatch {
...
-
-impl Eq for BearerToken {}
-
-impl Distribution<BearerToken> for Standard {
-    fn sample<R: rand::Rng + ?Sized>(&self, rng: &mut R) -> BearerToken {
-        BearerToken(URL_SAFE_NO_PAD.encode(rng.gen::<[u8; 16]>()))
-    }
-}
-
 /// Returns the given [`Url`], possibly modified to end with a slash.
```
divergentdave, line 603: I think we should hoist this to the top level of the crate, since it's not related to VDAFs at all.

divergentdave (approved):

PR #1985, opened 2023-09-25
Add `AuthenticationTokenHash`

tgeoghegan:
When acting as an HTTP server, Janus only needs a hash of the aggregator authentication token to validate incoming requests, and can avoid storing the actual token. This commit adds
`janus_core::auth_tokens::AuthenticationTokenHash` which will be used to implement that.

Part of #1509

### DIFF
--- a/aggregator_core/src/task.rs
+++ b/aggregator_core/src/task.rs
@@ -4,7 +4,7 @@ use crate::SecretBytes;
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use derivative::Derivative;
 use janus_core::{
-    auth_tokens::AuthenticationToken,
+    auth_tokens::{AuthenticationToken, AuthenticationTokenHash},
     hpke::{generate_hpke_config_and_private_key, HpkeKeypair},
     time::TimeExt,
     url_ensure_trailing_slash,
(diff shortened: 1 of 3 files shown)


aggregator_core/src/task.rs
```diff
@@ -400,22 +400,35 @@ impl Task {
...
+        incoming_auth_token: Option<&AuthenticationToken>,
+    ) -> bool {
+        // TODO(#1509): leader should hold only an AuthenticationToken and refuse to use it for
+        // incoming token validation. Helper should hold only an AuthenticationTokenHash, making the
+        // AuthenticationTokenHash::from call here unnecessary.
+        self.aggregator_auth_token()
+            .map(AuthenticationTokenHash::from)
+            .zip(incoming_auth_token)
+            .map(|(own_token_hash, incoming_token)| own_token_hash.validate(incoming_token))
+            .unwrap_or(false)
```
inahga, line 416: TIL you can zip an `Option`.

inahga (approved):

inahga:
Note that I only reviewed https://github.com/divviup/janus/pull/1985/commits/2577838372f8c3fb190dccce78c08638d76186a8.

aggregator_core/src/task.rs
branlwyd: (options work as `Iterator`s which will always have either 0 (if `None`) or 1 (if `Some`) elements)

core/src/auth_tokens.rs
```diff
@@ -0,0 +1,484 @@
...
+    }
+
+    /// Validate that a bearer token value matches the format in
+    /// https://datatracker.ietf.org/doc/html/rfc6750#section-2.1.
+    fn validate(value: &str) -> Result<(), anyhow::Error> {
+        let mut iter = value.chars();
+        let mut any_non_equals = false;
+        // First loop: consume "normal" characters, stop when we see an equals sign for padding or
+        // reach the end of the input.
+        for c in &mut iter {
```
branlwyd, line 211: Any reason not to use a regex here? I lean away from them normally but a small value defined by BNF-like grammar is a perfect place for a regex IMO.

I think it would be something like `[-a-zA-Z0-9._~+/]+=*`.

branlwyd: (we could also just try decoding as base64, since I suppose that's the intent -- though that does lose some non-base64 values that are allowed by the BNF, such as `A=======`.)

branlwyd (approved):

core/src/auth_tokens.rs
divergentdave: I introduced this in #1978. It got moved in #1984, and it'll disappear from here once it's rebased on main. A regex sounds good, it may be able to run faster too. As described in #1978, base64-decoding would be more strict than necessary, both in terms of the character set allowed, the choice of padding vs. no padding, and limitations on length (excluding padding). I can make a follow-up PR to swap in a regex here.

core/src/auth_tokens.rs
```diff
@@ -330,4 +439,46 @@ mod tests {
...
+    #[case::dap_auth(r#"{ type: "DapAuth", hash: "MJOoBO_ysLEuG_lv2C37eEOf1Ngetsr-Ers0ZYj4vdQ" }"#)]
+    #[test]
+    fn serde_aggregator_token_hash_valid(#[case] yaml: &str) {
+        serde_yaml::from_str::<AuthenticationTokenHash>(yaml).unwrap();
+    }
+
+    #[rstest::rstest]
+    #[case::bearer_token_invalid_encoding(r#"{ type: "Bearer", hash: "+" }"#)]
+    #[case::bearer_token_wrong_length(
+        r#"{ type: "Bearer", hash: "MJOoBO_ysLEuG_lv2C37eEOf1Ngetsr-Ers0ZYj4" }"#
```
divergentdave, line 454: This case will fail base64url decoding, but I think the intent was to check the conversion from a byte slice to a byte array. I recommending adding another test case with an encoded hash that's one character longer or shorter.

core/src/auth_tokens.rs
tgeoghegan: Sorry, I should have called out that this PR is stacked on #1984.

divergentdave (approved):

core/src/auth_tokens.rs
tgeoghegan: `basenc(1)` seems to think it's 30 bytes of valid Base64URL:
```
> echo "MJOoBO_ysLEuG_lv2C37eEOf1Ngetsr-Ers0ZYj4" | basenc -d --base64url | xxd
00000000: 3093 a804 eff2 b0b1 2e1b f96f d82d fb78  0..........o.-.x
00000010: 439f d4d8 1eb6 cafe 12bb 3465 88f8       C.........4e..
> echo "MJOoBO_ysLEuG_lv2C37eEOf1Ngetsr-Ers0ZYj4" | basenc -d --base64url | wc -c
30
```

divergentdave: Whoops, I forgot to convert from base64url to base64 when I checked, and Python's base64.b64decode apparently skipped over the underscores and hyphens. The encoded value is in fact 40 bytes long, which is correct.

PR #1998, opened 2023-09-25
Use regex for bearer token validation

divergentdave:
This implements the suggestion in https://github.com/divviup/janus/pull/1985/files#r1336411760.

### DIFF
--- a/core/src/auth_tokens.rs
+++ b/core/src/auth_tokens.rs
@@ -2,12 +2,13 @@ use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use derivative::Derivative;
 use http::{header::AUTHORIZATION, HeaderValue};
 use rand::{distributions::Standard, prelude::Distribution};
+use regex::Regex;
 use ring::{
     constant_time,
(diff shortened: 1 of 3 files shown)


core/src/auth_tokens.rs
```diff
@@ -201,36 +202,15 @@ impl BearerToken {
...
-        for c in &mut iter {
-            match c {
-                '=' => {}
-                _ => {
-                    return Err(anyhow::anyhow!(
-                        "bearer token may only contain '=' at the end"
-                    ))
-                }
-            }
+        static REGEX: OnceLock<Regex> = OnceLock::new();
```
inahga, line 205: Neat-o, `OnceLock` was stabilized in 1.70, which is now our MSRV. We can probably drop `lazy_static`, with a bit of effort.

inahga (approved):
I'm partial to either approach. This regex turns out to be not unreadable.

core/src/auth_tokens.rs
divergentdave: Already done in #1852 :grin:

inahga: Aha, while I was on vacation :laughing:

Issue #1977, opened 2023-09-25
cargo test fail


c02-student:
"Nice to meet you.

I am very interested in DAP and VDAF, and I would like to try running Janus in my local environment.

The OS in my local environment is Ubuntu 22.04.
I have installed Docker, Rust (rustc 1.72.1), and Kind (version 0.17.0) in my local environment.
I executed 'docker buildx bake --load,' and the Docker image was successfully created.
When I ran 'cargo test,' the following tests failed with the following error message:

```
failures:
    e2e_prio3_count
    e2e_prio3_count_fixed_size
    e2e_prio3_count_vec
    e2e_prio3_fixed16vec_fixed_size
    e2e_prio3_fixed32vec_fixed_size
    e2e_prio3_fixed64vec_fixed_size
    e2e_prio3_histogram

test result: FAILED. 5 passed; 7 failed; 0 ignored; 0 measured; 0 filtered out; finished in 71.90s


---- e2e_prio3_count_vec stdout ----
thread 'e2e_prio3_count_vec' panicked at 'called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(127.0.0.1)), port: Some(32948), path: "/internal/test/ready", query: None, fragment: None }, source: hyper::Error(IncompleteMessage) }', /root/src/janus/interop_binaries/src/lib.rs:521:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- e2e_prio3_count stdout ----
thread 'e2e_prio3_count' panicked at 'called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Ipv4(127.0.0.1)), port: Some(32950), path: "/internal/test/ready", query: None, fragment: None }, source: hyper::Error(IncompleteMessage) }', /root/src/janus/interop_binaries/src/lib.rs:521:10

---- e2e_prio3_fixed16vec_fixed_size stdout ----
[...]

PR #2006, opened 2023-09-26
Interop binaries: wait for database to be up

divergentdave:
This waits for the Postgres database to come up before starting `janus_interop_aggregator` etc. This is for #1977. I tried this out locally, and saw the following on stdout from the `setup` job. (the first two lines are from `pg_isready`)

```
/var/run/postgresql:5432 - no response
/var/run/postgresql:5432 - accepting connections
janus_interop_aggregator: started
aggregation_job_creator: started
aggregation_job_driver: started
collection_job_driver: started
```

### DIFF
--- a/interop_binaries/setup.sh
+++ b/interop_binaries/setup.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+until pg_isready -U postgres; do sleep 1; done
 /usr/bin/supervisorctl -c /etc/janus/supervisord.conf start janus_interop_aggregator
 /usr/bin/supervisorctl -c /etc/janus/supervisord.conf start aggregation_job_creator
 /usr/bin/supervisorctl -c /etc/janus/supervisord.conf start aggregation_job_driver
--- a/interop_binaries/setup.sh
+++ b/interop_binaries/setup.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+set -e
+
 timeout 5m bash -c 'until pg_isready -U postgres; do sleep 1; done'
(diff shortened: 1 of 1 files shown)


tgeoghegan:
Nice find, David. @c02-student, since none of us have been able to reproduce this, do you think you could check whether this patch fixes your issue?

interop_binaries/setup.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+until pg_isready -U postgres; do sleep 1; done
```
tgeoghegan, line 2: This will wait for postgres forever, right? I think we should give up after some delay so that e.g. postgres crashing on startup due to misconfiguration doesn't cause a test to hang forever.

tgeoghegan (changes requested):

interop_binaries/setup.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+timeout 5m bash -c 'until pg_isready -U postgres; do sleep 1; done'
```
inahga, line 2: Consider running this script with `set -e` or perhaps better adding `|| "timeout waiting for postgres to be ready" && exit 1`. Otherwise the script will fall through on failure.

This is a nit since we'll be able to see the results in the logs anyways, can take it or leave it.

inahga (approved):

interop_binaries/setup.sh
c02-student: I was able to resolve the issue with this change !
I have tested it in my local environment for both release/0.5 and 0.6.0-prerelease-1, and both passed successfully.

Thank you .

tgeoghegan (approved):

PR #2011, opened 2023-09-27
Integration/E2E test logging improvements 

inahga:
Some improvements that came out of me debugging https://github.com/divviup/janus/issues/1977.

- Pass through the RUST_LOG environment variable to test containers, so we can directly control log level at the CLI. We still default to `info` logging level.
- Have each test deposit logs to a folder whose name is the test name, so that logs from a run that invoked multiple tests can be differentiated.
- Always output logs if JANUS_E2E_LOGS_PATH is set, in case we need help debugging a test that should fail, but doesn't.

### DIFF
--- a/interop_binaries/src/lib.rs
+++ b/interop_binaries/src/lib.rs
@@ -23,7 +23,6 @@ use std::{
     process::{Command, Stdio},
     str::FromStr,
     sync::Arc,
-    thread::panicking,
 };
 use testcontainers::{Container, Image};
 use tokio::sync::Mutex;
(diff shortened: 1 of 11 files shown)


interop_binaries/src/lib.rs
```diff
@@ -531,6 +538,15 @@ impl Default for Keyring {
     }
 }
 
+/// Returns the environment variable RUST_LOG. If it's unset or otherwise invalid, return the
+/// default value of "debug".
```
tgeoghegan, line 542:
```suggestion
/// default value of "info".
```

tgeoghegan (approved):

PR #2012, opened 2023-09-27
Gracefully handle unrecognized HPKE algorithms

### LINKED ISSUE #857: `janus_client` should have preferences for HPKE configurations
As of DAP-03, aggregators may advertise multiple HPKE configurations for a given task. `janus_client` needs to decide what HPKE config to use when presented with a list. I think by default, we could choose whichever one we think is fastest (though that will vary based on e.g. whether AES is accelerated on the platform), but we might want to allow clients to configure allow- or deny-lists. For example, some clients might only allow NIST cryptography and thus will want to stick to P-256 and AES.

divergentdave:
This closes #857. The message enums for algorithm IDs are expanded to include catch-all variants, so that we can parse an `HpkeConfigList` even if it contains unsupported or unrecognized algorithms. I added a function to `janus_core` that probes whether `hpke_dispatch` supports a given config's suite of algorithms. Then, I used that in `janus_client` to filter through the received `HpkeConfigList` and select a `HpkeConfig` with supported algorithms.

This includes some Rust-level breaking changes: The message enums now have a value carrying variant, so they can't be cast to a primitive anymore. The associated types `<HpkeKemId as TryFrom<u16>>::Error` etc. changed from `num_enum::TryFromPrimitiveError` to `std::convert::Infallible`. I removed a now-unused variant from `janus_messages::Error`. I also removed a redundant argument from `janus_client::aggregator_hpke_config()` while I was in there.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -5,7 +5,7 @@ use derivative::Derivative;
 use http::header::CONTENT_TYPE;
 use itertools::Itertools;
 use janus_core::{
-    hpke::{self, HpkeApplicationInfo, Label},
+    hpke::{self, is_hpke_config_supported, HpkeApplicationInfo, Label},
     http::HttpErrorResponse,
     retries::{http_request_exponential_backoff, retry_http_request},
     time::{Clock, TimeExt},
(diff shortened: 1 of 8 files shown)


client/src/lib.rs
```diff
@@ -157,17 +156,27 @@ pub async fn aggregator_hpke_config(
...
+    let mut first_error = None;
+    for config in hpke_configs.hpke_configs() {
+        match is_hpke_config_supported(config) {
+            Ok(()) => return Ok(config.clone()),
+            Err(e) => {
+                if first_error.is_none() {
+                    first_error = Some(e);
+                }
+            }
+        }
```
inahga, line 175:
```suggestion
            Err(e) => first_error.get_or_insert(e),
        };
```
nit, can take it or leave it: [`get_or_insert()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.get_or_insert) inserts when the contained value is `None`. Might be dubious whether this is a readability improvement though.

inahga (approved):

client/src/lib.rs
divergentdave: Hmm, I think I'll pass, since we're not using the `&mut T` it returns in this case, but just the side-effect.

PR #2013, opened 2023-09-28
Encapsulate HttpApiProblem in new type

### LINKED ISSUE #1621: Breaking change from upgrading http_api_problem
When upgrading from Janus version 0.5.9 to 0.5.10, the version of http-api-problem used by both the `janus_core` and `janus_client` crates incremented from 0.56.0 to 0.57.0. The type `http_api_problem::HttpApiProblem` is part of `janus_core`'s public API through an error enum variant, and thus upgrading either Janus crate alone resulted in a broken build. To avoid issues like this in the future, we should encapsulate this type, so we can upgrade this dependency without issue.

divergentdave:
This encapsulates our use of `http_api_problem::HttpApiProblem` in a new struct in `janus_core`. This serves two purposes. First, it isolates breaking changes due to future upgrades of `http_api_problem` from downstream users, which fixes #1621. Second, it deduplicates logic previously in `janus_client` and `janus_collector` to check a problem type URI against DAP's list of problem types, and optionally extract this into our `DapProblemType` enum.


tgeoghegan (approved):

aggregator/src/aggregator/error.rs
```diff
@@ -103,11 +103,8 @@ pub enum Error {
     #[error("HTTP client error: {0}")]
     HttpClient(#[from] reqwest::Error),
     /// HTTP server returned an error status code.
-    #[error("HTTP response status {problem_details}")]
-    Http {
-        problem_details: Box<HttpApiProblem>,
-        dap_problem_type: Option<DapProblemType>,
-    },
+    #[error("HTTP response status {0}")]
+    Http(Box<HttpErrorResponse>),
```
inahga, line 107: Q: why does this need to be `Box`ed?

inahga (approved):

aggregator/src/aggregator/error.rs
divergentdave: Note that we previously boxed `HttpApiProblem`. It's a big struct, since it holds four `String`s, a `HashMap`, and a `u16`, for a total of 152 bytes including padding. Our various error enums got flagged with `clippy::large_enum_variant` at one time before we boxed this. Keeping the `Error` enum smaller means it can be copied around with a few inline moves as it's returned in results, rather than needing to call out to memcpy.

aggregator/src/aggregator/problem_details.rs
```diff
@@ -251,8 +251,8 @@ mod tests {
                     // Confirm that post_to_helper() correctly parsed the error type from error_handler().
                     assert_matches!(
                         actual_error,
-                        Error::Http { dap_problem_type: problem_type, .. } => {
-                            assert_eq!(problem_type, test_case.expected_problem_type);
+                        Error::Http (error_response) => {
```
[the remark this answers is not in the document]
divergentdave: `rustfmt` stayed away since it's in the body of an `assert_matches!()` macro call.

PR #2015, opened 2023-09-28
Set a default log level in Dockerfile.interop_aggregator

divergentdave:
This is a follow-up to #2011 to set the default `RUST_LOG` level in a Dockerfile environment variable. Docker allows environment variables declared in a container image to be overridden when creating a container, so the new code in #2011 will still be able to pass through `RUST_LOG` from integration test processes. This change just affects other use cases of the images (such as in the divviup-ts test suite), so that they still get a good default amount of logging.


inahga (approved):
:+1:

tgeoghegan (approved):

PR #2016, opened 2023-09-28
Task rewrite: introduce new task types

tgeoghegan:
We want to enable aggregators to only store authentication token hashes if they don't need the actual token (e.g., a helper only receives aggregation sub-protocol messages). With the existing `janus_aggregator_core::task::Task` structure, this is awkward, because that struct was used both as a view into all of a task's parameters and secrets, useful for test code that orchestrates the provisioning and execution of a task, as well as a specific DAP participant's view of the task, which is useful in Janus itself.

To resolve that tension, this commit introduces distinct structures for those distinct views. `janus_aggregator_core::task::test_util::Task` contains all of a task's parameters and secrets, regardless of any protocol role. `janus_aggregatore_core::task::test_util::NewTaskBuilder` is used in test contexts to build `test_util::Task` values, and then `test_util::Task::{leader_view, helper_view}` then allow obtaining the leader or helper specific view of the task as a
`janus_aggregator_core::task::AggregatorTask`. This holistic view of a task is only used in test code, and thus we confine it to the `test_util` module.

`AggregatorTask` is an aggregator's view of the task, and uses an enum to represent the leader, helper or taskprov-provisioned helper specific parameters. This task representation is what gets used by non-test configurations of Janus. We also add `SerializediAggregatorTask`, which much like `SerializedTask` is an adapter so that `AggregatorTask` values can be serialized to and from YAML, for use in `janus_cli`.

The goal is to migrate all of Janus from the existing `janus_aggregator_core::task::Task` and
`janus_aggregator_core::task::test_util::TaskBuilder` onto `task::test_util::Task`, `task::test_util::NewTaskBuilder` and `task::AggregatorTask`. Doing this all in one go would yield a huge diff that is difficult to review, so in order to enable incremental adoption of the new structures throughout Janus, we have them co-exist for now. Once all usages of `task::Task` and `task::test_util::TaskBuilder` have been removed, those types can be deleted and `NewTaskBuilder` can be renamed to just `TaskBuilder`.

To see where all this is going, check out #2014.

Part of #1524

### DIFF
--- a/aggregator_core/src/task.rs
+++ b/aggregator_core/src/task.rs
@@ -481,6 +481,486 @@ impl Task {
     }
 }
 
+/// Task parameters common to all views of a DAP task.
+#[derive(Clone, Derivative, PartialEq, Eq)]
+#[derivative(Debug)]
+struct CommonTaskParameters {
+    /// Unique identifier for the task.
+    task_id: TaskId,
+    /// The query type this task uses to generate batches.
+    query_type: QueryType,
+    /// The VDAF this task executes.
+    vdaf: VdafInstance,
+    /// Secret verification key shared by the aggregators.
+    vdaf_verify_key: SecretBytes,
+    /// The maximum number of times a given batch may be collected.
+    max_batch_query_count: u64,
+    /// The time after which the task is considered invalid.
+    task_expiration: Option<Time>,
+    /// The age after which a report is considered to be "expired" and will be considered a
+    /// candidate for garbage collection. A value of `None` indicates that garbage collection is
+    /// disabled.
+    report_expiry_age: Option<Duration>,
+    /// The minimum number of reports in a batch to allow it to be collected.
+    min_batch_size: u64,
+    /// The duration to which clients should round their reported timestamps to. For time-interval
+    /// tasks, batch intervals must be multiples of this duration.
+    time_precision: Duration,
+    /// How much clock skew to allow between client and aggregator. Reports from
+    /// farther than this duration into the future will be rejected.
+    tolerable_clock_skew: Duration,
+}
+
+impl CommonTaskParameters {
+    /// Create a new [`CommonTaskParameters`] with the provided values.
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        task_id: TaskId,
+        query_type: QueryType,
+        vdaf: VdafInstance,
+        vdaf_verify_key: SecretBytes,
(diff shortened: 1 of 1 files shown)


aggregator_core/src/task.rs
```diff
@@ -481,6 +481,486 @@ impl Task {
...
+        hpke_keys: I,
+        aggregator_parameters: AggregatorTaskParameters,
+    ) -> Result<Self, Error> {
+        // Compute hpke_configs mapping cfg.id -> (cfg, key).
+        let hpke_keys: HashMap<HpkeConfigId, HpkeKeypair> = hpke_keys
+            .into_iter()
+            .map(|keypair| (*keypair.config().id(), keypair))
+            .collect();
+
+        common_parameters.validate()?;
```
divergentdave, line 645: It seems out of place that this is in a constructor instead of `AggregatorTask::validate()`, or instead of having the constructor check all validation constraints.

It looks like our old `Task` struct didn't validate on construction largely because it was internally used as storage inside `TaskBuilder`, so we intentionally initialized it with inconsistent values, wrote to fields without re-checking invariants, etc., and then only validated parameters on conversion from a `TaskBuilder` into the inner `Task`. Since `AggregatorTask` doesn't get used for that purpose anymore, we could make the `validate()` methods private, make both constructors fallible, and always validate before returning a `CommonTaskParameters` or `AggregatorTask`.

aggregator_core/src/task.rs
```diff
@@ -481,6 +481,486 @@ impl Task {
...
+    }
+
+    /// Returns the collector authentication token for this task, if this aggregator is the leader.
+    /// TODO(#1509): make this an AuthenticationTokenHash
+    pub fn collector_auth_token(&self) -> Option<&AuthenticationToken> {
+        self.aggregator_parameters.collector_auth_token()
+    }
+
+    /// Return the HPKE keypairs used by this aggregator to decrypt client reports, or `None` for
+    /// taskprov tasks.
```
divergentdave, line 835: This should instead say that the HashMap will be empty for taskprov tasks.

aggregator_core/src/task.rs
tgeoghegan: I folded the validate methods into the constructors, since validation never needs to be done independently of instantiation of a `CommonTaskParameters` or `AggregatorTask` value.

divergentdave (approved):

PR #2017, opened 2023-09-28
Task rewrite: adopt `AggregatorTask` in datastore

tgeoghegan:
Adopts `janus_aggregator_core::task::test_util::NewTaskBuilder` and `janus_aggregator_core::task::AggregatorTask` in the `janus_aggregator_core::datastore` module. Much as the previous change provides two kinds of `Task` structure, we now provide two sets of methods for reading and writing tasks: one that deals in the new `AggregatorTask` and the other which deals in the old `Task`.

We add routines for converting between `task::Task` and `task::AggregatorTask` to make it easier for these two paths through the datastore to co-exist. This conversion is lossy because `AggregatorTask` only retains one of the aggregator endpoints, but this doesn't cause substantial problems in Janus, and we can live it transitionally.

Part of #1524

### DIFF
--- a/aggregator_core/src/task.rs
+++ b/aggregator_core/src/task.rs
@@ -499,15 +499,15 @@ impl Task {
                 aggregator_auth_token: self
                     .aggregator_auth_token
                     .clone()
-                    .ok_or_else(|| Error::InvalidParameter("no aggregator auth token in task"))?,
+                    .ok_or(Error::InvalidParameter("no aggregator auth token in task"))?,
                 collector_auth_token: self
                     .collector_auth_token
                     .clone()
-                    .ok_or_else(|| Error::InvalidParameter("no collector auth token in task"))?,
+                    .ok_or(Error::InvalidParameter("no collector auth token in task"))?,
                 collector_hpke_config: self
                     .collector_hpke_config
                     .clone()
-                    .ok_or_else(|| Error::InvalidParameter("no collector HPKE config in task"))?,
+                    .ok_or(Error::InvalidParameter("no collector HPKE config in task"))?,
             },
         )
     }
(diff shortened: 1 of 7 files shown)


aggregator_core/src/task.rs
```diff
@@ -479,6 +479,168 @@ impl Task {
...
+    pub fn view_for_role(&self) -> Result<AggregatorTask, Error> {
+        match self.role {
+            Role::Leader => self.leader_view(),
+            Role::Helper => self.helper_view().or_else(|_| self.taskprov_helper_view()),
+            _ => Err(Error::InvalidParameter("role is not an aggregator")),
+        }
+    }
+}
+
+impl From<AggregatorTask> for Task {
```
inahga, line 576: I'm cool with this for transitional purposes, but think we should do away with it ASAP (since it's lossy and contains false data). Looks like you're well on your way to doing so though.

inahga (approved):

aggregator_core/src/task.rs
```diff
@@ -479,6 +479,168 @@ impl Task {
...
+            self.vdaf.clone(),
+            self.vdaf_verify_key.clone(),
+            self.max_batch_query_count,
+            self.task_expiration,
+            self.report_expiry_age,
+            self.min_batch_size,
+            self.time_precision,
+            self.tolerable_clock_skew,
+            self.hpke_keys.values().cloned().collect::<Vec<_>>(),
+            AggregatorTaskParameters::TaskProvHelper,
```
inahga, line 558: [Elsewhere we've used `Taskprov`](https://github.com/divviup/janus/blob/main/aggregator_api/src/models.rs#L210) instead of `TaskProv` (i.e. lowercase P).

Whichever one works for me, but we should be consistent. It also wasn't introduced in this PR so doesn't have to be resolved here. Can proceed, then maybe a PR after the rest that does `find | sed`.

aggregator_core/src/task.rs
tgeoghegan: Yes, this is going to go away along with `janus_aggregator_core::task::Task` as this series of PRs progresses.

aggregator_core/src/task.rs
tgeoghegan: That makes sense. I'm going to change the name in a later PR so that I don't have to rebase the existing stack.

PR #2019, opened 2023-09-29
Task rewrite: adopt `AggregatorTask` in taskprov

tgeoghegan:
# Stacked on #2018 

Adopts `janus_aggregator_core::task::AggregatorTask` in the `taskprov` module and code that uses it in `janus_aggregator`.

Of particular interest is that we do away with
`janus_aggregator_core::taskprov::Task` and instead represent taskprov tasks as a `janus_aggregator_core::task::Task` with `AggregatorTaskParameters::TaskProvHelper`. Hopefully we'll be able to further unify handling of taskprov and regular tasks in the future.

Part of #1524

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -34,7 +34,7 @@ use janus_aggregator_core::{
     },
     query_type::AccumulableQueryType,
     task::{self, AggregatorTask, Task, VerifyKey},
-    taskprov::{self, PeerAggregator},
+    taskprov::PeerAggregator,
 };
 #[cfg(feature = "test-util")]
 use janus_core::test_util::dummy_vdaf;
@@ -645,7 +645,7 @@ impl<C: Clock> Aggregator<C> {
         task_config: &TaskConfig,
         aggregator_auth_token: Option<&AuthenticationToken>,
     ) -> Result<(), Error> {
-        let (peer_aggregator, leader_url, helper_url) = self
+        let (peer_aggregator, leader_url, _) = self
             .taskprov_authorize_request(peer_role, task_id, task_config, aggregator_auth_token)
             .await?;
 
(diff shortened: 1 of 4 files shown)


aggregator_core/src/taskprov.rs
```diff
@@ -266,77 +263,6 @@ impl KeyType for VdafVerifyKeyLength {
     }
 }
 
-/// Newtype for [`task::Task`], which omits certain fields that aren't required for taskprov tasks.
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct Task(pub(super) task::Task);
```
inahga, line 271: Good riddance!

aggregator/src/aggregator.rs
```diff
@@ -661,38 +661,29 @@ impl<C: Clock> Aggregator<C> {
...
-            our_role,
             vdaf_verify_key,
             task_config.query_config().max_batch_query_count() as u64,
             Some(*task_config.task_expiration()),
             peer_aggregator.report_expiry_age().cloned(),
             task_config.query_config().min_batch_size() as u64,
             *task_config.query_config().time_precision(),
             *peer_aggregator.tolerable_clock_skew(),
+            // Taskprov task has no per-task HPKE keys
+            [],
```
inahga, line 679: It seems then the HPKE keys should be part of the (normal) Helper and Leader variants of AggregatorTaskParameters. I'm generally not a fan of passing false data (although I suppose an empty slice is not really _false_ per se). WDYT?

tgeoghegan: I tried that for the reasons you say. The issue is that then you can't have `AggregatorTask::hpke_keys(&self) -> &HashMap<HpkeConfigId, HpkeKeypair>`, because not all variants have HPKE keys. From there there are two options:

- return `Option<&HashMap<HpkeConfigId, HpkeKeypair>>` and return `None` for task prov tasks. This forces the caller to do annoying `Option` unwrapping or iterator flattening gymnastics despite `None` not being any more expressive than an empty hash.
- return an owned `HashMap<...>` value and return an empty hash for the taskprov case. This is unfortunate because it forces an allocation on one path, or a clone (and hence allocation) on the other (you can't `return &HashMap::new()` because the value wouldn't live long enough).

All three options are kind of awkward, but I think this is the least awkward. I do think we can do better here, by further unifying the taskprov and non-taskprov cases such that code holding an `AggregatorTask` value doesn't need to know where the task came from. `AggregatorTask::hpke_keys` would yield a map of keys, which would have been constructed from the `global_hpke_keys` table in the taskprov case or `task_hpke_keys` otherwise. But I don't propose to do that refactoring now.

inahga (approved):

PR #2023, opened 2023-09-29
Taskprov: fix authorization bug in aggregate init call

inahga:
Calls to /aggregate subsequent to taskprov task provisioning would fail for the same task, because /aggregate was not using the right authorization tokens (that is, it was attempting to use non-existent authorization tokens and failing).

The first commit is just a refactoring PR to make the test implementation cleaner. The second commit is the operative one.

### DIFF
--- a/aggregator/src/aggregator/taskprov_tests.rs
+++ b/aggregator/src/aggregator/taskprov_tests.rs
@@ -78,238 +78,250 @@ pub struct TaskprovTestCase {
     datastore: Arc<Datastore<MockClock>>,
     handler: Box<dyn Handler>,
     peer_aggregator: PeerAggregator,
-    report_metadata: ReportMetadata,
-    transcript: VdafTranscript<16, TestVdaf>,
-    report_share: ReportShare,
     task: Task,
     task_config: TaskConfig,
     task_id: TaskId,
-    aggregation_param: Poplar1AggregationParam,
+    vdaf: TestVdaf,
+    global_hpke_key: HpkeKeypair,
 }
 
-async fn setup_taskprov_test() -> TaskprovTestCase {
-    install_test_trace_subscriber();
+impl TaskprovTestCase {
+    async fn new() -> Self {
+        install_test_trace_subscriber();
 
-    let clock = MockClock::default();
-    let ephemeral_datastore = ephemeral_datastore().await;
-    let datastore = Arc::new(ephemeral_datastore.datastore(clock.clone()).await);
+        let clock = MockClock::default();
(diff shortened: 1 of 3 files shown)


tgeoghegan:
I think you'll need to rebase now that #2019 has landed, and maybe some more as I will be landing changes to `aggregator/src/aggregator.rs`.

inahga:
No problem, I can hang onto this until the PR churn is done.

inahga:
Rebase complete, should be ready for review.

aggregator/src/aggregator/taskprov_tests.rs
```diff
@@ -78,238 +78,276 @@ pub struct TaskprovTestCase {
...
-    assert_eq!(prepare_step.report_id(), test.report_share.metadata().id());
-    assert_matches!(prepare_step.result(), &PrepareStepResult::Continue { .. });
+        assert_eq!(test_conn.status(), Some(Status::Ok), "{}", name);
+        assert_headers!(
+            &test_conn,
+            "content-type" => (AggregationJobResp::MEDIA_TYPE)
+        );
+        let aggregate_resp: AggregationJobResp = decode_response_body(&mut test_conn).await;
+
+        assert_eq!(aggregate_resp.prepare_resps().len(), 1, "{}", name);
```
tgeoghegan, line 336: nit:
```suggestion
        assert_eq!(aggregate_resp.prepare_resps().len(), 1, "{name}");
```
and in a few other places

aggregator/src/aggregator/taskprov_tests.rs
```diff
@@ -78,238 +78,276 @@ pub struct TaskprovTestCase {
...
+            peer_aggregator,
+            task,
+            task_config,
+            task_id,
+            vdaf,
+            global_hpke_key,
+        }
+    }
+
+    fn generate_helper_report_share(
```
tgeoghegan, line 202: nit: this looks a lot like what `aggregator::aggregate_init_tests::PrepareInitGenerator` does. It goes and constructs a `PrepareInit`, but on the way there it does generate a VDAF transcript and generate a report share. I wonder if you could add a `PrepareInitGenerator::next_report_share_with_metadata` and use that here.

tgeoghegan (approved):

aggregator/src/aggregator/taskprov_tests.rs
inahga: WDYT about the latest commit? It's not really a simplification IMO, but I like that it expands on PrepareInitGenerator to make other kinds of tests easier to write.

tgeoghegan (approved):

PR #2028, opened 2023-09-29
Task rewrite: `janus_cli`

tgeoghegan:
# Stacked on #2027

Adopts `AggregatorTask` in `janus_cli`. There's a functional change here: the YAML task accepted by `janus_cli` is now an `AggregatorTask` (via `SerializedAggregatorTask`), containing an aggregator's specific view of a task, meaning it contains just the peer aggregator's endpoint URL. In the near future, there will be more changes to the task YAML format so that for instance leader tasks only get the collector auth token _hash_ and helper tasks only get the aggregator auth token _hash_.

Part of #1524

### DIFF
--- a/docs/samples/tasks.yaml
+++ b/docs/samples/tasks.yaml
@@ -6,9 +6,8 @@
   # DAP's recommendation.
   task_id: "G9YKXjoEjfoU7M_fi_o2H0wmzavRb2sBFHeykeRhDMk"
 
-  # HTTPS endpoints of the leader and helper aggregators.
-  leader_aggregator_endpoint: "https://example.com/"
-  helper_aggregator_endpoint: "https://example.net/"
+  # HTTPS endpoints of the peer aggregator.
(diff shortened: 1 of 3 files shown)


docs/samples/tasks.yaml
```diff
@@ -6,9 +6,8 @@
   # DAP's recommendation.
   task_id: "G9YKXjoEjfoU7M_fi_o2H0wmzavRb2sBFHeykeRhDMk"
 
-  # HTTPS endpoints of the leader and helper aggregators.
-  leader_aggregator_endpoint: "https://example.com/"
-  helper_aggregator_endpoint: "https://example.net/"
+  # HTTPS endpoints of the peer aggregator.
```
divergentdave, line 9: nit: grammar
```suggestion
  # HTTPS endpoint of the peer aggregator.
```

divergentdave (approved):

PR #2029, opened 2023-09-29
Task rewrite: integration tests

tgeoghegan:
# Stacked on #2028 

Adopts `AggregatorTask`, `test_util::Task` and `NewTaskBuilder` across the integration test and interop binaries modules.

Part of #1524

### DIFF
--- a/integration_tests/tests/daphne.rs
+++ b/integration_tests/tests/daphne.rs
@@ -1,11 +1,12 @@
-use common::{submit_measurements_and_verify_aggregate, test_task_builders};
-use janus_aggregator_core::task::{QueryType, Task};
+use common::{submit_measurements_and_verify_aggregate, test_task_builder};
+use janus_aggregator_core::task::QueryType;
 use janus_core::{
     test_util::{install_test_trace_subscriber, testcontainers::container_client},
     vdaf::VdafInstance,
 };
 use janus_integration_tests::{client::ClientBackend, daphne::Daphne, janus::Janus};
 use janus_interop_binaries::test_util::generate_network_name;
+use janus_messages::Role;
 
 mod common;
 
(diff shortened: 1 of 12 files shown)


interop_binaries/Cargo.toml
```diff
@@ -30,7 +30,7 @@ futures = { version = "0.3.28", optional = true }
 fixed = { version = "1.23", optional = true }
 fixed-macro = { version = "1.1.1", optional = true }
 hex = { version = "0.4", optional = true }
-janus_aggregator_core.workspace = true
+janus_aggregator_core = { workspace = true, features = ["test-util"] }
```
divergentdave, line 33: Note that this will turn on the janus_aggregator_core/test-util feature in workspace-wide builds. This shouldn't be a huge problem, as we do `-p janus_aggregator` builds as part of CI, which would catch any mistakes missed in the workspace build.

integration_tests/tests/daphne.rs
```diff
@@ -58,26 +53,21 @@ async fn janus_daphne() {
...
-                .build()
-        })
-        .collect::<Vec<_>>()
-        .try_into()
-        .unwrap();
+    task_parameters.endpoint_fragments.leader_endpoint_path = "/v04/".to_string();
+    let mut leader_aggregator_endpoint = task_builder.leader_aggregator_endpoint().clone();
+    leader_aggregator_endpoint.set_path("/v04/");
+    let task = task_builder
+        .with_leader_aggregator_endpoint(leader_aggregator_endpoint)
```
divergentdave, line 65: This should be applied to the helper endpoint instead.

interop_binaries/Cargo.toml
tgeoghegan: `integration_tests/Cargo.toml` already had `janus_aggregator_core = { workspace = true, features = ["test-util"] }` so this was already true, wasn't it? In any case I agree that CI makes sure that builds without `test-util` work, as should the Docker image builds.

integration_tests/tests/daphne.rs
```diff
@@ -58,26 +53,20 @@ async fn janus_daphne() {
...
-        .map(|task| {
-            let mut helper_aggregator_endpoint = task.helper_aggregator_endpoint().clone();
-            helper_aggregator_endpoint.set_path("/v04/");
-            task.with_helper_aggregator_endpoint(helper_aggregator_endpoint)
-                .build()
-        })
-        .collect::<Vec<_>>()
-        .try_into()
-        .unwrap();
+    task_parameters.endpoint_fragments.leader_endpoint_path = "/v04/".to_string();
```
divergentdave, line 60: I think this needs to change as well, so the collector doesn't get a 404 when talking to the Janus leader.

divergentdave (approved):

PR #2030, opened 2023-09-29
Task rewrite: remove temporary adapters

tgeoghegan:
# Stacked on #2029

Removes the old `Task`, `SerializedTask` and `TaskBuilder` structures now that everything uses the new stuff. `NewTaskBuilder` is renamed to `TaskBuilder` since it's the only `TaskBuilder` we've got.

Part of #1524


divergentdave (approved):
FWIW this is much easier to review locally, passing the `--patience` flag to `git diff`, as that correctly identifies that this PR is deleting huge contiguous chunks.

PR #2031, opened 2023-10-02
Remove Clock trait from janus_client API

### LINKED ISSUE #1441: Simplify janus_client's time handling
We should remove the `Clock` trait from `janus_cllient`'s public interface. Instead, we should have two methods to submit measurements, one that uses the current time, and one that uses a provided report timestamp. Note that this would be a breaking change.

divergentdave:
This closes #1441, getting rid of the `Clock` generic from the `janus_client` API, and adding a separate method that takes an explicit timestamp.


client/src/lib.rs
```diff
@@ -217,15 +214,13 @@ impl<V: vdaf::Client<16>, C: Clock> Client<V, C> {
 
     /// Shard a measurement, encrypt its shares, and construct a [`janus_core::message::Report`]
     /// to be uploaded.
-    fn prepare_report(&self, measurement: &V::Measurement) -> Result<Report, Error> {
+    fn prepare_report(&self, measurement: &V::Measurement, time: &Time) -> Result<Report, Error> {
```
tgeoghegan, line 217: Could we take `std::time::Instant` here to further reduce this interface's dependency on Janus concepts? But maybe that's not a good idea because `Instant` isn't flexible enough.

tgeoghegan (approved):

client/src/lib.rs
divergentdave: Instant wouldn't work, because that's often a monotonic counter and not wall clock time. We could take a `SystemTime` or a `chrono` type.

jbr: SystemTime wouldn't require reexporting a particular version of chrono or time

tgeoghegan (approved):

PR #2062, opened 2023-10-02
Only store aggregator auth token hash in helper

tgeoghegan:
When acting as the helper, Janus only needs the aggregator auth token hash to validate requests. It never makes aggregation sub-protocol requests and thus doesn't need the actual token.

In this commit we change the schema of the `tasks` table so that helpers may store an auth token hash instead of an auth token. These changes are reflected in `janus_aggregator_core::task::AggregatorTask`. Finally we make changes to the aggregator API to accommodate this: the helper reports the aggregator auth token it generates during task provisioning, but doesn't store that token and cannot subsequently provide it when responding to `GET /tasks/:task_id` requests. Fortunately that field in `TaskResp` was already optional, and divviup-api already ignored it except in the creation case.

Part of #1509, 1524

### DIFF
--- a/aggregator_api/src/routes.rs
+++ b/aggregator_api/src/routes.rs
@@ -14,7 +14,9 @@ use janus_aggregator_core::{
     taskprov::PeerAggregator,
     SecretBytes,
 };
-use janus_core::{hpke::generate_hpke_config_and_private_key, time::Clock};
+use janus_core::{
+    auth_tokens::AuthenticationTokenHash, hpke::generate_hpke_config_and_private_key, time::Clock,
+};
 use janus_messages::HpkeConfigId;
 use janus_messages::{
     query_type::Code as SupportedQueryType, Duration, HpkeAeadId, HpkeKdfId, HpkeKemId, Role,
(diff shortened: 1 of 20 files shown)


tgeoghegan:
Downgrading to draft while I fix tests.

aggregator_api/src/routes.rs
```diff
@@ -194,9 +204,15 @@ pub(super) async fn post_task<C: Clock>(
...
-    Ok(Json(
-        TaskResp::try_from(task.as_ref()).map_err(|err| Error::Internal(err.to_string()))?,
-    ))
+    let mut task_resp =
+        TaskResp::try_from(task.as_ref()).map_err(|err| Error::Internal(err.to_string()))?;
+
+    // When creating a new task in the helper, we must put the unhashed aggregator auth token in the
+    // response so that divviup-api can later provide it to the leader, but the helper doesn't store
+    // the unhashed token and can't later provide it.
+    task_resp.aggregator_auth_token = aggregator_auth_token;
```
inahga, line 213: Somewhere we ought to document the expected usage of the aggregator API for provisioning tasks, i.e. provision the helper first, then the leader using the generated auth token. Right now it's not clear that's the intended usage.

Doesn't have to be done in this PR.

inahga (approved):

divergentdave (approved):

PR #2064, opened 2023-10-02
only store collector auth token hash in leader

tgeoghegan:
# Stacked on #2062

The leader only needs the hash of the collector auth token in order to authenticate collection sub-protocol requests and thus we can avoid the risk of storing the unhashed token.

This commit changes the `tasks` table to store a hash instead of the token, and makes the corresponding change to
`janus_aggregator_core::task::AggregatorTask`. There are corresponding changes to the aggregator API: the leader must now be provided the collector auth token hash during task creation, and `TaskResp` now only contains a collector auth token hash.

Part of #1509, #1524

### DIFF
--- a/aggregator_api/src/routes.rs
+++ b/aggregator_api/src/routes.rs
@@ -49,7 +49,7 @@ pub(super) async fn get_config(
             SupportedQueryType::FixedSize,
         ],
         // Unconditionally indicate to divviup-api that we support collector auth token hashes
-        features: &["token-hash"],
+        features: &["TokenHash"],
     })
 }
 
--- a/aggregator_api/src/routes.rs
+++ b/aggregator_api/src/routes.rs
@@ -48,6 +48,8 @@ pub(super) async fn get_config(
             SupportedQueryType::TimeInterval,
             SupportedQueryType::FixedSize,
         ],
+        // Unconditionally indicate to divviup-api that we support collector auth token hashes
+        features: &["token-hash"],
     })
 }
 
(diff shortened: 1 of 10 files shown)


aggregator_api/src/routes.rs
```diff
@@ -48,6 +48,8 @@ pub(super) async fn get_config(
             SupportedQueryType::TimeInterval,
             SupportedQueryType::FixedSize,
         ],
+        // Unconditionally indicate to divviup-api that we support collector auth token hashes
+        features: &["token-hash"],
```
inahga, line 52: I have mild apprehension about introducing a feature flag for this, as if we have permanent/long-lived feature flags we risk combinatorial explosion of features down the line ([among other problems](https://bpapillon.com/post/feature-flags-theory-vs-reality/)).

I get the purpose is to avoid having to backport this feature (and we should avoid it IMO), but we should revisit this when we drop DAP-04. Perhaps can defer it with an issue for "remove `token-hash` feature flag?".

aggregator_api/src/tests.rs
```diff
@@ -86,7 +86,8 @@ async fn get_config() {
         concat!(
             r#"{"protocol":"DAP-07","dap_url":"https://dap.url/","role":"Either","vdafs":"#,
             r#"["Prio3Count","Prio3Sum","Prio3Histogram","Prio3CountVec","Prio3SumVec"],"#,
-            r#""query_types":["TimeInterval","FixedSize"]}"#
+            r#""query_types":["TimeInterval","FixedSize"],"#,
+            r#""features":["token-hash"]}"#,
```
inahga, line 90: nit: the rest of our features are camel case. No opinion on whether that's right or not, but I think it stands to be consistent.

inahga (approved):

aggregator_api/src/routes.rs
tgeoghegan: divviup-api needs to dynamically figure out whether a Janus aggregator API it is working with needs collector auth token hashes or tokens. @jbr suggested not to infer this from `"protocol": "DAP-07"` since we might deploy this feature on other Janus versions, which I agree with. But you're right that we can eventually drop the feature: #2066

aggregator_api/src/tests.rs
tgeoghegan: Sure, I changed it to `TokenHash`. @jbr do you have a strong feeling about the string here?

divergentdave (approved):

PR #2067, opened 2023-10-03
subscriber-01: Implement draft-wang-ppm-dap-taskprov-02

inahga:
Adopts draft-wang-ppm-dap-taskprov-02 for the subscriber-01 branch.

The primary difference between this revision and the latest is that the taskprov configuration is expressed in the extension field of each report share. A helper must parse the extension in each AggregateInitRequest and use it to provision the task.

The implementation of this is unclean. Due to how AggregateInitRequest is structured, we need to know the query type ahead of time. We can't decode it unless we know the query type, but we won't know the query type until we decode it. To work around this, decode by trial.

To simplify implementation I made all helper requests use the main aggregator auth token established in https://github.com/divviup/janus/pull/1976. Leader requests are unmodified, but are effectively inert/useless now. We can consider removing leader endpoints in a future PR.

The first commit contains the operative logic, the remaining commits address tests.

Tests underwent considerable modification. The aggregate init harness needed to be adapted to provide the taskprov extension and aggregator authentication token. I accepted the large amount of change on the basis that we will likely just want to revert this PR to adopt taskprov-04+ again.

I disabled all integration tests as this branch can no longer service non-taskprov leader->helper requests.

### DIFF
--- a/aggregator/src/aggregator/aggregate_init_tests.rs
+++ b/aggregator/src/aggregator/aggregate_init_tests.rs
@@ -6,76 +6,96 @@ use janus_aggregator_core::{
         test_util::{ephemeral_datastore, EphemeralDatastore},
         Datastore,
     },
-    task::{test_util::TaskBuilder, QueryType, Task},
+    task::{QueryType, Task},
+    taskprov::VerifyKeyInit,
     test_util::noop_meter,
+    SecretBytes,
 };
 use janus_core::{
-    task::{AuthenticationToken, VdafInstance, DAP_AUTH_HEADER},
+    hpke::{test_util::generate_test_hpke_config_and_private_key, HpkeKeypair},
+    task::{AuthenticationToken, DAP_AUTH_HEADER},
     test_util::{dummy_vdaf, install_test_trace_subscriber, run_vdaf, VdafTranscript},
-    time::{Clock, MockClock, TimeExt as _},
+    time::{Clock, DurationExt, MockClock, TimeExt as _},
 };
 use janus_messages::{
-    query_type::TimeInterval, AggregateInitializeReq, PartialBatchSelector, ReportMetadata,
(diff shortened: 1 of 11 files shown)


aggregator/src/aggregator/aggregate_init_tests.rs
```diff
@@ -103,73 +128,150 @@ pub(super) async fn setup_aggregate_init_test() -> AggregationJobInitTestCase {
...
         clock.clone(),
         &noop_meter(),
-        Config::default(),
+        config.clone(),
     )
     .await
     .unwrap();
 
-    let aggregation_param = dummy_vdaf::AggregationParam(0);
+    let vdaf = Prio3Aes128Count::new_aes128_count(2).unwrap();
```
inahga, line 179: I'm unhappy about using Prio3 as the test VDAF as that requires us to discard tests that exercise varying aggregation parameters (unless I'm mistaken?).

Unsure if it's really worth addressing in this PR. We can either adapt the taskprov messages to accept `dummy_vdaf` when compiled under tests, or I can try harder to use Poplar1 (the API is quite different in libprio-rs 0.10 compared to the latest).

tgeoghegan (approved):
I'm skeptical of the value of investing more work into this branch without more clarity on the requirements here. Thus far we've been making the guesses and divining what to do here from reading tea leaves, and it's already led to wasted effort (going to the taskprov that puts the config in a header and now back to per-report-extensions). Or in the context of this PR, we could simplify it somewhat if we knew which of the two query modes the subscriber is going to use exclusively.

That said, this particular piece of work is already done and this accomplishes what it sets out to do, so I think we can take this change.

inahga:
Agreed that we ought to freeze changes to this branch here until we have more clarity on requirements (or better yet, direct time with their engineering).

Issue #230, opened 2022-06-13
HPKE interface improvements


branlwyd:
Potential improvements to our HPKE interface library (roughly, `janus/src/hpke.rs`):

- [ ] refactor HpkePrivateKey, HpkePublicKey to carry around a decoded key rather than a byte string (more challenging now that we are routing through the `hpke_dispatch` library) to avoid deserializing on every cryptographic operation
- [x] everywhere that uses an `HpkePrivateKey` also requires an `HpkeConfig` for context; refactor to use e.g. `HpkePrivateConfig` which is equivalent to `(HpkeConfig, HpkePrivateKey)` to reduce the number of parameters we're passing around & reduce the chances of a mismatch bug (#2072)

branlwyd:
Closing as not planned -- the remaining decoding step does not cause a performance issue.

PR #2072, opened 2023-10-03
Use HpkeKeypair instead of config and private key

divergentdave:
This changes `janus_core::hpke::open()` and `janus_collector::CollectorParameters::new()` to take a single `HpkeKeypair` instead of separate `HpkeConfig` and `HpkePrivateKey` arguments. This addresses part of #230.


tgeoghegan (approved):
Good stuff! This was bothering me when I was rewriting the task structures but I opted not to do anything about it that already large changeset, so I'm glad to see this addressed.

PR #2074, opened 2023-10-03
Include `task_id` in effective primary keys for all tables.

### LINKED ISSUE #1550: Add index on `collection_jobs.collection_job_id`.
Currently, none of the indexes on `collection_jobs` refers to `collection_job_id`. But e.g. `get_collection_job` (and perhaps other operations) do lookups by collection job ID. I suspect this means retrieving a collection job by ID is currently a table-scan (have not analyzed); this will likely impact scalability for larger deployments and should be fixed.

branlwyd:
This updates the `aggregation_jobs` & `collection_jobs` tables; previously, only the direct ID (`aggregation_job_id` or `collection_job_id`) was treated as the effective primary key. For both tables, this will allow an indexed lookup where previously a scan was required (in `get_aggregation_job`/`get_collection_job`, respectively).

Closes #1550.

### DIFF
--- a/db/00000000000001_initial_schema.up.sql
+++ b/db/00000000000001_initial_schema.up.sql
@@ -167,7 +167,7 @@ CREATE TABLE aggregation_jobs(
     lease_token              BYTEA,                                             -- a value identifying the current leaseholder; NULL implies no current lease
     lease_attempts           BIGINT NOT NULL DEFAULT 0,                         -- the number of lease acquiries since the last successful lease release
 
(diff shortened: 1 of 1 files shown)


branlwyd:
One day I will remember to add the `allow-changed-migrations` label before sending the PR. Today... is not that day.

tgeoghegan (approved):

tgeoghegan:
Does this change collide with #2073 at all?

PR #2077, opened 2023-10-04
Collector API redesign

### LINKED ISSUE #762: Expose collect job deletion in collector crate
`janus_collector` should expose a way to delete a collect job, as [required by the specification](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.5.1-15) and also implemented in Janus.

### LINKED ISSUE #1796: Collector API redesign

divergentdave:
This overhauls the public API of janus_collector. `CollectorParameters` is removed, and functionally replaced with the new `CollectorBuilder`. `CollectionJob` and `PollResult` are made public, along with methods for starting and polling a job. A new method is added to delete a job.

`CollectionJob` has a public constructor and `into_fields()` method, so that it can be serialized and reconstructed in stateful collectors, but I chose not to define serde implementations (or `prio::codec` implementations) in this PR, since the struct has two generic parameters, mixing VDAF-level and DAP-level concepts. Implementing serialization in an opinionated way may require additional trait bounds. The constructor and access to fields at least makes it possible for applications to do their own serialization that suits their needs.

Closes #1796 and closes #762.

### DIFF
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -171,13 +171,19 @@ impl<P, Q: QueryType> CollectionJob<P, Q> {
         }
     }
 
-    /// Destructure a collection job into its fields.
-    pub fn into_fields(self) -> (CollectionJobId, Query<Q>, P) {
-        (
-            self.collection_job_id,
-            self.query,
-            self.aggregation_parameter,
-        )
+    /// Gets this collection job's identifier.
+    pub fn collection_job_id(&self) -> &CollectionJobId {
+        &self.collection_job_id
+    }
+
+    /// Gets the query used to create this collection job.
+    pub fn query(&self) -> &Query<Q> {
+        &self.query
+    }
+
+    /// Gets the aggregation parameter used to create this collection job.
+    pub fn aggregation_parameter(&self) -> &P {
+        &self.aggregation_parameter
     }
 }
 
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -7,7 +7,7 @@
 //! # Examples
 //!
 //! ```no_run
-//! use janus_collector::{AuthenticationToken, Collector, CollectorParameters, default_http_client};
+//! use janus_collector::{AuthenticationToken, Collector};
 //! use janus_core::{hpke::generate_hpke_config_and_private_key};
 //! use janus_messages::{
 //!     Duration, HpkeAeadId, HpkeConfig, HpkeConfigId, HpkeKdfId, HpkeKemId, Interval, TaskId,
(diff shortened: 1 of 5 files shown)


collector/src/lib.rs
```diff
@@ -334,59 +275,190 @@ where
...
+        }
+    }
+
+    /// Finalize construction of a [`Collector`].
+    pub fn build(self) -> Result<Collector<V>, Error> {
+        let http_client = if let Some(http_client) = self.http_client {
+            http_client
+        } else {
+            default_http_client()?
+        };
```
inahga, line 333: Q: Is having the default inserted here just so that `new()` isn't fallible?

inahga (approved):
Nicely written!

collector/src/lib.rs
divergentdave: It was moreso to avoid setting up and tearing down synchronization primitives, etc. for the connection pool if a separate client is supplied, and the default goes unused.

collector/src/lib.rs
```diff
@@ -135,99 +134,32 @@ static COLLECTOR_USER_AGENT: &str = concat!(
...
-            "tasks/{}/collection_jobs/{collection_job_id}",
-            self.task_id
-        ))?)
-    }
-}
-
 /// Construct a [`reqwest::Client`] suitable for use in a DAP [`Collector`].
 pub fn default_http_client() -> Result<reqwest::Client, Error> {
     Ok(reqwest::Client::builder()
-        .redirect(reqwest::redirect::Policy::none())
```
tgeoghegan, line 208: What's the effect of removing the redirect policy?

collector/src/lib.rs
```diff
@@ -238,21 +170,30 @@ impl<P, Q: QueryType> CollectionJob<P, Q> {
             aggregation_parameter,
         }
     }
+
+    /// Destructure a collection job into its fields.
+    pub fn into_fields(self) -> (CollectionJobId, Query<Q>, P) {
```
tgeoghegan, line 175: This method seems like it's roughly equivalent to making the three fields on the struct `pub`. I get that the intent is to enable clients to work out their own serialization scheme, but I don't know if this is much better than three different getter methods. Given that this is public API, I think it might be wiser not to add this and see about adding serialization later, since that wouldn't be a breaking change.

tgeoghegan (approved):

collector/src/lib.rs
divergentdave: It was only needed for pre-draft-04, when we expected 303 See Other responses, and needed to capture the Location header for multiple subsequent requests. Since we don't expect any redirects from DAP servers, it should have no impact, but if a sufficiently weird DAP aggregator returned a redirect response, we would now follow it instead of trying to parse an empty response or bailing on an unexpected status code.

tgeoghegan: Yep, makes sense!

collector/src/lib.rs
divergentdave: I swapped this out for normal getters

tgeoghegan (approved):

PR #2080, opened 2023-10-05
Update to testcontainers-rs 0.15

divergentdave:
This updates to the just-released version 0.15.0 of testcontainers. This version removed all pre-defined container images to reduce maintenance burden, so I vendored the `Postgres` image into `janus_core` and simplified it.

### DIFF
--- a/core/src/test_util/testcontainers.rs
+++ b/core/src/test_util/testcontainers.rs
@@ -1,7 +1,10 @@
 //! Testing functionality that interacts with the testcontainers library.
 
-use std::sync::{Arc, Mutex, OnceLock, Weak};
-use testcontainers::clients::Cli;
+use std::{
+    collections::HashMap,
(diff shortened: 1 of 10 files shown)


core/src/test_util/testcontainers.rs
```diff
@@ -19,3 +22,47 @@ pub fn container_client() -> Arc<Cli> {
         client
     })
 }
+
+/// A [`testcontainers::Image`] that provides a Postgres database cluster.
```
tgeoghegan, line 26: nit: a single container is not really a cluster

tgeoghegan (approved):

divergentdave:
This is now failing due to Rust 1.73.0 Clippy changes, I'll address those separately and rebase.

core/src/test_util/testcontainers.rs
tgeoghegan: Oh, neat, TIL.

PR #2091, opened 2023-10-09
Fix cli test for `fpvec_bounded_l2` feature flag.

MxmUrw:
It seems that the `collect.trycmd` file was updated, while the `collect_fpvec_bounded_l2.trycmd` file was not. The result was that the relevant test didn't pass when running `cargo test --features fpvec_bounded_l2`. This commit fixes it.


inahga (approved):
Thanks!

PR #2112, opened 2023-10-10
Test all features in CI

inahga:
Follow up to https://github.com/divviup/janus/pull/2091, this error should be caught in CI.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -73,13 +73,13 @@ jobs:
       run: cargo build --profile ci --package janus_core
     # Note: keep Build & Test steps consecutive, and match flags other than `--no-run`.
     - name: Build
-      run: cargo test --profile ci --no-run --locked --all-targets
+      run: cargo test --profile ci --no-run --locked --all-targets --features fpvec_bounded_l2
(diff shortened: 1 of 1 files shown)


inahga:
Hmm, this CI should have actually failed since it wasn't based on https://github.com/divviup/janus/pull/2091.

.github/workflows/ci-build.yml
```diff
@@ -73,7 +73,7 @@ jobs:
       run: cargo build --profile ci --package janus_core
     # Note: keep Build & Test steps consecutive, and match flags other than `--no-run`.
     - name: Build
-      run: cargo test --profile ci --no-run --locked --all-targets
+      run: cargo test --profile ci --no-run --locked --all-targets --all-features
```
divergentdave, line 76: This needs to be repeated on the line below without `--no-run`

inahga:
This turns out to not be as straightforward as a fix as I had hoped. There are mutually exclusive tests between the default featureset and `fpvec_bounded_l2`, so we need to run tests twice for each featureset. This needs some careful thought on how not to blow up the build time. I will consider this in https://github.com/divviup/janus/issues/2115.

PR #2117, opened 2023-10-12
Bump main version to 0.6.0-prerelease-4

inahga:


inahga:
Note that I will not be deploying this version this week, as we're awaiting divviup-api support. https://github.com/divviup/divviup-api/issues/542

tgeoghegan (approved):

PR #2122, opened 2023-10-13
Differential Privacy Cleanups

### LINKED ISSUE #1658: Remove Prio3FixedPoint64BitBoundedL2VecSum
The instantiation of the fixed point vector VDAF with 64-bit fixed point numbers is of limited use, because it can only handle one- two-, or three-dimensional vectors.

```
failed to construct Prio3FixedPoint64BitBoundedL2VecSum VDAF

Caused by:
    0: flp error: value error: number of entries (4) not compatible with field size
    1: value error: number of entries (4) not compatible with field size
```

This is because the vector coordinates each get encoded in 63 bits, so the square of the vector's L2-norm gets encoded in 126 bits. Adding four squared coordinates together would bring you to $\approx 2^{128}$, which would exceed the prime modulus of `Field128`.

Given that the intended use case of this VDAF is federated learning, where low dimensionality and 64-bit fixed point precision are uncommon, I think we should remove this VDAF, leaving just the 16-bit and 32-bit versions. If need be, we can always add another specialization at, say, 48 bits down the line.

cc @ooovi @MxmUrw

divergentdave:
This cleans up a couple items as a follow-up to #1892. First, we close #1658 by removing the 64-bit variant of Prio3FixedPointBoundedL2VecSum. Due to length constraints, this could only be used on vectors with maximum dimension of three. Lower-precision variants will be more practically useful. Second, I removed the DP strategy generic type parameter from `compute_aggregate_share()`, and relaxed the bound on the VDAF type parameter from `AggregatorWithNoise` to `Aggregator`. This simplifies callsites and gets rid of some extra monomorphizations.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -16,8 +16,8 @@ use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use bytes::Bytes;
 #[cfg(feature = "fpvec_bounded_l2")]
 use fixed::{
-    types::extra::{U15, U31, U63},
-    FixedI16, FixedI32, FixedI64,
+    types::extra::{U15, U31},
+    FixedI16, FixedI32,
 };
 use futures::future::try_join_all;
 use http::{header::CONTENT_TYPE, Method};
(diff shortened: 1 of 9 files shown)


tgeoghegan (approved):
LGTM, provided losing the 64bit variant doesn't break @ooovi and @MxmUrw

MxmUrw:
> LGTM, provided losing the 64bit variant doesn't break @ooovi and @MxmUrw

Yes, 64bit isn't needed for realistic use cases. Fine with us.

Issue #2081, opened 2023-10-05
Proposed design for removing container image blobs


divergentdave:
Currently,  we build docker images inside `interop_binaries/build.rs` and incorporate the resulting images into this crate and `janus_integration_tests`, its dependent, via `include_bytes!()`. This has multiple negative knock-on effects, including space amplification from including the image blobs in many intermediate compiler files and test binaries, increased build times both from processing the blobs and putting an additional optimized build in the critical path, need for workarounds in IDE configuration, increased GitHub Actions cache pressure, and barriers for new contributors. I propose the following plan to remove these blobs, while keeping the same overall test coverage. The main downside is that `cargo test` would no longer perform all tests (except `in_cluster`), instead one would need to run two commands, `cargo test` and `cargo xtask name_of_xtask`, in order to replicate the same test coverage.

* Move most of the contents of src/bin/ into new modules in the library crate. The binary files should just call into functions defined in the libraries, ideally.
  * Bonus: this will allow reusing `Config` structs in tests like `graceful_shutdown`, rather than constructing YAML files longhand.
* Add a way for the aggregator's entry point to optionally send back the port it is bound on to its caller, to enable use with the socket address "127.0.0.1:0".
* Create a new set of integration tests that run two aggregators in-process. We would define some new helpers that construct all the `Config` and `Option` structs needed, and then spawn each aggregator component on a separate Tokio task. These tests will tell us if the whole DAP process is working end-to-end, which is 80% of the value we get out of the existing integration tests.
* Move the existing integration tests that depend on Janus's interop test containers behind a new Cargo feature, (perhaps in a separate crate too) and take container image names in via environment variables rather than importing container images from compiled-in blobs (this would work akin to the in-cluster feature and tests we have now). We would keep these integration tests around so that we can confirm our container images themselves don't have any problems not caught by the new single-process tests.
* Follow the [cargo-xtask](https://github.com/matklad/cargo-xtask) pattern to create a command that invokes `docker buildx bake interop_binaries --load` to build interop binary container images, collect the resulting image IDs, and pass them via environment variables to `cargo test -p janus_integration_tests --feature new-feature -- filter`.
  * We could also remove `PROFILE=small` variants from `docker-bake.hcl`, since they will no longer be needed.
[...]

PR #2123, opened 2023-10-13
Add single process integration tests

divergentdave:
This is part of #2081. This PR adds new integration tests that run both Janus aggregator instances in-process, relying on the existing ephemeral database support for Postgres. This also includes one test with an in-process Janus leader and a Daphne helper running in a container.

Note that the divviup-ts integration tests, or tests with Daphne as the leader, are not supported with in-process aggregators yet. That will require making connections from inside a container to TCP servers on the host, which may be doable using `host.docker.internal` on Windows/Mac and `--add-host` with `host-gateway` on Linux, FWIW.

### DIFF
--- a/integration_tests/src/janus.rs
+++ b/integration_tests/src/janus.rs
@@ -1,13 +1,39 @@
 //! Functionality for tests interacting with Janus (<https://github.com/divviup/janus>).
 
 use crate::interop_api;
-use janus_aggregator_core::task::test_util::Task;
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
+use janus_aggregator::{
+    binaries::{
+        aggregation_job_creator::{
+            self, Config as AggregationJobCreatorConfig, Options as AggregationJobCreatorOptions,
+        },
+        aggregation_job_driver::{
+            self, Config as AggregationJobDriverConfig, Options as AggregationJobDriverOptions,
+        },
+        aggregator::{self, Config as AggregatorConfig, Options as AggregatorOptions},
+        collection_job_driver::{
+            self, Config as CollectionJobDriverConfig, Options as CollectionJobDriverOptions,
+        },
+    },
+    binary_utils::{BinaryContext, CommonBinaryOptions},
+    config::{CommonConfig, DbConfig, JobDriverConfig, TaskprovConfig},
+    metrics::MetricsConfiguration,
+    trace::{TokioConsoleConfiguration, TraceConfiguration},
+};
+use janus_aggregator_core::{
+    datastore::test_util::{ephemeral_datastore, EphemeralDatastore},
+    task::test_util::Task,
+    test_util::noop_meter,
+};
+use janus_core::time::RealClock;
 use janus_interop_binaries::{
     get_rust_log_level, test_util::await_http_server, testcontainer::Aggregator,
     ContainerLogsDropGuard,
 };
 use janus_messages::Role;
+use std::net::{Ipv4Addr, SocketAddr};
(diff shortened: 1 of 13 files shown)


integration_tests/src/janus.rs
```diff
@@ -56,3 +85,191 @@ impl<'a> Janus<'a> {
...
+            .await
+            .expect("task provisioning failed");
+
+        // Construct configuration and options for each component.
+        let common_binary_options = CommonBinaryOptions {
+            config_file: PathBuf::new(),
+            database_password: None,
+            datastore_keys,
+            otlp_tracing_metadata: Vec::new(),
+            otlp_metrics_metadata: Vec::new(),
```
tgeoghegan, line 125: nit: could we provide `Default` on `CommonBinaryOptions` and do `CommonBinaryOptions { datastore_keys, ..Default::default() }`?

integration_tests/src/lib.rs
```diff
@@ -24,47 +24,89 @@ pub struct TaskParameters {
...
+    /// Provides the URL for the aggregator's endpoint from the perspective of the host. If the
+    /// aggregator is in a virtual network, this will use a port forward, with the provided port
+    /// number. If the aggregator itself is running on the host, it will use the provided port
+    /// number as well.
+    pub fn endpoint_for_host(&self, port: u16) -> Url {
+        let path = match self {
+            AggregatorEndpointFragments::VirtualNetwork { path, .. } => path,
+            AggregatorEndpointFragments::Localhost { path } => path,
+        };
+        Url::parse(&format!("http://127.0.0.1:{port}{path}")).unwrap()
```
tgeoghegan, line 50: nit: should we use `http://localhost:{port}{path}` here just in case someone only has an IPv6 localhost? Though I notice that this was hard-coded to 127.0.0.1 in the existing code this refactors, so maybe we deliberately forced this to use IPv4 to avoid port collisions across the protocols...

integration_tests/src/janus.rs
```diff
@@ -56,3 +85,191 @@ impl<'a> Janus<'a> {
             .get_host_port_ipv4(Aggregator::INTERNAL_SERVING_PORT)
     }
 }
+
+/// Represents a running Janus test instance in this process.
+pub struct JanusInProcess {
```
tgeoghegan, line 90: There's loads of other tests (such as in `janus_aggregator::aggregator::http_handlers`) where we set up an in-process instance of Janus by calling `janus_aggregator::aggregator::http_handlers::aggregator_handler`. I wonder if we should adapt those to use this entry point, or otherwise unify them? Though I don't think that needs to happen in this PR.

tgeoghegan (approved):

integration_tests/src/lib.rs
divergentdave: Using `localhost` leads to problems when connecting to Docker container ports, because the same exposed container port may be published at different ephemeral ports on 127.0.0.1 and ::1. When you combine that with our HTTP client doing happy eyeballs on both addresses for `localhost`, that resulted in wires getting crossed between different concurrently running tests. Sticking with 127.0.0.1 here and `Container::get_host_port_ipv4()` elsewhere keeps things straight.

integration_tests/src/janus.rs
divergentdave: Using the handler directly in those tests means they can run without binding to a port, skip both sides of the HTTP protocol, and manipulate requests and responses in memory. (using `trillium-testing`) I think it makes sense to keep things as they are, because they each have a different area of interest.

Issue #2124, opened 2023-10-13
Cap the number of retries that a database tx can undergo


inahga:
As discussed in slack https://isrg.slack.com/archives/C0167LT4C73/p1697139388022519.

It is possible for a client to attempt to insert a task whose ID collides with an expired task. When this happens, an infinite loop of retries is triggered.

The trigger is illustrated with this example client https://github.com/divviup/janus/commit/538a244744c3b5ccec3b2a72cbc5afdd8b0dc860#diff-9cbf3f0a31c9bbe4a567f128a2f4fd96fb100fe794131ff92e42400779ba85f1.

Besides needing to fix this bug, we also identified several Janus improvements that help with troubleshooting and limiting the impact of this class of bug:
- [ ] Cap the number of retries that a tx can undergo, 
- [x] and add associated logging and metrics.
- [x] ~Add exponential backoff to retries~ we don't think this is necessary at this time as retries should be very rare, but it's something we can keep in mind.
- [x] ~https://github.com/divviup/janus/issues/2120, this bug was originally discovered because the Google crawler was scraping a Janus subscriber's site. It has deterministic RNG and thus was sending the same report ID on every site crawl~. Not going to implement this in janus, this is a front-end LB concern.
- [x] Add the ability to change log level at runtime https://docs.rs/tracing-subscriber/latest/tracing_subscriber/reload/index.html

tgeoghegan:
It seems like the only thing left to do in Janus is to cap the retries on DB transactions. Retitling to make it clear that we have addressed the bug that caused the infinite retries.

PR #2125, opened 2023-10-13
Fix infinite retry bug

inahga:
An infinite retry could be triggered if a report with the same ID as an expired report was submitted.

Supports https://github.com/divviup/janus/issues/2124.

I didn't add a separate method for `get_client_reports_including_expired_ones` because naming things is hard, and I think this is the only situation where we care about expired reports.

I altered the error message to provide more detail as to what's happening in this branch of code. The error here does matter because it is logged by `tracing`.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -1473,12 +1473,43 @@ impl<C: Clock> Transaction<'_, C> {
         // Slow path: no rows were affected, meaning a row with the new report ID already existed
         // and we hit the query's ON CONFLICT DO NOTHING clause. We need to check whether the new
         // report matches the existing one.
-        let existing_report = match self
-            .get_client_report(vdaf, new_report.task_id(), new_report.metadata().id())
+        let existing_report = {
+            // We intentionally don't use `get_client_report` because it omits expired reports. It
+            // is possible that we have conflicted with an expired report that hasn't been fully
+            // GC'd yet.
+            let stmt = self
+                .prepare_cached(
+                    "SELECT
+                        client_reports.client_timestamp,
+                        client_reports.extensions,
+                        client_reports.public_share,
+                        client_reports.leader_input_share,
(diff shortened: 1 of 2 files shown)


aggregator_core/src/datastore.rs
```diff
@@ -1497,11 +1528,13 @@ impl<C: Clock> Transaction<'_, C> {
                 // `None` (since all reads in the same transaction are from the same snapshot), so
                 // so it can't evaluate idempotency. All it can do is give up on this transaction
                 // and try again, by calling `retry` and returning an error; once it retries, it
-                // will be able to read the report written by the successful writer. (It doesn't
-                // matter what error we return here, as the transaction will be retried.)
+                // will be able to read the report written by the successful writer.
                 self.retry();
-                return Err(Error::MutationTargetAlreadyExists);
-            }
+                Error::DbState(
```
divergentdave, line 1533: It might be worth making a separate error variant for this sort of message, because `DbState` is currently used for unrecoverable corruption (i.e. wrong schema version, byte array of the wrong length, application-level invariant violated, etc.)

aggregator_core/src/datastore/tests.rs
```diff
@@ -7317,3 +7317,73 @@ async fn roundtrip_taskprov_peer_aggregator(ephemeral_datastore: EphemeralDatast
...
+            })
+        })
+        .await
+        .unwrap();
+
+    // Advance the clock well past the report expiry age.
+    clock.advance(&report_expiry_age.add(&report_expiry_age).unwrap());
+
+    // Insert a client report with the same ID, but a more current timestamp.
+    assert_matches!(
```
divergentdave, line 7366: nit: to minimize the amount of unformattable code, I recommend moving the transaction execution to a `let res = ...;` binding. Then the `assert_matches!()` macro call can be a one-liner.

divergentdave (approved):

Issue #2071, opened 2023-10-03
Janus 0.6.0 blockers


tgeoghegan:
Thus far we have been releasing the Janus 0.6 series as `0.6.0-prerelease` because we have been making breaking changes to Janus API and the SQL schema and don't yet want to commit to stability. This issue tracks the remaining changes we want to make before we can release Janus 0.6.0.

- [x] https://github.com/divviup/janus/issues/1858
- [x] https://github.com/divviup/janus/issues/230
- [x] https://github.com/divviup/janus/issues/1796
- [x] https://github.com/divviup/janus/issues/2073
- [x] https://github.com/divviup/janus/issues/1997 (@inahga)
- [x] https://github.com/divviup/janus/issues/1524
- [x] https://github.com/divviup/janus/issues/1521
- [x] https://github.com/divviup/janus/issues/1509
- [x] https://github.com/divviup/divviup-api/issues/542

tgeoghegan:
I tagged #230 as a 0.6.0 blocker because we wanted to take #2072, which affects the public API of Janus crates. The unresolved part of that issue is about doing fewer deserializations when working with HPKE keys. That's still interesting, but I don't think we should block 0.6.0 on it. Long story short: #230 stays open, but it's not a blocker anymore.

tgeoghegan:
All of this is done and we're ready to release Janus 0.6.0 (#2167).

PR #2151, opened 2023-10-19
release/0.5: adopt new `divviup-api` API

tgeoghegan:
The PRs related to https://github.com/divviup/divviup-api/issues/542 implement the new concept of "collector credential". This entails changes to the API for provisioning tasks via divviup-api that we use in e2e integration tests. On `release/0.5`, we don't use the `divviup-client` crate, so we update the client baked into Janus.

Another curiosity is that while `divviup-api` will mint a collector auth token in response to our request to create a collector auth token, that auth token won't be used and instead the one generated by the Janus aggregator API gets plumbed back to us in the task creation request.

Part of #2071


tgeoghegan:
I empirically verified that we can run the Janus integration tests against a `divviup-api` that includes https://github.com/divviup/divviup-api/pull/583. But will that work in the website? In particular, if a subscriber creates a task in a Janus 0.5.x aggregator via the website, will they be handed the collector auth token generated by the Janus aggregator API, or the collector auth token generated by divviup-api?

divergentdave (approved):

PR #2168, opened 2023-10-23
Add endpoint for configuring tracing filter at runtime

inahga:
Supports https://github.com/divviup/janus/issues/2124, and may be necessary for helping us debug issues without blowing up our cloud logging bill.

I chose to add this endpoint to the same `/healthz` handler we use for k8s status checks. I didn't think it fit on the aggregator API since we support deployments where the aggregator API is disabled, and the control plane will never be able to make full use of it since it only operates on a single replica of Janus.

### DIFF
--- a/aggregator/src/binary_utils.rs
+++ b/aggregator/src/binary_utils.rs
@@ -381,6 +381,9 @@ async fn get_traceconfigz(
     }))
 }
 
+/// Allows modifying the runtime tracing filter. Accepts a request with a body corresponding to
+/// [`TraceconfigzBody`]. If the `filter` field is empty, the filter will fallback to `error`.
+/// See [`EnvFilter::try_new`] for details.
 async fn put_traceconfigz(
     conn: &mut trillium::Conn,
     (State(trace_filter_handler), Json(req)): (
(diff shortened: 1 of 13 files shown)


inahga:
@jbr I could use your feedback here as well since I intend to add this to divviup-api.

docs/DEPLOYING.md
```diff
@@ -48,6 +48,22 @@ human-readable or structured JSON log format will be chosen automatically. This
 can be overridden with the `force_json_output` and `stackdriver_json_output`
 configuration parameters under `logging_config`.
 
+The environment variable can be overridden at runtime using the `/traceconfigz`
```
divergentdave, line 51: For clarity, since the last sentence changed topics
```suggestion
The `RUST_LOG` environment variable can be overridden at runtime using the `/traceconfigz`
```

aggregator/src/binary_utils.rs
```diff
@@ -347,6 +353,70 @@ async fn health_endpoint_server(address: SocketAddr) {
...
+) -> Result<Json<TraceconfigzBody>, Status> {
+    trace_filter_handler
+        .reload(EnvFilter::new(req.filter))
+        .map_err(|err| {
+            conn.set_body(format!("failed to update filter: {}", err));
+            // Note: reload will accept malformed filter directives, and fall back to the `error`
+            // filter. An error indicating malformation is never propagated up, so we can't give
+            // a proper HTTP status code. The error state will be revealed to the operator in the
+            // response body, however, since it will show the fallback log filter rather than their
+            // desired one.
```
divergentdave, line 399: We can improve this by using `EnvFilter::try_new()`, which only falls back to `error` on empty inputs, not invalid inputs. The errors returned by `reload()` itself indicate that an internal lock was poisoned by a panic or the reloadable subscriber was dropped, so a 500 status code sounds good for them.

divergentdave (approved):

jbr:
My feedback is subjective, but I find the mix of trillium::Conn::set_body and returning Status as Result::Err confusing. I think this would either work as a trillium_api handler that returns `Result<impl Handler, (String, Status)>` or use a Conn -> Conn handler that returns `conn.with_body(...).with_status(...)`

inahga:
Odd, the tuple strategy was my first attempt, but it wasn't working for me (always returning 200). Working for me now and the tests prove it.

divergentdave:
This is probably because `impl Handler for String` also sets the status to 200, so the order of the tuple matters. I think something similar came up in a previous discussion when working on the aggregator API.

inahga:
Indeed :point_up:. In fact now I can't reproduce my success when writing a unit test.

Fundamentally I don't think a tuple with `String` or `&'static str` can work, because they invoke `conn.ok()`, which does the following:
```rust
pub fn ok(self, body: impl Into<Body>) -> Self {
    self.with_status(200).with_body(body).halt()
}
```
So with `(String, Status)`, we `halt()` before we can handle the `Status`, and with `(Status, String)`, `String` truncates the status.

Let me see how the Conn -> Conn strategy looks.

inahga:
Ah, I don't think the Conn -> Conn approach is an improvement. We can't use the `api()` wrapper, which decodes the request body for us. We also lose the ability to ? to shortcut errors, and use the Status handler. To do these manually is a lot of code. (Let me know if I'm missing something--I don't think a Conn -> Conn function can be a Conn -> Result<Conn, _> function).

PR #2169, opened 2023-10-24
release Janus 0.5.26

tgeoghegan:


divergentdave (approved):
FYI 0.5.25 hasn't been used in a release yet, so we could use that instead. Six of one, half a dozen of the other.

tgeoghegan:
> FYI 0.5.25 hasn't been used in a release yet, so we could use that instead. Six of one, half a dozen of the other.

Oh, good point. Well, I'll just release 0.5.25 and save myself ~40 minutes of waiting for this PR's tests to pass.

PR #2173, opened 2023-10-25
Add robots.txt to discourage web scraping

inahga:
Supports https://github.com/divviup/janus/issues/2124.

We found that GoogleBot triggers report upload from one of our subscriber's TypeScript client. This
is somewhat benign, but still wasteful.

GoogleBot in particular has deterministic RNG, so the report ID it derives is always the same for
each scrape, contributing to log noise. Luckily, before following requests, it will attempt to
read robots.txt before proceeding.

### DIFF
--- a/aggregator/src/aggregator/http_handlers.rs
+++ b/aggregator/src/aggregator/http_handlers.rs
@@ -208,6 +208,8 @@ pub(crate) static AGGREGATION_JOB_ROUTE: &str =
 pub(crate) static COLLECTION_JOB_ROUTE: &str = "tasks/:task_id/collection_jobs/:collection_job_id";
 pub(crate) static AGGREGATE_SHARES_ROUTE: &str = "tasks/:task_id/aggregate_shares";
 
+static ROBOTS_TXT: &str = "User-agent: *\nDisallow: /\n";
+
 /// Constructs a Trillium handler for the aggregator.
 pub async fn aggregator_handler<C: Clock>(
     datastore: Arc<Datastore<C>>,
@@ -228,6 +230,12 @@ async fn aggregator_handler_with_aggregator<C: Clock>(
         metrics("janus_aggregator").with_route(|conn| conn.route().map(ToString::to_string)),
         Router::new()
             .without_options_handling()
+            // Discourage web scrapers from indexing Janus. DAP clients which submit reports on
+            // page load may be triggered by a web scraper.
+            .get(
+                "robots.txt",
+                |conn: Conn| async move { conn.ok(ROBOTS_TXT) },
+            )
             .get("hpke_config", instrumented(api(hpke_config::<C>)))
             .with_route(
                 trillium::Method::Options,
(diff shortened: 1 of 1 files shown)


aggregator/src/aggregator/http_handlers.rs
```diff
@@ -208,6 +208,8 @@ pub(crate) static AGGREGATION_JOB_ROUTE: &str =
 pub(crate) static COLLECTION_JOB_ROUTE: &str = "tasks/:task_id/collection_jobs/:collection_job_id";
 pub(crate) static AGGREGATE_SHARES_ROUTE: &str = "tasks/:task_id/aggregate_shares";
 
+static ROBOTS_TXT: &str = "User-agent: *\nDisallow: /\n";
```
branlwyd, line 211: I suppose the risk here is that we may disallow some bots that our eventual subscribers _want_ to have trigger report uploads. (concrete example: operational metrics like page load time are plausibly interesting even if the page load is triggered by a bot.) These bots would have to *not* use a deterministic RNG.

The choice we make here will affect all subscribers -- I can't think of a way to make this task-specific (or account-specific, etc).

WDYT about blocking Googlebot specifically, since we know it won't work, but allowing access from others?

tgeoghegan:
I don't know if Janus is the right place to do this. I can see that Divvi Up wants to do this, but is this something that every deployment of Janus wants?

Further, why do we need Janus to execute code just to serve up a static file? Isn't this a problem we could offload to a cloud load balancer or a CDN?

aggregator/src/aggregator/http_handlers.rs
inahga: > I can't think of a way to make this task-specific (or account-specific, etc).

We can change the contents based on the referrer, although this is clumsy at best. I don't think it's worth pursuing.

> WDYT about blocking Googlebot specifically, since we know it won't work, but allowing access from others?

That sounds fine to me.

inahga:
> I don't know if Janus is the right place to do this. I can see that Divvi Up wants to do this, but is this something that every deployment of Janus wants?

Unsure without subscriber feedback, my personal reaction is that there is no useful data that a scraper could contribute (nor is it necessary to preserve its privacy).

I'm fine to take a conservative approach and leave it out of Janus to leave it for future feedback. Or perhaps we can make it a configuration flag?

> Further, why do we need Janus to execute code just to serve up a static file? Isn't this a problem we could offload to a cloud load balancer or a CDN?

A CDN or LB will always need some upstream to point to to get its content, in the case of our deployment this is Janus. The CDN will cache Janus' response.

There are ways around this, but none I can see that don't involve introducing an additional service (e.g. nginx serving a static page). I don't think the few crawler requests that come down the pipe justify that kind of additional complexity.

In short: if we concur that this doesn't belong in Janus, I think we ought to just not worry about robots.txt for the time being.

divergentdave:
I get the impression that setting up something like nginx just for serving static files is fairly common in Kubernetes generally. We could do so for this case by adding a route for `/robots.txt` to our Ingress. The `ingress-gce` controller that we use sets up a deployment and service in the `kube-system` namespace that always returns 404s, so that it can be used as a default backend, if no other is provided.

branlwyd:
> my personal reaction is that there is no useful data that a scraper could contribute (nor is it necessary to preserve its privacy).

IMO (two separate points):
(a) there are more bots than just scrapers
(b) one of the value propositions is for Divvi Up to provide operational metrics, e.g. average time to complete some operation -- site owners may well want to capture this for bots as well. I think it would be unwise for us to block this use case.

In any case, I agree with Tim's point that this doesn't need to live in Janus itself.

inahga:
Sounds good. I'll leave this out of Janus and try for an nginx based approach in janus-ops.

PR #2177, opened 2023-10-26
Add `janus_db_migrator` container

tgeoghegan:
Janus now builds and ships a container image that bundles `sqlx` and the current Janus version's SQL migration scripts. This makes it easier for deployments to apply SQL migrations, as they no longer need to build a `sqlx` image themselves and (more importantly) figure out how to provide the appropriate set of migration scripts to whatever is applying the migrations.

While updating the READMEs, I also fixed a few Markdown hyperlinks to make some long lines less long.


branlwyd (approved):

inahga (approved):
:+1:

PR #2179, opened 2023-10-26
Add support for TLS postgres connections

### LINKED ISSUE #231: Support connecting to DB over TLS
Currently, we never use TLS when connecting to the backing DB. We should add configuration support for connecting to the DB over TLS.

divergentdave:
This adds support for TLS postgres connections, and closes #231. I used `tokio-postgres-rustls`, which is a straightforward glue crate connecting the two libraries, and about two pages long. This adds a new configuration option to provide a trust store as a PEM file. I added a test that connects to a database with TLS enabled, with sslmode=require added to the client's configuration. Getting the keys into the database container took some extra setup.


aggregator/src/config.rs
```diff
@@ -62,16 +62,21 @@ pub struct DbConfig {
...
+
     /// If false, the program will not check whether the database's current
     /// schema version is supported.
     #[serde(default = "DbConfig::default_check_schema_version")]
     pub check_schema_version: bool,
-    // TODO(#231): add option for connecting to database over TLS, if necessary
+
+    /// Path to a PEM file with root certificates to trust for TLS database connections.
+    #[serde(default)]
+    pub tls_trust_store_path: Option<String>,
```
tgeoghegan, line 79: nit: would `Option<PathBuf>` work?

tgeoghegan (approved):

inahga (commented):
Code looks good.

Q: why not use the tls-rustls feature of sqlx? https://docs.rs/sqlx/latest/sqlx/#tls-support. I haven't actually tried using it yet so LMK if there's something obviously wrong with it.

divergentdave:
Switching from tokio-postgres to sqlx would be a big change, and we'd lose the ability to pipeline queries, which we currently take advantage of. The sqlx-cli binary we distribute already has the native-tls feature flag enabled by default, but we could swap that to rustls for consistency.

aggregator/src/binary_utils.rs
```diff
@@ -572,4 +599,74 @@ mod tests {
...
+                "ssl_key_file=/etc/ssl/postgresql/127.0.0.1-key.pem".to_string(),
+            ]),
+        ))
+        .with_volume((volume.name(), "/etc/ssl/postgresql"));
+        let db_container = client.run(image);
+        const POSTGRES_DEFAULT_PORT: u16 = 5432;
+        let port = db_container.get_host_port_ipv4(POSTGRES_DEFAULT_PORT);
+
+        let db_config = DbConfig {
+            url: format!("postgres://postgres@127.0.0.1:{port}/postgres?sslmode=require")
```
inahga, line 661: I think we would want `sslmode=verify-full` to fully test that the trust store code is working https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION. Otherwise the client will ignore the validity of the root CA cert.

divergentdave: With this client, the only options available are `disable`, `prefer`, or `require`. See https://docs.rs/tokio-postgres/latest/tokio_postgres/config/struct.Config.html. I can confirm that the `require` mode is performing full certificate verification, because it returned errors when I was passing the wrong certificate into the trust store. Effectively, I think this parameter only controls policy around `StartupMessage` versus `SSLRequest` in the wire protocol, and `tokio-postgres` entirely delegates verification to the `MakeTlsConnect` implementation. The `MakeTlsConnect` implementations only get a hostname, and no configuration. Since we're using rustls, the controlling API is `DangerousClientConfig::set_certificate_verifier()`, so we're OK.

inahga (approved):

PR #2181, opened 2023-10-26
Update cache configuration for sqlx container

divergentdave:
This is a follow-up to #2177 to tweak the docker buildx cache configuration. Currently, the layers aren't being cached, and sqlx is rebuilt (with the release profile) on each commit. This PR changes the `cache-from` lines to read from a different cache key (distinct from that used for Janus) and adds a `cache-to` line that writes back layers under that new cache key.

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -180,11 +180,12 @@ target "janus_db_migrator" {
   }
   dockerfile = "Dockerfile.sqlx"
   cache-from = [
-    "type=gha,scope=main-janus",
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):

divergentdave:
Success!

```
#29 [janus_db_migrator builder 3/3] RUN cargo install sqlx-cli     --version 0.7.2     --no-default-features --features rustls,postgres
#29 CACHED
```

PR #2208, opened 2023-11-01
Use simpler body for trace config endpoint

inahga:
From brief discussion in https://github.com/divviup/divviup-api/pull/607#discussion_r1378180279, the trace config endpoint can be made simpler.

(From discussion in https://isrg.slack.com/archives/C0167LT4C73/p1698787638814279, the previous code is subtly wrong or otherwise cumbersome to properly use, because of how trillium intends to do content negotiation).


tgeoghegan:
Does this need a corresponding divviup-api change to be deployable?

tgeoghegan (approved):

PR #2218, opened 2023-11-02
Scrub report data once it is no longer required.

### LINKED ISSUE #1278: Opportunistic deletion of report shares
It seems likely that database storage capacity will be the chief cost driver. We could reduce storage usage if we deleted report shares and preparation messages/shares once a report has been successfully aggregated and its query lifetime has been exhausted. Past that point, we only need to keep metadata about the report for some time, for anti-replay enforcement, until the report's timestamp is old enough that we can forget about the report entirely. This would reduce the amount of time for which we store the largest blobs, in the common case of successful aggregation and collection, in exchange for some additional writes. We could accomplish this either by overwriting report share columns, prepare state columns, and prepare message columns with NULL, or by separating report contents and report metadata into separate tables (similarly for report aggregation) and deleting rows from the table that holds the contents.

branlwyd:
Specifically, we scrub report data once aggregation has been started, at the end of the first aggregation step. This will considerably reduce the amount of data we store, at the cost of additional writes to the datastore. We also change the semantics of report-upload idempotency; see below.

Implementing this required changing the semantics of report-upload idempotency. Specifically, we can no longer detect when an uploaded report matches an already-stored report if the already-stored report has been scrubbed, since the data required to perform this check has been deleted. The new semantics are to do idempotency-checking on (task_id, report_id) only; any task that sees a duplicate report ID will treat it as a duplicated upload & return OK. I think these semantics are acceptable: in real deployments, a repeated report ID will most commonly mean a client is retrying an upload that actually succeeded the first time, and returning OK is fine here. There are some scenarios (e.g. bad client RNG) where the old duplicate-detection semantics would be better than the new semantics, but I think these are marginal enough that we can ignore them. The new semantics would also help with implementing the object-store design, which also doesn't store reports longer than necessary to perform aggregation.

(We could maintain the existing duplicate-checking logic by e.g. keeping a hash of the report around rather than the full report. If folks think it's worthwhile to maintain the existing semantics, let's chat about it in review.)

The current logic around scrubbing reports works for VDAFs which permit only a single aggregation per report. We will need to expand it once we support VDAFs permitting more than one aggregation per report by scrubbing each report at the beginning of the final aggregation for that report.

Closes #1278.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -3327,28 +3327,23 @@ mod tests {
         assert_eq!(task.id(), got_report.task_id());
         assert_eq!(report.metadata(), got_report.metadata());
 
-        // Report uploads are idempotent
+        // Report uploads are idempotent.
         aggregator
             .handle_upload(task.id(), &report.get_encoded())
             .await
             .unwrap();
 
-        // Reports may not be mutated
+        // Even if the report is modified, it is still reported as a duplicate.
         let mutated_report = create_report_custom(
             &leader_task,
             clock.now(),
             *report.metadata().id(),
             leader_task.current_hpke_key(),
         );
-        let error = aggregator
+        aggregator
             .handle_upload(task.id(), &mutated_report.get_encoded())
             .await
-            .unwrap_err();
-        assert_matches!(error.as_ref(), Error::ReportRejected(task_id, report_id, timestamp) => {
-            assert_eq!(task.id(), task_id);
-            assert_eq!(mutated_report.metadata().id(), report_id);
-            assert_eq!(mutated_report.metadata().time(), timestamp);
-        });
+            .unwrap();
     }
 
     #[tokio::test]
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -3300,13 +3300,12 @@ mod tests {
     async fn upload() {
         install_test_trace_subscriber();
 
-        let (vdaf, aggregator, clock, task, datastore, _ephemeral_datastore) =
-            setup_upload_test(Config {
-                max_upload_batch_size: 1000,
-                max_upload_batch_write_delay: StdDuration::from_millis(500),
-                ..Default::default()
-            })
-            .await;
+        let (vdaf, aggregator, clock, task, ds, _ephemeral_datastore) = setup_upload_test(Config {
+            max_upload_batch_size: 1000,
+            max_upload_batch_write_delay: StdDuration::from_millis(500),
+            ..Default::default()
+        })
+        .await;
         let leader_task = task.leader_view().unwrap();
         let report = create_report(&leader_task, clock.now());
 
@@ -3315,17 +3314,17 @@ mod tests {
             .await
             .unwrap();
 
-        let got_report = datastore
+        let got_report = ds
             .run_unnamed_tx(|tx| {
-                let (vdaf, task_id, report_id) =
-                    (vdaf.clone(), *task.id(), *report.metadata().id());
+                let vdaf = vdaf.clone();
+                let task_id = *task.id();
+                let report_id = *report.metadata().id();
                 Box::pin(async move { tx.get_client_report(&vdaf, &task_id, &report_id).await })
             })
             .await
             .unwrap()
             .unwrap();
-        assert_eq!(task.id(), got_report.task_id());
-        assert_eq!(report.metadata(), got_report.metadata());
+        assert!(got_report.eq_report(&vdaf, leader_task.current_hpke_key(), &report));
 
         // Report uploads are idempotent.
         aggregator
@@ -3333,7 +3332,8 @@ mod tests {
             .await
             .unwrap();
 
-        // Even if the report is modified, it is still reported as a duplicate.
+        // Even if the report is modified, it is still reported as a duplicate. The original report
+        // is stored.
         let mutated_report = create_report_custom(
             &leader_task,
             clock.now(),
@@ -3344,6 +3344,19 @@ mod tests {
             .handle_upload(task.id(), &mutated_report.get_encoded())
             .await
             .unwrap();
+
+        // Verify that the original report, rather than the modified report, is stored.
+        let got_report = ds
+            .run_unnamed_tx(|tx| {
+                let vdaf = vdaf.clone();
+                let task_id = *task.id();
+                let report_id = *report.metadata().id();
+                Box::pin(async move { tx.get_client_report(&vdaf, &task_id, &report_id).await })
+            })
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(got_report.eq_report(&vdaf, leader_task.current_hpke_key(), &report));
     }
 
     #[tokio::test]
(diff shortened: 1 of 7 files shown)


divergentdave (approved):
The duplicate report checks sound good to me.

One thing to note about the changes here: `get_client_report()` will panic if called on a row that's had `scrub_client_report()` applied to it. This is only called from the aggregation job driver, so I think we will be okay because the aggregation job state is set in the same transaction that scrubs client reports. Advancing the report state keeps us from calling `get_client_report()` again, and REPEATABLE READ means we will only ever see both or neither of these changes.

aggregator_core/src/datastore.rs
```diff
@@ -1470,81 +1470,48 @@ impl<C: Clock> Transaction<'_, C> {
...
-                // writer's write, and fall into this codepath.
-                //
-                // The failing writer can't do anything about this problem while in its current
-                // transaction: further attempts to read the client report will continue to return
-                // `None` (since all reads in the same transaction are from the same snapshot), so
-                // so it can't evaluate idempotency. All it can do is give up on this transaction
-                // and try again, by calling `retry` and returning an error; once it retries, it
-                // will be able to read the report written by the successful writer.
-                self.retry();
-                Error::Concurrency(
```
inahga, line 1533: We can remove this error variant since this was the only usage of it.

aggregator_core/src/datastore.rs
```diff
@@ -1470,81 +1470,48 @@ impl<C: Clock> Transaction<'_, C> {
...
+                "UPDATE client_reports SET
+                extensions = NULL,
+                public_share = NULL,
+                leader_input_share = NULL,
+                helper_encrypted_input_share = NULL,
+                updated_at = $1,
+                updated_by = $2
+            FROM tasks
+            WHERE tasks.id = client_reports.task_id
+              AND tasks.task_id = $3 AND client_reports.report_id = $4",
```
inahga, line 1500: Does pg actually immediately reclaim the space gained by setting these NULL? Does it require a vacuum?

I will experiment/research, but curious if you know the answer off hand.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -435,6 +442,7 @@ impl AggregationJobDriver {
             aggregation_job,
             &stepped_aggregations,
             report_aggregations_to_write,
+            report_ids_to_scrub,
             resp.prepare_resps(),
         )
         .await
```
inahga, line 448: somewhat nit: Adding more functionality to `process_response_from_helper` makes me a bit uncomfortable, since it overloads that function's purpose and introduces an argument that is only necessary for `step_aggregation_job_aggregate_init`.

Can we instead scrub reports at the end of `step_aggregation_job_aggregate_init()`? Or was the intent of having it done in `process_response_from_helper` was to be concurrent with the other tx's?

inahga (commented):
This seems safe to deploy normally--any new aggregation jobs will scrub their reports, and client reports associated with old aggregation jobs will eventually get GC'd.

aggregator_core/src/datastore.rs
divergentdave: Since Postgres uses MVCC, this UPDATE results in new tuples being inserted. The existing tuple becomes a dead tuple once enough transactions have finished, and the space occupied by the dead tuple gets freed by autovacuum workers. (FWIW, DELETE'd tuples need help from autovacuum workers to complete their lifecycle as well, the only difference is that there's no new tuple inserted in the table)

The story gets more complicated if rows are big enough that they don't fit in a page, as report shares may get stored via the TOAST system. In that case, the row is represented by both a tuple in the main table, and one or more tuples in TOAST tables. Autovacuuming applies to the TOAST table as well.

branlwyd: David's explanation matches my understanding. This behavior is why I'm nervous about increasing our rate of writes -- it's hard to reason about how Postgres' autovacuum will handle this. (That is, we may be saving a lot of space, but incurring a bunch of extra work for vacuum.)

aggregator/src/aggregator/aggregation_job_driver.rs
branlwyd: We need to scrub reports at the end of the initial step, in the same transaction that updates the aggregation state. If we scrubbed reports in an earlier transaction, we could end up succeeding at scrubbing the reports, but ultimately failing to step the aggregation job -- this would leave the aggregation job stuck, as we would no longer have the reports required to complete initialization on the next attempt.

branlwyd:
> get_client_report() will panic if called on a row that's had scrub_client_report() applied to it.

I didn't realize this -- this is also true of report shares written by the Helper, since `put_report_share` leaves the same columns NULL'ed out. I think you're correct that our current application logic will never run afoul of this, but I changed it to return an error instead of panicking.

tgeoghegan (changes requested):

aggregator_core/src/datastore.rs
```diff
@@ -1112,20 +1112,28 @@ impl<C: Clock> Transaction<'_, C> {
     {
         let time = Time::from_naive_date_time(&row.get("client_timestamp"));
 
-        let encoded_extensions: Vec<u8> = row.get("extensions");
+        let encoded_extensions: Vec<u8> = row
+            .get::<_, Option<_>>("extensions")
+            .ok_or_else(|| Error::IncompleteReport)?;
```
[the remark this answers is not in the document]
branlwyd: But I want to panic if the value can't be decoded, as this represents programmer error: either "asking for a column I didn't actually read" or "trying to read into an inconsistent type". (The specific programmer error before this PR was "trying to decode a `NULL`able column into a `Vec<u8>` rather than an `Option<Vec<u8>>`".) Generally, `Row::get` should be used with an appropriate output type, as long as the output type is known.

aggregator/src/aggregator.rs
```diff
@@ -3327,28 +3327,23 @@ mod tests {
...
 
-        // Report uploads are idempotent
+        // Report uploads are idempotent.
         aggregator
             .handle_upload(task.id(), &report.get_encoded())
             .await
             .unwrap();
 
-        // Reports may not be mutated
+        // Even if the report is modified, it is still reported as a duplicate.
```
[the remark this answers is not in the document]
branlwyd: Good idea -- I did both.

aggregator_core/src/datastore.rs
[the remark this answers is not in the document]
tgeoghegan: Yeah, that makes sense. I was concerned about the case where there is a value in the `extensions` column but it's malformed, but you're correct that this case will panic and not get flattened into `IncompleteReport`. This is fine as-is.

aggregator_core/src/datastore/tests.rs
```diff
@@ -597,36 +597,34 @@ async fn roundtrip_report(ephemeral_datastore: EphemeralDatastore) {
         ),
     );
 
-    // Write a report twice to prove it is idempotent
```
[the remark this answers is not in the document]
branlwyd: This doesn't work anymore: the second attempt to write will always return `MutationTargetAlreadyExists` (which is handled & tested as part of the aggregator implementation).

tgeoghegan (approved):

PR #2220, opened 2023-11-03
collect: hide sensitive environment variable values

inahga:
Don't print the values of these variables when using `--help`, to avoid exposing secrets.


tools/tests/cmd/collect.trycmd
divergentdave, line 1: `tools/tests/cmd/collect_fpvec_bounded_l2.trycmd` should be updated as well

divergentdave (approved):

tools/tests/cmd/collect.trycmd
inahga: Ah nice catch, forgot to run with `cargo test --all-features`.

PR #2221, opened 2023-11-03
collect: Accept the newest divviup-api credentials file

inahga:
Since we've adopted https://github.com/divviup/divviup-api/pull/565/files#diff-5e36eb06fc0c6a6dab5d267db1ed4c554858a2a3419a200f21153b689dfcc837, the collector credentials yielded by `divviup collector-credential generate` is like so:

```json
{
  "id": "b17f6ef9-2696-4793-8af2-377b77668c59",
  "hpke_config": {
    "id": 241,
    "kem_id": "X25519HkdfSha256",
    "kdf_id": "HkdfSha256",
    "aead_id": "Aes128Gcm",
    "public_key": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
  },
  "created_at": "2023-11-03T20:28:39.796684Z",
  "deleted_at": null,
  "updated_at": "2023-11-03T20:28:39.796685Z",
  "name": "inahga",
  "token_hash": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
  "token": "aaaaaaaaaaaaaaaaaaaaaa"
}
{
  "aead": "AesGcm128",
  "id": 241,
  "kdf": "Sha256",
  "kem": "X25519HkdfSha256",
  "private_key": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
  "public_key": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
  "token": "aaaaaaaaaaaaaaaaaaaaaa"
}
```

The additional top level object chokes `collect.rs`, if we're supplying the verbatim output of `divviup`. IMO we shouldn't expect users to have to mangle the output of divviup, in realistic usage I think people will want to do `divviup collector-credential generate >credential.json`.

Change `collect.rs`'s JSON parsing logic so that it searches all top level objects for the one it wants.

Also change references of `divviup hpke-config` to `divviup collector-credential`, since that's what that subcommand is called now.

### DIFF
--- a/core/src/hpke.rs
+++ b/core/src/hpke.rs
@@ -261,7 +261,7 @@ impl HpkeKeypair {
     }
 }
 
-/// HPKE configuration compatible with the output of `divviup hpke-config generate`.
+/// HPKE configuration compatible with the output of `divviup collector-credential generate`.
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
 pub struct DivviUpHpkeConfig {
     id: HpkeConfigId,
--- a/tools/src/bin/collect.rs
+++ b/tools/src/bin/collect.rs
@@ -369,9 +369,7 @@ impl Options {
                 // Iterate over each one and return the first one that is valid.
                 match Deserializer::from_reader(reader)
                     .into_iter::<Value>()
-                    .collect::<Result<Vec<_>, _>>()?
-                    .into_iter()
-                    .find_map(|partial| DivviUpHpkeConfig::deserialize(partial).ok())
+                    .find_map(|partial| DivviUpHpkeConfig::deserialize(partial.ok()?).ok())
                 {
                     Some(divviup_hpke_config) => HpkeKeypair::try_from(divviup_hpke_config)
                         .context("could not convert HPKE config"),
--- a/tools/src/bin/collect.rs
+++ b/tools/src/bin/collect.rs
@@ -1,4 +1,4 @@
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine};
 use clap::{
     builder::{NonEmptyStringValueParser, StringValueParser, TypedValueParser},
@@ -23,6 +23,8 @@ use prio::{
     codec::Decode,
     vdaf::{self, prio3::Prio3},
 };
+use serde::Deserialize;
+use serde_json::{Deserializer, Value};
 use std::{fmt::Debug, fs::File, path::PathBuf};
 use tracing_log::LogTracer;
 use tracing_subscriber::{prelude::*, EnvFilter, Registry};
@@ -317,7 +319,7 @@ struct Options {
     #[derivative(Debug = "ignore")]
     hpke_private_key: Option<HpkePrivateKey>,
     /// Path to a JSON document containing the collector's HPKE configuration and private key, in
-    /// the format output by `divviup hpke-config generate`.
+    /// the format output by `divviup collector-credential generate`.
     #[clap(
         long,
         help_heading = "DAP Task Parameters",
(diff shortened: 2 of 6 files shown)


core/src/hpke.rs
```diff
@@ -261,7 +261,7 @@ impl HpkeKeypair {
     }
 }
 
-/// HPKE configuration compatible with the output of `divviup hpke-config generate`.
+/// HPKE configuration compatible with the output of `divviup collector-credential generate`.
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
 pub struct DivviUpHpkeConfig {
```
tgeoghegan, line 266: I'm pretty sure the answer is "no", but I just want to check: is there a crate shared across Janus and `divviup-api` where type definitions like this could live?

tools/src/bin/collect.rs
```diff
@@ -362,9 +364,19 @@ impl Options {
...
-                let divviup_hpke_config: DivviUpHpkeConfig =
-                    serde_json::from_reader(reader).context("could not parse HPKE config file")?;
-                HpkeKeypair::try_from(divviup_hpke_config).context("could not convert HPKE config")
+
+                // Divvi Up credential files come with multiple different top level JSON objects.
+                // Iterate over each one and return the first one that is valid.
+                match Deserializer::from_reader(reader)
+                    .into_iter::<Value>()
+                    .collect::<Result<Vec<_>, _>>()?
+                    .into_iter()
```
tgeoghegan, line 373: Are these two lines necessary? IIUC you get an `Iterator<Item = Value>` from the first call to `into_iter`, then collect it into `Vec<Value>`, then get another `Iterator<Item = Value>` from the second `into_iter`.

tgeoghegan (approved):
I think this change is fine (with a couple nits), but I'm concerned that putting Divvi Up specific concepts (like the credentials format or the UUID of a Divvi Up collector credential) into Janus is a layering violation. Maybe this stuff should be moved to a `divviup collect` subcommand?

core/src/hpke.rs
inahga: I think janus_core is the closest there is, although I don't think divviup-api relies on it for its definition of the HPKE config (seeing as divviup-api provides a `token` field).

This topic has also come up recently wrt sharing tracing/metrics subscribers. https://isrg.slack.com/archives/C0167LT4C73/p1698861551769459. I like the idea but I see it as a nice-to-have cleanup.

inahga:
> I'm concerned that putting Divvi Up specific concepts (like the credentials format or the UUID of a Divvi Up collector credential) into Janus is a layering violation.

I considered the same thing, I ended up viewing it from this lens: The format we're accepting is not _necessarily_ DU specific, even though we're doing with DU in mind. I see this as relaxing the file format to "somewhere inside this file is a structure we are interested in" rather than "this file must only consist of the structure we are interested in". This is also why I didn't bother structifying the entire DU file format.

Your point though is very relevant for https://github.com/divviup/janus/issues/2224 where I think there is a strong argument for it being a layering violation.

tools/src/bin/collect.rs
inahga: Hmm, I constructed this nasty iterator with Friday brain, let me take another stab at it.

inahga: From `StreamDeserializer`, you're actually getting an `Iterator<Item = Result<Value>>`, where the Result is the result of deserializing the next Value. It's tricky to code-golf this into a neat iterator, but I landed on:
```
                match Deserializer::from_reader(reader)
                    .into_iter::<Value>()
                    .find_map(|partial| DivviUpHpkeConfig::deserialize(partial.ok()?).ok())
```
This swallows an error if the JSON is malformed, but serde gives bad error messages anyway. The supplied file should also be tool-generated, so if there's something wrong with it there isn't any action for the user to take other than opening a bug report.

inahga:
@tgeoghegan checking in if you're comfortable with my answers to your questions, before I merge.

inahga:
From discussion in https://isrg.slack.com/archives/C0167LT4C73/p1699311235409159, a better approach would be to make the `divviup collector-credential generate` semantics more user friendly.

PR #2222, opened 2023-11-04
Fix repeated collection for time-interval.

branlwyd:
This was caused by the batch_aggregations soft-delete condition not respecting batches.batch_interval, which is used in time-interval tasks to allow coherent GC semantics even for batches including 0 reports (which time-interval must allow). This bug could specifically be triggered only if a collection request was made for a time interval including a minimal time interval with 0 reports, and then repeated. In that case, the empty batch aggregations would not be read due to the GC condition, which would then cause us to try writing them again, leading to a MutationTargetAlreadyExists error which would fail the collection job step until the collection job was abandoned.

This bug could only occur in time-interval: fixed-size tasks will always have more than 0 reports per batch.

I also augmented our integration tests to try collecting the same batch twice, to ensure we do not regress on this behavior.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -2963,7 +2963,7 @@ impl<C: Clock> Transaction<'_, C> {
                   AND batch_aggregations.batch_identifier = $2
                   AND batch_aggregations.aggregation_param = $3
                   AND ord = $4
-                  AND UPPER(batches.client_timestamp_interval) >= COALESCE($5::TIMESTAMP - tasks.report_expiry_age * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)",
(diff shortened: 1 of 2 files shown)


inahga (approved):

tgeoghegan (approved):
Should we backport this to release/0.5?

branlwyd:
tl;dr: I think we can skip backporting this.

The `release/0.5` branch supports a deployment used by a single subscriber (and we would route new subscribers to a newer version of the protocol, so this number is unlikely to increase). They currently have a single time-interval task (as well as a number of fixed-size tasks), which they have not uploaded to for at least the two weeks covered by our task metrics. Given that this bug only impacts time-interval, I think backporting is unnecessary. [we should keep this bug in mind in case our DAP-04 subscriber decides they want to use time-interval]

branlwyd:
(I'm gonna go ahead and merge this, but if folks have thoughts w.r.t. backporting, let's continue discussion here or in Slack.)

PR #2223, opened 2023-11-06
client: derive clone

inahga:
In case, e.g., a client needs to be used inside a `move` closure.


divergentdave (approved):

tgeoghegan (approved):
LGTM. I suspect the derived `Debug` impl could be improved but let's not do that here: #2225.

PR #2229, opened 2023-11-06
Don't use rustls if no trust store is provided

### LINKED ISSUE #2226: v0.6.2 requires setting database.tls_trust_store_path
Hi there, I was testing out https://github.com/divviup/janus/commit/f85c31cab60681267190e34e15c17a46195ed36e in v0.6.2 and I found that if I deployed without setting `database.tls_trust_store_path` janus would exit with the following error:
```
Starting up
Error: couldn't create database connection pool Caused by:
0: couldn't make connection to database
1: Error occurred while creating a new object: error performing TLS handshake: invalid peer certificate: UnknownIssuer
2: error performing TLS handshake: invalid peer certificate: UnknownIssuer
3: invalid peer certificate: UnknownIssuer
```
Looking at the code adding this feature, it will always connect via TLS. Without setting the `tls_trust_store_path` parameter it'll load an empty root CA store, which I think means no certificate will ever work?

divergentdave:
This closes #2226. #2179 introduced optional support for TLS in database connections. For database servers that don't support TLS, this resulted in no change to default behavior, but for database servers that do support TLS, our client would always try to use TLS, which fails without having a trust store. This PR fixes the default behavior without a trust store by conditionally using `tokio_postgres::NoTls` in that situation. This struct has a special implementation of a trait, with an override of a `#[doc(hidden)]` method that can't be overridden from third-party crates, that changes the client's default behavior to skip attempting TLS connections.

### DIFF
--- a/docs/DEPLOYING.md
+++ b/docs/DEPLOYING.md
@@ -29,6 +29,13 @@ Each component must be provided a connection string/URL. Note that the database
 password may be separately passed on the command line or through an environment
 variable, to override (or fill in) the password set in the configuration file.
 
+##### TLS
+
+By default, TLS is never used for database connections. Opportunistic TLS
+connections can be enabled by providing a file with PEM-format CA certificates
+to trust. Furthermore, TLS connections can be required by setting
+`sslmode=require` in the database connection string.
+
 #### Health Check
 
 Each binary starts an HTTP server to service health check requests from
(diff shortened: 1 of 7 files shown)


tgeoghegan (approved):

inahga (approved):
We ought to document TLS usage in `docs/DEPLOYING.md`, as I think if I were a newcomer I would find this behavior of `tls_trust_store_path` surprising. That is, we ought to explicitly mention that connections are made in plain unless `tls_trust_store_path` is present and loaded with the CA certificate of the server.

docs/DEPLOYING.md
inahga, line 1: Excellent, thanks!

branlwyd (approved):

PR #2230, opened 2023-11-06
collect: improve polling feedback and timing; re-export ExponentialBackoff

inahga:
A few things happening here, I can break this up if we want to evaluate them separately.

The main goal is to choose exponential backoff parameters that are better suited to my impatience. IMO at the CLI, more polling aggressiveness is justified since the tool is more likely being used interactively. The previous parameters yielded backoff timings of:
```
>>> [15*(1.2**t) for t in range(100) if 15*(1.2**t) < 300]
[15.0, 18.0, 21.599999999999998, 25.919999999999995, 31.104, 37.32479999999999, 44.78975999999999, 53.747711999999986, 64.49725439999997, 77.39670527999998, 92.87604633599996, 111.45125560319995, 133.74150672383993, 160.4898080686079, 192.58776968232948, 231.10532361879538, 277.3263883425545]
```

The new parameters yield: 
```
>>> [3*(1.2**t) for t in range(100) if 3*(1.2**t) < 300]
[3.0, 3.5999999999999996, 4.32, 5.183999999999999, 6.2208, 7.464959999999998, 8.957951999999999, 10.749542399999997, 12.899450879999996, 15.479341055999996, 18.57520926719999, 22.29025112063999, 26.748301344767988, 32.09796161372158, 38.5175539364659, 46.22106472375908, 55.46527766851089, 66.55833320221306, 79.86999984265569, 95.84399981118682, 115.01279977342418, 138.015359728109, 165.6184316737308, 198.74211800847692, 238.49054161017233, 286.1886499322068]
```

I also provided more `info` level logging to provide feedback to the user when polling. Finally, I reexport ExponentialBackoff as part of the collector API, so that users of the `janus_collector` library don't have to import `backoff` themselves.

### DIFF
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -51,7 +51,8 @@
 
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-use backoff::{backoff::Backoff, ExponentialBackoff};
+use backoff::backoff::Backoff;
+pub use backoff::ExponentialBackoff;
 use chrono::{DateTime, Duration, TimeZone, Utc};
 use derivative::Derivative;
 pub use janus_core::auth_tokens::AuthenticationToken;
@@ -83,6 +84,7 @@ use std::{
     time::{Duration as StdDuration, SystemTime},
 };
 use tokio::time::{sleep, Instant};
+use tracing::info;
 use url::Url;
 
 /// Errors that may occur when performing collections.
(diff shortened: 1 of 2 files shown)


collector/src/lib.rs
```diff
@@ -620,7 +622,14 @@ impl<V: vdaf::Collector> Collector<V> {
             // poll_once() already retries upon server and connection errors, so propagate any error
             // received from it and return immediately.
             let retry_after = match self.poll_once(job).await? {
-                PollResult::CollectionResult(aggregate_result) => return Ok(aggregate_result),
+                PollResult::CollectionResult(aggregate_result) => {
+                    info!(
```
branlwyd, line 626: IMO, we shouldn't log at `info` for happy-path operations in a library. Drop this to a lower level?

collector/src/lib.rs
```diff
@@ -659,6 +668,13 @@ impl<V: vdaf::Collector> Collector<V> {
             } else {
                 backoff_duration
             };
+
+            info!(
```
branlwyd, line 672: Drop this to a lower level too.

branlwyd (approved):
LGTM, just some basic nits. Making "interactive" collection more aggressive sounds good to me.

branlwyd:
(the reason I think we want to not log by default is `collector` is a library, not a CLI tool, and while it's used in our CLI tool it is also intended for use in a stateful collector implementation.)

PR #2231, opened 2023-11-07
collect: 'divviup hpke-config' -> 'divviup collector-credential'

inahga:
Update doccomments and help texts to reflect what this subcommand is now called.

### DIFF
--- a/core/src/hpke.rs
+++ b/core/src/hpke.rs
@@ -261,7 +261,7 @@ impl HpkeKeypair {
     }
 }
 
-/// HPKE configuration compatible with the output of `divviup hpke-config generate`.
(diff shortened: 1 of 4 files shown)


jbr:
A slight tweak to the framing of this: I think the api specification of collector secrets should live in the collector, which is in janus, and divviup-api should "save a janus-compatible collector secret" instead of janus accepting "the format that the divviup cli spits out"

branlwyd (approved):

inahga:
@jbr I think it's a good idea but I'd like to pursue it separately since it would be a more widespread change--this PR just makes sure our help texts are at least _correct_.

Issue #2219, opened 2023-11-03
collect.rs: Subcommands for creating and polling collection jobs


inahga:
Add subcommands to `collect.rs` that allow creation and polling of collection jobs. This lets users poll collection jobs on their own interval, rather than waiting for our prescribed ones. It also allows recovery from a failed run of the tool, since the collection job ID would no longer be opaque.

PR #2234, opened 2023-11-08
collect: add subcommands for creating and polling jobs

inahga:
Closes https://github.com/divviup/janus/issues/2219.

Adds subcommand support and two subcommands.

`new-job` creates a new collection job and returns its job ID for later polling.

`poll-job` polls the previously created collection job. Note that all parameters between the two subcommands must be identical to correctly decode the aggregate shares.

Without any subcommands, collect.rs behaves exactly as it does before.


inahga:
This implementation still needs considerable refactoring.

@simon-friedberger I could use your feedback on the semantics of these new subcommands, whether they match your expectations.

simon-friedberger:
LGTM!

tools/src/bin/collect.rs
```diff
@@ -498,6 +615,52 @@ async fn run_collection_generic<V: vdaf::Collector, Q: QueryTypeExt>(
 where
     V::AggregateResult: Debug,
 {
+    let collection = new_collector(options, vdaf, http_client)?
+        .start_collection(query, agg_param)
+        .await
+        .map_err(|err| Error::Anyhow(err.into()))?;
+    println!("Job ID: {}", collection.collection_job_id());
```
branlwyd, line 622: Could we provide machine-parseable output here? Either JSON as discussed elsewhere, or just a flag which causes this command to only output the collection job ID as it would be used as a later input, i.e. without the `Job ID: ` prefix.

I think this is important as these two new commands would be used together, most likely in a scripted way.

If you're planning on doing this as part of the eventual JSON-output-for-everything change, IMO we can wait 'til then.

tools/src/bin/collect.rs
```diff
@@ -273,15 +312,46 @@ struct QueryOptions {
...
+enum Subcommands {
+    /// Create a new collection job and poll it to completion
+    ///
+    /// This is the default action when no subcommand is provided
+    Run,
+    /// Initialize a new collection job
+    ///
+    /// Outputs the newly generated collection job ID to stdout
+    NewJob,
+    /// Poll an existing job once
```
branlwyd, line 325:
```suggestion
    /// Poll an existing collection job once
```

(nitpick)

branlwyd (changes requested):

tools/src/bin/collect.rs
```diff
@@ -273,15 +312,46 @@ struct QueryOptions {
     current_batch: bool,
 }
 
+#[derive(Debug, Derivative, PartialEq, Eq, Subcommand)]
+enum Subcommands {
+    /// Create a new collection job and poll it to completion
```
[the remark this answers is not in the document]
inahga: For the new code, it's for consistency with existing doccomments in collect.rs.

However, this style is inconsistent with the rest of Janus (spot checking janus_cli, and config.rs), so I will add punctuation to all help texts.

tools/src/bin/collect.rs
inahga: > If you're planning on doing this as part of the eventual JSON-output-for-everything change, IMO we can wait 'til then.

This was my plan, yes.

tools/src/bin/collect.rs
```diff
@@ -273,15 +312,46 @@ struct QueryOptions {
...
+    ///
+    /// This is the default action when no subcommand is provided
+    Run,
+    /// Initialize a new collection job
+    ///
+    /// Outputs the newly generated collection job ID to stdout
+    NewJob,
```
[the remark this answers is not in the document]
inahga: This required introduction of a new method for the collector client `Collector::start_collection_with_id()`, since the existing `start_collection()` randomly generates the ID for us.

I wonder if for the next round of breaking changes if we should replace `start_collection()` to have the same behavior as `start_collection_with_id()`.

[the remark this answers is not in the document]
inahga: There is also an argument to be had about having job deletion support, but I will leave that for a future PR to keep this one small.

tools/src/bin/collect.rs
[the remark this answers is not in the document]
divergentdave: I avoided periods at the end on structs used with Clap's derive macro because the periods got stripped off from the output of `--help`. At one point I put two sentences in a doc comment, and it looked funny with only one period in the middle. Experimenting with this some more now, it looks like Clap removes a trailing period if the doc comment is one line, but does nothing if there are multiple paragraphs. When passing `-h` instead of `--help`, then only the first paragraph is shown, and its trailing period is stripped.

Based on this, I think it would make sense to have the first line not include a period, try to keep the description brief so that it doesn't run on too long in the compressed `-h` format, and then use full sentences with periods below a blank line, so they don't appear in the short help, but appear as-is in long help.

tools/src/bin/collect.rs
```diff
@@ -389,115 +466,208 @@ async fn main() -> anyhow::Result<()> {
...
+    println!("Job ID: {}", collection.collection_job_id());
+    Ok(())
+}
+
+async fn run_poll_job<V: vdaf::Collector, Q: QueryTypeExt>(
+    options: Options,
+    vdaf: V,
+    http_client: reqwest::Client,
+    query: Query<Q>,
+    agg_param: V::AggregationParam,
```
divergentdave, line 638: We may as well standardize the arguments to the different run functions:
```suggestion
    agg_param: &V::AggregationParam,
```

tools/src/bin/collect.rs
```diff
@@ -1142,4 +1318,148 @@ mod tests {
...
+            &format!("--hpke-config={encoded_hpke_config}"),
+            &format!("--hpke-private-key={encoded_private_key}"),
+            "--vdaf",
+            "count",
+            "--batch-interval-start",
+            "1000000",
+            "--batch-interval-duration",
+            "1000",
+            "new-job",
+            &format!("{collection_job_id}"),
```
divergentdave, line 1396: If the random collection job ID starts with a hyphen when encoded, this test would flake. Since this is a positional argument, we can't use the --flag=value workaround, as above, but we can use `--` to indicate that all subsequent arguments are positional arguments, not flags. (I haven't tested this myself, but hopefully one of these works)
```suggestion
            "--",
            "new-job",
            &format!("{collection_job_id}"),
```

```suggestion
            "new-job",
            "--",
            &format!("{collection_job_id}"),
```

tools/src/bin/collect.rs
```diff
@@ -1142,4 +1318,148 @@ mod tests {
...
+            &format!("--hpke-config={encoded_hpke_config}"),
+            &format!("--hpke-private-key={encoded_private_key}"),
+            "--vdaf",
+            "count",
+            "--batch-interval-start",
+            "1000000",
+            "--batch-interval-duration",
+            "1000",
+            "poll-job",
+            &collection_job_id.to_string(),
```
divergentdave, line 1451: See above, we'll need to delimit this to avoid flakes.

tools/src/bin/collect.rs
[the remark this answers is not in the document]
inahga: That seems reasonable. I'll adopt that approach for collect.rs and check the rest of Janus for consistency in another PR.

tools/src/bin/collect.rs
inahga: I had it this way because of this confusing error:
```
error[E0308]: mismatched types
   --> tools/src/bin/collect.rs:646:20
    |
646 |         .poll_once(&collection_job)
    |          --------- ^^^^^^^^^^^^^^^ expected associated type, found `&<... as Vdaf>::AggregationParam`
    |          |
    |          arguments to this method are incorrect
    |
    = note: expected reference `&CollectionJob<<V as prio::vdaf::Vdaf>::AggregationParam, _>`
               found reference `&CollectionJob<&<V as prio::vdaf::Vdaf>::AggregationParam, Q>`
```

But with a fresh brain I see now I can just `agg_param.clone()` when I construct the `collection_job`.

tools/src/bin/collect.rs
inahga: Nice catch! `new-job -- collection_job_id` works.

tools/src/bin/collect.rs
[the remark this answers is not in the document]
branlwyd: I like the idea of `Collector`'s interface changing to always take a collection job ID with the next round of breaking changes -- it's easy enough to generate a random ID that I don't think this is much of a burden on consumers of the library.

branlwyd (approved):

Issue #2252, opened 2023-11-14
Poll for jobs on fixed interval


inahga:
Aggregation and collection job acquisition happens by polling the associated job table with an exponential backoff strategy (default 1-60s, 2x multiplier without jitter).

This introduces an artificial spin-up time for job acquisition when at low load. For instance, if a collection job hasn't been created in a while, a job might have to wait up to 60 seconds before it is acquired.

The impact is that this is an annoyance--interactive collection with e.g. `collect.rs` is slower.

I'll also note that we should introduce jitter, since there is a high likelihood of job drivers coming up at the same time in a kubernetes environment, i.e. during deployment or HPA scale-up.

There's a few ways around this
1. In deployed environments, set `min_job_discovery_delay = max_job_discovery_delay` and change Janus to introduce jitter.
2. Change Janus to poll on a fixed interval with jitter; replace `{min,max}_job_discovery_delay` with `job_discovery_delay` and `job_discovery_delay_randomization_factor`.
3. Make exponential backoff factor configurable, so we can set it to e.g. `1.05` to increase time before reaching the max.
4. Keep exponential backoff strategy but introduce [LISTEN/NOTIFY with psql to wake waiting job drivers](https://www.postgresql.org/docs/current/sql-notify.html).

I lean towards option 2. We have strict control over the number of job driver replicas that poll the DB, so I'm not worried about thundering herd. It would also result in some minor simplification of the job driver logic, and keep us from having to have vestigial `{min,max}_job_discovery_delay` config parameters around that we won't use in production.

This can be done as a non-breaking change--the new parameters can be introduced and the old ones will do nothing.

inahga:
The strategy we use for the `aggregator_job_creator` looks even simpler https://github.com/divviup/janus/blob/main/aggregator/src/aggregator/aggregation_job_creator.rs#L219, i.e. it has no randomization factor and instead selects a wait randomly from `[0, poll_time)`. This seems fine to me.

PR #2253, opened 2023-11-14
Do not use exponential backoff when acquiring jobs

inahga:
Replaces the exponential backoff strategy used in job drivers with the fixed interval with random jitter strategy used by the job creator.

The configuration parameters min_job_discovery_delay_secs and max_job_discovery_delay_secs are deprecated and replaced with the parameter job_discovery_interval_secs. The configuration changes are backwards compatible--if job_discovery_interval_secs is not present we use the value max_job_discovery_delay_secs in its place.

Closes https://github.com/divviup/janus/issues/2252

### DIFF
--- a/aggregator/src/binary_utils/job_driver.rs
+++ b/aggregator/src/binary_utils/job_driver.rs
@@ -8,6 +8,7 @@ use opentelemetry::{
     metrics::{Histogram, Meter, Unit},
     KeyValue,
 };
+use rand::{thread_rng, Rng};
 use std::{
     fmt::{Debug, Display},
     future::Future,
@@ -33,10 +34,8 @@ pub struct JobDriver<C: Clock, R, JobAcquirer, JobStepper> {
     stopper: Stopper,
 
     // Configuration values.
-    /// Minimum delay between datastore job discovery passes.
-    min_job_discovery_delay: Duration,
-    /// Maximum delay between datastore job discovery passes.
-    max_job_discovery_delay: Duration,
+    /// The maximum amount of time to wait between job acquisition attempts.
+    job_discovery_interval: Duration,
     /// How many jobs to step at the same time in this process.
     max_concurrent_job_workers: usize,
     /// Allowable clock skew between datastore and job driver, used when determining if a lease has
@@ -76,8 +75,7 @@ where
         runtime: R,
         meter: Meter,
         stopper: Stopper,
-        min_job_discovery_delay: Duration,
-        max_job_discovery_delay: Duration,
+        job_discovery_interval: Duration,
         max_concurrent_job_workers: usize,
         worker_lease_clock_skew_allowance: Duration,
         incomplete_job_acquirer: JobAcquirer,
@@ -90,8 +88,7 @@ where
             runtime,
             meter,
             stopper,
-            min_job_discovery_delay,
-            max_job_discovery_delay,
+            job_discovery_interval,
             max_concurrent_job_workers,
             worker_lease_clock_skew_allowance,
             incomplete_job_acquirer,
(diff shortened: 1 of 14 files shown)


tgeoghegan (approved):

inahga:
I could use a re-review. Upon closer inspection of the job creator code, I realize I misunderstood how it was working. It polls on a fixed interval, but the first iteration is delayed by some random duration.

My first stab at this was choosing a random duration in the interval [0, job_discovery_interval), which is not how the job creator works. The latest commit fixes the behavior to match that of the job creator and minor refactor to make it more closely mimic its structure. https://github.com/divviup/janus/blob/main/aggregator/src/aggregator/aggregation_job_creator.rs#L212

aggregator/src/config.rs
```diff
@@ -147,6 +147,38 @@ pub struct JobDriverConfig {
     pub maximum_attempts_before_failure: usize,
 }
 
+// TODO(#2252): This custom deserializer is for backwards-compatibility with Janus 0.6.3 and below.
+// Remove it for the next major version of Janus.
+impl<'de> Deserialize<'de> for JobDriverConfig {
```
tgeoghegan, line 152: This is fine, but I wonder if either [`serde(rename)`](https://serde.rs/field-attrs.html#rename) or [`serde(alias)`](https://serde.rs/field-attrs.html#alias) could do the trick with less code?

aggregator/src/binary_utils/job_driver.rs
```diff
@@ -158,14 +159,44 @@ where
             // (we'll pick up any additional jobs on the next iteration of this loop). We can't
             // overestimate since this task is the only place that permits are acquired.
```
tgeoghegan, line 160: nit: I think this should say "lease" instead of "permit". While you're in here, could you check whether we're being consistent about that term in this file?

aggregator/src/binary_utils/job_driver.rs
```diff
@@ -117,13 +114,17 @@ where
 
         // Set up state for the job driver run.
         let sem = Arc::new(Semaphore::new(self.max_concurrent_job_workers));
-        let mut job_discovery_delay = Duration::ZERO;
+
+        let mut next_run_instant = Instant::now();
+        if !self.job_discovery_interval.is_zero() {
+            next_run_instant += thread_rng().gen_range(Duration::ZERO..self.job_discovery_interval);
```
tgeoghegan, line 120: I think we should apply jitter in a tighter range. If I configure job discovery interval to be 60 seconds, I'd be surprised to learn that an interval of 1 ms is just as likely as 60 s. Maybe choose a random value in the range `[0.9 * discovery_interval, 1.1 * discovery_interval)`?

branlwyd: This isn't jitter, though: this is effectively selecting a uniformly random time in the the job discovery window for this task to attempt to acquire jobs. This will (probably) distribute job-acquiration attempts from different instances of this deployment uniformly across the job discovery interval.

Though maybe jitter would be the better solution? (IMO, I'm OK with this approach.)

branlwyd (approved):

aggregator/src/binary_utils/job_driver.rs
tgeoghegan: OK, I got confused because the PR description says "random jitter strategy used by the job creator". However I notice that the implementation here is what `aggregation_job_creator` already does. That's been working fine (AFAIK) so I'm comfortable using that strategy in the job drivers, too. I guess let's just make sure not to call it jitter.

inahga: @branlwyd is correct--sorry I didn't update the PR description when I realized the aggregator job creator's strategy is not to jitter but to do offset.

aggregator/src/config.rs
inahga: `#[serde(alias = "max_job_discovery_delay_secs")]` works great, but we then lose the ability to `warn!` when falling back to a deprecated config. IMO the warning is important as a heads up for the next minor release (or in our case, heads up that the environment is missing the updated configuration).

aggregator/src/binary_utils/job_driver.rs
inahga: I just changed it here. It's a bit tricky--in this context most occurrences of "permit" are referring to acquisition of SemaphorePermits.

PR #2271, opened 2023-11-16
Enable logging when running tests in CI

divergentdave:
This enables logging when running `cargo test`. This is particularly helpful for the in-process integration tests.


inahga (approved):

tgeoghegan (approved):
The change LGTM, and the test failure at least proves that we get logs. Does this need to be rebased on #2272?

PR #2273, opened 2023-11-16
Remove direct configuration of OTLP gRPC metadata

divergentdave:
This takes advantage of support in the latest version of `opentelemetry-otlp` for reading additional metadata/headers from [standard environment variables](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.25.0/specification/protocol/exporter.md#specifying-headers-via-environment-variables). Support for setting these is removed from our configuration files and from our command-line interface, and documentation is updated with an example of how to pass them in via the environment. This is mainly relevant when using Honeycomb.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,7 +26,6 @@ updates:
           - opentelemetry
           - opentelemetry-*
           - opentelemetry_*
-          - tonic
       trillium:
         patterns:
           - trillium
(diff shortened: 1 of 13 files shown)


tgeoghegan (approved):
I don't really understand the finer points of configuring Honeycomb, but I trust you, and the diff overwhelmingly deletes code, so I'm for it.

PR #2274, opened 2023-11-16
Replace most Row::try_get() calls with Row::get()

divergentdave:
This replaces most `try_get()` calls with `get()`, to follow our convention in this file. I left the ones of type `Json<T>` alone for now: in addition to mismatched column names and unsupported Postgres data types, those can also error/panic if the serde deserialization fails, which is a less clear signal of a source code-level mistake. We could leave them be, or do `row::get<_, serde_json::Value>("...")` and `serde_json::from_value(...).map_err(|e| Error::DbState(...))?`, at the expense of some allocations, in order to more tightly follow our conventions.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -525,8 +525,8 @@ impl<C: Clock> Transaction<'_, C> {
             .await?;
         let row = self.query_one(&stmt, &[]).await?;
 
-        let version = row.try_get("version")?;
-        let description = row.try_get("description")?;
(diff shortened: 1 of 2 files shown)


tgeoghegan (approved):
The convention in question should be documented in `/docs/DEVELOPMENT.md`, including the rationale for it.

divergentdave:
Done, see https://github.com/divviup/janus/pull/2218#discussion_r1382153960 for reference.

tgeoghegan (approved):

PR #2285, opened 2023-11-17
[experimental] Replace Postgres with CockroachDB for storage.

branlwyd:


tgeoghegan:
I think this experiment has concluded: we're not adopting CockroachDB at this time.

PR #2307, opened 2023-11-28
collect: allow specifying HPKE JSON config via environment variable

inahga:
Stacked on #2302.

Allows specifying the `HPKE_CONFIG_JSON_CONTENTS` environment variable, which is the contents of the file generated by `divviup collector-credential generate`. Suitable for integration with systems that don't have neat support for files.

Also removes needless use of `Derivative`--all our structs already use `Derivative` to hide their sensitive contents.

### DIFF
--- a/tools/src/bin/collect.rs
+++ b/tools/src/bin/collect.rs
@@ -5,7 +5,6 @@ use clap::{
     error::ErrorKind,
     ArgAction, Args, CommandFactory, FromArgMatches, Parser, ValueEnum,
 };
-use derivative::Derivative;
 #[cfg(feature = "fpvec_bounded_l2")]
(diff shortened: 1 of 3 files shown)


inahga:
@simon-friedberger how do the semantics of this new argument look?

simon-friedberger:
Seems good to me.
Nit: Maybe "hpke_config_json" and "hpke_config_json_file" are easier to understand than "hpke_config_json" and "hpke_config_json_contents". I was a bit confused by the latter.

inahga:
Agreed that's more clear, but they would be breaking changes so we'd want to save it for the next semver major release.

divergentdave (approved):

Issue #2293, opened 2023-11-21
Expose report error counts in app Task metrics


moztcampbell:
[Feature request]
For a given task, can the DivviUp task metics show a count of requests to the `/task/XXXX/reports` endpoint that were rejected and not included in report count. This would be primarily for rough detection of configuration issues and more.

At a minimum, including reports that result in malformed payload (typically due to protocol or task configuration mismatch).
Errors around wrong HTTP method, MIME type, etc could be included or excluded based on what is simplest to implement since these are less likely to be the issues we encounter.

jbr:
This is something we've discussed. I've transferred the issue to janus, which would need to track the statistics for display, as divviup-api does not see reports directly

inahga:
Looking at our current report upload handling logic, there are several cases that we return client success for:
1. Public share decoding failed
2. Report encrypted with outdated or unknown HPKE config
3. Report decryption failed
4. Leader input share decoding failed

1 and 4 could be grouped together as their causes are similar (VDAF disagreement, bugs in client-side code).

I suggest we create monotonic counters in the database per task. Roughly this schema:
```sql
CREATE TABLE task_upload_counters (
    task_id                BYTEA UNIQUE NOT NULL, -- 32-byte TaskID as defined by the DAP specification
    report_decode_failure  BIGINT,
    report_decrypt_failure BIGINT,
    report_unknown_key     BIGINT,
    report_success         BIGINT

    CONSTRAINT fk_task_id FOREIGN KEY(task_id) REFERENCES tasks(id) ON DELETE CASCADE
);
```

Rationale for separate table: Prevent UPDATEs from being too costly, as IIUC an UPDATE has similar cost to an INSERT, and the `tasks` table is large.

Then, we extend the ReportWriteBatcher to also batch writes to this table, depending on the error (or lack thereof) encountered by the report upload handler. We will incur an additional database write for situations where we originally short-circuit, but the queries should be low-cost and fit in with the existing query batching logic. We will also need to carry up these counters through the aggregator API, likely on the existing task metrics endpoint.

These counters are not subject to garbage collection.

We need to be concious of deadlocks when incrementing counters, see https://blog.pjam.me/posts/atomic-operations-in-sql/. When committing a batch, we should sort the counter increment statements by task ID.

The follow up work in divviup-api would be to expose these counters, and preferably provide help text as to the meaning of these counters (I can assist with the copywriting for this).

[...]

PR #2311, opened 2023-12-02
Add additional context to report upload rejection errors

inahga:
From discussion in Slack and somewhat alluded to in https://github.com/divviup/janus/issues/2293.

Client upload errors are opaque at the moment. They don't provide much detail as to went wrong, and some classes of errors don't return errors at all. This makes troubleshooting a misconfigured client cumbersome.

First, introduce usage of the `detail` field in HTTP problem documents as defined in [RFC 7807](https://www.rfc-editor.org/rfc/rfc7807.html#page-5). 

The plumbing is relatively straightforward. `aggregator::Error`'s variants can contain an additional `Reason` enum which is provided to `ProblemDetailsConnExt::with_problem_details()`. This pattern will be useful for taskprov as well, where opt-out rationale is opaque because of the lack of a `detail` field.

Then, expose all errors encountered in `handle_upload_generic` to the client, with appropriate reasons.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -3,7 +3,9 @@
 pub use crate::aggregator::error::Error;
 use crate::{
     aggregator::{
+        accumulator::Accumulator,
         aggregate_share::compute_aggregate_share,
+        error::{handle_ping_pong_error, ReportRejectedReason},
         error::{BatchMismatch, OptOutReason},
         query_type::{CollectableQueryType, UploadableQueryType},
         report_writer::{ReportWriteBatcher, WritableReport},
@@ -90,8 +92,6 @@ use std::{
 use tokio::{sync::Mutex, try_join};
 use tracing::{debug, info, trace_span, warn};
 use url::Url;
-
-use self::{accumulator::Accumulator, error::handle_ping_pong_error};
 
 pub mod accumulator;
 #[cfg(test)]
@@ -1410,6 +1410,7 @@ impl VdafOps {
                     *task.id(),
                     *report.metadata().id(),
                     *report.metadata().time(),
+                    ReportRejectedReason::TaskExpired,
                 )));
             }
         }
@@ -1426,6 +1427,7 @@ impl VdafOps {
                     *task.id(),
                     *report.metadata().id(),
                     *report.metadata().time(),
+                    ReportRejectedReason::TooOld,
                 )));
             }
         }
@@ -3166,7 +3168,7 @@ async fn send_request_to_helper<T: Encode>(
 
 #[cfg(test)]
 mod tests {
-    use crate::aggregator::{Aggregator, Config, Error};
+    use crate::aggregator::{error::ReportRejectedReason, Aggregator, Config, Error};
     use assert_matches::assert_matches;
     use futures::future::try_join_all;
     use janus_aggregator_core::{
@@ -3535,11 +3537,18 @@ mod tests {
             .unwrap();
 
         // Try to upload the report, verify that we get the expected error.
-        assert_matches!(aggregator.handle_upload(task.id(), &report.get_encoded()).await.unwrap_err().as_ref(), Error::ReportRejected(err_task_id, err_report_id, err_time) => {
-            assert_eq!(task.id(), err_task_id);
-            assert_eq!(report.metadata().id(), err_report_id);
-            assert_eq!(report.metadata().time(), err_time);
-        });
+        let error = aggregator
+            .handle_upload(task.id(), &report.get_encoded())
+            .await
+            .unwrap_err();
+        assert_matches!(
+            error.as_ref(),
+            Error::ReportRejected(err_task_id, err_report_id, err_time, ReportRejectedReason::IntervalAlreadyCollected) => {
+                assert_eq!(task.id(), err_task_id);
+                assert_eq!(report.metadata().id(), err_report_id);
+                assert_eq!(report.metadata().time(), err_time);
+            }
+        );
     }
 
     #[tokio::test]
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1387,6 +1387,16 @@ impl VdafOps {
         C: Clock,
         Q: UploadableQueryType,
     {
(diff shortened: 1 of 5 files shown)


aggregator/src/aggregator.rs
```diff
@@ -1502,18 +1506,24 @@ impl VdafOps {
...
+                    plaintext_input_share.payload(),
+                )?,
+            ))
+        });
 
-        let leader_input_share = match A::InputShare::get_decoded_with_param(
-            &(&vdaf, Role::Leader.index().unwrap()),
-            leader_plaintext_input_share.payload(),
-        ) {
+        let (extensions, leader_input_share) = match decoded_leader_input_share {
```
inahga, line 1526: Notice that I've chained these operations together. Previously, for the first decoding phase of turning a decrypted Vec into a `PlaintextInputShare`, we were simply wrapping the error in `aggregator::Error`. 

This seemed incorrect to me. Since it's an error decoding the input share, it is the same class of error as the later `InputShare` decoding operation, where we provide report rejection context and make note of it with an OTEL counter.

I've chained them so that they have the same behavior under error.

aggregator/src/aggregator.rs
```diff
@@ -1387,6 +1387,16 @@ impl VdafOps {
         C: Clock,
         Q: UploadableQueryType,
     {
+        // Shorthand function for generating an Error::ReportRejected with proper parameters.
+        let reject_report = |reason| {
+            Arc::new(Error::ReportRejected(
```
tgeoghegan, line 1392: Why does this need to be wrapped in an `Arc`?

aggregator/src/aggregator/error.rs
```diff
@@ -26,8 +26,8 @@ pub enum Error {
     Message(#[from] janus_messages::Error),
     /// Corresponds to `reportRejected`, §3.2
     #[error("task {0}: report {1} rejected: {2}")]
-    ReportRejected(TaskId, ReportId, Time),
-    /// Corresponds to `reportTooEarly`, §3.2. A report was rejected becuase the timestamp is too
+    ReportRejected(TaskId, ReportId, Time, ReportRejectedReason),
+    /// Corresponds to `reportTooEarly`, §3.2. A report was rejected because the timestamp is too
     /// far in the future, §4.3.2.
```
tgeoghegan, line 31: I'd be rather surprised if these section references were still valid, several DAP drafts later. I think we should delete them. We could replace them with links to the relevant text in the current DAP draft, because at least then it's clear which draft of the document is being referred to. n.b. I'm not proposing cleaning up all section references in Janus or even in this enum in this PR, but we can improve these lines since they have been touched anyway.

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -44,50 +44,58 @@ impl Handler for Error {
...
+                conn.with_problem_details(DapProblemType::InvalidMessage, None, None)
             }
+            Error::ReportRejected(task_id, _, _, reason) => conn.with_problem_details(
+                DapProblemType::ReportRejected,
+                Some(task_id),
+                Some(reason.detail()),
+            ),
             Error::InvalidMessage(task_id, _) => {
-                conn.with_problem_details(DapProblemType::InvalidMessage, task_id.as_ref())
+                conn.with_problem_details(DapProblemType::InvalidMessage, task_id.as_ref(), None)
```
tgeoghegan, line 55: nit: I don't love having the majority of these callsites take `None`. Maybe we could have a problem details builder where you would optionally chain a call to `.with_detail(&str)`? Or you could even specialize it further for `ReportRejectedReason`.

aggregator/src/aggregator.rs
inahga: In the context of this code it's to match the existing return type.

For the overall context, it relates to this change https://github.com/divviup/janus/commit/a24d5cbcc075321352b2603dd4b983ab39c2c121, in particular here https://github.com/divviup/janus/blob/main/aggregator/src/aggregator/report_writer.rs#L149. It's to work around the fact that `aggregator::Error` is not `Clone`. We need multiple "copies" of it to fan out transaction errors to a bunch of listening channels.

However I think this can possibly be simplified. I will play around with it and open a PR if we can drop the Arc.

tgeoghegan: Ah, OK. Sometimes I wish that Rust structs were `Clone` by default, but then that's probably a bad idea. Anyway, if we can improve this then I'm all for it, but it's probably not worth too much effort.

aggregator/src/aggregator/http_handlers.rs
inahga: WDYT about https://github.com/divviup/janus/pull/2311/commits/1b6d22e2398f86d40dd94cb707de9898ae1d5f96? Not much of a LOC savings but it is nice having named arguments.

tgeoghegan: Much better! I like to see a bunch of `None` arguments disappear.

aggregator/src/aggregator.rs
branlwyd: I was pretty disappointed at having to use an `Arc` here, too -- it added some complexity to the types we work throughout the aggregator implementation.

IIRC, the reason I couldn't just make `aggregator::Error` `Clone` is that some of the included types weren't `Clone`, and they came from other crates so we could not easily make them so. If we want to simplify here, we could try making all included types `Clone`. Alternatively, we could follow the suggestion from the commit Ameer links to & make `aggregator::Error` a wrapper around an `Arc<InnerError>`, where the `InnerError` is equivalent to what `aggregator::Error` is today -- that would let us clone the `aggregator::Error` type without needing the internal types to be `Clone`, but will complicate logic where we discriminate the error type.

aggregator/src/aggregator/error.rs
inahga: For this feedback https://github.com/divviup/janus/pull/2311/commits/c46141f1846cb8ac44158cba9a3ef9579e76fde8, I cleaned up some comments to use updated URLs and drop section references.

branlwyd (approved):

divergentdave (approved):

PR #2324, opened 2023-12-05
DivviUpHpkeConfig -> CollectorCredential

inahga:
From comment in https://github.com/divviup/janus/pull/2231#issuecomment-1799278456.

Generally we'd like to avoid Divvi Up specific entities inside Janus. Deprecate the `DivviUpHpkeConfig` struct and introduce  `CollectorCredential`. Going forward, this should be the authoritative definition of `CollectorCredential` (i.e. we should probably make `divviup-api` use this struct).

Move it to the `collector` crate. I believe it's going to always be a collector-only struct so it best fits there.

Also add the `token` field, which is now included by collector credentials generated by `divviup collector-credential generate`.


collector/src/credential.rs
```diff
@@ -0,0 +1,105 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PrivateCollectorCredential {
+    id: HpkeConfigId,
+    kem: Kem,
+    kdf: Kdf,
+    aead: Aead,
+    public_key: HpkePublicKey,
+    private_key: HpkePrivateKey,
+    // TODO(inahga): Turn this non-optional for the next breaking change cycle.
+    token: Option<BearerToken>,
```
inahga, line 23: Unsure if this should be Option or not.

This comes from an iteration of `divviup` where it output credentials without an associated token. I don't think that made it to a release and therefore I'm not sure if any such credentials are out in the wild. In the strictest of terms it should be Option for backwards-compatibility, but unsure if being too strict here is counterproductive.

tools/src/bin/collect.rs
inahga, line 1: I have collect.rs cleanups in another PR.

collector/src/credential.rs
branlwyd: IMO: even if we didn't generate such credentials, I suppose we can't be sure that no one did. I think we should make the conservative choice to keep this optional for now, and write up an issue tracking our desire to make this non-optional with the next time we make breaking changes.

collector/src/credential.rs
```diff
@@ -0,0 +1,105 @@
...
+/// shares.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PrivateCollectorCredential {
+    id: HpkeConfigId,
+    kem: Kem,
+    kdf: Kdf,
+    aead: Aead,
+    public_key: HpkePublicKey,
+    private_key: HpkePrivateKey,
+    // TODO(inahga): Turn this non-optional for the next breaking change cycle.
```
branlwyd, line 22: process nit: for any TODOs, we generally reference a GH issue rather than a specific person or anything else. (there may still be a few errant TODOs floating around the codebase, but let's avoid introducing any new ones)

inahga: Yep, I'll be replacing it with an issue number before merging. I didn't do it ahead of time pending the discussion of whether it was necessary or not.

Issue #2224, opened 2023-11-06
collect.rs: Use bearer token in divviup credentials


inahga:
The Divvi Up API provides the bearer token used for collection along with the HPKE config, e.g.
```
{
  "aead": "AesGcm128",
  "id": 241,
  "kdf": "Sha256",
  "kem": "X25519HkdfSha256",
  "private_key": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
  "public_key": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
  "token": "aaaaaaaaaaaaaaaaaaaaaa"
}
```

collect.rs should use use token automatically when using DU credentials, so the user doesn't have to copy out the bearer token into `AUTHORIZATION_BEARER_TOKEN`.

PR #2330, opened 2023-12-06
collect: use authentication token provided in collector credential

inahga:
Closes https://github.com/divviup/janus/issues/2224.

At long last, we only need one credential for this CLI to work:
```
$ collect --task-id rU9XlLAicxOIikqRzwtx8y0hLEuDnRNoQ7ODLe6uAaY --leader https://staging-dap-07.api.divviup.org/ --vdaf count --current-batch --collector-credential-file ~/Projects/credentials.json
```

### DIFF
--- a/tools/src/bin/collect.rs
+++ b/tools/src/bin/collect.rs
@@ -250,12 +250,10 @@ fn private_collector_credential_parser(
 }
 
 #[derive(Debug, Args, PartialEq, Eq)]
-#[group(required = true)]
 struct AuthenticationOptions {
     /// Authentication token for the DAP-Auth-Token HTTP header
     #[clap(
         long,
-        required = false,
         value_parser = StringValueParser::new().try_map(AuthenticationToken::new_dap_auth_token_from_string),
         env,
         hide_env_values = true,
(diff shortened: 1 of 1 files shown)


tools/src/bin/collect.rs
```diff
@@ -450,31 +447,78 @@ struct Options {
...
-            &self.hpke_config.hpke_private_key,
-            &self.hpke_config.collector_credential_file,
             &self.hpke_config.collector_credential,
+            &self.hpke_config.collector_credential_file,
         ) {
-            (Some(config), Some(private), _, _) => {
-                Ok(HpkeKeypair::new(config.clone(), private.clone()))
-            }
-            (None, None, Some(collector_credential_file), _) => {
+            (Some(collector_credential), _) => Ok(Some(collector_credential.clone())),
```
branlwyd, line 455: nit: specifying a collector credential directly is mutually-exclusive with specifying a collector credential file (enforced by Clap config); this code might read a bit easier if the patterns were written such that they match only if exactly one of the given options is specified, with a default match arm panicking with some useful error message.

Same issue with the match block in the `authentication_token` function below. While touching things, consider updating the `unreachable!` arm in `hpke_keypair` to also panic with a useful message.

branlwyd (approved):

PR #2331, opened 2023-12-07
Fix global mismatch with trillium-opentelemetry

divergentdave:
This PR upgrades trillium-opentelemetry and enlists the type system in making sure that it is using the correct MeterProvider, by passing a Meter directly. We currently are depending on two versions of opentelemetry transitively, which results in duplicate globals, and ultimately results in lost metrics. I also made use of the new `.with_error_type()` feature to use our error codes in labels on the built-in histogram metrics, in addition to our custom counter metric elsewhere.

### DIFF
--- a/aggregator_api/src/lib.rs
+++ b/aggregator_api/src/lib.rs
@@ -13,7 +13,7 @@ use janus_core::{auth_tokens::AuthenticationToken, hpke, http::extract_bearer_to
 use janus_messages::{HpkeConfigId, RoleParseError, TaskId};
 use opentelemetry::metrics::Meter;
 use routes::*;
-use std::{str::FromStr, sync::Arc};
+use std::{borrow::Cow, str::FromStr, sync::Arc};
 use tracing::error;
 use trillium::{
     Conn, Handler,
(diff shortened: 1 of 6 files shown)


aggregator_api/src/lib.rs
```diff
@@ -67,13 +68,20 @@ impl Handler for ReplaceMimeTypes {
...
+) -> impl Handler {
     (
         // State used by endpoint handlers.
         State(ds),
         State(Arc::new(cfg)),
         // Metrics.
-        metrics("janus_aggregator").with_route(|conn| conn.route().map(ToString::to_string)),
+        metrics(meter).with_route(|conn| {
+            conn.route()
+                .map(|route_spec| Cow::Owned(route_spec.to_string()))
```
jbr, line 83: is there an appropriate error type for this `metrics` invocation?

divergentdave: Not currently. As a future enhancement, we could make this module's Error type store some state, as we do in the aggregator module currently. We would need to define the set of error label values as well.

branlwyd (approved):

PR #2334, opened 2023-12-07
Integration test for rate limits

tgeoghegan:
Adds tests to exercise Divvi Up's rate limiting to the existing `in_cluster` test module. The test setup reads in rate limit configuration from a file whose path is specified via environment variable, akin to the existing `JANUS_E2E_*` variables.

These tests don't actually exercise any part of Janus itself, but they rely on the test support for forwarding ports and otherwise working with a pair of Januses in a Kubernetes cluster, so this is a convenient place to define them.

### DIFF
--- a/integration_tests/tests/integration/in_cluster.rs
+++ b/integration_tests/tests/integration/in_cluster.rs
@@ -333,3 +333,217 @@ async fn in_cluster_fixed_size() {
     )
     .await;
 }
+
+mod rate_limits {
+    #![cfg(feature = "in-cluster-rate-limits")]
+
+    use super::InClusterJanusPair;
+    use http::Method;
+    use janus_aggregator_core::task::QueryType;
+    use janus_core::{test_util::install_test_trace_subscriber, vdaf::VdafInstance};
+    use janus_messages::{AggregationJobId, CollectionJobId, TaskId};
+    use rand::random;
+    use reqwest::StatusCode;
+    use serde::Deserialize;
+    use std::{env, fs::File};
+    use url::Url;
+
+    /// Configuration for the rate limit test. We need to know the QPS and the window over which
+    /// it is enforced so that we can send the appropriate number of requests. We load this config
+    /// from a file so that an integration test harness that knows what rate limits have been
+    /// deployed to the cluster we test against can set the values.
+    #[derive(Deserialize)]
+    struct TestConfig {
+        window: u64,
+        upload_qps: u64,
+        aggregation_job_qps: u64,
+        collection_job_qps: u64,
+    }
+
+    impl TestConfig {
+        fn load() -> Self {
+            serde_json::from_reader(
+                File::open(
+                    env::var("JANUS_E2E_RATE_LIMIT_TEST_CONFIG")
+                        .unwrap()
+                        .as_str(),
+                )
+                .unwrap(),
+            )
+            .unwrap()
+        }
+    }
+
+    async fn run_rate_limit_test(
+        request_url_maker: &dyn Fn(&Url, &Url, &TaskId, &TaskId) -> (Url, Url),
+        rate_limit_picker: &dyn Fn(&TestConfig) -> u64,
+        method: Method,
+    ) {
+        install_test_trace_subscriber();
+        let test_config = TestConfig::load();
+
+        let janus_pair =
+            InClusterJanusPair::new(VdafInstance::Prio3Count, QueryType::TimeInterval).await;
+
+        let (leader_url, helper_url) = janus_pair
+            .task_parameters
+            .endpoint_fragments
+            .endpoints_for_host_client(janus_pair.leader.port(), janus_pair.helper.port());
+        let other_task_id = random();
+
+        // Send requests to two different tasks, to prove that rate limits are per-task ID. It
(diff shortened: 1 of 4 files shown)


integration_tests/tests/integration/in_cluster.rs
```diff
@@ -333,3 +333,217 @@ async fn in_cluster_fixed_size() {
     )
     .await;
 }
+
+mod rate_limits {
+    #![cfg(feature = "in-cluster-rate-limits")]
```
tgeoghegan, line 338: This test module is gated on a non-default feature so that downstream environments that don't have rate limits set up can still run the other integration tests.

integration_tests/tests/integration/in_cluster.rs
```diff
@@ -333,3 +333,244 @@ async fn in_cluster_fixed_size() {
...
+        let rate_limit = rate_limit_picker(&test_config);
+        let rate_limit_excess = 0.1;
+        let total_requests_count =
+            ((1.0 + rate_limit_excess) * (rate_limit * test_config.window) as f64) as u64;
+
+        let mut acceptable_status_count = 0;
+        let mut too_many_requests_count = 0;
+
+        // Send ten requests at a time, because otherwise the kubectl port-forward gets overwhelmed
+        for _ in 0..(total_requests_count / 10) {
```
branlwyd, line 418: This block works, but it's kind of awkward: we generate groups of 10 requests at a time, execute them concurrently, then move on to the next group of 10. Other than running an incorrect number of requests if the `total_request_count` is not a multiple of 10, this also unnecessarily splits processing of each request into two pieces & otherwise requires slightly more complicated code. (It also reduces throughput somewhat by introducing more waiting on requests than necessary, but that is probably extremely marginal given that this is a test.)

I suggest instead introducing a [Semaphore](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html) with 10 permits. Have each spawned async task acquire a permit before making its request, and release the permit once the response is received. Spawn all tasks at once, and have each task be responsible for updating `acceptable_status_count` & `too_many_requests_count` (either by updating them atomically, hiding access behind a mutex, or just by returning a value from each async task indicating which variable to update).

(An alternative to using a semaphore would be to start 10 worker tasks, then have them execute requests until enough requests have been executed, synchronizing on a counter or perhaps a queue. This would scale a bit better than one-task-per-request but the additional complexity probably isn't justified in a test.)

integration_tests/tests/integration/in_cluster.rs
```diff
@@ -333,3 +333,244 @@ async fn in_cluster_fixed_size() {
...
+        second_task_id: &'a TaskId,
+    }
+
+    async fn run_rate_limit_test(
+        request_url_maker: &dyn Fn(&RequestUrlParameters) -> (Url, Url),
+        rate_limit_picker: &dyn Fn(&TestConfig) -> u64,
+        method: Method,
+    ) {
+        install_test_trace_subscriber();
+        let test_config = TestConfig::load();
```
branlwyd, line 390: nit: could we cache the result of loading the test config? Since this is a test & errors lead to a panic, we could use a static `Once` in the `TestConfig::load` method to load the test config, with the `Once` roughly wrapping the code that currently exists in `TestConfig::load`.

(I'm slightly curious about why we need to read this information from a file -- I suppose we can't hardcode it in the tests because ultimately the rate-limit config is determined outside of this code. Is that about correct?)

branlwyd (approved):

integration_tests/tests/integration/in_cluster.rs
tgeoghegan: Check out https://github.com/divviup/janus-ops/pull/1205 to see how Terraform will drop the JSON test config file that these tests then load. Since the rate limits are actually governed by the .tfvars files in janus-ops, I don't want to duplicate those values here, because then we'd have to make two changes across two projects every time we tweak the rate limits in Kind deployments.

This test is defined here in Janus, but as I note in the PR description, this isn't actually about Janus but rather exercising the Caddy rate limiting reverse proxy we set up over in janus-ops. It's just that this crate, since it already knows how to deal with a pair of Janus aggregators in a Kind cluster, is a convenient place to do that. I think maybe we could kick all of the `integration_tests::in_cluster` module out of the Janus repository, but I'm not sure I want to deal with that now: we already had a bunch of Kubernetes-specific goop here that isn't strictly essential to running or testing Janus, and instead reflects how Divvi Up happens to manage it.

branlwyd: Agreed that we can wait to move integration tests out of the `janus` repo -- arguably we should have put them in `janus_ops` (or their own repository) to start, but at this point we have integration tests already defined in `janus`, so this is the right place to put this code unless/until we decide to move all integration tests.

integration_tests/tests/integration/in_cluster.rs
tgeoghegan: This is much better with a semaphore, thank you for that suggestion. I considered updating `acceptable_status_count` etc. the way you suggested, but we need the test to block until all the spawned tasks are done anyway, meaning we'd have to `await` all the handles to get at least a `()` value, so getting a `(HeaderValue, HttpStatus)` as we currently do doesn't seem any worse.

PR #2352, opened 2023-12-12
dependabot-automerge: only attempt approval of unapproved PRs.

branlwyd:


divergentdave:
Hmm, I'm concerned this won't work. Here are the docs pages I'm looking at:

- https://docs.github.com/en/actions/learn-github-actions/contexts#github-context
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request
- https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=opened#pull_request

This runs on `pull_request` events, so `github.event` will contain `pull_request` webhook payloads, and by default we'll get payloads for `opened`, `synchronize`, and `reopened` events. These do not contain a `review` key.

Maybe instead of adding to the conditional here, we could instead set the `types` argument on the trigger, to only run this workflow on `opened` events.

branlwyd:
Thanks, I think you are right. I am not a fan of GH actions...

divergentdave (approved):

PR #2353, opened 2023-12-13
db: remove untrue comment

inahga:
This comment is no longer true, see https://github.com/divviup/janus/blob/main/aggregator/src/aggregator.rs#L2069.


divergentdave:
Note that this will make sqlx unhappy, since we're changing an existing migration. After cutting a release that includes this, any future 0.6 deploys will fail until we manually intervene and rewrite the _sqlx_migrations table. If we want to take this change, we'll need a plan for that.

inahga:
Ah, you're right, I forgot sqlx checksums the verbatim schema file, including comments. I'll close this PR and open an issue for us to address for the next round of breaking changes.

divergentdave:
We can take this change now that we're making breaking changes for 0.7.

PR #2355, opened 2023-12-13
Fix idempotency bug in `handle_aggregate_init_generic`

inahga:
Calls to the helper's `/aggregate` endpoint were not actually idempotent.

This is because we invoked `accumulator.flush_to_datastore(tx, &vdaf)`, before checking whether the request was replayed. This function makes changes, and we would `Ok` on replayed requests, causing the changes to be committed. 

The overall effect was repeated `/aggregate` requests would count again against the underlying batch aggregations. This would render the batch uncollectable, since the checksums between helper and leader wouldn't match.

This was particularly noticeable when the helper was under heavy load. If the upstream LB times out the request, the request isn't cancelled and the changes are committed. Since the leader sees `502 Bad Gateway`, it retries the request.

`flush_to_datastore()` was at that spot because we were mostly interested in the reports that couldn't be applied to a batch because the batch has been collected. In this PR, introduce a function `unwritable_reports()` that yields those reports without the side effect of writing the batch aggregation.

I will backport this change to 0.5.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1953,7 +1953,7 @@ impl VdafOps {
                 let accumulator = Arc::clone(&accumulator);
 
                 Box::pin(async move {
-                    let unwritable_reports = accumulator.flush_to_datastore(tx, &vdaf).await?;
+                    let unwritable_reports = accumulator.unwritable_reports(tx, &vdaf).await?;
 
                     for report_share_data in &mut report_share_data {
                         // Verify that we haven't seen this report ID and aggregation parameter
@@ -2034,7 +2034,8 @@ impl VdafOps {
 
                     if !replayed_request {
                         // Write report shares, aggregations, and batches.
-                        (report_share_data, _) = try_join!(
+                        let (resulting_unwriteable_reports, resulting_report_share_data, _) = try_join!(
+                            accumulator.flush_to_datastore(tx, &vdaf),
                             try_join_all(report_share_data.into_iter().map(|mut rsd| {
                                 let task = Arc::clone(&task);
                                 async move {
@@ -2088,6 +2089,12 @@ impl VdafOps {
                                 Ok(())
                             }}))
                         )?;
+                        report_share_data = resulting_report_share_data;
+
+                        // Sanity check: we should receive the same set of unwritable reports after
+                        // writing as we did in the beginning of this transaction. If not, this is
+                        // a bug and we shouldn't commit this transaction.
+                        assert_eq!(unwritable_reports, resulting_unwriteable_reports);
                     }
 
                     Ok(Self::aggregation_job_resp_for(
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1942,188 +1942,193 @@ impl VdafOps {
         let interval_per_batch_identifier = Arc::new(interval_per_batch_identifier);
         let accumulator = Arc::new(accumulator);
 
-        Ok(datastore
-            .run_tx("aggregate_init", |tx| {
-                let vdaf = vdaf.clone();
-                let task = Arc::clone(&task);
-                let req = Arc::clone(&req);
-                let aggregation_job = Arc::clone(&aggregation_job);
-                let mut report_share_data = report_share_data.clone();
-                let interval_per_batch_identifier = Arc::clone(&interval_per_batch_identifier);
-                let accumulator = Arc::clone(&accumulator);
+        Ok(
+            datastore
+                .run_tx("aggregate_init", |tx| {
+                    let vdaf = vdaf.clone();
+                    let task = Arc::clone(&task);
+                    let req = Arc::clone(&req);
+                    let aggregation_job = Arc::clone(&aggregation_job);
+                    let mut report_share_data = report_share_data.clone();
+                    let interval_per_batch_identifier = Arc::clone(&interval_per_batch_identifier);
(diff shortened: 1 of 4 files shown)


aggregator/src/aggregator/accumulator.rs
```diff
@@ -121,6 +121,52 @@ impl<const SEED_SIZE: usize, Q: AccumulableQueryType, A: vdaf::Aggregator<SEED_S
...
+                .await
+                .map(|batch_aggregation| {
+                    batch_aggregation.map(|batch_aggregation| {
+                        if batch_aggregation.state() == &BatchAggregationState::Collected {
+                            // Unwrap safety: this only panics if the mutex is poisoned. If
+                            // it is, one of the other futures also panicked, so we can
+                            // panic too.
+                            unmergeable_report_ids
+                                .lock()
+                                .unwrap()
```
inahga, line 151: Do we generally want to comment on mutex poisoning? My impression is that `.unwrap()` on a mutex to panic on mutex poisoning is standard in Rust code.

tgeoghegan: I agree that panicking here is OK. I do still like to have an unwrap safety comment, but we could make each such comment smaller by adding a note to the code style section of `docs/DEVELOPMENT.md`, so that then you wouldn't have to explain it again each time.

tgeoghegan (approved):

aggregator/src/aggregator.rs
```diff
@@ -1953,7 +1953,7 @@ impl VdafOps {
                 let accumulator = Arc::clone(&accumulator);
 
                 Box::pin(async move {
-                    let unwritable_reports = accumulator.flush_to_datastore(tx, &vdaf).await?;
+                    let unwritable_reports = accumulator.unwritable_reports(tx, &vdaf).await?;
```
branlwyd, line 1956: `unwritable_reports` (the variable) is used to determine if we should mark an existing report aggregation as `BatchCollected`.

With this change, we introduce `unwritable_reports` (the accumulator method) to compute the set of reports that would/will be unwritable, without changing the datastore. I think this works, but it's less than ideal for a few reasons: we now have to maintain the same logic in two separate functions (`unwritable_reports` & `flush_to_datastore`); and we have to maintain more code.

Instead, WDYT about dropping the `unwritable_reports` method; instead, move the preexisting call to `flush_to_datastore` down to where you moved it in this PR, but execute it before the `try_join!` rather than as part of the `try_join!` and update the report aggregations to `BatchCollected` based on those changes.

docs/DEVELOPMENT.md
```diff
@@ -73,3 +73,27 @@ To create a new migration:
   so a panic is appropriate. (Note that some `FromSql` implementations may
   return errors for other reasons, such as out-of-range values or serde
   deserialization falures)
+
+* In production code, use of `.unwrap()` should be preceded by a comment
```
branlwyd, line 77: Good idea to document this!

branlwyd (changes requested):

aggregator/src/aggregator.rs
inahga: Nice suggestion!

I was able to have `flush_to_datastore()` still execute concurrently with the other insertion functions, by using some `.map()` trickery on the return value.

This lowered my confidence in the accumulator unit test, so I moved it further out to work at the HTTP level, with the other aggregate init tests in `http_handlers.rs`.

I also was also able to negotiate with `rustfmt` to format this code (the problem is the first `try_join!` in the closure, temporarily removing the `!` allows rustfmt to format the whole block). Apologies that the diff is somewhat mangled now.

inahga: Good gracious, my latest commit makes this block formattable, by declaring the futures outside of the `try_join!` statement and keeping their variable names short :shrug:

aggregator/src/aggregator.rs
```diff
@@ -1942,99 +1942,104 @@ impl VdafOps {
...
+                            let conflict_fut =
+                                Q::get_conflicting_aggregate_share_jobs::<SEED_SIZE, C, A>(
+                                    tx,
+                                    &vdaf,
+                                    task.id(),
+                                    req.batch_selector().batch_identifier(),
+                                    report_share_data.report_share.metadata(),
+                                );
+                            // These futures intentionally have short names and are defined separately
+                            // from the `try_join!` macro due to a `rustfmt` bug.
```
divergentdave, line 1977: FWIW, this isn't so much a bug in rustfmt as a limitation inherent to the language. Tokens inside a macro follow a much looser syntax, and in this context, many of the formatting rules are not safe to apply mechanically. A common example is `assert_eq!()` and `trailing_comma = "Vertical"`. If one were to break up an `assert_eq!()` into multiple lines, the usual rule would have you add a trailing comma after the second argument, but none of `assert_eq!()`'s rules matches that. If there's a comma after the second argument, it expects there to be at least a third message argument.

divergentdave (approved):

aggregator/src/aggregator.rs
inahga: I was thinking it a bug since the one `try_join!` breaks the ability to format the entire long closure. Point still valid though, I'll `s/bug/limitation/`'.

divergentdave: Another aspect of rustfmt is that it will do nothing to an entire "chain" if there's no way to format it that satisfies all rules. We've run into this before with long string literals or a combination of indentation, long method names, and a big turbofish.

branlwyd (approved):

PR #2356, opened 2023-12-13
Helper aggregate init: perform per-report-aggregation work concurrently.

branlwyd:
Previously, per-report-aggregation work was performed serially. This led to a number of serialized database round trips equal to the number of report aggregations. With this change, per-report-aggregation work is now done in a single serialized database round trip. This will likely be most impactful for large aggregation jobs.


branlwyd:
(note: this conflicts with currently-in-review #2355; that change should merge first since it is a bugfix, and I'll update this PR accordingly once I can. I don't stack this PR on #2355 since at time of writing, the approach may change somewhat.)

tgeoghegan (approved):

PR #2359, opened 2023-12-14
Downgrade some `tracing::info!` calls to `debug!`

tgeoghegan:
Janus was emitting traces at `info` every time it successfully handled a request (which yields one log statement per successful upload) as well as for a variety of client errors (e.g., undecodable input share). Downgrade those to `debug` to keep a lid of logging and log storage costs.


inahga (commented):
I think we can also downgrade `warn!` on 1444 and 1529 to `debug!` since it's they're client errors. This would be for consistency, i.e. all client error should be the same log level.

PR #2363, opened 2023-12-15
Fix deadlock in `Accumulator::flush_to_datastore()`

inahga:
A deadlock is possible in `Accumulator::flush_to_datastore()` because there is no deterministic order to shard updates. Fix this by sorting the batch aggregations by shard ID before calling `update_batch_aggregation()`.

The effect on the helper's `aggregate_init` is quite nice! Deadlocks turn out to be extremely costly.

I installed some metrics to categorize database retries by reason (i.e. serialization error or deadlock)[^1]. I then issued 31680 Prio3Count time interval reports at about 500 QPS. Here's the helper's retry metrics:
image

After this change:
image

Note there's still a few deadlocks in the aggregate_init's usage of `Transaction::update_batch()`, but the change has removed all deadlocks in `flush_to_datastore()`. I will investigate `update_batch()` in another PR.

The transaction duration stats also show massive improvement (before and after is shown in the same graph).
image

I repeated these results over 3 trials. I did not profile the leader's usage of `Accumulator`, but I imagine it's an improvement in any case.

[^1]: There is a metric `pg_stat_database_deadlocks`, accessible with `SELECT datname, deadlocks FROM pg_stat_database` but I was having trouble plumbing this through prometheus. Something to work on for ops.

### DIFF
--- a/aggregator/src/aggregator/accumulator.rs
+++ b/aggregator/src/aggregator/accumulator.rs
@@ -137,7 +137,22 @@ impl<const SEED_SIZE: usize, Q: AccumulableQueryType, A: vdaf::Aggregator<SEED_S
     ) -> Result<HashSet<ReportId>, datastore::Error> {
         let unmergeable_report_ids = Arc::new(Mutex::new(HashSet::new()));
 
-        try_join_all(self.aggregations.values().map(|data| {
+        // Prevent deadlocks when inserting into shards by providing a deterministic order to shard
+        // updates. Suppose two concurrent processes are attempting to update a batch aggregation.
+        // We want to avert this situation:
+        //
+        // A> BEGIN;
+        // A> UPDATE batch_aggregations WHERE ord = 1 ... -- Row with ord 1 is locked for update.
+        // B> BEGIN;
+        // B> UPDATE batch_aggregations WHERE ord = 2 ... -- Row with ord 2 is locked for update.
+        // A> UPDATE batch_aggregations WHERE ord = 2 ... -- A is now blocked waiting for B to finish.
+        // B> UPDATE batch_aggregations WHERE ord = 1 ... -- Kaboom!
+        //
+        // This situation cannot happen if the UPDATE statements are ordered by `ord`.
+        let mut aggregations: Vec<_> = self.aggregations.values().collect();
+        aggregations.sort_unstable_by_key(|data| data.batch_aggregation.ord());
+
+        try_join_all(aggregations.into_iter().map(|data| {
             let unmergeable_report_ids = Arc::clone(&unmergeable_report_ids);
             async move {
                 match tx
--- a/aggregator/src/aggregator/accumulator.rs
+++ b/aggregator/src/aggregator/accumulator.rs
@@ -148,9 +148,23 @@ impl<const SEED_SIZE: usize, Q: AccumulableQueryType, A: vdaf::Aggregator<SEED_S
         // A> UPDATE batch_aggregations WHERE ord = 2 ... -- A is now blocked waiting for B to finish.
         // B> UPDATE batch_aggregations WHERE ord = 1 ... -- Kaboom!
         //
-        // This situation cannot happen if the UPDATE statements are ordered by `ord`.
+        // Fully ameliorating this requires we sort by `batch_identifier` and `ord`.
+        //
+        // However, `try_join_all` executes futures concurrently thus there's no guarantee that the
+        // order will be respected, so there remains the possibility of deadlock. This is rare--in
+        // testing we have noticed that the probability of deadlock is drastically lower with a
+        // sorted list than with an unsorted list.
+        //
+        // There could be changes in the `Accumulator`'s usage patterns or changes in the tokio
+        // scheduler that cause this to regress, in which case the next possible solution is to
+        // execute batch aggregation updates serially.
         let mut aggregations: Vec<_> = self.aggregations.values().collect();
-        aggregations.sort_unstable_by_key(|data| data.batch_aggregation.ord());
+        aggregations.sort_unstable_by_key(|data| {
+            (
+                data.batch_aggregation.batch_identifier(),
+                data.batch_aggregation.ord(),
+            )
+        });
 
         try_join_all(aggregations.into_iter().map(|data| {
             let unmergeable_report_ids = Arc::clone(&unmergeable_report_ids);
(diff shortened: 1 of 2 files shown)


inahga:
> I did not profile the leader's usage of Accumulator, but I imagine it's an improvement in any case.

The leader turns out to be pretty much unaffected by deadlocks and has good performance on this query. I believe this is because the leader's usage of `flush_to_datastore()` is preceded by a network call to the helper, which naturally introduces some jitter to concurrent `flush_to_datastore()` calls. (Which lends a bit of credence to my theory that exponential backoff on database retries may be a performance win).

In short this PR should only be considered an improvement to a Janus helper.

tgeoghegan (approved):
Can't argue with graphs! You mentioned that you installed some metrics to categorize DB errors. Should we add those to Janus? It seems like it's interesting information.

inahga:
The [code I wrote only](https://github.com/divviup/janus/commit/ebe4922d) categorizes retry reasons--the existing `janus_database_transactions_total` has us covered for general errors but doesn't distill retries by reason.

I think we can get the same information out of the `pg_stat_database.deadlocks` row, I'll add a dashboard for our deployments. It won't tell us the relative rates of serialization retries vs deadlock retries. But deadlock IMO is the more interesting error anyways, as it represents a programming error. Serialization retries are just the cost of our transaction isolation level.

aggregator/src/aggregator/accumulator.rs
```diff
@@ -137,7 +137,22 @@ impl<const SEED_SIZE: usize, Q: AccumulableQueryType, A: vdaf::Aggregator<SEED_S
...
+        // B> BEGIN;
+        // B> UPDATE batch_aggregations WHERE ord = 2 ... -- Row with ord 2 is locked for update.
+        // A> UPDATE batch_aggregations WHERE ord = 2 ... -- A is now blocked waiting for B to finish.
+        // B> UPDATE batch_aggregations WHERE ord = 1 ... -- Kaboom!
+        //
+        // This situation cannot happen if the UPDATE statements are ordered by `ord`.
+        let mut aggregations: Vec<_> = self.aggregations.values().collect();
+        aggregations.sort_unstable_by_key(|data| data.batch_aggregation.ord());
+
+        try_join_all(aggregations.into_iter().map(|data| {
```
branlwyd, line 155: N.B. this isn't a total solution because [`try_join_all`](https://docs.rs/futures/latest/futures/future/fn.try_join_all.html) drives its futures concurrently, with no guaranteed order. Since we see a performance improvement, it seems that the current implementation will (usually?) drive the futures in the given order at least to the point that a deadlock would be created. If the implementation changes in a futures update, we might see a regression here.

All that said, a performance win is a performance win. I have no objection to merging this as-is, since the apparent "solution" to enforce ordering would be to run these transactions serially -- and that would be a big perf loss.

aggregator/src/aggregator/accumulator.rs
```diff
@@ -137,7 +137,22 @@ impl<const SEED_SIZE: usize, Q: AccumulableQueryType, A: vdaf::Aggregator<SEED_S
...
+        // A> BEGIN;
+        // A> UPDATE batch_aggregations WHERE ord = 1 ... -- Row with ord 1 is locked for update.
+        // B> BEGIN;
+        // B> UPDATE batch_aggregations WHERE ord = 2 ... -- Row with ord 2 is locked for update.
+        // A> UPDATE batch_aggregations WHERE ord = 2 ... -- A is now blocked waiting for B to finish.
+        // B> UPDATE batch_aggregations WHERE ord = 1 ... -- Kaboom!
+        //
+        // This situation cannot happen if the UPDATE statements are ordered by `ord`.
+        let mut aggregations: Vec<_> = self.aggregations.values().collect();
+        aggregations.sort_unstable_by_key(|data| data.batch_aggregation.ord());
```
branlwyd, line 153: nit: a single `Accumulator` instance handles a single `task_id` & `aggregation_param`, but in a time-interval query we might be accumulating into multiple batches at once. I think we should sort on a key that is `(batch_identifier, ord)` rather than just `ord`.

branlwyd (approved):

aggregator/src/aggregator/accumulator.rs
inahga: Good point! I'll fix my comment to point this out.

aggregator/src/aggregator/accumulator.rs
inahga: Nice catch!

inahga: This was a bit trickier than I had originally thought, since `QueryType::BatchIdentifier` is not `Ord`.

I don't think it makes sense to make it `Ord`. That'll be a breaking change in the `messages` crate. I also don't think there's a meaningful ordering for `TimeInterval` (indeed even `Range` is not `Ord`).

So I ended up implementing a `compare_batch_identifiers()` function on `AccumulableQueryType` with the noted caveat that it's really only for database ordering, which fits the purpose of the `aggregator_core` crate anyways.

I could use a peek at it since it feels somewhat hacky, maybe there's an easier way.

branlwyd: tl;dr: I think we should make `QueryType::BatchIdentifier` and `QueryType::PartialBatchIdentifier` `Ord` (but let's chat about why/how this is a breaking change). Let's make them `PartialOrd` too, for completeness.

> I don't think it makes sense to make it Ord. That'll be a breaking change in the messages crate.

Why is this breaking -- is it only because someone outside of Janus might have implemented the `QueryType` trait (and would now have to make sure their batch identifiers are `Ord`)? If so, IMO we should take this change without considering a major-version bump; I think `QueryType` is intended for Janus usage only, but is defined in `janus_messages` as a public type because (a) `janus_messages` needs to understand query types and (b) dependent application crates also need to understand query types. I don't think we really expect anyone to implement a new query type, and if they do, they making their batch identifier `Ord` seems like a small change even if technically breaking -- batch identifiers ought to be plain-old-datatypes that can be compared lexicographically, if no other comparison is feasible.

> I also don't think there's a meaningful ordering for Interval (indeed even Range is not Ord).

Any consistent definition of `Ord` will work here. For example, deriving `PartialOrd` and `Ord` on `Interval` gives a "lexicographic" comparison, i.e. start time is compared first, then duration is compared if the start times are equal.

inahga: Lexicographic `Ord` SGTM!

> is it only because someone outside of Janus might have implemented the QueryType trait (and would now have to make sure their batch identifiers are Ord)?

That was my thought, yes.

PR #2364, opened 2023-12-15
Helper aggregate init idempotency tweaks.

branlwyd:
1. Perform `flush_to_datastore`, computing `unwritable_reports`, and update report aggregations appropriately before writing report aggregations back to the datastore. Before this PR, these were done concurrently with the report aggregation writes; this meant that, if an unwritable report occurred, we would store a different set of report aggregations than we returned in the response. (This introduces a serialized round trip to the datastore, but this corresponds to a serialized round trip that was removed with #2355, so we are at net unchanged number of round trips to the datastore.)

2. When we are handling a replayed request, return the stored report aggregations, instead of the report aggregations we computed. This ensures that our response to a repeated request matches the original response, without requiring us to perfectly recompute the set of responses (which, in some cases, would require tripping the same race conditions as before e.g. for BatchCollected errors). I think this closes a few real holes in our idempotency handling: other than the bug described in point 1, we can also update a report aggregation to ReportReplayed when we try to write it, which we wouldn't recompute correctly on a retry. In general, I think this approach is less error-prone/requires less complicated logic to get right. But note that it does require reading some additional data from the datastore in the replayed-request path.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -2038,17 +2038,36 @@ impl VdafOps {
                             }
 
                             return Ok(Self::aggregation_job_resp_for(
-                                report_share_data
-                                    .into_iter()
-                                    .map(|data| data.report_aggregation),
+                                tx.get_report_aggregations_for_aggregation_job(
+                                    vdaf.as_ref(),
+                                    &Role::Helper,
+                                    task.id(),
+                                    aggregation_job.id(),
+                                )
+                                .await?,
                             ));
                         }
                         Err(e) => return Err(e),
                     };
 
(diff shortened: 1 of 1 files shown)


branlwyd:
Note that, in a replayed request, we still do all of the pre-transaction VDAF computation. This is seemingly unnecessary now that we're returning the report aggregations that were written to the datastore on the initial request. Unfortunately, it wouldn't be easy to remove this work: we'd either need to check idempotency in a new transaction before the VDAF computation work, or we'd need to move the VDAF computation work into the transaction; both of these seem less desirable than repeating the computational work.

inahga (approved):
Makes sense!

PR #2380, opened 2023-12-19
Remove `uuid` dependency

tgeoghegan:
`janus_aggregator` and `janus_aggregator_core` have been carrying around a dependency on `uuid` for a while, but neither crate actually uses them.


inahga (approved):

jbr:
we may want to add a `cargo udeps` gh action

PR #2381, opened 2023-12-19
Exponential backoff on database retries

inahga:
****


inahga:
With this the helper sees fewer retries on `aggregate_init` and lower database CPU usage. However, this turns out to be rather regressive for the leader. I've tried with a number of backoff parameters, and all yield unacceptable regressions in 50th pct transaction times and overall transaction rollback rate.

I think we'll get more optimization mileage out of redesigning queries to avoid serialization errors in the first place.

PR #2382, opened 2023-12-19
Run E2E integration tests against cloud env

tgeoghegan:
Extends the test harness in
`janus_integration_tests::integration::in_cluster` so that it can run
tests against remote instances of divviup-api, a Janus leader and a
Janus helper.

Concretely, this means connecting to those services at some remote URL
rather than over a `kubectl port-forward` to a local Kind cluster. Past
that, we also exercise less of the `divviup-api` surface.

In Kind, we do everything from creating an account through provisioning
a task. This works because the `divviup-api` used in that context has
authentication disabled and also because that entire environment is
understood to be ephemeral -- we can litter it with tasks and paired
aggregators and anything else we want and not clean up, because our
expectation is that the entire Kubernetes cluster and all its storage
will be destroyed.

But in this new configuration, we expect to run the test in long-running
staging environments, or even production. While I'm not concerned about
adding new tasks, reports and aggregatons there -- task expiry and
garbage collection will eventually clean those up -- I don't want to
create new accounts and paired aggregators on every test run. That means
that the test needs to be run with a handful of `JANUS_E2E_` environment
variables defined that provide an authentication token and IDs for
various resources to be used. See the doccomment on
`InClusterJanusPair::new_in_cloud` for details.

This change also adds yet more code to Janus that is very, very useful
for testing Divvi Up, but not necessarily relevant to Janus (after all,
someone could be running Janus without `divviup-api` or without
Kubernetes). I'm interested in moving this chunk of test harness
elsewhere, but for now this is the best way to get a reviewable diff.

Relevant to https://github.com/divviup/janus-ops/issues/86

### DIFF
--- a/integration_tests/Cargo.toml
+++ b/integration_tests/Cargo.toml
@@ -12,6 +12,7 @@ version.workspace = true
 in-cluster = ["dep:k8s-openapi", "dep:kube"]
 testcontainer = ["janus_interop_binaries/testcontainer"]
 in-cluster-rate-limits = []
+in-cluster-in-cloud = []
 
 [dependencies]
 anyhow.workspace = true
@@ -38,8 +39,10 @@ serde.workspace = true
 serde_json = "1.0.110"
 testcontainers.workspace = true
 tokio.workspace = true
-url = { version = "2.5.0", features = ["serde"] }
+trillium-rustls.workspace = true
 trillium-tokio.workspace = true
+url.workspace = true
+uuid.workspace = true
 
 [dev-dependencies]
 divviup-client = { version = "0.1", features = ["admin"] }
--- a/integration_tests/Cargo.toml
+++ b/integration_tests/Cargo.toml
@@ -12,13 +12,13 @@ version.workspace = true
 in-cluster = ["dep:k8s-openapi", "dep:kube"]
 testcontainer = ["janus_interop_binaries/testcontainer"]
 in-cluster-rate-limits = []
-in-cluster-in-cloud = []
 
 [dependencies]
 anyhow.workspace = true
 assert_matches.workspace = true
 backoff = { version = "0.4", features = ["tokio"] }
 base64.workspace = true
+clap.workspace = true
 futures = "0.3.30"
 hex = "0.4"
 http = "0.2"
@@ -45,6 +45,7 @@ url.workspace = true
 uuid.workspace = true
 
 [dev-dependencies]
+chrono.workspace = true
 divviup-client = { version = "0.1", features = ["admin"] }
 janus_collector = { workspace = true, features = ["test-util"] }
 rstest.workspace = true
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -4,6 +4,7 @@ use janus_aggregator_core::task::QueryType;
 use janus_collector::AuthenticationToken;
 use janus_core::{hpke::HpkeKeypair, vdaf::VdafInstance};
 use janus_messages::{Duration, TaskId};
+use std::time;
 use url::Url;
 
 pub mod client;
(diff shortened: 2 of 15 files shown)


tgeoghegan:
I'm using a cargo feature to govern whether the tests hit a local cluster or a remote set of services. I'm open to other suggestions on how to do this.

inahga:
It is possible to have additional CLI args for a Rust test:
```rust
    #[test]
    fn args() {
        std::env::args().for_each(|arg| println!("{}", arg));
    }
```
```
🐧 cargo test args -- --nocapture -- --foobar
[...]
args
--nocapture
--
--foobar
[...]
```
So we _could_ use `clap` argument parsing. While its invocation is ugly, it does have the benefit of being a bit more self-documenting, as we can use `cargo test -- -- --help` to get the site-specific arguments.

I feel moderately strongly that this would be an improvement, but as it's a rather large yak-shave I'd suggest we take it as a future improvement.

integration_tests/Cargo.toml
```diff
@@ -40,8 +41,10 @@ serde.workspace = true
 serde_json = "1.0.108"
 testcontainers.workspace = true
 tokio.workspace = true
-url = { version = "2.5.0", features = ["serde"] }
+trillium-rustls.workspace = true
```
inahga, line 44: What's the rationale for having `trillium-rustls` be a workspace dependency when it's only used once? I've actually been grappling with "when to use workspace dependency vs. not" so was curious as to your thoughts--I don't feel strongly one way or another.

integration_tests/tests/integration/common.rs
```diff
@@ -182,8 +236,8 @@ pub async fn submit_measurements_and_verify_aggregate_generic<V>(
     .with_collect_poll_backoff(
         ExponentialBackoffBuilder::new()
             .with_initial_interval(time::Duration::from_millis(500))
-            .with_max_interval(time::Duration::from_millis(500))
-            .with_max_elapsed_time(Some(time::Duration::from_secs(60)))
+            .with_max_interval(time::Duration::from_secs(30))
+            .with_max_elapsed_time(Some(time::Duration::from_secs(600)))
```
inahga, line 240: Shouldn't this use `task_parameters.max_interval` and `task_parameters.max_elapsed_time`?

inahga (changes requested):
Nicely written! One note, I think the collector interval/elapased_time are not fully plumbed through.

integration_tests/Cargo.toml
tgeoghegan: Hmm, this might be rebase carnage because we did happen to use `trillium-rustls` in more than one crate at some point. I don't feel very strongly about this one way or the other, either. My only thought is that we will probably end up using `trillium-rustls` in another Janus crate at some point, and so we'll have saved ourselves the trouble of moving the dependency by eagerly putting it in the top-level `Cargo.toml`.

integration_tests/tests/integration/common.rs
tgeoghegan: Yes, it absolutely should! Thanks.

tgeoghegan:
> It is possible to have additional CLI args for a Rust test:

Nice thing about this would be that we wouldn't have to worry about some functions showing up as dead code depending on feature enablement, so I kinda like it. I'm inclined to go this way.

integration_tests/tests/integration/in_cluster.rs
```diff
@@ -32,8 +38,185 @@ struct InClusterJanusPair {
...
+            leader_aggregator_dap_url,
+            helper_aggregator_dap_url,
+            Duration::from_secs(30),
+            Duration::from_secs(600),
+        );
+        let task = task_builder.with_min_batch_size(100).build();
+        task_parameters.min_batch_size = 100;
+
+        let provision_task_request = NewTask {
+            name: format!("Integration test task {}", Uuid::new_v4()),
```
divergentdave, line 152: Putting a UUID in the task's name seems like it could be confusing, when the task will get another UUID assigned to it in divviup-api's database. I suggest we insert a date and time here instead. (same for the other NewTask occurrence)

divergentdave (approved):

integration_tests/tests/integration/in_cluster.rs
tgeoghegan: Good idea -- I think name collisions don't matter, and timestamp will be sufficient to let us tell one from the other.

integration_tests/src/lib.rs
```diff
@@ -22,32 +23,42 @@ pub struct TaskParameters {
...
     pub fn endpoint_for_host(&self, port: u16) -> Url {
-        let path = match self {
-            AggregatorEndpointFragments::VirtualNetwork { path, .. } => path,
-            AggregatorEndpointFragments::Localhost { path } => path,
-        };
-        Url::parse(&format!("http://127.0.0.1:{port}{path}")).unwrap()
+        match self {
+            AggregatorEndpointFragments::VirtualNetwork { path, .. }
+            | AggregatorEndpointFragments::Localhost { path } => {
+                Url::parse(&format!("http://127.0.0.1:{port}{path}")).unwrap()
```
jbr, line 58: The tiniest of nits: Will this always be run on an ipv4 interface? `"localhost:{port}{path}"` might be preferable if not

divergentdave: Unfortunately we need to pick one localhost address and stick with it, because when Docker sets up a port forward, it may use different port numbers for 127.0.0.1 and ::1. This choice is paired with `get_host_port_ipv4()` vs. `get_host_port_ipv6()` elsewhere.

integration_tests/tests/integration/in_cluster.rs
```diff
@@ -32,8 +38,185 @@ struct InClusterJanusPair {
...
+                Uuid::parse_str(&leader_aggregator_id).unwrap(),
+                Uuid::parse_str(&helper_aggregator_id).unwrap(),
+                Uuid::parse_str(&collector_credential_id).unwrap(),
+                serde_json::from_str::<PrivateCollectorCredential>(&collector_credential_json)
+                    .unwrap(),
+            ),
+            _ => panic!("missing or invalid environment variables"),
+        };
+
+        let mut divviup_api = DivviupClient::new(
```
jbr, line 122: I'm not sure why multiline suggestions aren't working, but another entirely-equivalent but slightly more concise option would be

```rust
let divviup_api = DivviupClient::new(divviup_api_token, RustlsConfig::<ClientConfig>::default())
    .with_default_pool()
    .with_url(divviup_api_url);

```

inahga (approved):
:+1:

PR #2383, opened 2023-12-20
GC: add config to group GC in tx's, control concurrent tx's.

branlwyd:
Two new settings are introduced to the GC config:
 * tasks_per_tx: controls how many tasks are grouped into the same transaction for purposes of garbage collection. If left unset, each task's GC work is run in its own transaction.
 * concurrent_tx_limit: controls how many concurrent database transactions are allowed for GC. If left unset, there is no limit to the number of concurrent transactions used for GC.

If neither of these new options are configured, the behavior will be left as before -- each task will be GC'ed in its own transaction, with no limit to the number of concurrent transactions.

Note that grouping GC for different tasks into a single transaction risks failures (or retries) in the GC process for one task impacting GC for other, independent tasks.

### DIFF
--- a/aggregator/src/aggregator/garbage_collector.rs
+++ b/aggregator/src/aggregator/garbage_collector.rs
@@ -1,10 +1,14 @@
-use anyhow::{Context, Result};
-use futures::future::join_all;
-use janus_aggregator_core::{datastore::Datastore, task::AggregatorTask};
+use anyhow::{Context, Error, Result};
+use futures::future::{join_all, try_join_all, OptionFuture};
+use janus_aggregator_core::datastore::{self, Datastore};
 use janus_core::time::Clock;
+use janus_messages::TaskId;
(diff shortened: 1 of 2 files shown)


inahga (commented):
The implementation LGTM.

I get why we might want to limit the number of concurrent transactions. But why would we want to group more tasks for GC? I don't see the upside to it--in my head the general strategy is to keep transactions as small as reasonably possible.

branlwyd:
For efficiency reasons, we might decide that fewer, larger transactions are a good idea. Keeping transactions reasonably small & more importantly independent is a good idea to avoid conflicts. But transaction overhead means it makes sense to group work together sometimes -- e.g. we batch report uploads to avoid one-transaction-per-upload. I am genuinely unsure if we'll find value in grouping transactions (a GC transaction will tend to be much larger than a single-report-upload transaction), but I think having a knob to experiment with would be a good idea.

inahga (approved):
SGTM!

PR #2384, opened 2023-12-20
Metric for transaction retries

inahga:
This metric will help us identify transactions that excessively retry due to serialization errors or deadlock. Transactions that retry too much are candidates for optimization, e.g. by sharding or fixing the underlying deadlock.

Here's an example of how this looks:
```promql
label_replace(histogram_quantile(0.5, sum without(job, node, instance, service) (rate(janus_database_retries_bucket[10m]))), "quantile", "0.50", "", "") or label_replace(histogram_quantile(0.95, sum without(job, node, instance, service) (rate(janus_database_retries_bucket[10m]))), "quantile", "0.95", "", "") or label_replace(histogram_quantile(0.99, sum without(job, node, instance, service) (rate(janus_database_retries_bucket[10m]))), "quantile", "0.99", "", "")
```
image

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -166,12 +166,12 @@ impl<C: Clock> Datastore<C> {
         meter: &Meter,
     ) -> Datastore<C> {
         let transaction_status_counter = meter
-            .u64_counter("janus_database_transactions")
+            .u64_counter(TRANSACTION_METER_NAME)
             .with_description("Count of database transactions run, with their status.")
             .with_unit(Unit::new("{transaction}"))
             .init();
         let rollback_error_counter = meter
-            .u64_counter("janus_database_rollback_errors")
+            .u64_counter(TRANSACTION_ROLLBACK_METER_NAME)
             .with_description(concat!(
                 "Count of errors received when rolling back a database transaction, ",
                 "with their PostgreSQL error code.",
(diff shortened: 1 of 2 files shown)


tgeoghegan (approved):

divergentdave (approved):
FWIW we already get some coarse retry statistics from `janus_database_transactions_total{status="retry"}`, but I like that this histogram will give us the distribution of retries as well.

I'd recommend specifying a custom set of histogram buckets for this over in `metrics.rs`, because the seven bucket boundaries under 1.0 won't do us any good on this metric.

aggregator_core/src/datastore.rs
```diff
@@ -209,6 +216,7 @@ impl<C: Clock> Datastore<C> {
         for<'a> F:
             Fn(&'a Transaction<C>) -> Pin<Box<dyn Future<Output = Result<T, Error>> + Send + 'a>>,
     {
+        let mut retry_count = 0;
```
branlwyd, line 219: outside the scope of this PR: given that we've tripped over an infinite-retry on transactions at least once, I think we should eventually implement a maximum-retry-count to act as a circuit breaker. However, that can wait until at least we know what realistic retry counts look like, perhaps based on feedback from this metric once it's deployed. We should probably set our maximum retry count to at least an order of magnitude higher than that (and yell/alert as loudly as possible if we start to encounter it, since it either indicates that our maximum retry count is too low or that we've found a serious bug).

branlwyd (approved):

aggregator_core/src/datastore.rs
branlwyd: This idea is captured in https://github.com/divviup/janus/issues/2124.

inahga: Agreed. I was hoping that [introducing `backoff` to this](https://github.com/divviup/janus/pull/2381) logic would handle cancellation for us, but I found exponential backoff to be regressive.

inahga:
@divergentdave I could use another peek to ensure I didn't get anything subtly wrong (although my view of the metrics looks fine).

divergentdave (approved):

PR #2386, opened 2023-12-21
Return server error for abandoned collection jobs

### LINKED ISSUE #248: Decide how to handle collection attempts on abandoned collect jobs
In some cases (e.g. repeated failures) we will permanently abandon collect jobs. These collect jobs will never complete. However, the collector will have the collect job ID, and can thus request collection of these jobs.

We might consider specifying the "permanent failure" state in the DAP spec; the upside would be an interoperable way for an aggregator to communicate to a collector that a given collect job should be abandoned.

inahga:
Fixes #248.

I opted not to embed an abandonment reason into the DB. There is no action that a user can take for this error except to contact the operators, so we're going to have to go log crawling anyway. The current cases that cause this also do not have straightforward fixes (why is the collection job retrying so many times? why is there a batch mismatch?). Thus I don't think it's worth the additional complexity.

~Note that this PR is insufficient to stop collectors from infinite looping, [because our retry logic in `core.rs` retries on all server errors](https://github.com/divviup/janus/blob/main/core/src/retries.rs#L89). I will address this in a future PR. I think we need to not retry on 500 errors, but also give a pass through `aggregator/src/aggregator/http_handlers.rs` to provide `503 Service Unavailable` for retryable errors, e.g. DB pool timeouts.~ Nope, not quite. We do have a `max_elapsed_time` set for our default backoff config, so a collector will retry the 500 error for up to ten minutes before giving up.

### DIFF
--- a/aggregator/src/aggregator/http_handlers.rs
+++ b/aggregator/src/aggregator/http_handlers.rs
@@ -74,7 +74,7 @@ impl Handler for Error {
             Error::DeletedCollectionJob(_) => conn.with_status(Status::NoContent),
             Error::AbandonedCollectionJob(collection_job_id) => conn.with_problem_document(
                 &ProblemDocument::new(
-                    "https://docs.divviup.org/references/janus_errors#collection-job-abandoned",
+                    "https://docs.divviup.org/references/janus-errors#collection-job-abandoned",
                     "The collection job has been abandoned.",
                     Status::InternalServerError,
                 )
--- a/aggregator/src/aggregator/http_handlers.rs
+++ b/aggregator/src/aggregator/http_handlers.rs
@@ -44,59 +44,73 @@ impl Handler for Error {
         warn!(error_code, error=?self, "Error handling endpoint");
         match self {
             Error::InvalidConfiguration(_) => conn.with_status(Status::InternalServerError),
-            Error::MessageDecode(_) => {
-                conn.with_problem_document(&ProblemDocument::new(DapProblemType::InvalidMessage))
-            }
+            Error::MessageDecode(_) => conn
+                .with_problem_document(&ProblemDocument::new_dap(DapProblemType::InvalidMessage)),
             Error::ReportRejected(task_id, _, _, reason) => conn.with_problem_document(
-                &ProblemDocument::new(DapProblemType::ReportRejected)
+                &ProblemDocument::new_dap(DapProblemType::ReportRejected)
                     .with_task_id(task_id)
                     .with_detail(reason.detail()),
             ),
             Error::InvalidMessage(task_id, _) => {
-                let mut doc = ProblemDocument::new(DapProblemType::InvalidMessage);
+                let mut doc = ProblemDocument::new_dap(DapProblemType::InvalidMessage);
                 if let Some(task_id) = task_id {
                     doc = doc.with_task_id(task_id);
                 }
                 conn.with_problem_document(&doc)
             }
             Error::StepMismatch { task_id, .. } => conn.with_problem_document(
-                &ProblemDocument::new(DapProblemType::StepMismatch).with_task_id(task_id),
+                &ProblemDocument::new_dap(DapProblemType::StepMismatch).with_task_id(task_id),
             ),
             Error::UnrecognizedTask(task_id) => conn.with_problem_document(
-                &ProblemDocument::new(DapProblemType::UnrecognizedTask).with_task_id(task_id),
+                &ProblemDocument::new_dap(DapProblemType::UnrecognizedTask).with_task_id(task_id),
             ),
             Error::MissingTaskId => {
-                conn.with_problem_document(&ProblemDocument::new(DapProblemType::MissingTaskId))
+                conn.with_problem_document(&ProblemDocument::new_dap(DapProblemType::MissingTaskId))
             }
             Error::UnrecognizedAggregationJob(task_id, _) => conn.with_problem_document(
-                &ProblemDocument::new(DapProblemType::UnrecognizedAggregationJob)
+                &ProblemDocument::new_dap(DapProblemType::UnrecognizedAggregationJob)
                     .with_task_id(task_id),
             ),
             Error::DeletedCollectionJob(_) => conn.with_status(Status::NoContent),
+            Error::AbandonedCollectionJob(collection_job_id) => conn.with_problem_document(
+                &ProblemDocument::new(
+                    "https://docs.divviup.org/references/janus_errors#collection-job-abandoned",
+                    "The collection job has been abandoned.",
+                    Status::InternalServerError,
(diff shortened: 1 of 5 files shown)


aggregator/src/aggregator/http_handlers.rs
```diff
@@ -44,59 +44,73 @@ impl Handler for Error {
...
+                    "The collection job has been abandoned.",
+                    Status::InternalServerError,
+                )
+                .with_detail(concat!(
+                    "An internal problem has caused the server to stop processing this collection ",
+                    "job. The job is no longer collectable. Contact the server operators for ",
+                    "assistance."
+                ))
+                .with_collection_job_id(collection_job_id),
+            ),
```
inahga, line 87: Notice I've created a few new conventions here.

The first is that we can use `ProblemDocument::new()` for any manner of custom errors that don't necessarily fit into a DAP problem type. The existing DAP problem document creation is moved to `ProblemDocument::new_dap()`.

The second is that I used `https://docs.divviup.org/references/janus_errors#collection-job-abandoned` as the problem details `type`. [The RFC states that type should be a URI](https://www.rfc-editor.org/rfc/rfc7807#section-3.1) that is resolvable to some user readable documentation. [DAP also states that we must not use its namespace for custom errors](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-3.2-4) (IIUC? The phrasing is odd...), nor do I think we should define custom error types for Janus-specific concerns.

It seems the easiest place to put user-friendly error documentation is in our docs site. Anyone else have any other thoughts?

(n.b. the link doesn't exist atm--I will create the documentation before merging this PR).

tgeoghegan: This strikes me as a good way to represent this kind of error. Some disconnected thoughts:

- linking to docs.divviup.org makes sense for us, Divvi Up, but I'm not sure it's right for anyone else who might ever use Janus as a DAP aggregator. I suppose the ideal would be some kind of configuration file allowing a deployment to provide a map of error codes to documentation URIs, but I'm reluctant to commit to something that complicated. I think we could move ahead with this as is but maybe file an issue about at least having a toggle to disable this in config.
- If you want to make sure your usage of problem documents is valid, you could check what Boulder does. You could also ask for advice on the [httpbis](https://datatracker.ietf.org/wg/httpbis/about/) list.
- It's not that surprising to me that the language in DAP is unclear here. You should consider making a PR to disambiguate. I can promise the editors will be receptive to it.

aggregator/src/aggregator/problem_details.rs
```diff
@@ -36,19 +36,30 @@ pub struct ProblemDocument<'a> {
     taskid: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     detail: Option<&'a str>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    collection_job_id: Option<String>,
 }
 
 impl<'a> ProblemDocument<'a> {
-    pub fn new(error_type: DapProblemType) -> Self {
```
tgeoghegan, line 42: We should have doccomments explaining the distinction between `ProblemDocument::new` and `::new_dap` and explaining which should be used, when.

aggregator/src/aggregator/problem_details.rs
```diff
@@ -36,19 +36,30 @@ pub struct ProblemDocument<'a> {
...
+    pub fn new(type_: &'static str, title: &'static str, status: Status) -> Self {
```
tgeoghegan, line 44: This is a breaking change to a public export from crate `janus_aggregator`. However we don't ship that crate anywhere (unlike the client and collector crates), so I'm not sure if we observe semver for this kind of thing. I bet if you look we break this crate's API all the time, but could you double check?

aggregator/src/aggregator/http_handlers.rs
branlwyd: IMO, since we need to reference some real resolvable URL here, I think referencing some page under `docs.divviup.org` makes sense. I suppose the justification is "Janus, while separate from DU, is maintained by the DU team & therefore documented on the DU site." Or, to put it another way, I think we'd write advice for deploying/troubleshooting Janus into the `docs.divviup.org` domain, too.

The other options would seem to be:
1) Deploy a new domain for Janus-specific documentation. This seems like a lot of effort just to maintain DU/Janus separation.
2) Use a URL which references some GitHub page in the Janus repository. I'd prefer to use domains we maintain control of.

Notably, I believe `type` is the key that would be used for programmatic determination of the type of error -- all other fields are at the wrong level of specificity and/or intended for human consumption. (also, per the RFC: "Consumers MUST use the "type" string as the primary identifier for the problem type") So I think we don't want to change this often (outside of major-version breaks?), and don't want to let it be configured -- this would be equivalent to e.g. changing error codes, and may confuse software written to consume these codes.

aggregator/src/aggregator/problem_details.rs
branlwyd: If it's not in `crates.io`, as far as I'm concerned we don't have a public API to break. I think someone could manually clone the Janus repository & depend on `janus_aggregator` that way? But IMO that is not usage we want to support.

branlwyd (approved):

aggregator/src/aggregator/problem_details.rs
tgeoghegan: I think that's reasonable, too. We should document this semver policy in `docs/DEVELOPMENT.md` (not needed in this PR though).

inahga: Indeed I've been treating `aggregator`'s API boundary as its configuration and HTTP API endpoints. Agreed on writing docs, I can make a PR.

branlwyd: Yep, I agree with that: a breaking change to `janus_aggregator` would be a breaking change to deployability of the artifacts (binaries/containers), rather than to the APIs in the crate.

aggregator/src/aggregator/http_handlers.rs
inahga: Interestingly Boulder doesn't really run into this kind of need, because the ACME spec covers some generic errors like `serverInternal` and `unauthorized`.

Indeed, [our friend has run into this kind of situation](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/437) ahead of us. Given the discussion in that thread I do think having `type` set to a documentation page is the most helpful thing we can do for this particular error and still fully fits DAP's intended usage of problem details 

As for having the DU name attached, I'm inclined to proceed as-is and just be receptive to any future proposals to make it configurable (i.e. if another entity wants to run a DAP aggregator but not have it link to our site).

tgeoghegan (approved):

PR #2390, opened 2023-12-22
Add versioning and stability policy

inahga:
As discussed in https://github.com/divviup/janus/pull/2386#discussion_r1434372926.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,19 @@ supported.
 
 Janus is currently in active development.
 
+<!--toc:start-->
+- [janus](#janus)
+  - [Draft versions and release branches](#draft-versions-and-release-branches)
+  - [Building](#building)
+    - [Container image](#container-image)
+  - [Minimum Supported Rust Version (MSRV)](#minimum-supported-rust-version-msrv)
+  - [Running tests](#running-tests)
+    - [inotify limits](#inotify-limits)
+  - [Deploying Janus](#deploying-janus)
+  - [Cargo features](#cargo-features)
+  - [Versioning and Stability](#versioning-and-stability)
+<!--toc:end-->
+
 ## Draft versions and release branches
 
 The `main` branch is under continuous development and will usually be partway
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,3 +134,23 @@ its configuration and operation.
...
+Janus follows [semantic versioning](https://semver.org/). Because we are at major
+version 0, we increment the minor version number for breaking changes, and
+increment the patch version for new features and backwards-compatible bug fixes.
+
+What is considered a breaking change depends on the crate. The following crates
+are stable on their Rust API:
+- `janus_collector`
+- `janus_client`
+- `janus_messages`
+- `janus_core`
```
inahga, line 149: Note that I've defined `janus_core` and `janus_messages` as stable. I'm not sure that we actually want this though. I would ideally like to say something like
> `janus_core` and `janus_messages` are unstable and should not be directly depended on. If you find yourself needing to depend on these packages in the course of using `janus_client` and `janus_collector`, open a bug report.

i.e. we should reexport types required by `janus_client` and `janus_collector`.

README.md
```diff
@@ -121,3 +134,23 @@ its configuration and operation.
...
+
+The following crates are stable on their external configuration and HTTP API.
```
branlwyd, line 151: IMO, mention that these are unpublished (which helps justify the API instability)

README.md
branlwyd: I think we should say that `janus_messages` is stable: we originally split the `janus_messages` crate out so that one of our partners could use it in their own software, and AFAIK they are still doing so.

I'm agnostic as to whether `janus_core` should be stable -- I do not remember why we decided to publish it to crates.io. Is that information recorded anywhere?

tgeoghegan: `janus_client` and `janus_collector` depend on `janus_core`, so it has to be on crates.io.

tgeoghegan (approved):

Issue #2389, opened 2023-12-21
Implement `draft-ietf-ppm-dap-09`


tgeoghegan:
Janus should implement the latest draft of the standard.

- [x] update to a new libprio-rs that does VDAF-08 (#2477)
- [x] HPKE domain separation strings (#2479)
- [x] [Section 4.1.2](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-09#section-4.1.2-3): "The Leader MUST select a batch that has not yet been associated with a collection job." (#2391)
- [x] [Section 4.1.2](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-09#section-4.1.2-4): `max_batch_size` is now optional in fixed size tasks (#2551)
- [x] [Section 4.5.2.2](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-09#section-4.5.2.2-20): "If for whatever reason the Leader must abandon the aggregation job, it SHOULD send an HTTP DELETE request to the aggregation job URI so that the Helper knows it can clean up its state." (#2530, https://github.com/divviup/public-docs/pull/49)
- [x] [Section 4.6.1](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-09#section-4.6.1-16): Ensure we comply with the new MUST: "If the Leader has a pending aggregation job that overlaps with the batch and aggregation parameter for the collection job, the Leader MUST first complete the aggregation job before proceeding and requesting an aggregate share from the Helper. This avoids a race condition between aggregation and collection jobs that can yield trivial batch mismatch errors." (Already satisfied by Janus; [see comment below](https://github.com/divviup/janus/issues/2389#issuecomment-1900915729)
- [x] update versions table in README.md (#2468)
- [x] Branch for DAP-07 ([`release/0.6`](https://github.com/divviup/janus/tree/release/0.6))
- [x] update divviup-ts integration tests once that component supports -09
- [x] update Daphne integration tests once it supports -09
- [x] Implement [`draft-wang-ppm-dap-taskprov-06`](https://datatracker.ietf.org/doc/draft-wang-ppm-dap-taskprov/06/) (#2598)
- [x] ~Sync up with changes to [`draft-dcook-ppm-dap-interop-test-design`](https://github.com/divergentdave/draft-dcook-ppm-dap-interop-test-design/issues/43)~ No such changes should be needed

[...]

Issue #2254, opened 2023-11-15
Adapters for easy collection


inahga:
Issue transferred from internal discussion, so that we can get community input.

Originally proposed by @branlwyd on Aug 21, 2023
> We should provide some adapters to allow for easy collection of data into commonly-desired formats. This adapter would effectively act as a DAP Collector, and emit aggregates in a format desired by the operator of the adapter. Potential output formats include:
> 
>     * CSV
> 
>     * Prometheus/Grafana stack
> 
>     * Other formats, based on customer needs/feedback
> 
> 
> Note that a DAP Collector is a stateful service; the adapter will need some storage to track both outstanding collection jobs to be polled, as well as collected aggregates to be emitted.
> 
> We may want to consider running these adapters ourselves eventually, to provide a turnkey collection solution for low-complexity users. We should carefully consider the operational & privacy impacts to Divvi Up -- running these adapters ourselves would expose us to aggregates, which may not be acceptable to many subscribers; we would also be on the hook for configuring the adapter & storing data for the adapter, which will simplify the deployment process for subscribers.

inahga:
@branlwyd wrote on Aug 28, 2023

---
Some prior Slack discussion:
* https://isrg.slack.com/archives/C0167LT4C73/p1662059643587269?thread_ts=1662059519.781809&cid=C0167LT4C73
* https://isrg.slack.com/archives/C0167LT4C73/p1666033828750679?thread_ts=1666030501.793389&cid=C0167LT4C73

We decided to prioritize this work for a few reasons:
* This will simplify subscribers' usage of Divvi Up.
* We might eventually offer a managed version of these adapters as part of the Divvi Up service, to further simplify subscribers' usage of Divvi Up (at the cost of revealing the aggregates to us, which some subscribers may not want).
* This would be the first "full" open-source implementation of a Collector, which would be useful as a concrete example for both our practical Divvi Up planning & perhaps to further evolutions of the DAP protocol.

There's a design element to this work, I'm going to sketch a design in this issue next.

inahga:
@branlwyd wrote on Aug 28, 2023

---
Design considerations:
* There are basically two aspects to this "adapter": retrieving aggregates from a DAP Aggregator, and emitting aggregates in the appropriate format.
[...]

PR #2391, opened 2023-12-22
Fixed-size: return each batch to at most one current-batch query.

branlwyd:
Part of DAP-09 implementation (#2389). With this change, Collectors are expected to remember the collection job IDs they generate before sending the collection request. If the Collector encounters a transient failure, it can then resend the request (with the same collection job ID), and rely on request idempotency for things to work out.

The upside of this change is that Collectors no longer need to have logic to deduplicate multiple current-batch requests that are mapped to the same underlying batch, and they can generate current-batch requests at higher rates without risking more duplicated collections.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -2647,15 +2647,13 @@ impl VdafOps {
                             )
                         })?;
 
-                    let (batches, _) = try_join!(
-                        Q::get_batches_for_collection_identifier(
-                            tx,
-                            &task,
-                            collection_job.batch_identifier(),
-                            collection_job.aggregation_parameter()
-                        ),
-                        Q::acknowledge_collection(tx, task.id(), collection_job.batch_identifier()),
-                    )?;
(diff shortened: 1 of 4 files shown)


branlwyd:
(we should wait to merge this until we have created a DAP-07 branch & main is for DAP-09, unless we want to backport this change to DAP-07)

inahga:
I think whether we want to backport depends on whether we want our WIP stateful collector to support DAP-07. This change is necessary for it to be optimally implemented https://github.com/divviup/janus/issues/2254#issuecomment-1834483595.

The changeset looks simple enough to backport (i.e. merge as-is), although I still am giving it a close read.

inahga (approved):

tgeoghegan:
We should take this once #2468 lands

tgeoghegan:
We've now branched for `release/0.6`, meaning that `main` is tracking toward `draft-ietf-ppm-dap-09`. This can be merged (though it needs a rebase).

PR #2419, opened 2024-01-02
Fix Clippy lints

divergentdave:
This fixes some Clippy lints that surfaced with the latest Rust toolchain. Notably, this pointed out that `janus_collector::credential::PrivateCollectorCredential::hpke_keypair()` and `impl TryFrom<DivviUpHpkeConfig> for HpkeKeypair` have unnecessary error handling, ever since we made `janus_core::hpke::HpkeConfig` able to deserialize unsupported algorithms. (I didn't change the method's return value or the trait implemented since they are part of the public API)

### DIFF
--- a/collector/src/credential.rs
+++ b/collector/src/credential.rs
@@ -2,7 +2,7 @@ use crate::Error;
 use hpke_dispatch::{Aead, Kdf, Kem};
 use janus_core::{
     auth_tokens::{AuthenticationToken, BearerToken},
-    hpke::{self, HpkeKeypair, HpkePrivateKey},
+    hpke::{HpkeKeypair, HpkePrivateKey},
 };
 use janus_messages::{HpkeConfig, HpkeConfigId, HpkePublicKey};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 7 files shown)


collector/src/credential.rs
```diff
@@ -29,21 +29,15 @@ impl PrivateCollectorCredential {
         self.token.clone().map(AuthenticationToken::Bearer)
     }
 
-    /// Returns the [`HpkeKeypair`] necessary for decrypting aggregate shares. This errors if the
-    /// configuration references any HPKE parameters that are not supported by the application.
+    /// Returns the [`HpkeKeypair`] necessary for decrypting aggregate shares. This cannot fail
+    /// currently, but returns a `Result` for historical reasons.
     pub fn hpke_keypair(&self) -> Result<HpkeKeypair, Error> {
```
tgeoghegan, line 34: I'd say that we should provide a new, infallible version and mark this one deprecated, but I'm not sure what name we'd use for the new version.

tgeoghegan (approved):

collector/src/credential.rs
divergentdave: I gave it an ugly name, `hpke_keypair_infallible()`, and we can iron it out in 0.7.

tgeoghegan (approved):

Issue #2392, opened 2023-12-22
Automatically run `cargo udeps` on a cronjob


tgeoghegan:
[`cargo-udeps`](https://crates.io/crates/cargo-udeps) can automatically detect unused dependencies in `Cargo.toml`s. We could periodically run it in a GitHub Actions cronjob.

divergentdave:
Done in #2433 and #2436.

PR #2433, opened 2024-01-03
Fix unused dependencies, add periodic workflow to monitor

divergentdave:
This implements #2392.  We had a variety of small issues to clean up before `cargo-udeps` would run cleanly. I think we'll have to merge this PR before we can test the GHA workflow, but I added a workflow_dispatch trigger to speed that up.

I had the workflow create an issue because by default, only the last person to touch a scheduled workflow gets alerted when it fails.

### DIFF
--- a/.github/workflows/unused-dependencies.yml
+++ b/.github/workflows/unused-dependencies.yml
@@ -1,0 +1,70 @@
+name: unused-dependencies
+
+on:
+  schedule:
+    - cron: "21 4 * * mon"
+  workflow_dispatch:
+
+jobs:
+  cargo-udeps:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: "read"
+      issues: "write"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@nightly
+      - name: Install cargo-udeps
+        uses: taiki-e/install-action@cargo-udeps
+      - name: Run cargo-udeps, all features
+        run: cargo +nightly udeps --all-features | tee log_all_features.txt
+      - name: Run cargo-udeps, default features
+        if: success() || failure()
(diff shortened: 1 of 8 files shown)


.github/workflows/unused-dependencies.yml
```diff
@@ -0,0 +1,70 @@
...
+      issues: "write"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@nightly
+      - name: Install cargo-udeps
+        uses: taiki-e/install-action@cargo-udeps
+      - name: Run cargo-udeps, all features
+        run: cargo +nightly udeps --all-features | tee log_all_features.txt
```
inahga, line 22: Hmm, this construct is somewhat suspicious.

Without running the script with `-o pipefail`, this will always return success, e.g.
```bash
#!/usr/bin/env bash
false | cat -
echo $?
```
always outputs `0`, unless `-o pipefail` is set.

It looks like GHA on a Linux runner runs with `-o pipefail` though, IIUC https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell. So no action required, unless it doesn't work.

jbr:
This is really nice, I've never seen an action that opens an issue on failure, so I'll likely be cribbing from this for nightly clippy! In the specific case of udeps, why not run `aig787/cargo-udeps-action@v1` on PR? How could a dependency become unused other than through explicit addition in a PR?

inahga (approved):

divergentdave:
I was hesitant to run this on PRs, because we'd need to cache `.rlib`s built with a nightly toolchain for it to be performant, and we've already got a lot of pressure on our GitHub Actions cache from the other builds. Tim's suggestion to run it on a schedule seemed like a good compromise, because we can run it much less often, move it out of the critical path, and get away with not caching. The impact of missing an unused dependency for a few days is pretty minimal.

PR #2439, opened 2024-01-05
Leader: do not send empty aggregation initialization requests.

branlwyd:
Before this PR, this would happen if every report in the job was filtered (for example, because the report was GC'ed before the aggregation job was sent). DAP did not proscribe how to handle this; a Helper Janus would return an error, causing the aggregation job to become stuck.

Now, instead of sending an empty aggregation initialization request, the Leader sends no request at all and continues processing as if it received an empty response. This will cause the aggregation job to be finished; the Helper will not know about this aggregation job, but that's OK since the aggregation job will not contribute anything to the eventual aggregate.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_driver.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver.rs
@@ -408,31 +408,38 @@ impl AggregationJobDriver {
             }
         }
 
-        // Construct request, send it to the helper, and process the response.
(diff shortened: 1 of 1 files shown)


branlwyd:
Part of #2439.

tgeoghegan (approved):

divergentdave (approved):

inahga (approved):

PR #2440, opened 2024-01-05
Comment cleanup

divergentdave:
This cleans up some comments across the codebase.

- Update old references to draft-gpew-priv-ppm.
- Update references to pre-resource HTTP routes.
- Review TODOs, and delete them if completed or reference a more-specific issue if available.
  - Add a missing divviup-ts integration test for Prio3SumVec.
- Fix rustdoc warnings in private items.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -169,9 +169,11 @@ pub struct Aggregator<C: Clock> {
     task_aggregators: Mutex<HashMap<TaskId, Arc<TaskAggregator<C>>>>,
 
     // Metrics.
-    /// Counter tracking the number of failed decryptions while handling the /upload endpoint.
+    /// Counter tracking the number of failed decryptions while handling the
+    /// `tasks/{task-id}/reports` endpoint.
     upload_decrypt_failure_counter: Counter<u64>,
-    /// Counter tracking the number of failed message decodes while handling the /upload endpoint.
+    /// Counter tracking the number of failed message decodes while handling the
+    /// `tasks/{task-id}/reports` endpoint.
     upload_decode_failure_counter: Counter<u64>,
     /// Counters tracking the number of failures to step client reports through the aggregation
     /// process.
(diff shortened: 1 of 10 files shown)


aggregator/src/aggregator.rs
```diff
@@ -235,14 +237,18 @@ impl<C: Clock> Aggregator<C> {
 
         let upload_decrypt_failure_counter = meter
             .u64_counter("janus_upload_decrypt_failures")
-            .with_description("Number of decryption failures in the /upload endpoint.")
+            .with_description(
+                "Number of decryption failures in the tasks/{task-id}/reports endpoint.",
+            )
```
tgeoghegan, line 242: Will changing the description of Prometheus time series affect aggregation of the metric?

aggregator/src/aggregator.rs
```diff
@@ -2583,7 +2591,7 @@ impl VdafOps {
 
     /// Handle GET requests to a collection job URI obtained from the leader's `/collect` endpoint.
     /// The return value is an encoded `CollectResp<Q>`.
-    /// https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.5.1
+    /// <https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.5.1>
```
tgeoghegan, line 2594: The mention of `/collect` is obsolete, as is the reference to draft 2.

aggregator/src/aggregator.rs
```diff
@@ -2830,7 +2838,7 @@ impl VdafOps {
         Ok(())
     }
 
-    /// Implements the `/aggregate_share` endpoint for the helper, described in §4.4.4.3
+    /// Implements the `tasks/{task-id}/aggregate_shares` endpoint for the helper, described in §4.4.4.3
```
tgeoghegan, line 2841: The section reference is out of date (I would think), let's just delete it
```suggestion
    /// Implements the `tasks/{task-id}/aggregate_shares` endpoint for the helper
```

tgeoghegan (approved):

aggregator/src/aggregator.rs
divergentdave: No, I don't believe so. I think it just shows up in tooltips, etc.

aggregator/src/aggregator.rs
divergentdave: Good catch, I'll do another pass for this.

inahga (approved):

PR #2441, opened 2024-01-06
Redact database URL password

inahga:
Credit to @jbr for the discovery and suggested fix.

### DIFF
--- a/aggregator/src/config.rs
+++ b/aggregator/src/config.rs
@@ -100,7 +100,7 @@ fn format_database_url(url: &Url, fmt: &mut std::fmt::Formatter) -> Result<(), s
     match url.password() {
         Some(_) => {
             let mut url = url.clone();
-            url.set_password(Some("REDACTED")).ok();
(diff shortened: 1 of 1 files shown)


aggregator/src/config.rs
```diff
@@ -94,6 +94,19 @@ impl DbConfig {
     }
 }
 
+/// Makes a best-effort attempt to redact the password from the database URL, so that it is safe
+/// to display in logs.
+fn format_database_url(url: &Url, fmt: &mut std::fmt::Formatter) -> Result<(), std::fmt::Error> {
+    match url.password() {
+        Some(_) => {
+            let mut url = url.clone();
+            url.set_password(Some("REDACTED")).ok();
```
divergentdave, line 103: nit: I think this is the more idiomatic way to ignore a Result
```suggestion
            let _ = url.set_password(Some("REDACTED"));
```

divergentdave (approved):

tgeoghegan (approved):

Issue #2464, opened 2024-01-09
Leader: Batch GC can cause an aggregation job to become unwritable.


branlwyd:
In the presence of GC, an aggregation job can become unwritable if one of the relevant underlying batches is garbage collected. Specifically, the error occurs due to an underflow in `outstanding_aggregation_jobs`.

An example flow reproducing this issue:
1) Aggregation job is created, including a report that references some batch `B`. `B`'s `outstanding_aggregation_jobs` counter is incremented at this point.
2) `B` is GC'ed.
3) The aggregation job is written, for example because of an attempt to step it. The `aggregation_job_writer.rs` logic will not read `B`, will create a new batch with `outstanding_job_count` of 0, and will decrement that count, leading to an underflow to `u64::MAX`.
4) Upon attempting to write the batch, we will attempt to convert `u64::MAX` to an `i64`, which leads to an integer conversion error.

(h/t @inahga for analysis)

branlwyd:
The fix approach discussed so far: augment `get_batch` to return whether the batch is GC'ed, and use that to ignore batch writes to GC'ed batches.

This is tricky: we can determine this for time-interval (since the GC condition is based on the batch interval for the batch, which is used as the primary identifier to the batch); but for fixed-size, the GC condition is based on the maximum client timestamp contained in the batch, which we need to have the batch around to determine.

However, I think we don't need to care for fixed-size: in the fixed-size query type, the GC condition for both batches & aggregation jobs is based on the maximum of the included client timestamp interval, and all reports in an aggregation job are attached to the same batch. This implies that all aggregation jobs in a batch will be GC'ed at or before the batch is GC'ed.

So: I think this bug does not exist in the fixed-size query type, and we can use the discussed approach but only for the time-interval query type. I'm not totally happy with this because it's yet another time-interval-only hack; we're piling more query-type-specific complexity on.

branlwyd:
There may be a similar issue in the Helper codepath (`handle_aggregate_init_generic`, `handle_aggregate_continue_generic` in `aggregator.rs`), which should be investigated; but it's likely less severe since we do not count outstanding aggregation jobs in the Helper.

tgeoghegan:
[...]

PR #2467, opened 2024-01-10
Fix writing aggregation jobs touching GC'ed batches.

branlwyd:
This issue should only exist in the time-interval query type, as fixed-size is arranged such that aggregation jobs touching a given batch must be GC'ed before the batch. I include a guard to ensure that the new codepath is only taken in the expected case of an already-GC'ed batch for a time-interval query, as otherwise we might drop batch writes if we fell into it unexpectedly.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_driver.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver.rs
@@ -1896,6 +1896,314 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn step_time_interval_aggregation_job_init_partially_garbage_collected() {
+        // This is a regression test for https://github.com/divviup/janus/issues/2464.
+        const OLDEST_ALLOWED_REPORT_TIMESTAMP: Time = Time::from_seconds_since_epoch(1000);
+        const REPORT_EXPIRY_AGE: Duration = Duration::from_seconds(500);
+        const TIME_PRECISION: Duration = Duration::from_seconds(10);
+
+        // Setup: insert an "old" and "new" client report, and add them to a new aggregation job.
+        install_test_trace_subscriber();
+        let mut server = mockito::Server::new_async().await;
+        let clock = MockClock::new(OLDEST_ALLOWED_REPORT_TIMESTAMP);
+        let ephemeral_datastore = ephemeral_datastore().await;
+        let ds = Arc::new(ephemeral_datastore.datastore(clock.clone()).await);
+        let vdaf = Arc::new(Prio3::new_count(2).unwrap());
+
+        let task = TaskBuilder::new(QueryType::TimeInterval, VdafInstance::Prio3Count)
+            .with_helper_aggregator_endpoint(server.url().parse().unwrap())
+            .with_report_expiry_age(Some(REPORT_EXPIRY_AGE))
+            .with_time_precision(TIME_PRECISION)
+            .build();
+
+        let leader_task = task.leader_view().unwrap();
+
+        let first_time = OLDEST_ALLOWED_REPORT_TIMESTAMP
+            .sub(&Duration::from_seconds(3 * TIME_PRECISION.as_seconds()))
+            .unwrap()
+            .to_batch_interval_start(&TIME_PRECISION)
+            .unwrap();
+        let first_batch_identifier =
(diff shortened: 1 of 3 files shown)


branlwyd:
Closes #2464. (though I still need to evaluate the Helper codepath, which I should split out to a separate issue if there is work needed there)

I'm not totally happy with this fix as it's something of a hack, but I think it's about the best we can do -- the best other solution discussed was to return an `is_gc'ed` flag from `get_batch`, but we can't accurately compute that flag for the fixed-size query type.

I verified that the new test reproduces the issue by removing the fix & observing the test fail with an underflow error.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -1895,6 +1895,314 @@ mod tests {
...
+        let leader_task = task.leader_view().unwrap();
+
+        let first_time = OLDEST_ALLOWED_REPORT_TIMESTAMP
+            .sub(&Duration::from_seconds(3 * TIME_PRECISION.as_seconds()))
+            .unwrap()
+            .to_batch_interval_start(&TIME_PRECISION)
+            .unwrap();
+        let first_batch_identifier =
+            TimeInterval::to_batch_identifier(&leader_task, &(), &first_time).unwrap();
+        let first_report_metadata = ReportMetadata::new(random(), first_time);
```
tgeoghegan, line 1928: nit: rename `first_time`, `first_batch_identifier` and `first_report_metadata` to indicate the intent, making it more clear to future test maintainers what condition this exercises? For instance, I think the point here is that the `first_*` variables pertain to a GC eligible report while the `second_*` ones are for an uneligible report. So `gc_eligible_time`, `gc_eligible_report_metadata` and `gc_uneligible_time`, `gc_uneligible_report_metadata`, perhaps?

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -1895,6 +1895,314 @@ mod tests {
...
+            .run_unnamed_tx(|tx| {
+                let vdaf = Arc::clone(&vdaf);
+                let leader_task = leader_task.clone();
+                let first_report = first_report.clone();
+                let second_report = second_report.clone();
+
+                Box::pin(async move {
+                    tx.put_aggregator_task(&leader_task).await?;
+                    tx.put_client_report(vdaf.borrow(), &first_report).await?;
+                    tx.put_client_report(vdaf.borrow(), &second_report).await?;
```
tgeoghegan, line 1989: nit: I think it's better to call `unwrap()` here, because then if any of these three `put`s does fail, it's obvious which one it was. If we propagate the error then it blows up when we unwrap the result from `run_unnamed_tx` which is less helpful.

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -332,15 +333,53 @@ impl<const SEED_SIZE: usize, Q: CollectableQueryType, A: vdaf::Aggregator<SEED_S
...
+                        //
+                        // See https://github.com/divviup/janus/issues/2464 for more detail.
+                        if batch_op == Operation::Put {
+                            // Guard to ensure we are in the situation we think we're in:
+                            // `to_batch_interval` returns a value only for the time-interval query
+                            // type; this code effectively checks that the query type is
+                            // time-interval, and the batch interval is past the GC window.
+                            if !Q::to_batch_interval(batch_identifier)
+                                .map(|interval| interval.end() < tx.clock().now())
+                                .unwrap_or(false)
```
tgeoghegan, line 363: This works fine, but we have `self.task` in scope, so can't we more directly check the query type?

tgeoghegan (approved):

aggregator/src/aggregator/aggregation_job_driver.rs
branlwyd: Good call, I've been making this change as I touch code -- it is indeed much better to fail sooner rather than later in tests. I missed these as I adapted a lot of this code from a previously-existing test.

aggregator/src/aggregator/aggregation_job_writer.rs
branlwyd: I suppose we should check both -- I also want to guard against a time-interval task unexpectedly not finding a batch that shouldn't be GC'ed.

PR #2469, opened 2024-01-10
Fix `PrepareRespResult` wire encoding

### LINKED ISSUE #2466: PrepareResp format does not match DAP draft-07
The serialization and deserialization routines for [`PrepareResp`](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-07.html#section-4.5.1.2-5) fail to include a 4-byte length prefix before the payload, if the state is equal to `continue`.

Given that Janus 0.6 is already deployed, but not yet interoperating with any other aggregators, I think it would make sense to not fix this on our draft-07 branch, and instead fix it when we implement draft-09. We expect to do interoperation testing against Daphne on the draft-09 version of the protocol.

tgeoghegan:
Include the 4 byte length prefix when encoding ping-pong messages into a `PrepareRespResult`.

Closes #2466

### DIFF
--- a/messages/src/lib.rs
+++ b/messages/src/lib.rs
@@ -2280,7 +2280,8 @@ impl Encode for PrepareStepResult {
         match self {
             Self::Continue { message: prep_msg } => {
                 0u8.encode(bytes);
-                prep_msg.encode(bytes);
+                let encoded_prep_msg = prep_msg.get_encoded();
+                encode_u32_items(bytes, &(), &encoded_prep_msg);
             }
             Self::Finished => 1u8.encode(bytes),
             Self::Reject(error) => {
@@ -2292,7 +2293,7 @@ impl Encode for PrepareStepResult {
 
     fn encoded_len(&self) -> Option<usize> {
         match self {
-            Self::Continue { message: prep_msg } => Some(1 + prep_msg.encoded_len()?),
+            Self::Continue { message: prep_msg } => Some(1 + 4 + prep_msg.encoded_len()?),
             Self::Finished => Some(1),
             Self::Reject(error) => Some(1 + error.encoded_len()?),
         }
@@ -2304,7 +2305,8 @@ impl Decode for PrepareStepResult {
         let val = u8::decode(bytes)?;
         Ok(match val {
             0 => {
-                let prep_msg = PingPongMessage::decode(bytes)?;
+                let prep_msg_bytes = decode_u32_items(&(), bytes)?;
+                let prep_msg = PingPongMessage::get_decoded(&prep_msg_bytes)?;
                 Self::Continue { message: prep_msg }
             }
             1 => Self::Finished,
@@ -2900,7 +2902,8 @@ where
         let encoding = hex::decode(hex_encoding).unwrap();
         assert_eq!(
             encoded_val, encoding,
-            "Couldn't roundtrip (encoded value differs): {val:?}"
+            "Couldn't roundtrip (encoded value differs): {val:?}
+encoded:\t\t{encoded_val:02x?}\nexpected encoding:\t{encoding:02x?}"
         );
         let decoded_val = T::decode(&mut Cursor::new(&encoded_val)).unwrap();
         assert_eq!(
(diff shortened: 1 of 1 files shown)


messages/src/lib.rs
```diff
@@ -2900,7 +2902,8 @@ where
         let encoding = hex::decode(hex_encoding).unwrap();
         assert_eq!(
             encoded_val, encoding,
-            "Couldn't roundtrip (encoded value differs): {val:?}"
+            "Couldn't roundtrip (encoded value differs): {val:?}
+encoded:\t\t{encoded_val:02x?}\nexpected encoding:\t{encoding:02x?}"
```
tgeoghegan, line 2906: On failure, `assert_eq!` will print out the mismatched values as vectors of unpadded decimals, which was driving me nuts because the message encoding tests are written in padded (i.e. each byte is two characters) _hex_ vectors. So you had to convert from decimal to hex, and also the good/bad vectors wouldn't line up vertically. On failure, we now get this output:

```
---- tests::roundtrip_aggregation_job_resp stdout ----
thread 'tests::roundtrip_aggregation_job_resp' panicked at messages/src/lib.rs:2903:9:
assertion `left == right` failed: Couldn't roundtrip (encoded value differs): AggregationJobResp { prepare_resps: [PrepareResp { report_id: ReportId(AQIDBAUGBwgJCgsMDQ4PEA), result: Continue }, PrepareResp { report_id: ReportId(EA8ODQwLCgkIBwYFBAMCAQ), result: Finished }] }
encoded:		[00, 00, 00, 39, 01, 02, 03, 04, 05, 06, 07, 08, 09, 0a, 0b, 0c, 0d, 0e, 0f, 10, 00, 00, 00, 00, 13, 01, 00, 00, 00, 05, 30, 31, 32, 33, 34, 00, 00, 00, 05, 35, 36, 37, 38, 39, 10, 0f, 0e, 0d, 0c, 0b, 0a, 09, 08, 07, 06, 05, 04, 03, 02, 01, 01]
expected encoding:	[00, 00, 00, 35, 01, 02, 03, 04, 05, 06, 07, 08, 09, 0a, 0b, 0c, 0d, 0e, 0f, 10, 00, 00, 00, 00, 13, 01, 00, 00, 00, 05, 30, 31, 32, 33, 34, 00, 00, 00, 05, 35, 36, 37, 38, 39, 10, 0f, 0e, 0d, 0c, 0b, 0a, 09, 08, 07, 06, 05, 04, 03, 02, 01, 01]
  left: [0, 0, 0, 57, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 0, 0, 0, 0, 19, 1, 0, 0, 0, 5, 48, 49, 50, 51, 52, 0, 0, 0, 5, 53, 54, 55, 56, 57, 16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 1]
 right: [0, 0, 0, 53, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 0, 0, 0, 0, 19, 1, 0, 0, 0, 5, 48, 49, 50, 51, 52, 0, 0, 0, 5, 53, 54, 55, 56, 57, 16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 1]


failures:
    tests::roundtrip_aggregation_job_resp

test result: FAILED. 44 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
```

The left/right lines are redundant, but at least now you can see the hex bytes.

tgeoghegan:
I think we have the same problem in `struct PrepareInit` and `struct PrepareContinue`. I will fix that here as well.

messages/src/lib.rs
divergentdave: FWIW, we could get rid of the duplication by either using [hex_display](https://docs.rs/hex-display/latest/hex_display/) (which would get rid of the commas/spaces too) or with a similar newtype that calls `f.write_fmt(format_args!("{:02x?}", self.0)` in its `Debug` implementation.

divergentdave (approved):

messages/src/lib.rs
tgeoghegan: For me, having the bytes spaced out this way makes it easier to spot differences. I suppose the ideal would be some formatter or library that could diff them and then highlight the bytes that differ, maybe in colour, but I don't want to investigate that in this PR.

divergentdave: Sounds good, I'll open an issue with a proposal.

branlwyd (approved):

PR #2474, opened 2024-01-11
Log version information on startup

divergentdave:
This adds version information to the "Starting up" log messages, in addition to the `target_info` metric.

Sample output:

```
..."version":"0.6.9","git_revision":"0f7c650e-modified","rust_version":"1.75.0"...
```

### DIFF
--- a/aggregator/Cargo.toml
+++ b/aggregator/Cargo.toml
@@ -17,14 +17,12 @@ default = []
 fpvec_bounded_l2 = ["dep:fixed", "janus_core/fpvec_bounded_l2"]
 tokio-console = ["dep:console-subscriber"]
 otlp = [
-    "dep:git-version",
     "dep:opentelemetry-otlp",
     "dep:opentelemetry-semantic-conventions",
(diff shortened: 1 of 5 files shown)


aggregator/Cargo.toml
```diff
@@ -52,7 +50,7 @@ deadpool-postgres = "0.12.1"
 derivative.workspace = true
 fixed = { version = "1.24", optional = true }
 futures = "0.3.30"
-git-version = { version = "0.3.9", optional = true }
+git-version = "0.3.9"
```
inahga, line 53: Since we're only using a macro out of `git-version`, can it be a dev dependency?

inahga (approved):

aggregator/Cargo.toml
divergentdave: Yeah, no dice. I think the idea may be that `macro_rules!` must ride along in `.rlib` files, and thus we need Cargo to always build it and pass it along as a dependency to rustc.

tgeoghegan (approved):

PR #2476, opened 2024-01-11
Abandon collection jobs early when a fatal error is encountered

inahga:
This sets a collection job to immediately abandon on failure, except in a narrow set of retryable errors.

The `is_retryable_error()` function is relatively narrow. It basically only retries on particular HTTP status codes and database errors. It's tricky to enumerate all possible retryable rrors. I'm welcome to input if there are other errors that should be retried.

Note that it retries on 500 errors. I don't really want it to, but this is done to share the logic with the other HTTP retry logic we have. I will address whether we should generally retry on 500 errors in a future PR.

I will give the same treatment to aggregation jobs in another PR (see https://github.com/divviup/janus/issues/235)

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_driver.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver.rs
@@ -1500,7 +1500,7 @@ mod tests {
         assert_matches!(
             error.downcast().unwrap(),
             Error::Http(error_response) => {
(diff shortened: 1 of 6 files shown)


tgeoghegan (approved):
You'll need to backport this to `release/0.6`.

PR #2483, opened 2024-01-12
Include batch details in batch mismatch error

inahga:
This should make investigation of a BatchMismatch error a bit easier, by propagating error information from the helper into the leader. This would be to mainly to spot trends in report counts or checksums.

This also comes with the happy extension of supporting arbitrary fields in a `ProblemDocument` and slightly more test coverage on the helper's checksum/report count calculation.

### DIFF
--- a/aggregator/src/aggregator/http_handlers.rs
+++ b/aggregator/src/aggregator/http_handlers.rs
@@ -108,7 +108,8 @@ impl Handler for Error {
             ),
             Error::BatchMismatch(inner) => conn.with_problem_document(
                 &ProblemDocument::new_dap(DapProblemType::BatchMismatch)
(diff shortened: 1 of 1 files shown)


divergentdave:
FWIW, I think this extra information in JSON bodies would go against RFC 7807, as "extension members" are determined by the problem type definition, and DAP only defines the "task_id" extension member. It would be safe to template more information into the "detail" member. (It's possible this has changed in the more recent "problem details-bis" RFC, but I haven't looked over that yet, nor has DAP updated its reference.)

inahga:
Ah I think you're right--and this hasn't changed in RFC 9457. I will template the info into the `detail` field.

tgeoghegan (approved):

PR #2502, opened 2024-01-16
Abandon aggregation jobs early when a fatal error is encountered

inahga:
Closes https://github.com/divviup/janus/issues/235. Similar to https://github.com/divviup/janus/pull/2476.

I've duplicated the `is_retryable_error()` function because I strongly suspect that the interesting errors are going to differ between the two jobs.

I've removed usage of `anyhow::Error` from the aggregation job driver. While we can `downcast()` an `anyhow::Error`, it seemed cleaner just to go ahead and use our fully fleshed out `aggregator::Error` instead.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_driver.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver.rs
@@ -935,7 +935,7 @@ impl AggregationJobDriver {
                             // that's the more interesting error for debugging purposes.
                             //
                             // If cancellation fails, the job will be picked up again. This isn't
-                            // a big deal, since stepping a collection job is idempotent. It would
+                            // a big deal, since stepping an aggregation job is idempotent. It would
                             // just be some wasted work next time around.
                             warn!(
                                 %attempts,
--- a/aggregator/src/aggregator/aggregation_job_driver.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver.rs
@@ -4164,7 +4164,13 @@ mod tests {
             .unwrap();
         let batch_identifier = TimeInterval::to_batch_identifier(&leader_task, &(), &time).unwrap();
         let report_metadata = ReportMetadata::new(random(), time);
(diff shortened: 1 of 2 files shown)


aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -899,29 +900,80 @@ impl AggregationJobDriver {
         self: Arc<Self>,
         datastore: Arc<Datastore<C>>,
         maximum_attempts_before_failure: usize,
-    ) -> impl Fn(Lease<AcquiredAggregationJob>) -> BoxFuture<'static, Result<(), anyhow::Error>>
-    {
+    ) -> impl Fn(Lease<AcquiredAggregationJob>) -> BoxFuture<'static, Result<(), Error>> {
```
inahga, line 903: This function is basically the same as its collector counterpart, which makes it a candidate for hoisting out into the `JobDriver`, however I think it's just under the threshold where that refactor would be worth it, in terms of how much code needs to be touched.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -899,29 +900,80 @@ impl AggregationJobDriver {
...
+                        if !Self::is_retryable_error(&error) {
+                            // Make a best-effort attempt to immediately cancel the aggregation job.
+                            // on fatal errors. This protects the helper from performing wasted
+                            // work.
+                            //
+                            // Cancellation might fail, but we will return the first error, since
+                            // that's the more interesting error for debugging purposes.
+                            //
+                            // If cancellation fails, the job will be picked up again. This isn't
+                            // a big deal, since stepping a collection job is idempotent. It would
```
branlwyd, line 938:
```suggestion
                            // a big deal, since stepping an aggregation job is idempotent. It would
```

branlwyd (approved):

Issue #2442, opened 2024-01-06
Leader: GC can cause aggregation initialization to send different requests on retries.


branlwyd:
The Leader aggregation logic will prune reports from an aggregation job before sending the aggregation initialization request for several reasons, including an inability to load the report (typically due to GC).

If the first attempt to send an initialization request fails, on retry, the report-pruning logic will be re-run. Since GC logic is based on wallclock time, this means that the set of pruned reports might be different on retry. The Helper may have successfully processed the request; if so, the Helper will notice the difference and fail the request for idempotency reasons.

We must be able to send the same request on retry. The suggested fix is to move the report data into the aggregation job (or more specifically, the report aggregations) before sending the first request; I think the easiest time to do this would be during aggregation job creation. [If that doesn't work out, another approach would be to mark reports which are "in-process" as un-GC'able until the first aggregation step is successfully completed. But I think that would likely be more complicated.]

branlwyd:
The two PRs above effectively fix this issue, but I'm leaving it open to track the work necessary to backport. The backported version of this functionality will need to support both the report-data-in-`client_reports` and the report-data-in-`report_aggregations` cases.

PR #2516, opened 2024-01-22
Aggregation: repeat same request on retries.

branlwyd:
This is implemented by copying the necessary report information into the report aggregations at time of aggregation job creation. This means that client-report GC will not change the request that is made.

I plan to follow this up with a PR to move report scrubbing from the end of the first aggregation step to aggregation job creation. This would have been included in this PR, but this PR is already quite large.

See individual commits for more detail in some cases--aggregation job creation & driving test cases are enhanced somewhat.

Part of #2442.


inahga (approved):
Looks good. I think storing the report data at the time of job creation is the right idea.

db/00000000000001_initial_schema.up.sql
```diff
@@ -226,11 +226,23 @@ CREATE TABLE report_aggregations(
...
+    helper_encrypted_input_share  BYTEA,  -- encoded HPKE ciphertext of helper input share (opaque DAP message)
+
+    -- Additional data for state WaitingLeader.
+    leader_prep_transition  BYTEA,  -- the current VDAF prepare transition (opaque VDAF message)
+
+    -- Additional data for state WaitingHelper.
+    helper_prep_state  BYTEA,  -- the current VDAF prepare state (opaque VDAF message)
+
+    -- Additional data for state Failed.
+    error_code  SMALLINT,  -- error code corresponding to a DAP ReportShareError value
```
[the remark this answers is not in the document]
branlwyd: Indeed. The database columns aren't such a big deal (I reorganized in this PR, but the actual changes are just new columns), but I will need to implement the ability to handle both old-style (report data in `client_reports`) and new-style (report data in `report_aggregations`) aggregations, which will be novel to the backport.

aggregator_core/src/test_util/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+
+/// Returns a [`LeaderStoredReport`] with the given task ID & metadata values and encrypted
+/// input shares corresponding to the given HPKE configs & input shares.
+pub fn generate_report<const SEED_SIZE: usize, V>(
+    task_id: TaskId,
+    report_metadata: ReportMetadata,
+    helper_hpke_config: &HpkeConfig,
+    extensions: Vec<Extension>,
+    transcript: &VdafTranscript<SEED_SIZE, V>,
+) -> LeaderStoredReport<SEED_SIZE, V>
```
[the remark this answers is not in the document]
branlwyd: Good idea -- the only (very minor) downside is some of the types are harder to name due to import issues with the `test-util` feature, but the overall readability win is undeniable.

aggregator_core/src/datastore/models.rs
```diff
@@ -774,41 +838,54 @@ where
...
+#[derive(Clone, Derivative)]
+#[derivative(Debug)]
 pub enum ReportAggregationState<const SEED_SIZE: usize, A: vdaf::Aggregator<SEED_SIZE, 16>> {
-    Start,
-    WaitingLeader(
+    StartLeader {
+        /// Public share for this report.
+        public_share: A::PublicShare,
+        /// The sequence of extensions from the Leader's input share for this report.
+        leader_extensions: Vec<Extension>,
```
tgeoghegan, line 857: I think we will want to `Debug = "ignore"` these fields as well. It might be nice if the `Debug` impl on `Extension` printed the extension type but not the contents, but that's not needed in this change.

db/00000000000001_initial_schema.up.sql
```diff
@@ -226,11 +226,23 @@ CREATE TABLE report_aggregations(
...
-    leader_prep_transition BYTEA,                           -- the current VDAF prepare transition (opaque VDAF message, only if in state WAITING, only populated for leader)
-    error_code          SMALLINT,                           -- error code corresponding to a DAP ReportShareError value; null if in a state other than FAILED
     last_prep_resp      BYTEA,                              -- the last PrepareResp message sent to the Leader, to assist in replay (opaque DAP message, populated for Helper only)
+    state               REPORT_AGGREGATION_STATE NOT NULL,  -- the current state of this report aggregation
+
+    -- Additional data for state StartLeader.
+    public_share                  BYTEA,  -- the public share for the report (opaque VDAF message)
+    leader_extensions             BYTEA,  -- encoded sequence of Extension messages from Leader input share (opaque DAP messages)
+    leader_input_share            BYTEA,  -- encoded leader input share (opaque VDAF message)
+    helper_encrypted_input_share  BYTEA,  -- encoded HPKE ciphertext of helper input share (opaque DAP message)
```
tgeoghegan, line 236: WDYT about writing some `CONSTRAINT`s to ensure that the columns are non-NULL or NULL as expected given the value of `state`, akin to the ones on table `tasks` that enforce teh presence or absence of auth tokens?

tgeoghegan (approved):

db/00000000000001_initial_schema.up.sql
branlwyd: I'd like to keep as much functionality in our application servers (which are horizontally scalable) rather than our DB (which is a monolith which must scale vertically). That is, I prefer to keep these checks in our datastore code rather than encoding them in our DB directly, especially for something like `report_aggregations` which constitutes a large fraction of our total writes.

That said, this is based on ~vibes~; if you feel strongly I can change things.

tgeoghegan: I don't have a particularly strong argument for doing it in the database vs. code.

PR #2530, opened 2024-01-23
leader deletes abandoned aggregation job in helper

tgeoghegan:
Per DAP-09, when the leader gives up on an aggregation job, it SHOULD delete that job in the helper. Our leader (specifically `aggregation_job_driver`) now does this, and we wire up a simple handler for deleting aggregation jobs on the helper side. This handler does nothing but mark the relevant row in the aggregations job as deleted. This could eventually be used by Janus garbage collection to more aggressively evict data before its expiry.

Part of #2389

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -2413,7 +2413,7 @@ impl VdafOps {
         A::OutputShare: Send + Sync,
     {
         Ok(datastore
-            .run_tx("delete aggregation job", |tx| {
+            .run_tx("delete_aggregation_job", |tx| {
                 let (task_id, aggregation_job_id) = (*task.id(), *aggregation_job_id);
                 Box::pin(async move {
                     let aggregation_job = tx
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -475,6 +475,41 @@ impl<C: Clock> Aggregator<C> {
             .await
     }
 
+    async fn handle_aggregate_delete(
+        &self,
+        task_id: &TaskId,
+        aggregation_job_id: &AggregationJobId,
+        auth_token: Option<AuthenticationToken>,
+        taskprov_task_config: Option<&TaskConfig>,
+    ) -> Result<(), Error> {
+        let task_aggregator = self
+            .task_aggregator_for(task_id)
+            .await?
+            .ok_or(Error::UnrecognizedTask(*task_id))?;
+        if task_aggregator.task.role() != &Role::Helper {
+            return Err(Error::UnrecognizedTask(*task_id));
+        }
+
+        if self.cfg.taskprov_config.enabled && taskprov_task_config.is_some() {
+            self.taskprov_authorize_request(
+                &Role::Leader,
+                task_id,
+                taskprov_task_config.unwrap(),
+                auth_token.as_ref(),
+            )
+            .await?;
+        } else if !task_aggregator
+            .task
+            .check_aggregator_auth_token(auth_token.as_ref())
+        {
+            return Err(Error::UnauthorizedRequest(*task_id));
+        }
+
+        task_aggregator
+            .handle_aggregate_delete(&self.datastore, aggregation_job_id)
+            .await
+    }
+
     /// Handle a collection job creation request. Only supported by the leader. `req_bytes` is an
     /// encoded [`CollectionReq`].
     async fn handle_create_collection_job(
@@ -993,6 +1028,16 @@ impl<C: Clock> TaskAggregator<C> {
             .await
     }
 
+    async fn handle_aggregate_delete(
+        &self,
+        datastore: &Datastore<C>,
+        aggregation_job_id: &AggregationJobId,
+    ) -> Result<(), Error> {
+        self.vdaf_ops
+            .handle_aggregate_delete(datastore, Arc::clone(&self.task), aggregation_job_id)
+            .await
+    }
+
     async fn handle_create_collection_job(
         &self,
         datastore: &Datastore<C>,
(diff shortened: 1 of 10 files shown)


aggregator/src/aggregator.rs
```diff
@@ -465,6 +465,41 @@ impl<C: Clock> Aggregator<C> {
...
+        auth_token: Option<AuthenticationToken>,
+        taskprov_task_config: Option<&TaskConfig>,
+    ) -> Result<(), Error> {
+        let task_aggregator = self
+            .task_aggregator_for(task_id)
+            .await?
+            .ok_or(Error::UnrecognizedTask(*task_id))?;
+        if task_aggregator.task.role() != &Role::Helper {
+            return Err(Error::UnrecognizedTask(*task_id));
+        }
```
tgeoghegan, line 481: Like the other aggregation job request paths, DELETE is only supported on the helper. We could also allow it on the leader, since it also has an `aggregation_jobs` table with rows that could be marked deleted. That could be useful in the context of job queue management which we have discussed in broad terms recently. However I'd prefer not to support DELETE on the leader until we have a more coherent plan for what to use that for.

aggregator/src/aggregator.rs
```diff
@@ -3164,24 +3308,31 @@ async fn send_request_to_helper<T: Encode>(
     method: Method,
     url: Url,
     route_label: &'static str,
-    content_type: &str,
-    request: T,
+    request_body: Option<RequestBody<T>>,
```
tgeoghegan, line 3311: The `DELETE` request to aggregation jobs has no body, so I needed to make those parameters optional. I combined them into `struct RequestBody` so that (1) we could have a single `Option<>` in the argument list instead of separate `Option<&str>` and `Option<T>` which need to be `Some` or `None` in lockstep and (2) used a struct instead of a tuple for readability.

aggregator/src/aggregator/error.rs
```diff
@@ -63,15 +63,18 @@ pub enum Error {
     /// An attempt was made to act on an unknown aggregation job.
     #[error("task {0}: unrecognized aggregation job: {1}")]
     UnrecognizedAggregationJob(TaskId, AggregationJobId),
+    /// An attempt was made to act on a known but deleted aggregation job.
+    #[error("deleted aggregation job: {0}")]
+    DeletedAggregationJob(TaskId, AggregationJobId),
     /// An attempt was made to act on an unknown collection job.
     #[error("unrecognized collection job: {0}")]
-    UnrecognizedCollectionJob(CollectionJobId),
+    UnrecognizedCollectionJob(TaskId, CollectionJobId),
```
tgeoghegan, line 71: Noting that the `AggregationJob` error variants include `TaskId`, I opted to augment the `CollectionJob` errors to match.

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -68,25 +68,39 @@ impl Handler for Error {
...
+            Error::UnrecognizedAggregationJob(task_id, aggregation_job_id) => conn
+                .with_problem_document(
+                    &ProblemDocument::new_dap(DapProblemType::UnrecognizedAggregationJob)
+                        .with_task_id(task_id)
+                        .with_aggregation_job_id(aggregation_job_id),
+                ),
+            Error::DeletedAggregationJob(task_id, aggregation_job_id) => conn
+                .with_problem_document(
+                    &ProblemDocument::new(
+                        "https://docs.divviup.org/references/janus-errors#aggregation-job-deleted",
```
tgeoghegan, line 80: I'll make a change to `divviup/public-docs` so that this link resolves to something.

aggregator/src/aggregator.rs
```diff
@@ -2293,6 +2382,49 @@ impl VdafOps {
...
+        A: 'static + Send + Sync,
+        A::AggregationParam: Send + Sync,
+        A::AggregateShare: Send + Sync,
+        for<'a> A::PrepareState: Send + Sync + Encode + ParameterizedDecode<(&'a A, usize)>,
+        A::PrepareShare: Send + Sync,
+        A::PrepareMessage: Send + Sync,
+        A::OutputShare: Send + Sync,
+    {
+        Ok(datastore
+            .run_tx("delete aggregation job", |tx| {
```
inahga, line 2406:
```suggestion
            .run_tx("delete_aggregation_job", |tx| {
```
nit: consistent with our other tx names.

inahga (approved):
Nicely done.

branlwyd (changes requested):
AFAICT, there is nothing stopping our Helper implementation from stepping a Deleted (or Abandoned, but this is impossible) aggregation job. Now that the Helper can have deleted/abandoned aggregation jobs, I think we should not allow stepping these jobs.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -3703,6 +3738,122 @@ mod tests {
         assert_eq!(want_collection_job, got_collection_job);
     }
 
+    struct CancelAggregationJobTestCase {
```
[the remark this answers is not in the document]
tgeoghegan: Because I was also going to use it in another place, but I forgot to rewrite `cancel_aggregation_job`. Thanks!

aggregator_core/src/datastore/models.rs
```diff
@@ -451,6 +451,8 @@ pub enum AggregationJobState {
     Finished,
     #[postgres(name = "ABANDONED")]
     Abandoned,
+    #[postgres(name = "DELETED")]
```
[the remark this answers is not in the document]
tgeoghegan: This matches the [`COLLECTION_JOB_STATE` enum](https://github.com/divviup/janus/blob/2b64b2a13fd1fa3e4173ac56cab4dd0a900218ed/db/00000000000001_initial_schema.up.sql#L305-L311). My thinking is that if and when we take job deletion into account in garbage collection, we might want to distinguish between abandoned and deleted. For instance, maybe we want to aggressively clean up deleted jobs, but want to keep abandoned ones around for investigation, since abandonment indicates an unusual condition whereas deletion is more routine.

tgeoghegan:
> AFAICT, there is nothing stopping our Helper implementation from stepping a Deleted (or Abandoned, but this is impossible) aggregation job. Now that the Helper can have deleted/abandoned aggregation jobs, I think we should not allow stepping these jobs.

`handle_aggregate_continue_generic` checks for the deleted state [here](https://github.com/divviup/janus/blob/7a9b9c01a2f8e718938d631762b217742a0edcf8/aggregator/src/aggregator.rs#L2313-L2319). The check on the aggregate init path is hidden in [`check_aggregation_job_idempotence`](https://github.com/divviup/janus/blob/7a9b9c01a2f8e718938d631762b217742a0edcf8/aggregator/src/aggregator.rs#L1699-L1706) (because that's where we've got the existing aggregation job, if any loaded and in scope).

aggregator/src/aggregator/http_handlers.rs
tgeoghegan: https://github.com/divviup/public-docs/pull/49

branlwyd (approved):

PR #2544, opened 2024-01-24
Trillium security patches

jbr:
Context:
* [github security advisory](https://github.com/trillium-rs/trillium/security/advisories/GHSA-9f9p-cp3c-72jf)
* [rustsec advisory for client](https://rustsec.org/advisories/RUSTSEC-2024-0008)
* [rustsec advisory for server](https://rustsec.org/advisories/RUSTSEC-2024-0009)

rebased on #2542

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -134,7 +134,7 @@ jobs:
     - name: Document
       run: cargo doc --profile ci --workspace
     - name: cargo-deny
-      uses: EmbarkStudios/cargo-deny-action@v1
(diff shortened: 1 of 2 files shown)


branlwyd (approved):

tgeoghegan:
We already bumped to these versions via dependabot ([`trillium-client`](https://github.com/divviup/janus/blob/75c2325398dacfd250e959997dedf0da5030712e/Cargo.lock#L5310), [`trillium-http`](https://github.com/divviup/janus/blob/75c2325398dacfd250e959997dedf0da5030712e/Cargo.lock#L5342)) so I think we are good here.

PR #2551, opened 2024-01-26
Fixed size: make `max_batch_size` optional

tgeoghegan:
`draft-ietf-ppm-dap-09` makes the `max_batch_size` parameter for fixed-size query tasks optional. If it's absent, we aim to create batches of exactly `min_batch_size` reports.

Part of #2389


aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -2589,6 +2589,187 @@ mod tests {
...
+                            .await,
+                        ))
+                    })
+                })
+                .await
+                .unwrap();
+
+        // Verify outstanding batches.
+        let mut total_max_size = 0;
+        println!("{outstanding_batches:?}");
```
inahga, line 2723:
```suggestion
```
Straggler println.

inahga (approved):

divergentdave (approved):

PR #2554, opened 2024-01-27
[0.6] DB migrations for "Aggregation: repeat same request on retries."

branlwyd:


branlwyd:
(The DB migration is split from the functionality to allow it to be released first -- to ease rollbacks, the new columns will be added in a release before the functionality change.)

branlwyd:
oops, wrong base, remake is easier than rebase

PR #2555, opened 2024-01-27
[0.6] DB migrations for "Aggregation: repeat same request on retries."

branlwyd:
Backport of DB changes of #2516.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -99,7 +99,7 @@ macro_rules! supported_schema_versions {
 // version is seen, [`Datastore::new`] fails.
 //
 // Note that the latest supported version must be first in the list.
-supported_schema_versions!(2, 1);
+supported_schema_versions!(3, 2, 1);
 
 /// Datastore represents a datastore for Janus, with support for transactional reads and writes.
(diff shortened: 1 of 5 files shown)


branlwyd:
(This is split from the functionality PR as it should go out on a release prior to the functionality release.)

Part of #2442.

branlwyd:
Any idea what's doing on with the `janus_migrations` check? When I run the relevant bash snippet (below) on my copy of the repo, I do not encounter an error.

```
  for DIRECTION in {down,up}; do
    SEQUENCE=$(find db/ -name "*.$DIRECTION.sql" -printf "%f\n" | cut -d_ -f1 | sed 's/^0*//' | sort -n)
    if [ "$(seq 1 "$(wc -l <<<"$SEQUENCE")")" != "$SEQUENCE" ]; then
      echo "fatal: SQL migrations must be numbered sequentially"
      exit 1
    fi
  done
```

divergentdave:
The release branch already has a migration numbered 2, which explains why the check fails on the merge commit. This'll need to be rebased, and the migration will need to be numbered 3.

inahga (approved):

tgeoghegan (approved):

Issue #2462, opened 2024-01-08
Don't log serialization errors


inahga:
Serialization errors are a normal part of Janus operations and will be retried automatically. We shouldn't log them at ERROR level.

inahga:
Upon some experimentation, I think this is best addressed operationally, i.e. outside of Janus.

There's a few problems with doing this in Janus.
- `tracing::instrument` doesn't seem to have any ability to change the log level based on the exact error. For example, I might want `Db::MutationTargetAlreadyExists` to be logged at `Level::DEBUG`, while `Db::DbState` should probably be at Error.
- `tracing_subscriber` has filters, but I don't see a way to introspect the actual content of a trace. Either way, this is hacky, since it involves saying `if log.contains("serialization error") then filter out` for each log line.
- We can demote all Janus datastore errors to DEBUG level, but I think this risks us losing out on log context that has proven valuable in some production debug.
- `tokio-postgres::Error` doesn't appear to differentiate between serialization errors and other unspecified database errors.

inahga:
I raised a question https://github.com/tokio-rs/tracing/discussions/2849 to solicit some ideas.

divergentdave:
I think the cleanest answer would be to write our own `instrument` attribute macro, possibly wrapping `tracing::instrument`. We could then write an if-else that either calls `error!` or `debug!`. The various tracing event macros set up a static variable each place they are called, and one of the things it tracks is whether the callsite is always enabled, always disabled, or sometimes enabled, based on the subscriber's filter. Thus, we really need to create two callsites if we want to dynamically choose the severity level.

branlwyd:
There are quite a few errors of this form, too:

`DB error: db error: ERROR: current transaction is aborted, commands ignored until end of transaction block`

This can happen naturally in the case of any error or retry since we use Postgres' pipelined query mode to send many queries concurrently; whatever we do for serialization errors should probably be applied here, too, since these errors do not necessarily indicate a "real" error.

inahga:
This is still a problem, and somewhat annoying, but we're not going to prioritize it at this time, especially since it's tricky to at runtime.

PR #2557, opened 2024-01-27
Downgrade various logs to `DEBUG` level

tgeoghegan:
The Trillium handler for errors was unconditionally logging them at `WARN`, generating large volumes of logging for problems that aren't Janus' fault, like reports whose timestamps are in the future. Now, we check whether the respones indicates a server error before bothing to log it (keeping in mind that client errors will still appear in metrics).

Additionally, we've made liberal use of the [`tracing::instrument`][1] attribute macro to decorate various functions and methods with tracing spans, and in particular its `err` argument for logging anytime a function that returns `Result<T, E>` returns an `Err`. Mostly, logging those errors is handled at a higher level -- say, in a Trillium handler or in the top-level loop of `aggregation_job_driver` -- so logging them at level `ERROR` at the function call itself isn't that helpful. This commit further qualifies the `err` argument so that those errors are now logged at `DEBUG`, so we can opt back into them should they prove useful.

[1]: https://docs.rs/tracing/latest/tracing/attr.instrument.html

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -699,7 +699,7 @@ impl<C: Clock> Aggregator<C> {
     }
 
     /// Opts in or out of a taskprov task.
-    #[tracing::instrument(skip(self, aggregator_auth_token), err)]
+    #[tracing::instrument(skip(self, aggregator_auth_token), err(level = Level::DEBUG))]
     async fn taskprov_opt_in(
         &self,
         peer_role: &Role,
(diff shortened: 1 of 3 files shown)


tgeoghegan:
This will need backport to `release/0.6` should we go ahead with it.

inahga:
I think this will partially, if not wholly, address https://github.com/divviup/janus/issues/2462.

inahga (approved):
It's unfortunate to lose datastore logging because I think it does reveal some interesting information for troubleshooting, but I think it means we should be more aggressive about turning up the log level at runtime when we need to debug.

divergentdave (approved):

PR #2559, opened 2024-01-29
[0.6] Scrub reports during aggregation job creation, rather than the initial aggregation step.

branlwyd:
Other than being the earliest point that we can scrub reports, this avoids (briefly) duplicating report data in both the `client_reports` and `report_aggregations` tables. I think placing this logic in the aggregation job creator also makes more sense for when we'll eventually expand it to handle VDAFs requiring multiple aggregations per report.

Backport of #2547. Stacked on #2556.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -134,7 +134,7 @@ jobs:
     - name: Document
       run: cargo doc --profile ci --workspace
     - name: cargo-deny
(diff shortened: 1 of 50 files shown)


branlwyd:
(This is rebased & ready for review.)

tgeoghegan (approved):

inahga (approved):

divergentdave (approved):

PR #2560, opened 2024-01-29
Split delete_expired_client_reports() query

divergentdave:
While running a cloud load test, with a single task in the database, I observed behavior similar to our production incident last month, where garbage collection transactions were taking progressively longer to execute, and aggregation progress eventually came to a halt, as both database instances' CPUs were monopolized by the garbage collector. As before, this was caused by the `delete_expired_client_reports()` query being executed with a hash join that hashed a sequential scan of `client_reports`. The database instance had already had the `random_page_cost = 1.1` flag change applied. I determined that the estimated selectivity of the `client_timestamp < COALESCE(...)` condition was way off.

Currently, the query planner only has at its disposal statistics describing the distribution of each column of each table. Via `CREATE STATISTICS`, we could additionally provide statistics about the joint distribution of two columns in one table, but that would not be enough to plan this query well. Rather, it depends on the joint distribution of `tasks.report_expiry` and `client_reports.client_timestamp` (plus the task ID and the foreign key relationship). Thus, I think the query planner must be flying blind when it comes to the value of the `COALESCE(...)` expression. Here and in the production incident, it went with a too-high estimate of the `client_timestamp` condition's selectivity. In this case, it likely correctly determined that the selectivity of `task_id = tasks.id` was 1, as it's a foreign key column with a cardinality of 1, and in the production incident, it likely had a fairly good estimate of selectivity from looking at the most common value statistics. While fixing `random_page_cost` was able to address the production incident, it wasn't enough to flip the costs of index scanning versus sequential scanning in this case, because the selectivity of the `task_id` condition being 1 raised the index scan cost more. To fully address the risk of this bad plan, we need more accurate selectivity and row count estimates. It's possible there are other situations that would trigger the same bad plan again, aside from a database with a small number of tasks, such as a much bigger `client_reports` table.

This PR fixes the problem by splitting the `delete_expired_client_reports()` query into two queries, one to look up task parameters, and one to do the SELECT and DELETE on `client_reports`. Since the query planner can't see through joins to know what the report timestamp cutoff will be, or make planning decisions based on partial execution results, we instead compute that cutoff in one query, and then provide it as a parameter to the next query. This makes the troublesome subquery foolproof by turning it into `SELECT _ FROM _ WHERE _._ = $1 AND _._ < $2 LIMIT $3`, which clearly should be a Limit node followed by an Index Scan or Bitmap Heap Scan/Bitmap Index Scan, for nontrivial table sizes. The selectivity of `client_reports.client_timestamp < $2::timestamp` can now be estimated well using single-column statistics.

I also added a unit test to ensure that `-infinity::timestamp` is properly round-tripped between the queries, if the report expiry age is not set.

<details><summary>Query plans</summary>

Here's the plan for the original query.

`EXPLAIN WITH client_reports_to_delete AS (SELECT client_reports.id FROM client_reports JOIN tasks ON tasks.id = client_reports.task_id WHERE tasks.task_id = '\xc5fabadc755a90f1f12113dd0d515d42ab11ac5d2dbe814e3b94cca94c5a7123'::bytea AND client_reports.client_timestamp < COALESCE('2024-01-26 00:00:00'::timestamp - tasks.report_expiry_age * '1 second'::interval, '-infinity'::timestamp) LIMIT 5000) DELETE FROM client_reports USING client_reports_to_delete WHERE client_reports.id = client_reports_to_delete.id;`

```
                                                                                                                          QUERY PLAN                                                                                                                          
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Delete on client_reports  (cost=2.94..112297.44 rows=0 width=0)
   ->  Nested Loop  (cost=2.94..112297.44 rows=5000 width=38)
         ->  Subquery Scan on client_reports_to_delete  (cost=2.38..98417.94 rows=5000 width=40)
               ->  Limit  (cost=2.38..98367.94 rows=5000 width=8)
                     ->  Hash Join  (cost=2.38..1849235.64 rows=93998 width=8)
                           Hash Cond: (client_reports_1.task_id = tasks.id)
                           Join Filter: (client_reports_1.client_timestamp < COALESCE(('2024-01-26 00:00:00'::timestamp without time zone - ((tasks.report_expiry_age)::double precision * '00:00:01'::interval)), '-infinity'::timestamp without time zone))
                           ->  Seq Scan on client_reports client_reports_1  (cost=0.00..1706602.96 rows=53578896 width=24)
                           ->  Hash  (cost=2.36..2.36 rows=1 width=16)
                                 ->  Index Scan using task_id_index on tasks  (cost=0.14..2.36 rows=1 width=16)
                                       Index Cond: (task_id = '\xc5fabadc755a90f1f12113dd0d515d42ab11ac5d2dbe814e3b94cca94c5a7123'::bytea)
         ->  Index Scan using client_reports_pkey on client_reports  (cost=0.56..2.78 rows=1 width=14)
               Index Cond: (id = client_reports_to_delete.id)
(13 rows)
```

Next, I ran `SET enable_seqscan = false`, and produced the following counterfactual query plan. Note that the Bitmap Index Scan row estimate is almost exactly 1/3 of the table's size, while the actual number of rows scanned is zero.

`EXPLAIN ANALYZE WITH client_reports_to_delete AS (SELECT client_reports.id FROM client_reports JOIN tasks ON tasks.id = client_reports.task_id WHERE tasks.task_id = '\xc5fabadc755a90f1f12113dd0d515d42ab11ac5d2dbe814e3b94cca94c5a7123'::bytea AND client_reports.client_timestamp < COALESCE('2024-01-26 00:00:00'::timestamp - tasks.report_expiry_age * '1 second'::interval, '-infinity'::timestamp) LIMIT 5000) DELETE FROM client_reports USING client_reports_to_delete WHERE client_reports.id = client_reports_to_delete.id;`

```
                                                                                                                                    QUERY PLAN                                                                                                                                     
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Delete on client_reports  (cost=300669.01..449305.32 rows=0 width=0) (actual time=2.419..2.423 rows=0 loops=1)
   ->  Nested Loop  (cost=300669.01..449305.32 rows=5000 width=38) (actual time=2.418..2.420 rows=0 loops=1)
         ->  Subquery Scan on client_reports_to_delete  (cost=300668.45..435425.82 rows=5000 width=40) (actual time=2.417..2.419 rows=0 loops=1)
               ->  Limit  (cost=300668.45..435375.82 rows=5000 width=8) (actual time=2.416..2.418 rows=0 loops=1)
                     ->  Nested Loop  (cost=300668.45..2833113.16 rows=93998 width=8) (actual time=2.414..2.415 rows=0 loops=1)
                           ->  Index Scan using task_id_index on tasks  (cost=0.14..2.36 rows=1 width=16) (actual time=0.027..0.031 rows=1 loops=1)
                                 Index Cond: (task_id = '\xc5fabadc755a90f1f12113dd0d515d42ab11ac5d2dbe814e3b94cca94c5a7123'::bytea)
                           ->  Bitmap Heap Scan on client_reports client_reports_1  (cost=300668.30..2654514.48 rows=17859632 width=24) (actual time=2.376..2.377 rows=0 loops=1)
                                 Recheck Cond: ((task_id = tasks.id) AND (client_timestamp < COALESCE(('2024-01-26 00:00:00'::timestamp without time zone - ((tasks.report_expiry_age)::double precision * '00:00:01'::interval)), '-infinity'::timestamp without time zone)))
                                 ->  Bitmap Index Scan on client_reports_task_and_timestamp_index  (cost=0.00..296203.39 rows=17859632 width=0) (actual time=2.368..2.368 rows=0 loops=1)
                                       Index Cond: ((task_id = tasks.id) AND (client_timestamp < COALESCE(('2024-01-26 00:00:00'::timestamp without time zone - ((tasks.report_expiry_age)::double precision * '00:00:01'::interval)), '-infinity'::timestamp without time zone)))
         ->  Index Scan using client_reports_pkey on client_reports  (cost=0.56..2.78 rows=1 width=14) (never executed)
               Index Cond: (id = client_reports_to_delete.id)
 Planning Time: 0.898 ms
 Execution Time: 3.084 ms
(15 rows)
```

I didn't keep any query plans from the new queries, but I did confirm they do scan the `client_reports_task_and_timestamp_index` index.

</details>

Note that this adds a database round trip to garbage collection, which will negatively impact garbage collection throughput somewhat. An alternative approach to solve the same issue, without the extra round trip, would be to run the two queries from a PL/pgSQL function.

We could also consider eliminating `report_expiry_age` as a per-task parameter, and instead moving it to Janus's configuration. (assuming we don't want to make it configurable by users) This could simplify a lot of queries by lifting this same calculation to the aggregator process.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -32,7 +32,7 @@ use opentelemetry::{
     metrics::{Counter, Histogram, Meter, Unit},
     KeyValue,
 };
-use postgres_types::{FromSql, Json, ToSql};
+use postgres_types::{FromSql, Json, Timestamp, ToSql};
 use prio::{
     codec::{decode_u16_items, encode_u16_items, CodecError, Decode, Encode, ParameterizedDecode},
     topology::ping_pong::PingPongTransition,
@@ -4218,13 +4218,40 @@ impl<C: Clock> Transaction<'_, C> {
         task_id: &TaskId,
         limit: u64,
     ) -> Result<u64, Error> {
-        let stmt = self
+        // Calculation of a report timestamp threshold is split apart from the main body of the
+        // query so that the query planner can see concrete timestamp value, compare it against the
+        // client_timestamp column's histogram, and make an accurate row count estimate. If the
+        // threshold is determined in a single query via a join, the query planner is not able to
+        // predict the task's report_expiry_age, and the accuracy of the row count estimate suffers.
+        let stmt_1 = self
+            .prepare_cached(
+                "SELECT
+                    id,
+                    COALESCE(
+                        $2::TIMESTAMP - tasks.report_expiry_age * '1 second'::INTERVAL,
+                        '-infinity'::TIMESTAMP
+                    ) AS threshold
+                FROM tasks WHERE tasks.task_id = $1",
+            )
+            .await?;
+        let row = self
+            .query_one(
+                &stmt_1,
+                &[
+                    /* task_id */ &task_id.get_encoded()?,
+                    /* now */ &self.clock.now().as_naive_date_time()?,
+                ],
+            )
+            .await?;
+        let id = row.get::<_, i64>("id");
+        let threshold = row.get::<_, Timestamp<NaiveDateTime>>("threshold");
+
+        let stmt_2 = self
             .prepare_cached(
                 "WITH client_reports_to_delete AS (
                     SELECT client_reports.id FROM client_reports
-                    JOIN tasks ON tasks.id = client_reports.task_id
-                    WHERE tasks.task_id = $1
-                      AND client_reports.client_timestamp < COALESCE($2::TIMESTAMP - tasks.report_expiry_age * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)
+                    WHERE client_reports.task_id = $1
+                        AND client_reports.client_timestamp < $2::TIMESTAMP
                     LIMIT $3
                 )
                 DELETE FROM client_reports
@@ -4233,10 +4260,10 @@ impl<C: Clock> Transaction<'_, C> {
             )
             .await?;
         self.execute(
-            &stmt,
+            &stmt_2,
             &[
-                /* task_id */ &task_id.get_encoded()?,
-                /* now */ &self.clock.now().as_naive_date_time()?,
+                /* id */ &id,
+                /* threshold */ &threshold,
                 /* limit */ &i64::try_from(limit)?,
             ],
         )
--- a/aggregator_core/src/datastore/tests.rs
+++ b/aggregator_core/src/datastore/tests.rs
@@ -5564,6 +5564,80 @@ async fn delete_expired_client_reports(ephemeral_datastore: EphemeralDatastore) 
 
 #[rstest_reuse::apply(schema_versions_template)]
 #[tokio::test]
+async fn delete_expired_client_reports_noop(ephemeral_datastore: EphemeralDatastore) {
+    install_test_trace_subscriber();
+
+    let clock = MockClock::default();
+    let ds = ephemeral_datastore.datastore(clock.clone()).await;
+    let vdaf = dummy_vdaf::Vdaf::new();
+
+    // Setup.
+    let (task_id, new_report_id, old_report_id) = ds
+        .run_unnamed_tx(|tx| {
+            Box::pin(async move {
+                let task = TaskBuilder::new(task::QueryType::TimeInterval, VdafInstance::Fake)
+                    .with_report_expiry_age(None)
+                    .build()
+                    .leader_view()
+                    .unwrap();
+                tx.put_aggregator_task(&task).await?;
+
+                let old_report = LeaderStoredReport::new_dummy(
+                    *task.id(),
+                    OLDEST_ALLOWED_REPORT_TIMESTAMP
+                        .sub(&Duration::from_seconds(1))
+                        .unwrap(),
+                );
+                let new_report =
+                    LeaderStoredReport::new_dummy(*task.id(), OLDEST_ALLOWED_REPORT_TIMESTAMP);
+                tx.put_client_report(&dummy_vdaf::Vdaf::new(), &old_report)
+                    .await?;
+                tx.put_client_report(&dummy_vdaf::Vdaf::new(), &new_report)
+                    .await?;
+
+                Ok((
+                    *task.id(),
+                    *new_report.metadata().id(),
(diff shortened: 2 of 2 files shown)


aggregator_core/src/datastore.rs
```diff
@@ -4218,13 +4218,40 @@ impl<C: Clock> Transaction<'_, C> {
...
+
+        let stmt_2 = self
             .prepare_cached(
                 "WITH client_reports_to_delete AS (
                     SELECT client_reports.id FROM client_reports
-                    JOIN tasks ON tasks.id = client_reports.task_id
-                    WHERE tasks.task_id = $1
-                      AND client_reports.client_timestamp < COALESCE($2::TIMESTAMP - tasks.report_expiry_age * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)
+                    WHERE client_reports.task_id = $1
+                        AND client_reports.client_timestamp < $2::timestamp
```
inahga, line 4254:
```suggestion
                        AND client_reports.client_timestamp < $2::TIMESTAMP
```
nit: consistent with other queries

inahga (approved):
Excellent analysis, nice job catching this.

I'm surprised that this query isn't problematic for similar reasons:
```
                "WITH unaggregated_reports AS (
                    SELECT client_reports.id FROM client_reports
                    JOIN tasks ON tasks.id = client_reports.task_id
                    WHERE tasks.task_id = $1
                      AND client_reports.aggregation_started = FALSE
                      AND client_reports.client_timestamp >= COALESCE($2::TIMESTAMP - tasks.report_expiry_age * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)
                    ORDER BY client_timestamp DESC
                    FOR UPDATE OF client_reports SKIP LOCKED
                    LIMIT 5000
                )
                UPDATE client_reports SET
                    aggregation_started = TRUE, updated_at = $3, updated_by = $4
                WHERE id IN (SELECT id FROM unaggregated_reports)
                RETURNING report_id, client_timestamp, extensions, public_share, leader_input_share,
                          helper_encrypted_input_share",
```
But looking at the query plan in query insights in the load testing environment, it behaves as expected.
image

tgeoghegan (approved):

divergentdave:
Good point, my guess is that `client_reports.aggregation_started = FALSE` both pushes the overall estimated selectivity way down, and encourages the use of the `client_reports_task_and_timestamp_unaggregated_index` partial index, since its predicate partly matches.

branlwyd:
Excellent analysis.

PR #2575, opened 2024-01-29
Bump `divviup-client` to 0.1.7

tgeoghegan:
`cargo deny` is blocking PRs like #2558 because of a rustsec advisory on `trillium-client`, a vulnerable version of which gets pulled in via `divviup-client`. dependabot, for whatever reason, only wants to take us to 0.1.6 (#2570) so this PR cuts to the chase and takes 0.1.7 in the `janus_integration_tests` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1142,13 +1142,13 @@ dependencies = [
 
 [[package]]
 name = "divviup-client"
-version = "0.1.5"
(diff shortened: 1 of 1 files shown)


tgeoghegan:
David beat me to it in #2474. That change doesn't bump `divviup-client`, but it's OK, dependabot will get there eventually.

PR #2584, opened 2024-01-29
Rephrase batch-GC'ed check in a non-hacky way.

### LINKED ISSUE #2464: Leader: Batch GC can cause an aggregation job to become unwritable.

branlwyd:
Now that the change has been live for a few weeks with no negative consequences, I'm happy to rephrase this in a less-hacky way.

Closes #2464.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_writer.rs
+++ b/aggregator/src/aggregator/aggregation_job_writer.rs
@@ -11,7 +11,7 @@ use janus_aggregator_core::{
         },
         Error, Transaction,
     },
-    task::{AggregatorTask, QueryType},
+    task::AggregatorTask,
 };
 use janus_core::time::{Clock, IntervalExt};
 use janus_messages::{AggregationJobId, Interval, PrepareError, ReportId};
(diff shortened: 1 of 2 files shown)


branlwyd:
(Since this is only a semantics change, I'm OK not backporting this to the 0.6 branch.)

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -344,25 +344,18 @@ impl<const SEED_SIZE: usize, Q: CollectableQueryType, A: vdaf::Aggregator<SEED_S
...
                         // See https://github.com/divviup/janus/issues/2464 for more detail.
                         if batch_op == Operation::Put {
-                            // Guard to ensure we are in the situation we think we're in: we are in
-                            // a time-interval task, and the batch interval is past the GC window.
-                            if !matches!(self.task.query_type(), QueryType::TimeInterval)
-                                || !Q::to_batch_interval(batch_identifier)
-                                    .map(|interval| interval.end() < tx.clock().now())
-                                    .unwrap_or(false)
-                            {
+                            if Q::is_batch_garbage_collected(tx.clock(), batch_identifier) != Some(true) {
```
inahga, line 358: Hmm, there is a new functional change here--a fixed size task in the same situation would go down the panic, where it didn't before. We're no longer assuming that a fixed size batch is GC'd.

I'm not sure that we want this, it seems possible to run into this state with a fixed size batch being GC'd, where a panic seems inappropriate. Thoughts?

inahga (approved):
I think we should backport the change, if it's not too much trouble. We're going to support 0.6 in production for a while, and I don't sense this is the last we'll see of this code. So it would nice to have parity between them as we work between 0.7 and 0.6.

inahga:
Oh oops I mean to comment instead of approve.

aggregator/src/aggregator/aggregation_job_writer.rs
branlwyd: I think this is not a logic change. The old logic to trigger a panic was:

```
!matches!(self.task.query_type(), QueryType::TimeInterval)
    || !Q::to_batch_interval(batch_identifier)
        .map(|interval| interval.end() < tx.clock().now())
        .unwrap_or(false)
```

...which would panic if the query type was `FixedSize`, or if the query type was `TimeInterval` and the batch interval's end was not before now.

The new logic panics if `Q::is_batch_garbage_collected` returns something other than `Some(true)`. This will happen if the query type is `FixedSize` (since `FixedSize`'s implementation always returns `None`), or if the query type is `TimeInterval` and the batch interval's end is not before now.

(please double-check this, I've looked real hard at it a couple of times but maybe I'm missing something obvious)

divergentdave: Agreed, LGTM. Tangentially, I'll note that `!matches!(..., ...)` is easy to misread, I caught a bug elsewhere a while back from a related typo.

divergentdave (approved):

aggregator_core/src/query_type.rs
```diff
@@ -68,6 +68,15 @@ pub trait AccumulableQueryType: QueryType {
 
     /// Get the default value of the partial batch identifier, if applicable.
     fn default_partial_batch_identifier() -> Option<&'static Self::PartialBatchIdentifier>;
+
+    /// Determine if the batch is expected to be garbage-collected, based on the identifier.
+    /// `Some(true)` and `Some(false)`` indicate the expected result, and `None` indicates that the
```
divergentdave, line 73: nit: extra quote
```suggestion
    /// `Some(true)` and `Some(false)` indicate the expected result, and `None` indicates that the
```

aggregator/src/aggregator/aggregation_job_writer.rs
inahga: Ah yep with fresh eyes I see that you're right. Sorry for the misdirect!

inahga (approved):

Issue #293, opened 2022-07-06
Harden Rust crate releasing against human errors


tgeoghegan:
Currently, to release new versions of the `janus_core` and `janus_client` crates, you have to do the following:

1. bump version in `janus_core/Cargo.toml`
2. bump version in `janus_client/Cargo.toml`
3. bump version of `janus_core` dependency in `janus_client/Cargo.toml`
4. run a build to update `Cargo.lock`
5. Make a PR with all that
6. Merge PR to main
7. Make a GitHub release

It's very easy to forget to do any of steps 1-4 (I have at least once, #292). We could do better here. Some ideas (which may exclude each other);

- run `cargo publish --dry-run` in CI to prove that the crates aren't busted (though I'm not sure how to make this work if `janus_client` depends on an unpublished `janus_core`)
- run `cargo publish` with `--locked` in CI and/or release workflows to guarantee that `Cargo.lock` is up to date (or run a `cargo build` and assert that `Cargo.lock` is unchanged)
- write a much more elaborate release workflow that would handle rewriting versions in `Cargo.toml` based on the git tag

divergentdave:
The `set-version` subcommand from the [`cargo-edit`](https://github.com/killercup/cargo-edit) project can streamline step 3, FWIW. Running `cargo set-version --bump patch -p janus_core` does both step 1 and step 3 in one go.

branlwyd:
(3) is no longer necessary for non-major version updates: we now specify "internal" crate dependencies as major versions, so we should get the new version automatically.

The `Cargo.lock` update in (4) can be done by `cargo update --workspace`.

tgeoghegan:
@inahga, you're working on this currently, right?

inahga:
Accidentally, yes! Didn't realize this issue existed.

PR #2585, opened 2024-01-30
Add workflow for automatically generating release

inahga:
Automates the Janus release process.

This workflow, when triggered by workflow_dispatch, will push a version bump commit to each active branch, then create a release on that new commit. That release will then trigger our other workflows, like publishing SQL schema and to crates.io.

There are a few tradeoffs to this approach:

This bypasses status checks. An operator should check the status of the main branch before using this workflow. I omitted an automated check for the latest status check because I don't want to end up in a situation where we have to make arbitrary commits to get the latest status check passing. I generally think that with per-PR status checks, it's very unlikely to release code that is flat out broken--likely blocking status checks would involve new clippy lints when stable is updated, or rustsec violations.

This workflow elegantly but also inelegantly handles prerelease branches. `cargo-edit` bumps version `0.7.0-prerelease-1` to `0.7.0`. This is great for when we're ready to release it for real, but not so great when we need incremental prerelease versions. I don't think there's a tidy solution to this, so we'll just have to handle incremental prerelease versions manually.

There is no facility for supplying extra release notes. I figure we can just edit the created release if we need to add more detail to a set of release notes.

Even with these tradeoffs, this workflow should make releasing Janus a 1-click process (modulo prerelease branches).

### Deploying

I will need @branlwyd's help with creating the `DIVVIUP_GITHUB_AUTOMATION_RELEASE_PAT` secret. We will want to make this an organization secret, so we can roll out this approach to multiple repositories. It will need `repo` scopes, and it *must* have an expiration date (otherwise repo cloning won't work).

We will also need to assign `divviup-github-automation` to the bypass list in the branch protection rules. This is denoted `Allow specified actors to bypass required pull requests`. I will do this if this PR is approved.

### DIFF
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -17,6 +17,8 @@ jobs:
       matrix:
         target_branch: ${{ fromJSON(inputs.target_branch) }}
     runs-on: ubuntu-latest
+    env:
+      CARGO_EDIT_VERSION: 0.12.2
     steps:
       - uses: actions/checkout@v4
         with:
@@ -24,28 +26,21 @@ jobs:
           token: '${{ secrets.DIVVIUP_GITHUB_AUTOMATION_RELEASE_PAT }}'
 
       - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@master
-        with:
-          toolchain: stable
+        uses: dtolnay/rust-toolchain@stable
 
-      - name: Restore cargo-edit from cache
-        id: cache-cargo-edit-restore
-        uses: actions/cache/restore@v4
+      - name: Cache cargo edit
+        uses: actions/cache@v4
         with:
-          path: |
-            ~/.cargo/bin
-          key: ${{ runner.os }}-cargo-edit
-
-      - run: cargo install cargo-edit --no-default-features --features set-version
-        if: steps.cache-cargo-edit-restore.outputs.cache-hit != 'true'
-
-      - name: Save cargo-edit to cache
-        id: cache-cargo-edit-save
-        uses: actions/cache/save@v4
-        with:
-          path: |
-            ~/.cargo/bin
-          key: ${{ steps.cache-cargo-edit-restore.outputs.cache-primary-key }}
+          path: ${{ runner.tool_cache }}/cargo-edit
+          key: cargo-edit-${{ env.CARGO_EDIT_VERSION }}
+      - name: Add the tool cache directory to the search path
+        run: echo "${{ runner.tool_cache }}/cargo-edit/bin/" >>$GITHUB_PATH
+      - name: Ensure cargo-edit is installed
+        run: |
+          cargo install \
+            --root ${{ runner.tool_cache }}/cargo-edit \
+            --version ${{ env.CARGO_EDIT_VERSION }} \
+            cargo-edit
       
       - run: cargo set-version --bump patch
       - run: git diff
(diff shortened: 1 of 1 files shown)


inahga:
This workflow can be demo'd in https://github.com/divviup/inahga-test.

inahga:
Closes https://github.com/divviup/janus/issues/293

divergentdave (approved):
I think this looks good, it'll take care of some purely mechanical steps

branlwyd:
I generated a PAT & set the permissions appropriately (I think) -- all non-pictured perms are not checked:

image

This PAT is available via the `DIVVIUP_GITHUB_AUTOMATION_RELEASE_PAT` org secret.

.github/workflows/make-release.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Restore cargo-edit from cache
+        id: cache-cargo-edit-restore
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            ~/.cargo/bin
+          key: ${{ runner.os }}-cargo-edit
+
+      - run: cargo install cargo-edit --no-default-features --features set-version
+        if: steps.cache-cargo-edit-restore.outputs.cache-hit != 'true'
```
branlwyd, line 40: Barring cache eviction, what will allow us to upgrade to newer versions of `cargo-edit`? (If it's nothing, WDYT about just reinstalling `cargo-edit` each time -- at least on my machine, it only takes a few seconds to compile & install. This isn't such a good suggestion if the install takes much longer on GH's actions runners.)

branlwyd (approved):

.github/workflows/make-release.yml
```diff
@@ -0,0 +1,97 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      target_branch:
+        description: >-
+          JSON array of branches to target. The first branch in the array will
+          be marked as the latest release.
+        required: false
+        type: string
+        default: '["release/0.6","release/0.5"]'
```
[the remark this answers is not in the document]
inahga: Yep. When we're ready to promote it to a production-ready release, we can just run the workflow normally, but in the mean time we'll have to do it manually.

 The main blocker is that `cargo-edit` doesn't [really support incrementing prerelease versions](https://github.com/killercup/cargo-edit/issues/700). (FWIW I'm not sure if the available alternatives do, like `release-plz`.)

.github/workflows/make-release.yml
divergentdave: We could write the version of cargo-edit into the workflow, pass it to cargo install, and put it in the cache key, as we do with a couple other Rust utilities. I expect the main reason we'll want to upgrade cargo-edit is to keep up with Cargo manifest format additions.

inahga: It does take about 3 minutes on a runner, which is an annoying amount of time when iterating on the action, but I suppose not a big deal given the whole release CI will take about an hour.

Either way I was figuring we can just do `gh cache delete Linux-cargo-edit` when we want to flush the cache and get a new version.

inahga: That's a good idea, I see we do that in https://github.com/divviup/libprio-rs/blob/main/.github/workflows/supply-chain.yml. I'll adopt this.

branlwyd: Hm, yeah, I think I like the idea of fixing the version of `cargo-edit` (following the more general principle that the versions of dependencies should be fixed).

PR #2586, opened 2024-01-30
Helper: disallow reports from too far in the future.

### LINKED ISSUE #2539: Helper should reject reports with times far in the future
DAP says that all aggregators must check if report timestamps are too far in the future, and helpers SHOULD respond with the report_too_early PrepareError if so. Beyond DAP's requirement, it is important that we implement this check, because reports with timestamps in the distant future will take that much longer to be garbage collected, and we don't want to have to store their IDs that long. Currently, we only check this condition when operating as the leader while accepting uploaded reports; we need to add this to aggregate initialization as well. However, we should get #2538 widely deployed before we start sending `report_too_early` error codes, to avoid compatibility issues.

branlwyd:
Like with the Leader, this is based on the `tolerable_clock_skew` task parameter.

Closes #2539.

### DIFF
--- a/aggregator/src/aggregator/aggregate_init_tests.rs
+++ b/aggregator/src/aggregator/aggregate_init_tests.rs
@@ -29,7 +29,8 @@ use janus_core::{
 };
 use janus_messages::{
     query_type::TimeInterval, AggregationJobId, AggregationJobInitializeReq, AggregationJobResp,
-    HpkeConfig, PartialBatchSelector, PrepareInit, PrepareStepResult, ReportMetadata, ReportShare,
+    Duration, HpkeConfig, PartialBatchSelector, PrepareError, PrepareInit, PrepareStepResult,
+    ReportMetadata, ReportShare,
 };
 use prio::{
     codec::Encode,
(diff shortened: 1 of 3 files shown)


aggregator/src/aggregator/aggregate_init_tests.rs
```diff
@@ -161,9 +162,8 @@ pub(super) struct AggregationJobInitTestCase<
     pub(super) clock: MockClock,
     pub(super) task: Task,
     pub(super) prepare_init_generator: PrepareInitGenerator<VERIFY_KEY_SIZE, V>,
-    pub(super) prepare_inits: Vec<PrepareInit>,
```
branlwyd, line 164: I removed this field since it duplicated data from `aggregation_job_init_req`, and these could fall out of sync if the test updated one but not the other (as I discovered while implementing this PR).

branlwyd (commented):
Thoughts on whether this should be backported? I think no, as our only (likely) deployments of earlier versions of DAP are Janus/Janus pairings, so we can trust the leader to check the report times (and the report time check is not a security check which _must_ be implemented by both aggregators for collusion-avoidance reasons).

inahga (approved):
Similar to my rationale in https://github.com/divviup/janus/pull/2584, I think we should take the backport to keep 0.6 and 0.7 from drifting too far apart.

divergentdave (approved):

PR #2597, opened 2024-01-31
Avoid script injection in make-release workflow

inahga:
Interpolating GitHub user-controlled variables directly into a shell script is unsafe, since they're interpolated verbatim, leading to the possibility of script injection. https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections

This change is just out of an abundance of caution. At the moment, only people with `write` access can trigger a workflow_dispatch, which should be a tightly controlled set of users who are exceedingly unlikely to execute injection attacks.

### DIFF
--- a/.github/workflows/make-release.yml
+++ b/.github/workflows/make-release.yml
@@ -45,10 +45,12 @@ jobs:
       - run: cargo set-version --bump patch
       - run: git diff
       - name: Push changes
+        env:
+          ACTOR: '${{ github.actor }}'
         run: |
           git config user.email "divviup-github-automation@divviup.org"
           git config user.name "divviup-github-automation"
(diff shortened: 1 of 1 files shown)


.github/workflows/make-release.yml
```diff
@@ -71,11 +73,13 @@ jobs:
           toolchain: stable
 
       - name: Create release
+        env:
+          GH_TOKEN: '${{ secrets.INAHGA_TEST }}'
```
branlwyd, line 77: Incorrect secret name?

inahga: Whoops!

divergentdave (approved):

PR #2598, opened 2024-01-31
Update taskprov for taskprov-06.

branlwyd:
There are several message changes, and a requirement for a new empty extension in each report share indicating that taskprov is in use.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -780,22 +780,15 @@ impl<C: Clock> Aggregator<C> {
         task_config: &TaskConfig,
         aggregator_auth_token: Option<&AuthenticationToken>,
     ) -> Result<(&PeerAggregator, Url, Url), Error> {
-        let aggregator_urls = task_config
-            .aggregator_endpoints()
-            .iter()
-            .map(|url| url.try_into())
-            .collect::<Result<Vec<Url>, _>>()?;
-        if aggregator_urls.len() != 2 {
-            return Err(Error::InvalidMessage(
-                Some(*task_id),
-                "taskprov configuration is missing one or both aggregators",
-            ));
-        }
-        let peer_aggregator_url = &aggregator_urls[peer_role.index().unwrap()];
+        let peer_aggregator_url = match peer_role {
+            Role::Leader => task_config.leader_aggregator_endpoint(),
+            Role::Helper => task_config.helper_aggregator_endpoint(),
+            _ => panic!("Unexpected role {peer_role}"),
+        }.try_into()?;
 
         let peer_aggregator = self
             .peer_aggregators
-            .get(peer_aggregator_url, peer_role)
+            .get(&peer_aggregator_url, peer_role)
             .ok_or(Error::InvalidTask(
                 *task_id,
                 OptOutReason::NoSuchPeer(*peer_role),
@@ -820,8 +813,8 @@ impl<C: Clock> Aggregator<C> {
         );
         Ok((
             peer_aggregator,
-            aggregator_urls[Role::Leader.index().unwrap()].clone(),
-            aggregator_urls[Role::Helper.index().unwrap()].clone(),
+            task_config.leader_aggregator_endpoint().try_into()?,
+            task_config.helper_aggregator_endpoint().try_into()?,
         ))
     }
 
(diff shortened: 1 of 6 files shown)


aggregator/src/aggregator.rs
```diff
@@ -1892,8 +1899,35 @@ impl VdafOps {
...
+                if require_taskprov_extension {
+                    if !extensions.get(&ExtensionType::Taskprov).map(|data| data.is_empty()).unwrap_or(false) {
+                        debug!(
+                            task_id = %task.id(),
+                            metadata = ?prepare_init.report_share().metadata(),
+                            "Taskprov task received report with missing or malformed taskprov extension",
+                        );
+                        aggregate_step_failure_counter
+                            .add(1, &[KeyValue::new("type", "missing_or_malformed_taskprov_extension")]);
+                        taskprov_error = Some(Error::InvalidMessage(Some(*task.id()), "missing or malformed taskprov extension"));
```
branlwyd, line 1912: The layout of this logic means that, if we receive a valid taskprov header in an aggregation initialization where some ReportShares don't contain the new taskprov extension, we will abort the aggregation but leave the task in place.

I think that's OK -- tasks can still only be created by authorized parties, and even if the task is created in error, the cost of an unused task sitting around is minimal.

More practically, it would be challenging to do otherwise: either we would need to check the extensions before we created the task (which would imply parsing the request, which is query-type specific, and decrypting the plaintext input shares), or we would need to unwind the task creation on failure (and think through possible race conditions with other aggregations).

aggregator/src/aggregator.rs
```diff
@@ -1892,8 +1899,35 @@ impl VdafOps {
...
+                    }
+                } else if extensions.contains_key(&ExtensionType::Taskprov) {
+                    // taskprov not enabled, but the taskprov extension is present.
```
branlwyd, line 1915: This isn't required by the taskprov spec, but I think it is a good idea.

branlwyd (commented):
I am creating this as a draft for now, `taskprov-06` is imminent -- I will update to that version in the same PR.

branlwyd:
(This PR now matches `HEAD` of the taskprov repo, which is likely to become `taskprov-06`. I'm going to wait until the new taskprov version is cut before I send for review.)

branlwyd:
I think this is ready for review: `taskprov-06` is still not cut, but chatting with the authors indicated there wasn't more work to be done. There may be some (hopefully) small last-minute changes, but I will fix them up in a follow-on PR. For now, I don't want to risk this PR bitrotting.

branlwyd:
`taskprov-06` has been cut: https://datatracker.ietf.org/doc/draft-wang-ppm-dap-taskprov/06/

aggregator/src/aggregator.rs
```diff
@@ -1904,6 +1910,30 @@ impl VdafOps {
...
+                        aggregate_step_failure_counter
+                            .add(1, &[KeyValue::new("type", "missing_or_malformed_taskprov_extension")]);
+                        return Err(PrepareError::InvalidMessage);
+                    }
+                } else if extensions.contains_key(&ExtensionType::Taskprov) {
+                    // taskprov not enabled, but the taskprov extension is present.
+                        debug!(
+                            task_id = %task.id(),
+                            metadata = ?prepare_init.report_share().metadata(),
+                            "Taskprov task received report with unexpected taskprov extension",
```
divergentdave, line 1930: This message is incorrect
```suggestion
                            "Non-Taskprov task received report with unexpected taskprov extension",
```

messages/src/taskprov.rs
```diff
@@ -254,6 +232,46 @@ impl Query {
     const FIXED_SIZE: u8 = 2;
 }
 
+// XXX: write roundtrip_query (to test encoded_len)
```
divergentdave, line 235: This stray comment can be deleted, as the test is done below

divergentdave: Separately, this part of a comment above can be deleted now

```
///   - The parent mod decoding logic assumes that the query type is encoded
///     directly adjacent to its associated parameters. This is not the case
///     in taskprov.
```

divergentdave (approved):

PR #2601, opened 2024-02-01
[0.6] Helper: disallow reports from too far in the future.

branlwyd:
Like with the Leader, this is based on the `tolerable_clock_skew` task parameter.

### DIFF
--- a/aggregator/src/aggregator/aggregate_init_tests.rs
+++ b/aggregator/src/aggregator/aggregate_init_tests.rs
@@ -27,7 +27,8 @@ use janus_core::{
 };
 use janus_messages::{
     query_type::TimeInterval, AggregationJobId, AggregationJobInitializeReq, AggregationJobResp,
-    HpkeConfig, PartialBatchSelector, PrepareInit, PrepareStepResult, ReportMetadata, ReportShare,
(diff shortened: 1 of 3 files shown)


aggregator/src/aggregator/aggregate_init_tests.rs
```diff
@@ -487,6 +489,80 @@ async fn aggregation_job_mutation_report_aggregations() {
...
+        dummy_vdaf::Vdaf::new(),
+        VdafInstance::Fake,
+        dummy_vdaf::AggregationParam(0),
+        (),
+        AuthenticationToken::Bearer(random()),
+    )
+    .await;
+
+    test_case.aggregation_job_init_req = AggregationJobInitializeReq::new(
+        test_case.aggregation_param.get_encoded().unwrap(),
```
divergentdave, line 504: Note that this needs some rebase cleanup to roll back to infallible Encode.

divergentdave (approved):

aggregator/src/aggregator/aggregate_init_tests.rs
branlwyd: D'oh, thanks!

PR #2602, opened 2024-02-01
[0.6] Rephrase batch-GC'ed check in a non-hacky way.

branlwyd:
Now that the change has been live for a few weeks with no negative consequences, I'm happy to rephrase this in a less-hacky way.

### DIFF
--- a/aggregator_core/src/query_type.rs
+++ b/aggregator_core/src/query_type.rs
@@ -70,7 +70,7 @@ pub trait AccumulableQueryType: QueryType {
     fn default_partial_batch_identifier() -> Option<&'static Self::PartialBatchIdentifier>;
 
     /// Determine if the batch is expected to be garbage-collected, based on the identifier.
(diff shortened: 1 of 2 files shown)


aggregator_core/src/query_type.rs
```diff
@@ -68,6 +68,15 @@ pub trait AccumulableQueryType: QueryType {
 
     /// Get the default value of the partial batch identifier, if applicable.
     fn default_partial_batch_identifier() -> Option<&'static Self::PartialBatchIdentifier>;
+
+    /// Determine if the batch is expected to be garbage-collected, based on the identifier.
+    /// `Some(true)` and `Some(false)`` indicate the expected result, and `None` indicates that the
```
divergentdave, line 73:
```suggestion
    /// `Some(true)` and `Some(false)` indicate the expected result, and `None` indicates that the
```

divergentdave (approved):

PR #2607, opened 2024-02-05
Time-interval: create aggregation jobs touching fewer batches.

### LINKED ISSUE #2603: Time-interval: create aggregation jobs touching as few batches as possible
If the Leader & Helper disagree on the set of reports in the batch, the batch is effectively uncollectable (i.e. lost). This can happen e.g. if the Helper believes an aggregation job to be successfully completed, but the Leader believes the aggregation job did not succeed due to a transient communication failure.

This failure mode is currently inherent to DAP. Therefore, we should do what we can to mitigate it in our implementation.

In the time-interval query type, aggregation jobs can include reports associated with an arbitrary number of batches (up to the aggregation job size). This means that a single aggregation job encountering the failure mode above will potentially lead to an arbitrary number of collection jobs becoming uncollectable/lost.

We should minimize the number of batches associated with a given time-interval aggregation job to reduce the blast radius of a failing aggregation job. One way to do this would be to generate aggregation jobs with reports ordered by time.

branlwyd:
The implementation strategy is to consider only a given batch at a time, generating as many aggregation jobs as possible. Only once there are too few reports from the batch remaining to generate any more aggregation jobs do we move on to the next batch, rolling in the leftovers from the current batch (effectively smearing across batches when necessary).

Note that this will not necessarily avoid multi-batch aggregation jobs in all cases. For example, if aggregation jobs are generated with sizes in [5, 10], and there are 11 reports available for aggregation in the current batch, this logic will generate a single batch of size 10, rolling the leftover report into the next batch. It is instead possible to generate two aggregation jbos of (say) 5 and 6, with no leftovers to be rolled into the next batch. I'm OK with this, as we generally want to create larger aggregation jobs when possible, but we may want to lean even more towards avoiding multi-batch aggregation jobs in the future.

Closes #2603.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -1242,16 +1242,12 @@ impl<C: Clock> Transaction<'_, C> {
     /// should generally only be called on report IDs returned from
     /// `get_unaggregated_client_report_ids_for_task`, as part of the same transaction, for any
     /// client reports that are not added to an aggregation job.
-    #[tracing::instrument(skip(self, report_ids), err(level = Level::DEBUG))]
-    pub async fn mark_reports_unaggregated(
+    #[tracing::instrument(skip(self), err(level = Level::DEBUG))]
+    pub async fn mark_report_unaggregated(
         &self,
         task_id: &TaskId,
-        report_ids: &[ReportId],
+        report_id: &ReportId,
     ) -> Result<(), Error> {
-        let report_ids: Vec<_> = report_ids
-            .iter()
-            .map(ReportId::get_encoded)
-            .collect::<Result<Vec<_>, _>>()?;
         let stmt = self
             .prepare_cached(
                 "UPDATE client_reports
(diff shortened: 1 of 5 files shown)


branlwyd:
This needs to be backported to the 0.6 branch, but I'm going to wait to generate the backport PR until this looks like it's in shape to be merged.

I'm interested in thoughts as to whether we should even more aggressively generate aggregation jobs touching fewer batches -- the implementation is a little tricky, but I think the only real downside is we might end up with more, smaller aggregation jobs (touching fewer batches) rather than fewer, larger aggregation jobs (touching more batches).

inahga:
> but I think the only real downside is we might end up with more, smaller aggregation jobs (touching fewer batches) rather than fewer, larger aggregation jobs (touching more batches)

This might be an upside. I imagine the corresponding helper-side transactions would become smaller, and less likely to run into serialization conflicts. It would be interesting to see the CPU load on the helper before/after the change you suggest.

aggregator_core/src/datastore.rs
```diff
@@ -1242,43 +1242,36 @@ impl<C: Clock> Transaction<'_, C> {
     /// should generally only be called on report IDs returned from
     /// `get_unaggregated_client_report_ids_for_task`, as part of the same transaction, for any
     /// client reports that are not added to an aggregation job.
-    #[tracing::instrument(skip(self, report_ids), err(level = Level::DEBUG))]
-    pub async fn mark_reports_unaggregated(
+    #[tracing::instrument(skip(self), err(level = Level::DEBUG))]
+    pub async fn mark_report_unaggregated(
```
inahga, line 1246: The doccomment for this needs to be updated.

inahga (approved):

branlwyd:
> This might be an upside. I imagine the corresponding helper-side transactions would become smaller, and less likely to run into serialization conflicts. It would be interesting to see the CPU load on the helper before/after the change you suggest.

I think you're correct about the effects on individual transactions, except maybe the serialization piece: my concern with more, smaller aggregation jobs is that that naturally creates more chances for contention between concurrently-stepped jobs (i.e. more chances of serialization failure). Of course it's hard to be sure analytically -- it could be that if the transactions are small enough, they'll complete quickly enough that the odds of contention are lowered.

inahga:
> it could be that if the transactions are small enough, they'll complete quickly enough that the odds of contention are lowered.

That was my thought. Indeed, I think we'll have to profile the code to be sure.

Issue #2606, opened 2024-02-02
Helper aggregation: do not perform VDAF computations, writes for repeated requests


branlwyd:
Currently, on a repeated request, the Helper will recompute the VDAF steps etc before checking whether the request is a repeated, idempotent request. We should move this check to happen before the computation; this will likely drastically reduce the cost of repeated requests.

The reason we do not do this currently is that we don't open a DB transaction until after all of the VDAF computation is complete. We should investigate if we can perform the necessary reads piggybacking off of some other DB transaction. (I think there may not be one to piggyback off of, however.)

branlwyd:
Unassigning, I need to prioritize cost-optimization work for now. I don't think we have evidence this will have a large positive effect, and we might have to introduce a new DB transaction to implement it, so it's plausible it might have a negative effect on the DB as well.

inahga:
We do this now. https://github.com/divviup/janus/pull/3135.

PR #2632, opened 2024-02-05
Set HTTP client timeouts

### LINKED ISSUE #426: Set reasonable timeouts in HTTP clients
By default, `reqwest` applies no timeout to requests it makes. Callers are expected to configure this using either [`reqwest::ClientBuilder::timeout`](https://docs.rs/reqwest/0.11.11/reqwest/struct.ClientBuilder.html#method.timeout) or [`reqwest::RequestBuilder::timeout`](https://docs.rs/reqwest/0.11.11/reqwest/struct.RequestBuilder.html#method.timeout). We should do so wherever we make HTTP requests using `reqwest`.

tgeoghegan:
Applies HTTP client timeouts in the `janus_client` and `janus_collector` crates, as well as in leader->helper HTTP requests.

Closes #426

### DIFF
--- a/docs/samples/advanced_config/aggregation_job_driver.yaml
+++ b/docs/samples/advanced_config/aggregation_job_driver.yaml
@@ -87,6 +87,17 @@ worker_lease_clock_skew_allowance_secs: 60
 # (required)
 maximum_attempts_before_failure: 10
 
+# Timeout to apply when establishing connections to the helper for HTTP requests, in seconds. See
+# https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connect_timeout for
+# details. (optional)
+http_request_connection_timeout_secs: 10
+
+# Timeout to apply to HTTP requests overall (including connection establishment) when communicating
+# with the helper. See
+# https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.timeout for details.
+# (optional)
+http_request_timeout_secs: 30
+
 # Number of sharded database records per batch aggregation. Must not be greater
 # than the equivalent setting in the collection job driver. (required)
 batch_aggregation_shard_count: 32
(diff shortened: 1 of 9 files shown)


divergentdave (commented):
The new janus_aggregator configuration settings should be documented in the sample configuration files.

When it comes time to deploy this, we should keep the timeouts generous (i.e. close to our helper's existing reverse proxy timeouts) until we make aggregation optimistically check for idempotent requests (#2606).

docs/samples/advanced_config/aggregation_job_driver.yaml
```diff
@@ -87,6 +87,17 @@ worker_lease_clock_skew_allowance_secs: 60
 # (required)
 maximum_attempts_before_failure: 10
 
+# Timeout to apply when establishing connections to the helper for HTTP requests, in seconds. See
+# https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connect_timeout for
+# details. (optional)
+http_request_connection_timeout_secs: 10
```
branlwyd, line 93: This may be getting ahead of ourselves, but what do you think about naming these settings based on the type of request they apply to? We can kinda-sorta get there since each job type has one type of request[1], but that requires operators to know the type of request each job sends.

[1] Sort of -- I suppose we might eventually view "aggregate initialization" as separate from "aggregate continuation", giving the agg job driver two kinds of requests. This practically doesn't matter right now, since we aren't deplying multi-step VDAFs at the moment.

branlwyd (approved):

docs/samples/advanced_config/aggregation_job_driver.yaml
```diff
@@ -87,6 +87,17 @@ worker_lease_clock_skew_allowance_secs: 60
 # (required)
 maximum_attempts_before_failure: 10
 
+# Timeout to apply when establishing connections to the helper for HTTP requests, in seconds. See
+# https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.connect_timeout for
+# details. (optional)
```
divergentdave, line 92: We should mention the default values here (i.e. omitting the option won't disable timeouts)

divergentdave (approved):

docs/samples/advanced_config/aggregation_job_driver.yaml
tgeoghegan: I think it's better this way. The nice thing about this layout is that we can add the new options to `struct JobDriverConfig` and get them for free in both the collection and aggregation job drivers. Also, these values are currently used to create a `reqwest::Client` that is then used for all HTTP requests made by the job driver. So saying the timeouts apply to all HTTP requests is actually accurate.

PR #2633, opened 2024-02-05
Add customized Prio3SumVec VDAF

divergentdave:
This adds support for a customized variant of Prio3. It uses a smaller field and more FLP proofs, and uses a different XOF. I chose the same algorithm ID that Daphne uses, from the private use range, so we can interoperate. This VDAF needs a different verify key length, but fortunately this required just a bit more plumbing to support. Unlike in Daphne, I hardcoded the number of proofs to 3 (the minimum allowed number when using Field64) to avoid misuse with choices that could compromise robustness.

I also noticed that the `vdaf_dispatch!()` macro no longer gets called in the mode that doesn't construct the VDAF, so I dropped a bunch of code before adding this VDAF to it.

### DIFF
--- a/core/src/vdaf.rs
+++ b/core/src/vdaf.rs
@@ -1,10 +1,27 @@
 use derivative::Derivative;
 use janus_messages::taskprov;
+use prio::{
+    field::Field64,
+    flp::{
+        gadgets::{Mul, ParallelSum},
+        types::SumVec,
+    },
+    vdaf::{prio3::Prio3, xof::XofHmacSha256Aes128, VdafError},
+};
 use serde::{Deserialize, Serialize};
 use std::str;
 
-/// The length of the verify key parameter for Prio3 & Poplar1 VDAF instantiations.
+/// The length of the verify key parameter for Prio3 and Poplar1 VDAF instantiations using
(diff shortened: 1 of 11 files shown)


core/src/vdaf.rs
```diff
@@ -1,11 +1,27 @@
...
-/// The length of the verify key parameter for Prio3 & Poplar1 VDAF instantiations.
+/// The length of the verify key parameter for Prio3 and Poplar1 VDAF instantiations using
+/// [`XofTurboShake128`][prio::vdaf::xof::XofTurboShake128].
 pub const VERIFY_KEY_LENGTH: usize = 16;
 
+/// Private use algorithm ID for a customized version of Prio3SumVec. This value was chosen for
+/// interoperability with Daphne.
+const ALGORITHM_ID_PRIO3_SUM_VEC_FIELD64_MULTIPROOF_HMACSHA256_AES128: u32 = 0xFFFF_1003;
+
+/// The length of the verify key parameter when using [`XofHmacSha256Aes128`]. This XOF is not part of the VDAF specification.
```
tgeoghegan, line 22: nit: wrap at 100 lines

tgeoghegan (approved):

PR #2637, opened 2024-02-05
Remove Prio3CountVec

### LINKED ISSUE #2608: Remove `Prio3CountVec`
Janus implements `Prio3CountVec` as a 1-bit `Prio3SumVec`, and the `Prio3CountVec` VDAF isn't specified or required anywhere. We should use the Janus 0.7 breaking change window to remove `Prio3CountVec`.

divergentdave:
This removes Prio3CountVec throughout. It was formerly polyfilled using Prio3SumVec inside the aggregator. Prio3CountVec is removed from the list of supported VDAFs exposed via the aggregator API, so the control plane should handle this change gracefully. I replaced Prio3CountVec with Prio3SumVec when it was incidentally used in tests.

Closes #2608. Stacked on #2633.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -858,15 +858,6 @@ impl<C: Clock> TaskAggregator<C> {
                 VdafOps::Prio3Count(Arc::new(vdaf), verify_key)
             }
 
-            VdafInstance::Prio3CountVec {
-                length,
-                chunk_length,
-            } => {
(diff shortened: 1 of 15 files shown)


tgeoghegan (approved):
Looks good! Just confirming: we won't need any divviup-api changes because we've updated the list of supported VDAFs, so creating a `Prio3CountVec` task will be impossible in aggregators with this change?

divergentdave:
The control plane server checks new task requests against the list of supported VDAFs, and the UI currently only shows "sum", "histogram", and "count" if both aggregators support them.

PR #2639, opened 2024-02-05
[0.6] Time-interval: create aggregation jobs touching fewer batches.

branlwyd:
The implementation strategy is to consider only a given batch at a time, generating as many aggregation jobs as possible. Only once there are too few reports from the batch remaining to generate any more aggregation jobs do we move on to the next batch, rolling in the leftovers from the current batch (effectively smearing across batches when necessary).

Note that this will not necessarily avoid multi-batch aggregation jobs in all cases. For example, if aggregation jobs are generated with sizes in [5, 10], and there are 11 reports available for aggregation in the current batch, this logic will generate a single batch of size 10, rolling the leftover report into the next batch. It is instead possible to generate two aggregation jobs of (say) 5 and 6, with no leftovers to be rolled into the next batch. I'm OK with this, as we generally want to create larger aggregation jobs when possible, but we may want to lean even more towards avoiding multi-batch aggregation jobs in the future.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_creator.rs
+++ b/aggregator/src/aggregator/aggregation_job_creator.rs
@@ -5,6 +5,7 @@ use fixed::{
     FixedI16, FixedI32,
 };
 use futures::future::try_join_all;
+use itertools::Itertools as _;
 use janus_aggregator_core::{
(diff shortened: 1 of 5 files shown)


branlwyd:
Part of #2603.

divergentdave (approved):

PR #2641, opened 2024-02-06
remove direct dependency on routefinder

jbr:


branlwyd (approved):
Thanks for this PR!

BTW, I'm curious why a direct dependency on `routefinder` is not intended usage -- is this effectively a "private" part of Trillium's API that Trillium reserves the right to change in a backwards-incompatible way?

jbr:
> Thanks for this PR!
> 
> BTW, I'm curious why a direct dependency on `routefinder` is not intended usage -- is this effectively a "private" part of Trillium's API that Trillium reserves the right to change in a backwards-incompatible way?

Exactly that. https://github.com/trillium-rs/trillium/pull/551 hides routefinder so it's fully protected as an internal implementation detail

jbr:
trillium-router should even be able to wholesale-replace routefinder with a different implementation as long as the resolved routes remain the same. But more topically, the trillium-router interface is failing if it's necessary to access Captures directly

divergentdave (approved):

PR #2644, opened 2024-02-06
Cancel on disconnect

jbr:
This uses two different approaches to the new cancel-on-disconnect api. I'm interested in feedback on which of them y'all prefer, and I can convert whichever to the other or leave them as is.

aggregator-api uses trillium-api's cancel on disconnect handler, which requires extracting all context from the conn prior to running the handler.

The DAP endpoints use Conn::cancel_on_disconnect for a less invasive change.

Since these endpoints really only have one await point each, they are entirely equivalent in terms of cancellation behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,7 +63,7 @@ trillium-head = "0.2.1"
 trillium-opentelemetry = "0.4.0"
 trillium-router = "0.3.6"
 trillium-rustls = "0.4.2"
-trillium-testing = "0.5.0"
+trillium-testing = "0.5.4"
 trillium-tokio = "0.3.4"
 url = { version = "2.5.0", features = ["serde"] }
 uuid = { version = "1.7.0", features = ["v4"] }
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1172,6 +1172,7 @@ enum VdafOps {
 /// length, also for explicitly specifying type parameters.
 macro_rules! vdaf_ops_dispatch {
     ($vdaf_ops:expr, ($vdaf:pat_param, $verify_key:pat_param, $Vdaf:ident, $VERIFY_KEY_LENGTH:ident, $dp_strategy:ident, $DpStrategy:ident) => $body:tt) => {
+        #[allow(unused_braces)]
         match $vdaf_ops {
             crate::aggregator::VdafOps::Prio3Count(vdaf, verify_key) => {
                 let $vdaf = vdaf;
--- a/aggregator/src/aggregator/error.rs
+++ b/aggregator/src/aggregator/error.rs
@@ -144,7 +144,7 @@ pub enum Error {
     /// An error occurred when trying to ensure differential privacy.
     #[error("differential privacy error: {0}")]
     DifferentialPrivacy(VdafError),
-    /// the http client disconnected while we were still executing
+    /// The http client disconnected while we were still executing.
     #[error("http client disconnect")]
     ClientDisconnected,
 }
--- a/aggregator/src/aggregator/error.rs
+++ b/aggregator/src/aggregator/error.rs
@@ -144,6 +144,9 @@ pub enum Error {
     /// An error occurred when trying to ensure differential privacy.
     #[error("differential privacy error: {0}")]
     DifferentialPrivacy(VdafError),
+    /// the http client disconnected while we were still executing
+    #[error("http client disconnect")]
+    ClientDisconnected,
 }
 
 /// Contains details that describe the report and why it was rejected.
@@ -307,6 +310,7 @@ impl Error {
             Error::BadRequest(_) => "bad_request",
             Error::InvalidTask(_, _) => "invalid_task",
             Error::DifferentialPrivacy(_) => "differential_privacy",
+            Error::ClientDisconnected => "client_disconnected",
         }
     }
 }
--- a/aggregator/src/aggregator/http_handlers.rs
+++ b/aggregator/src/aggregator/http_handlers.rs
@@ -27,7 +27,7 @@ use std::{borrow::Cow, time::Duration as StdDuration};
 use std::{io::Cursor, sync::Arc};
 use tracing::warn;
 use trillium::{Conn, Handler, KnownHeaderName, Status};
-use trillium_api::{api, State};
+use trillium_api::{api, State, TryFromConn};
 use trillium_caching_headers::CacheControlDirective;
 use trillium_opentelemetry::metrics;
 use trillium_router::{Router, RouterConnExt};
@@ -152,6 +152,7 @@ impl Handler for Error {
                 &ProblemDocument::new_dap(DapProblemType::InvalidTask).with_task_id(task_id),
             ),
             Error::DifferentialPrivacy(_) => conn.with_status(Status::InternalServerError),
+            Error::ClientDisconnected => conn,
         };
 
         if matches!(conn.status(), Some(status) if status.is_server_error()) {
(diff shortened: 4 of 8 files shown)


aggregator/src/aggregator.rs
```diff
@@ -1172,6 +1172,7 @@ enum VdafOps {
 /// length, also for explicitly specifying type parameters.
 macro_rules! vdaf_ops_dispatch {
     ($vdaf_ops:expr, ($vdaf:pat_param, $verify_key:pat_param, $Vdaf:ident, $VERIFY_KEY_LENGTH:ident, $dp_strategy:ident, $DpStrategy:ident) => $body:tt) => {
+        #[allow(unused_braces)]
```
jbr, line 1175: This was unrelated but the warnings seemed spurious

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -168,7 +168,7 @@ impl Handler for Error {
 /// Firefox has the highest Max-Age cap, at 24 hours, so we use that. Our CORS preflight handlers
 /// are tightly scoped to relevant endpoints, and our CORS settings are unlikely to change.
 /// See: <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age>.
-const CORS_PREFLIGHT_CACHE_AGE: u32 = 24 * 60 * 60;
+const CORS_PREFLIGHT_CACHE_AGE: &'static str = "86400"; //24 * 60 * 60;
```
jbr, line 171: This change is a infinitesimal small performance improvement, but also avoids calling `.to_string()`

Cargo.toml
```diff
@@ -63,7 +63,7 @@ trillium-head = "0.2.1"
 trillium-opentelemetry = "0.4.0"
 trillium-router = "0.3.6"
 trillium-rustls = "0.4.2"
-trillium-testing = "0.5.0"
+trillium-testing = "0.5.4"
```
jbr, line 66: This is necessary to make trillium-testing's TestConn not report closure when the body has been read to completion

aggregator_api/src/lib.rs
```diff
@@ -220,27 +229,3 @@ impl Handler for Error {
         .halt()
     }
 }
-
-trait ConnExt {
-    fn task_id_param(&self) -> Result<TaskId, Error>;
```
jbr, line 225: These both have been replaced by TryFromConn implementations

aggregator_api/src/routes.rs
```diff
@@ -52,18 +52,24 @@ pub(super) async fn get_config(
...
+            .transpose()
+            .map(Self)
+            .map_err(|err| Error::BadRequest(format!("Couldn't parse pagination_token: {:?}", err)))
+    }
+}
+
 pub(super) async fn get_task_ids<C: Clock>(
-    conn: &mut Conn,
-    State(ds): State<Arc<Datastore<C>>>,
+    (LowerBound(lower_bound), ds): (LowerBound, Store<C>),
```
jbr, line 71: as this is the first instance of this: I replaced a lot of restructuring of State and Json anywhere the contents don't need to be moved out. It's less verbose to dereference through the newtype

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -376,40 +385,37 @@ async fn hpke_config<C: Clock>(
 
 /// Handler for CORS preflight requests to "/hpke_config".
 async fn hpke_config_cors_preflight(mut conn: Conn) -> Conn {
-    conn.headers_mut().insert(KnownHeaderName::Allow, "GET");
     if let Some(origin) = conn.request_headers().get(KnownHeaderName::Origin) {
         let origin = origin.clone();
-        let request_headers = conn.headers_mut();
```
jbr, line 382: Fixed a typo

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -168,7 +168,7 @@ impl Handler for Error {
 /// Firefox has the highest Max-Age cap, at 24 hours, so we use that. Our CORS preflight handlers
 /// are tightly scoped to relevant endpoints, and our CORS settings are unlikely to change.
 /// See: <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age>.
-const CORS_PREFLIGHT_CACHE_AGE: u32 = 24 * 60 * 60;
+const CORS_PREFLIGHT_CACHE_AGE: &str = "86400"; //24 * 60 * 60;
```
tgeoghegan, line 171: Let's delete the trailing comment and fold it into the doccomment, like
> Firefox has the highest Max-Age cap, at 34 hours, so we use that (86400 = 24 * 60 * 60).

tgeoghegan (approved):
Seems fine to me. It'd be nice if we had a test to exercise this but it might be hard to write. Can `trillium_testing` simulate a client closing a connection?

aggregator/src/aggregator/http_handlers.rs
jbr: Fixed in 4d1ba6e03e74964e9c97d1c936aefc3dcce2d365

jbr:
> Seems fine to me. It'd be nice if we had a test to exercise this but it might be hard to write. Can `trillium_testing` simulate a client closing a connection?

Yes, that's possible but we'd need to close it at the right time, which is difficult to arrange without injecting test specific code into the handlers. If the transport is closed before the handler starts, we wouldn't be able to read and deserialize the request body and it would fail earlier. We could potentially test it specifically for an endpoint that doesn't expect a request body, but that will only exercise the first await point. There may be any number of other await points inside of the top level future that might get canceled in reality.

The behavior itself is tested [here](https://github.com/trillium-rs/trillium/blob/main/trillium/tests/liveness.rs) and can be manually live-tested with [this example](https://github.com/trillium-rs/trillium/blob/main/trillium/examples/liveness.rs) and [this example](https://github.com/trillium-rs/trillium/blob/main/api/examples/cancel_on_disconnect.rs), but this is entirely new functionality and suggestions of further testing approaches are very welcome.

The best way to exercise this in janus would be to set up a test harness that disconnects at various times (or after a random number of bytes), and then asserts that everything failed in an app-specific okay way (transactions rolled back, no deadlocks, etc). I'd happily set up the first part but I don't have context to make the right assertions to ensure that everything is left in an okay state. This could even try failing at every await point in sequence to fully cover the failure points, for extra confidence.

tgeoghegan:
Yeah, I wouldn't want to block this change on the necessary test support.

tgeoghegan:
Any blockers to merging this @jbr? If not I can click the button.

jbr:
No blockers, although stylistically I was figuring we might want further iteration since the two apis implement this with different approaches and I wasn't sure which y'all would prefer. That could also be done as a subsequent PR, as the behavior would be identical

tgeoghegan:
I guess I prefer the `instrumented(cancel_on_disconnect(whatever))` pattern as it intrudes less on application logic.

aggregator/src/aggregator/error.rs
```diff
@@ -144,6 +144,9 @@ pub enum Error {
     /// An error occurred when trying to ensure differential privacy.
     #[error("differential privacy error: {0}")]
     DifferentialPrivacy(VdafError),
+    /// the http client disconnected while we were still executing
```
branlwyd, line 147:
```suggestion
    /// The http client disconnected while we were still executing.
```

(nit, matching the style of other doccomments)

branlwyd (commented):
I also prefer the instrumented version pretty strongly -- other than not imposing on application logic, it provides a stronger guarantee that later changes to the program to introduce new `await` points can't forget to call `cancel_on_disconnect`.

branlwyd:
Obsoleted by #3119.

PR #2650, opened 2024-02-08
Fix clippy lints

divergentdave:


tgeoghegan (approved):
The lint this fixes is https://rust-lang.github.io/rust-clippy/master/index.html#/blocks_in_conditions? I'm not sure I agree with clippy here, but the new code is fine, too, and it's not worth opting out of the lint.

PR #2657, opened 2024-02-09
Fix fanning out transaction error from ReportWriteBatcher

### LINKED ISSUE #2655: "cannot call init when there are already clones of an Arc<Handler>"
We're seeing async tasks panic with "cannot call init when there are already clones of an Arc<Handler>" intermittently. This panic is inside trillium, at `trillium-0.2.15/src/handler.rs:218:14`. The panic happens repeatedly, but is intermixed with log messages from requests handled successfully. It's possible this is interrupting incoming requests, and causing liveness checks to fail.

divergentdave:
This fixes #2655. Transaction-level errors in ReportWriteBatcher were resulting in panics during response handling. I think that database connection pool timeouts were the most likely cause of such errors, during periods of high load. This PR adds a unit test simulating this condition, and fixes the panic in two ways, by upgrading trillium, and by using a newtype around `Arc<Error>` as a `Handler`, allowing us to specify our own (no-op) init method.

Reviewer note: I recommend flipping "ignore whitespace" on, because I turned an existing method into a function, and an existing function into a method.

### DIFF
--- a/aggregator/src/aggregator/error.rs
+++ b/aggregator/src/aggregator/error.rs
@@ -12,6 +12,8 @@ use prio::{topology::ping_pong::PingPongError, vdaf::VdafError};
 use std::{
     fmt::{self, Display, Formatter},
     num::TryFromIntError,
+    ops::Deref,
+    sync::Arc,
 };
 use tracing::info;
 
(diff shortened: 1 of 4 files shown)


aggregator/src/aggregator/error.rs
```diff
@@ -146,6 +148,25 @@ pub enum Error {
     DifferentialPrivacy(VdafError),
 }
 
+/// A newtype around `Arc<Error>`. This is needed to host a customized implementation of
+/// `trillium::Handler`.
+#[derive(Clone)]
+pub(crate) struct ArcError(Arc<Error>);
```
jbr, line 154: Is there any reason to delegate Display to the inner Error and impl Error for ArcError?

jbr: Would a more descriptive name explain to future readers _why_ it's wrapped in an ArcError instead of just using the Error directly? Along the lines of `BatchFailureFanoutError`? I was very confused when I encountered `Arc<Error>` initially and wouldn't be any less confused about `ArcError`

divergentdave: I don't think it would matter much, I only convert to this in one place (the `upload` ApiHandler function), and the codebase cares more about the Debug output of errors than Display. (This also doesn't delegate Debug, but nothing requires that)

tgeoghegan (approved):

Issue #2605, opened 2024-02-02
Leader: improve retry strategy for aggregation jobs


branlwyd:
Currently, aggregation jobs will be retried a fixed number of times. If the Helper succeeds at writing the results of the aggregation job back to its datastore, but the Leader does not receive the response (e.g. due to transient communication failures), the Leader & Helper will be out-of-sync and the batches associated with the aggregation job will be uncollectable/lost.

If we can detect when a request to the Helper fails due to timeout or other transient conditions (rather than due to a DAP-level problem with the aggregation job itself), we can retry forever to effectively solve the issue. We would need to be very careful to only infinitely retry on transient failures, and not permanent failures, however.

Before implementing this, we should consider improving our retry strategy to avoid overloading the Helper with retries.

divergentdave:
Beyond retrying aggregation jobs, we may want to stop marking collection jobs as collectable when abandoning related aggregation jobs. Currently, changing from IN_PROGRESS to ABANDONED results in a decrement of outstanding_aggregation_jobs, which eventually results in the batch becoming CLOSING/CLOSED. If we have ever sent a request for an aggregation job, but have not received an explicit success or failure response from the helper, then the helper's accumulators are in an indeterminate state, from the leader's perspective.

branlwyd:
Unassigning for now: retries for Leader -> Helper requests will help a lot. We might want to do more (such as bumping the number of retries, or improving our transient-failure detection, or adding backoff to job retries). But I think we should do so only if we observe further problems.

PR #2683, opened 2024-02-13
Add retries to Leader -> Helper requests.

branlwyd:
This affects requests made by the aggregation & collection job drivers. These retries are enabled by default, but can be disabled by setting `retry_max_elapsed_time_millis` to 0.

Also, update our `retry_http_request` functionality to retry if there is a retryable failure while reading the response body. Before, all uses of `retry_http_request` would not retry on any failure encountered while reading the request body.

Also, update the semantics of `retry_http_request` to match the documented semantics:
 * On success, return the response (now represented as an `HttpResponse`).
 * On an HTTP-level error, return an `HttpErrorResponse` (this used to be a `reqwest::Response`, but all callers parsed it into an `HttpErrorResponse`).
 * On a network-level error, return the `reqwest::Error` directly.

Part of #2605.


aggregator/src/aggregator.rs
```diff
@@ -3360,16 +3361,74 @@ fn empty_batch_aggregations<
...
+
+struct RequestTimer<'a> {
+    // Mutable state.
+    start: SyncMutex<Option<Instant>>,
+
+    // Immutable state.
+    http_request_duration_histogram: &'a Histogram<f64>,
+    domain: String,
+    endpoint: String,
+    method: String,
```
tgeoghegan, line 3378: nit: I suspect we could have `endpoint: &'static str` and avoid a copy. Harder for `method`, but since the value comes from [`Method::as_str`](https://docs.rs/reqwest/latest/reqwest/struct.Method.html#method.as_str), this could be `method: &'a str` without too much trouble.

aggregator/src/config.rs
```diff
@@ -162,16 +168,49 @@ pub struct JobDriverConfig {
...
+                self.retry_max_elapsed_time_millis,
+            )))
+            .build()
+    }
+
     fn default_http_connection_timeout() -> u64 {
         10
     }
 
     fn default_http_request_timeout() -> u64 {
```
tgeoghegan, line 199:
```suggestion
    fn default_http_connection_timeout_secs() -> u64 {
        10
    }

    fn default_http_request_timeout_secs() -> u64 {
```

I don't love that some values are expressed in seconds and some are in milliseconds, but I'm also not sure it's worth changing. However, we should rename these existing functions to disambiguate.

collector/src/lib.rs
```diff
@@ -124,10 +124,9 @@ pub enum Error {
 }
 
 impl Error {
-    /// Construct an error from an HTTP response's status and problem details document, if present
-    /// in the body.
-    async fn from_http_response(response: Response) -> Error {
-        Error::Http(Box::new(HttpErrorResponse::from_response(response).await))
+    /// Construct an error from an HTTP error response.
+    fn from_http_error_response(http_error_response: HttpErrorResponse) -> Error {
```
tgeoghegan, line 128: I'm not sure, but I suspect the only reason the previous version was a custom impl is because it was async. Now that this is an infallible, sync conversion from `HttpErrorResponse` to `Error`, can we make it `impl From<HttpErrorResponse> for Error`?

core/Cargo.toml
```diff
@@ -35,6 +35,7 @@ anyhow.workspace = true
 assert_matches = { version = "1", optional = true }
 backoff = { version = "0.4.0", features = ["tokio"] }
 base64.workspace = true
+bytes = "1"
```
tgeoghegan, line 38: nit: I'd prefer to see the dep hoisted up into the top-level `Cargo.toml` and replaced with `bytes.workspace = true` here and in `aggregator/Cargo.toml`.

core/src/retries.rs
```diff
@@ -37,83 +40,165 @@ pub fn http_request_exponential_backoff() -> ExponentialBackoff {
...
-pub fn test_http_request_exponential_backoff() -> ExponentialBackoff {
-    ExponentialBackoff {
-        initial_interval: Duration::from_nanos(1),
-        max_interval: Duration::from_nanos(30),
-        multiplier: 2.0,
-        max_elapsed_time: Some(Duration::from_millis(100)),
-        ..Default::default()
+/// HttpResponse represents an HTTP response. It will typically be returned from
+/// [`retry_http_request`].
+#[derive(Debug)]
```
tgeoghegan, line 45: Any reason not to derive `Clone`? I feel like I never regret deriving `Clone` if it doesn't require any extra work.

core/src/retries.rs
```diff
@@ -142,28 +288,25 @@ mod tests {
             .mock("GET", "/")
             .with_status(StatusCode::NOT_FOUND.as_u16().into())
             .with_header("some-header", "some-value")
-            .with_body("some-body")
+            .with_body("some-body") // once told me
```
tgeoghegan, line 291: going to pretend I didn't see this

tgeoghegan (approved):

aggregator/src/config.rs
branlwyd: I'm also a little uncomfortable with different units, but I do think millis is appropriate for a retry interval for an HTTP request.

We could freely change it in this version, but since I'm going to backport this to `release/0.6` things would get tricky there.

Issue #2360, opened 2023-12-14
get_task_metrics() is expensive and must be destroyed


divergentdave:
Here's a recently executed query plan:

![image](https://github.com/divviup/janus/assets/181772/04cfe214-7a77-4fb1-b2c8-302ef77e89a3)

inahga:
From discussion [out of band](https://isrg.slack.com/archives/C0167LT4C73/p1706030221975669), we probably will want to delete this function as-it-is. While it has information that might be useful, it is slow for large tasks even with index scans, there may be better ways to get the same information, and it isn't currently useful for billing purposes.

tgeoghegan:
Elsewhere we have work underway (#2293, #2508, #2537, #2553, #2513) to introduce a superior task metrics solution. So I think this issue only tracks deleting the old solution once the new one is deployed.

Issue #2588, opened 2024-01-30
Breaking changes for Janus 0.7.0


tgeoghegan:
We still have the opportunity to make breaking changes to Janus 0.7.0 as we have only released prerelease versions of it so far. Let's break public Rust API, database schema, emitted binaries!

- [x] <s>https://github.com/divviup/janus/issues/2312</s>
- [x] https://github.com/divviup/janus/issues/2236
- [x] https://github.com/divviup/janus/issues/2251
- [x] https://github.com/divviup/janus/issues/2329
- [x] <s>https://github.com/divviup/janus/issues/2463</s>
- [x] https://github.com/divviup/janus/issues/2608

inahga:
I landed a few breaking changes. Here's some summaries for the release notes:

- `DivviUpHpkeConfig` has been removed. Use `PrivateCollectorCredential` instead.
- `PrivateCollectorCredential::hpke_keypair()` has been made infallible. `PrivateCollectorCredential::hpke_keypair_infallible()` has been removed, replace it with `PrivateCollectorCredential::hpke_keypair()`.
- The `token` field of `PrivateCollectorCredential` has been made non-optional. If your credentials are missing this field, you should regenerate the credentials with a newer version of `divviup-cli`. Alternatively, you can edit the credential manually to supply the `token` field.
- The aggregator configuration field `max_job_discovery_delay_secs` used by job drivers has been fully removed. Replace it with `job_discovery_interval_secs`.

divergentdave:
Now would be a good opportunity to clean up Prio3CountVec support, since Prio3SumVec has been properly specified for a while.

tgeoghegan:
We're keen to release Janus 0.7.0, and have agreed to punt #2463 to a later breaking period. #2312 can be done in a backward compatible way, so it's not a breaking change and can be taken at any time during the 0.7.0 release cycle.

tgeoghegan:
We released [Janus 0.7.0](https://github.com/divviup/janus/releases/tag/0.7.0), so we're done here.

PR #2686, opened 2024-02-14
Remove get_task_metrics()

inahga:
Closes https://github.com/divviup/janus/issues/2360. Supports https://github.com/divviup/janus/issues/2588.

n.b. I won't backport this to release/0.6 to avoid breaking changes.


inahga:
There are no compatibility issues with divviup-api, since it doesn't use that API any more.

tgeoghegan (approved):

Issue #2654, opened 2024-02-09
Fix and reenable task upload counters


tgeoghegan:
We deployed accurate per-task metrics using the `task_upload_counters` table and found that it created performance problems for report uploads. We made a hotfix in #2656 that removes the calls to `Transaction::increment_task_upload_counter`. We should diagnose the problem and re-enable the feature. This time, we should make sure to wire up a feature flag/configuration option that would let us disable this without requiring code changes.

---
So it seems these improvements are necessary:
- [ ] Reduce statement count by batching increments.
- [ ] Adjust `fillfactor` to leave per-page space for HOT updates.
- [ ] Lock task stats behind a feature flag so we can turn it off in a pinch if need be.

tgeoghegan:
What we observed is that the `INSERT INTO task_upload_counters ...` query was top of the list in GCP query insights. It's spending all its time waiting to acquire locks. The query in question is:

```
INSERT INTO
  task_upload_counters (task_id,
    ord,
    report_success)
VALUES
  ((
    SELECT
      id
    FROM
      tasks
    WHERE
      task_id = $1), $2, $3)
ON
  CONFLICT (task_id,
    ord) DO
UPDATE
SET
  report_success = task_upload_counters.report_success + $4
```

I guess the issue is that if you have many uploads coming in at the same time for a single task, then they'll all try to write to the same row of `task_upload_counters`, and will lock it, whereas the corresponding writes to `report_uploads` or `report_aggregations` only affect a single row. @divergentdave suggested batching up writes to the task counters (i.e. increment them by 500 or 1,000 at once) so that we do fewer database transactions.

divergentdave:
Each batch of reports will only write to upload counter rows with one particular `ord` value, and all of the counter increments happen in the same transaction as the report insertions. My suggestion was more aimed at reducing the number of statements per transaction, and the number of times we update (and possibly lock) the same tuples.

inahga:
I think @divergentdave is right about heap-only tuple chains becoming too long (thx for the breadcrumb in slack).

Looking at some charts of the incident.
<img width="1775" alt="image" src="https://github.com/divviup/janus/assets/57697428/116294c2-0f34-4e14-b786-bb8c3b7435c9">

In the beginning, HOT updates dominate row updates. At a certain threshold, non-HOT updates become more frequent.

[...]

PR #2693, opened 2024-02-16
Fix and reenable task upload counters (#2684, main forward-port)

inahga:
Supports https://github.com/divviup/janus/issues/2654.

This is not locked behind a feature flag, as it doesn't seem necessary to me since this a prerelease branch, and we don't have any production instances of it.


db/00000000000001_initial_schema.up.sql
```diff
@@ -151,7 +151,7 @@ CREATE TABLE task_upload_counters(
 
     CONSTRAINT fk_task_id FOREIGN KEY(task_id) REFERENCES tasks(id) ON DELETE CASCADE,
     CONSTRAINT task_upload_counters_unique UNIQUE(task_id, ord)
-);
+) WITH (fillfactor = 50);
```
tgeoghegan, line 154: Let's copy the comment from https://github.com/divviup/janus/blob/release/0.6/db/00000000000004_task_upload_counters_fillfactor.up.sql explaining the `fillfactor` value here.

tgeoghegan (approved):

PR #2727, opened 2024-02-21
Copy report shares within database when creating aggregation jobs

### LINKED ISSUE #2689: Aggregation job creator: copy report data directly from `client_reports` to `report_aggregations`, rather than reading full reports
Currently, the aggregation job creator will read a large number of reports (5000, at time of writing) in order to create aggregation jobs. For VDAFs with large reports, this can require a significant amount of memory.

Instead, the aggregation job creator could read report IDs & other (small) metadata required to generate reports, and use a SQL query which causes Postgres to directly copy the data from the relevant `client_reports` row into the relevant `report_aggregations` row. This would decouple the memory usage of the aggregation job creator from the report size of the relevant VDAFs.

divergentdave:
This refactors the aggregation job writer, providing separate APIs for creating versus updating aggregation jobs, and changes aggregation job creation so that it does not read entire reports into memory, or write them back out from meory, but instead copies directly between the two database tables. This should improve the performance of the aggregation job creator, which we haven't yet scaled horizontally, and will mainly benefit tasks with very large VDAF messages.

Behavior differences between new aggregation jobs and updated aggregation jobs are controlled through a combination of static dispatch with traits and conditionals. There's one trait for updating a batch to reflect aggregation job changes, and another that abstracts over `ReportAggregation` and the new `ReportAggregationMetadata`. Amidst the aggregation job writer refactor, I moved up the check against updates to a garbage-collected batch -- this now happens at the top of the loop over batch identifiers, rather than within the nested loop over aggregation job updates. This can happen because we now require that all aggregation job "operations" are the same, and so we know whether to expect all referenced batches to already exist.

Closes #2689.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_writer.rs
+++ b/aggregator/src/aggregator/aggregation_job_writer.rs
@@ -70,14 +70,11 @@ where
     /// operation (aggregation into a collected batch is not allowed). These report aggregations
     /// will be written with a `Failed(BatchCollected)` state, and the associated report IDs will be
     /// returned.
-    pub async fn write<'a, C: Clock>(
+    pub async fn write<C: Clock>(
         &self,
         tx: &Transaction<'_, C>,
         vdaf: Arc<A>,
-    ) -> Result<(), Error>
-    where
-        A: 'a,
-    {
+    ) -> Result<(), Error> {
         self.updates
             .write(tx, vdaf, NewAggregationJobBatchUpdate)
             .await?;
--- a/aggregator/src/aggregator/aggregation_job_writer.rs
+++ b/aggregator/src/aggregator/aggregation_job_writer.rs
@@ -690,7 +690,7 @@ where
             .collect();
 
         // Read all relevant batches and report counts from the datastore.
-        let (batches, batches_with_reports) = try_join!(
+        let (batches, batches_with_unaggregated_reports) = try_join!(
             try_join_all(
                 updates
                     .by_batch_identifier_index
@@ -725,8 +725,10 @@ where
             })
             .collect();
 
-        let batches_with_unaggregated_reports: HashSet<_> =
-            batches_with_reports.into_iter().flatten().collect();
+        let batches_with_unaggregated_reports: HashSet<_> = batches_with_unaggregated_reports
+            .into_iter()
+            .flatten()
+            .collect();
 
         Ok(Self {
             task: &updates.task,
--- a/aggregator/src/aggregator/aggregation_job_writer.rs
+++ b/aggregator/src/aggregator/aggregation_job_writer.rs
@@ -76,7 +76,7 @@ where
         vdaf: Arc<A>,
     ) -> Result<(), Error> {
         self.updates
-            .write(tx, vdaf, NewAggregationJobBatchUpdate)
+            .write::<C, NewAggregationJobBatchUpdate>(tx, vdaf)
             .await?;
         Ok(())
     }
@@ -88,7 +88,6 @@ struct NewAggregationJobBatchUpdate;
 
 impl BatchUpdateCallback for NewAggregationJobBatchUpdate {
     fn update_batch<'a, const SEED_SIZE: usize, Q, A, RA>(
-        &self,
         batch: Batch<SEED_SIZE, Q, A>,
         aggregation_jobs: impl Iterator<
             Item = (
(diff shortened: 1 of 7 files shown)


inahga (approved):
Nice refactor, the writer is a lot easier to reason about now.

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -593,6 +570,442 @@ impl<const SEED_SIZE: usize, Q: CollectableQueryType, A: vdaf::Aggregator<SEED_S
...
+/// Generic callback used in the internals of aggregation job writers. Different implementations are
+/// used when creating new aggregation jobs versus updating existing aggregation jobs.
+trait BatchUpdateCallback {
+    /// This takes one batch and an iterator of aggregation jobs, and returns that batch updated to
+    /// reflect changes due to the aggregation job. Particularly, it will update the
+    /// `outstanding_aggregation_jobs` counter and the `client_timestamp_interval`.
+    fn update_batch<'a, const SEED_SIZE: usize, Q, A, RA>(
+        &self,
+        batch: Batch<SEED_SIZE, Q, A>,
+        aggregation_jobs: impl Iterator<
```
tgeoghegan, line 586: nit: I can't find a Rust style guide that discusses it but something bugs me about mixing `impl Trait` notation with a `where` clause in an item. I feel like we should use one or the other.

tgeoghegan (approved):

aggregator/src/aggregator/aggregation_job_writer.rs
divergentdave: I think I'd prefer to keep the `impl Iterator` for two reasons, firstly because we already have several generic parameters, and I'd like to avoid another. Secondly, the alternative would introduce a type variable on one line, use it on a second line, and bound the type variable on a third, which adds extra indirection when trying to read and understand the method's signature.

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -593,6 +570,442 @@ impl<const SEED_SIZE: usize, Q: CollectableQueryType, A: vdaf::Aggregator<SEED_S
...
+            ),
+        >,
+    ) -> Result<Batch<SEED_SIZE, Q, A>, Error>
+    where
+        Q: CollectableQueryType,
+        A: vdaf::Aggregator<SEED_SIZE, 16> + 'a,
+        RA: ReportAggregationUpdate + Clone + 'a;
+
+    /// Returns a flag indicating whether aggregation jobs are being created or updated.
+    fn creating_aggregation_jobs(&self) -> bool;
```
branlwyd, line 600: nit: I think this would be slightly easier to follow if we returned an enum rather than a bool -- something like `AggregationJobWriteType{ Create, Update }`. Otherwise, the reader has to remember that `true` is creation, `false` is update.

branlwyd (approved):
Nice refactor! Just a few nitpicky comments about code style, the functionality & overall design looks good.

divergentdave:
I put together a quick harness that shards 50 large reports (~800KiB each), puts them in the database, and then runs the aggregation job creator once. I then ran that in DHAT, before and after this change. There is a 13x reduction in allocated memory at t-gmax, plus t-gmax shifted from the end of the program (aggregation job creator) to near the beginning (sharding). Thus, this should resolve our OOM issues with large VDAF parameter sets, and we should be good to increase the "report window" that the aggregation job creator considers.

Before:

```
==73134== Total:     1,738,723,201 bytes in 216,731 blocks
==73134== At t-gmax: 182,086,752 bytes in 2,713 blocks
==73134== At t-end:  446,256 bytes in 812 blocks
==73134== Reads:     17,163,954,389 bytes
==73134== Writes:    16,613,213,709 bytes
```

```
Times {
  t-gmax: 71,882,402,481 instrs (99.92% of program duration)
  t-end:  71,937,071,675 instrs
}
```

After:

```
==74645== Total:     1,269,640,938 bytes in 213,518 blocks
==74645== At t-gmax: 13,599,602 bytes in 2,688 blocks
==74645== At t-end:  445,960 bytes in 810 blocks
==74645== Reads:     16,692,297,395 bytes
==74645== Writes:    16,086,090,205 bytes
```

```
Times {
  t-gmax: 8,474,753,610 instrs (12% of program duration)
  t-end:  70,644,425,617 instrs
}
```

PR #2728, opened 2024-02-22
Add option to ignore unrecognized DpMechanism

divergentdave:
This adds a configuration option inside the `TaskprovConfig` struct to ignore unrecognized DpMechanisms for testing purposes. `janus_messages` is updated to round-trip such mechanisms, rather than fail to decode them.


tgeoghegan (approved):

inahga (approved):
Nice. Glad that having `taskprov_config` its own struct paid off.

PR #2754, opened 2024-03-01
Write upload counters in separate transaction from uploads

inahga:
Supports https://github.com/divviup/janus/issues/2654.

This is a half-measure of decoupling report upload metrics from the uploads themselves, that I'd like to try in production. We can still do more with this decoupling, i.e. batching the uploads separately from `ReportUploadBatcher`.

I've observed high latency spikes in production that are ostensibly coincident with serialization conflicts. I suspect two or more transactions having to retry many INSERTs is proving problematic. Some of the spikes are severe, so I don't think scaling out the shard count is not going ameliorate the 99th percentile as much as I'd like.

image

(I have the narrow hope that this will finally fully fix the HOT update problem, but this is wishful thinking).

### DIFF
--- a/aggregator/src/aggregator/report_writer.rs
+++ b/aggregator/src/aggregator/report_writer.rs
@@ -27,7 +27,7 @@ use tokio::{
     sync::{mpsc, oneshot},
     time::{sleep_until, Instant},
 };
-use tracing::debug;
+use tracing::{debug, error};
 
 type ReportResult<C> = Result<Box<dyn ReportWriter<C>>, ReportRejection>;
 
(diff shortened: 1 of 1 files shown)


inahga:
Keeping this in draft, I want to try messing with `task_counter_shard_count` first.

inahga:
> Keeping this in draft, I want to try messing with task_counter_shard_count first.

This has a moderately positive effect in reducing update contention, but has a negative effect of increasing the number of raw updates, therefore increasing table bloat.

I'd like to take this change, since it's easy. It should have the same effect on update contention and without the table bloat, since a smaller+quicker tx should leave more opportunity for HOT updates to occur and less costly serialization retries.

aggregator/src/aggregator/report_writer.rs
```diff
@@ -185,19 +185,31 @@ impl<C: Clock> ReportWriteBatcher<C> {
...
+                    Box::pin(
+                        async move { task_upload_counters.write(counter_shard_count, tx).await },
+                    )
+                })
+                .await
+                .map_err(|err| {
+                    error!(?err, "Failed to write upload metrics");
+                    err
+                })
+                .ok();
```
divergentdave, line 211: nit: I'd prefer using a `let _res = ...` over `.ok()` to get out of `#[must_use]`.

divergentdave (approved):

PR #2769, opened 2024-03-04
Integration test using dummy VDAF with agg param

tgeoghegan:
Enable Janus to run the dummy VDAF from crate `prio` with arbitrary rounds
(by extending the representation of `VdafInstance::Fake`) and with multiple
aggregation parameters chosen by the test.

To that end, this resurrects some functionality previously deleted in #1508.
At the time, we had that code gated on `#[cfg(test)]`, with some cautions
about its likely performance problems. I'm bringing it back in the release
configuration of Janus. I don't think we need a feature flag for this: the
various `VdafInstance::Fake` variants that would use these codepaths are not
currently compiled into release Janus builds. Even if they were, deployments
can avoid this risky code by simply not using those VDAFs.

This wires up some tests for both time interval and fixed size tasks, running
multiple collections against a single batch with varying aggregation
parameters. However, a few tests are disabled because of stuff that doesn't
work yet:

- creating fixed size aggregation jobs w/ agg param
- using `Vdaf.is_valid` to check whether subsequent collections against a
  single batch are valid

Part of #225

- [x] Release a `prio` with https://github.com/divviup/libprio-rs/pull/970 and depend on that in this PR
- [x] Rebase after we take #2782

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,8 @@ k8s-openapi = { version = "0.20.0", features = ["v1_26"] }  # keep this version 
 kube = { version = "0.87.2", default-features = false, features = ["client", "rustls-tls"] }
 opentelemetry = { version = "0.22", features = ["metrics"] }
 opentelemetry_sdk = { version = "0.22", features = ["metrics"] }
-prio = { version = "0.16.1", features = ["multithreaded", "experimental"] }
+# TODO(timg): go back to released prio
+prio = { git = "https://github.com/divviup/libprio-rs", features = ["multithreaded", "experimental"], rev = "8774573dee1932c19c28ab83173b358b42a52b78" }
 serde = { version = "1.0.197", features = ["derive"] }
 serde_json = "1.0.114"
 serde_test = "1.0.175"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,8 +47,7 @@ k8s-openapi = { version = "0.20.0", features = ["v1_26"] }  # keep this version 
 kube = { version = "0.87.2", default-features = false, features = ["client", "rustls-tls"] }
 opentelemetry = { version = "0.22", features = ["metrics"] }
 opentelemetry_sdk = { version = "0.22", features = ["metrics"] }
-# TODO(timg): go back to released prio
-prio = { git = "https://github.com/divviup/libprio-rs", features = ["multithreaded", "experimental"], rev = "8774573dee1932c19c28ab83173b358b42a52b78" }
+prio = { version = "0.16.2", features = ["multithreaded", "experimental"] }
 serde = { version = "1.0.197", features = ["derive"] }
 serde_json = "1.0.114"
 serde_test = "1.0.175"
--- a/aggregator/src/aggregator/aggregation_job_creator.rs
+++ b/aggregator/src/aggregator/aggregation_job_creator.rs
@@ -19,6 +19,7 @@ use janus_aggregator_core::{
         Datastore,
     },
     task::{self, AggregatorTask},
+    VdafHasAggregationParameter,
 };
 #[cfg(feature = "fpvec_bounded_l2")]
 use janus_core::vdaf::Prio3FixedPointBoundedL2VecSumBitSize;
@@ -31,7 +32,8 @@ use janus_core::{
     },
 };
 use janus_messages::{
-    query_type::TimeInterval, AggregationJobStep, Duration as DurationMsg, Interval, Role, TaskId,
+    query_type::TimeInterval, AggregationJobStep, Duration as DurationMsg, Interval,
+    ReportMetadata, Role, TaskId,
 };
 use opentelemetry::{
     metrics::{Histogram, Meter, Unit},
@@ -387,6 +389,15 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
                 }
             },
 
+            #[cfg(feature = "test-util")]
+            (task::QueryType::TimeInterval, VdafInstance::Fake { rounds }) => {
+                let vdaf = Arc::new(prio::vdaf::dummy::Vdaf::new(*rounds));
+                self.create_aggregation_jobs_for_time_interval_task_with_param::<
+                    0,
+                    prio::vdaf::dummy::Vdaf,
+                >(task, vdaf).await
+            }
+
             (
                 task::QueryType::FixedSize {
                     max_batch_size,
@@ -526,6 +537,18 @@ impl<C: Clock + 'static> AggregationJobCreator<C> {
                             >(task, vdaf, max_batch_size, batch_time_window_size).await
                     }
                 }
+            }
+
+            #[cfg(feature = "test-util")]
+            (
+                task::QueryType::FixedSize {
+                    max_batch_size: _max_batch_size,
+                    batch_time_window_size: _batch_time_window_size,
+                },
+                VdafInstance::Fake { rounds },
+            ) => {
+                let _vdaf = prio::vdaf::dummy::Vdaf::new(*rounds);
+                todo!("wire up call to self.create_aggregation_jobs_for_fixed_size_task_with_param")
             }
 
             _ => {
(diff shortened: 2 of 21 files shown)


aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -2965,23 +3400,36 @@ mod tests {
...
+                tx.verify_client_report_scrubbed(task_id, report_id).await;
+            }
         }
 
-        // Verify that all reports we did not see a report aggregation for are not scrubbed. (We do
-        // so by reading the report, since reading a report will fail if the report is scrubbed.)
-        for report_id in want_ra_states.keys() {
-            if all_seen_report_ids.contains(report_id) {
-                continue;
+        // Verify that if the aggregation parameter is (), all reports we saw a report aggregation
```
branlwyd, line 3420: I think this comment goes above the block above which checks if the report is scrubbed.

aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -2965,23 +3400,36 @@ mod tests {
...
+                        (agg_job.aggregation_parameter() as &A::AggregationParam).clone(),
+                        ra.report_id(),
+                    )
+                })
+            })
             .collect();
-        for report_id in &all_seen_report_ids {
-            tx.verify_client_report_scrubbed(task_id, report_id).await;
+        for (agg_param, report_id) in &all_seen_report_ids {
+            if agg_param.type_id() == ().type_id() {
```
branlwyd, line 3415: nit: a cleaner way to implement this would be with a trait (probably an extension of `VdafHasAggregationParameter`), but I think it's OK for test code.

but maybe an easier-to-follow check would be: `TypeId::of::<A::AggregationParam>() == TypeId::of::<()>()`? (https://doc.rust-lang.org/std/any/struct.TypeId.html) That would make it clearer this is a function of the VDAF in use, rather than something that might change per-iteration.

aggregator_core/src/datastore.rs
```diff
@@ -1204,6 +1204,87 @@ impl<C: Clock> Transaction<'_, C> {
...
+                    ON report_aggs.client_report_id = client_reports.report_id  -- this join is very inefficient, fix before deploying in non-test scenario
+                    AND report_aggs.aggregation_param = collection_jobs.aggregation_param
+                    WHERE client_reports.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                    AND collection_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                    AND collection_jobs.state = 'START'
+                    AND report_aggs.id IS NULL
+                    LIMIT $2::BIGINT
+                ),
+                updated_client_reports AS (
+                    UPDATE client_reports SET aggregation_started = TRUE
```
branlwyd, line 1258: Q: is there a functional need to set these as `aggregation_started`? or is this bookkeeping?

messages/Cargo.toml
```diff
@@ -20,7 +20,9 @@ hex = "0.4"
 num_enum = "0.7.2"
 # We can't pull prio in from the workspace because that would enable default features, and we do not
 # want prio/crypto-dependencies
-prio = { version = "0.16.1", default-features = false, features = ["multithreaded", "experimental"] }
+# TODO(timg) go back to released prio
```
branlwyd, line 23: Note to self: needs to be resolved before this can merge.

Cargo.toml
```diff
@@ -47,7 +47,8 @@ k8s-openapi = { version = "0.20.0", features = ["v1_26"] }  # keep this version
 kube = { version = "0.87.2", default-features = false, features = ["client", "rustls-tls"] }
 opentelemetry = { version = "0.22", features = ["metrics"] }
 opentelemetry_sdk = { version = "0.22", features = ["metrics"] }
-prio = { version = "0.16.1", features = ["multithreaded", "experimental"] }
+# TODO(timg): go back to released prio
```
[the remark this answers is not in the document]
tgeoghegan: Yup, I put a task list in the PR description.

aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -2965,23 +3400,36 @@ mod tests {
...
 
         // Verify that all reports we saw a report aggregation for are scrubbed.
-        let all_seen_report_ids: HashSet<_> = agg_jobs_and_report_ids
+        let all_seen_report_ids: HashSet<(A::AggregationParam, _)> = agg_jobs_and_report_ids
             .iter()
-            .flat_map(|(_, report_aggs)| report_aggs.iter().map(|ra| ra.report_id()))
+            .flat_map(|(agg_job, report_aggs)| {
+                report_aggs.iter().map(|ra| {
+                    (
+                        (agg_job.aggregation_parameter() as &A::AggregationParam).clone(),
```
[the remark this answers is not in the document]
tgeoghegan: You are correct that we don't need the type annotation on the `HashSet<_>` on L3403. However, just as on L3383, I need to cast to `&A::AggregationParam` or the compiler won't let me call `clone`. I don't really understand why, TBH.

aggregator_core/src/datastore.rs
tgeoghegan: No, this just got resurrected from #1508, but it doesn't do anything anymore. I simplified the query considerably.

aggregator_core/src/datastore.rs
```diff
@@ -1204,6 +1204,87 @@ impl<C: Clock> Transaction<'_, C> {
...
+                    AND client_reports.client_timestamp <@ collection_jobs.batch_interval
+                    LEFT JOIN (
+                        SELECT report_aggregations.id, report_aggregations.client_report_id,
+                            aggregation_jobs.aggregation_param
+                        FROM report_aggregations
+                        INNER JOIN aggregation_jobs
+                        ON aggregation_jobs.id = report_aggregations.aggregation_job_id
+                        WHERE aggregation_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
+                    ) AS report_aggs
+                    ON report_aggs.client_report_id = client_reports.report_id  -- this join is very inefficient, fix before deploying in non-test scenario
```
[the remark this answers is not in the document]
tgeoghegan: Good point. I left a comment suggesting that.

aggregator_core/src/datastore.rs
tgeoghegan: I spoke too soon: the integration test fails if we don't set `aggregation-started = TRUE`. I believe I'll need to revisit this as I continue in this area (because in the presence of varying aggregation parameters, a single boolean doesn't suffice) but for now I'll leave the query as-is.

aggregator_core/src/datastore/models.rs
```diff
@@ -362,7 +362,6 @@ pub struct AggregationJob<const SEED_SIZE: usize, Q: QueryType, A: vdaf::Aggrega
     /// The ID of this aggregation job.
     aggregation_job_id: AggregationJobId,
     /// The aggregation parameter this job is run with.
-    #[derivative(Debug = "ignore")]
```
branlwyd, line 365: I think we made this change because the aggregation parameter was potentially sensitive (e.g., IIUC, for Poplar1/Mastic it would store prefixes of measurement values); I think we want to avoid logging aggregation parameter values.

If we really need to log these, maybe we could:
a) log some safe value derived from the aggregation parameter, e.g. a hash of the value
b) require that the `Debug` implementations of aggregation parameters take responsibility for producing non-sensitive debug values in a VDAF-specific way

(b) is the best IMO but also the hardest to implement, and we'd need to be careful not to backslide as new VDAF implementations come around.

aggregator/src/aggregator/aggregation_job_creator.rs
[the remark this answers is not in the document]
branlwyd: I can reproduce this. Very strange, but minor enough not to be worth spending too much time on.

aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -3078,29 +3538,46 @@ mod tests {
...
+                tx.verify_client_report_scrubbed(task_id, report_id).await;
+            }
         }
 
-        // Verify that all reports we did not see a report aggregation for are not scrubbed. (We do
-        // so by reading the report, since reading a report will fail if the report is scrubbed.)
-        for report_id in want_ra_states.keys() {
-            if all_seen_report_ids.contains(report_id) {
-                continue;
+        // Verify that reports aggregated with a non-unit aggregation parameter do not get scrubbed.
```
branlwyd, line 3560: This comment doesn't match the code logic. The current code logic does nothing on each iteration unless the aggregation param is the unit type.

aggregator/src/aggregator/aggregation_job_creator.rs
```diff
@@ -3078,29 +3538,46 @@ mod tests {
...
+                        (agg_job.aggregation_parameter() as &A::AggregationParam).clone(),
+                        ra.report_id(),
+                    )
+                })
+            })
             .collect();
-        for report_id in &all_seen_report_ids {
-            tx.verify_client_report_scrubbed(task_id, report_id).await;
+        for (_, report_id) in &all_seen_report_ids {
+            if is_unit_type::<A::AggregationParam>() {
```
branlwyd, line 3555: nit: hoist these aggregation-param-is-unit-type checks outside of the loops they are contained in, since these are effectively constant values.

aggregator_core/src/datastore/models.rs
tgeoghegan: I did this because I wanted to see what the aggregation param value being used with the dummy VDAF is, and that type is simple enough that including it in Debug output wasn't too noisy. I agree that (b) is desirable, but currently `Poplar1AggregationParam`'s debug output contains too much noise. So I'm going to revert this.

aggregator/src/aggregator/aggregation_job_creator.rs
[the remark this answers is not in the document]
tgeoghegan: If you have a small reproduction case, we should file a bug on `rustc`, because this doesn't make a lot of sense to me.

branlwyd (approved):
LGTM once we have a released prio.

PR #2770, opened 2024-03-04
expose `TaskBuilder` to integration tests

tgeoghegan:
Refactors integration test harness so that the top-level test functions can instantiate and manipulate a `TaskBuilder` value instead of having it constructed several calls deep in `test_task_builder_inner`.

This is useful because soon, we will want to run tests using VDAFs with non-trivial aggregation parameters. That means collecting a single batch multiple times, which means we want to set `max_batch_query_count` higher than 1. Exposing the `TaskBuilder` is an easier way to do that (or set other values) than to plumb more arguments into the `task_builder_*` functions.

### DIFF
--- a/integration_tests/tests/integration/common.rs
+++ b/integration_tests/tests/integration/common.rs
@@ -22,125 +22,85 @@ use std::{iter, time::Duration as StdDuration};
 use tokio::time::{self, sleep};
 use url::Url;
 
-/// Returns a tuple of [`TaskParameters`] and a task builder. This is suitable for test aggregators
-/// running in virtual networks (a Docker network or a Kind cluster).
-pub fn test_task_builder(
-    vdaf: VdafInstance,
-    query_type: QueryType,
-    collector_max_interval: time::Duration,
-    collector_max_elapsed_time: time::Duration,
-) -> (TaskParameters, TaskBuilder) {
-    let endpoint_random_value = hex::encode(random::<[u8; 4]>());
-    let endpoint_fragments = EndpointFragments {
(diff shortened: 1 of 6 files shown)


integration_tests/tests/integration/common.rs
```diff
@@ -22,133 +22,92 @@ use std::{iter, time::Duration as StdDuration};
...
-    )
-}
+    task_builder = task_builder
+        .with_leader_aggregator_endpoint(leader_endpoint)
+        .with_helper_aggregator_endpoint(helper_endpoint)
+        .with_min_batch_size(46)
+        // TODO(timg): is it OK to unconditionally set these? I think the remote case will ignore
+        // the values anyway
+        .with_dap_auth_aggregator_token()
+        .with_dap_auth_collector_token();
```
inahga, line 97: I think you're right, although it's non-obvious. I'd just leave a comment to the effect of "we expect that the remote case ignores these tokens and substitutes its own".

inahga (approved):

integration_tests/tests/integration/common.rs
tgeoghegan: Oh, whoops, I meant to run `cloud-integration-test` against this branch, see what happens and update the comment accordingly. Thanks for flagging!

https://github.com/divviup/janus-ops/actions/runs/8149361321

tgeoghegan: The test run failed, but that's because this DAP-09 Janus branch won't work with the DAP-07 staging environment. The test got far enough to fail at report decryption, which means that the authentication tokens were validated, and we were right about being able to unconditionally generate the tokens.

PR #2775, opened 2024-03-05
Upgrade dependencies to address security advisories, 0.6 backport

divergentdave:
Backport of #2774.


tgeoghegan:
Looks like we also need to bump `mio`?

divergentdave:
Ah indeed, Dependabot only gives us security updates on the main branch

tgeoghegan (approved):

Issue #2751, opened 2024-02-28
Handle repeated collection jobs in one database transaction


divergentdave:
Currently collection jobs for new and repeated batches share some code paths. As a result, we write back batch aggregations to the database unchanged, and split work across two database transactions, though no HTTP request happens between the two. This could be rewritten to reduce writes and handle repeated requests in only one transaction, but we probably don't get that many repeated collection requests.

branlwyd:
This performance optimization was implemented as part of #2782 & then improved in #2957 -- now, repeated collection jobs are handled in one DB transaction, without writing anything back except the completed collection job.

Issue #1392, opened 2023-05-22
Clean up batch/batch aggregation/collection job state tracking


branlwyd:
#1354 introduced explicit state-tracking for batches & collection jobs, along with a new `aggregation_job_writer` abstraction used to write aggregation jobs (by the Leader).

There is further cleanup work to be done after this commit:
  * The Helper codepaths should really be using `aggregation_job_writer`, too, for consistency.
  * We now track "closure" of a given batch in two ways: the batch state, and the batch aggregation state. But for the Leader, the batch aggregation state always closes after the batch state, so it is redundant. The Helper currently relies on only the batch aggregation state for its own protection against aggregating into a closed batch. Ultimately, I think we should get rid of batch aggregation state entirely and rely only on batch state to track "closure" of a batch for both the Leader & the Helper.
  * We now have an "accumulator" (which writes to batch aggregations) and an "aggregation job writer" (which writes to aggregation jobs, report aggregations, and batches). There isn't a particularly good reason for this split, and indeed there is a bit of awkward code where we have to check that the unwritable reports reported by the accumulator are a subset of the unwritable reports reported by the aggregation job writer. Once the previous points are handled, it makes sense to merge the accumulator and aggregation job writer functionality.

branlwyd:
See also #1393: some of this cleanup work would fit well into the performance-improvement work suggested by that issue.

branlwyd:
All of this has been done (for the second point, batches no longer exist separately from batch aggregations).

PR #2782, opened 2024-03-08
Remove batches table.

branlwyd:
The goal of this change is to reduce write contention, particularly on creation or stepping of aggregation jobs. Previously, every aggregation job creation or finalization required updating the relevant (unsharded) batches row. This was a source of significant write contention.

All uses of the unsharded batches table have been migrated to the sharded batch aggregations table. In particular:
 * Batches updates, particularly the state value, were used to avoid the possibility of aggregating into a collected batch even in the face of concurrent aggregation/collection. Now, the batch aggregations' state value is used.
 * Batches included an outstanding_aggregation_jobs value which was used to track how many aggregation jobs still needed to be completed. This is replaced by sharded aggregation_jobs_{created,terminated} counters in batch aggregations.
 * The client_timestamp_interval contained in the batch was used for GC. Now, a sharded client timestamp interval is included in batch aggregations, and is used for this purpose.

Note that, due to the particulars of Postgres' concurrency model, we now have to poll collection jobs for readiness in the collection job driver; I am not aware of a method of implementing a sharded counter where we can be guaranteed that at least one transaction updating the counter will realize when the counter has reached a certain value.

I also fixed a few bugs and cleaned up some technical debt along the way. In particular:
 * As Leader & Helper aggregation job writes are now identical, they both use aggregation_job_writer. The accumulator previously used for Helper writes has been removed.
 * Helper aggregation init could accumulate shares, then fail the report (e.g. on replay), then still write the accumulated value including the failed report's contribution. This was a privacy bug.
 * The check for all-batches-complete for a given collection job was racy. This was a correctness bug.
 * Abandoning an aggregation job could mark report aggregations as BatchCollected if they fell into a collected batch. Now they are left untouched, as intended, to allow investigation of abandoned aggregation jobs.
 * Collection job interval computation is moved to collection job stepping; this removes the need for collection job retrieval to read the batch aggregations, and imposes no additional reads on the collection job stepping code as it was already reading the batch aggregations.
 * Several tables had an unused "trace_context" column. This column was never read or written, so I removed it.


branlwyd:
A few thoughts on next steps:
* Polling collection jobs for readiness can be expensive if the batch aggregations are numerous/large. We can make this cheaper if we implement a DB method that sums the relevant aggregation_jobs_{created,terminated} counters and tells us if they are equal, before we go to the effort of reading the aggregate shares & other batch-aggregation data. I think if we are clever we can even do so in a way that doesn't require additional DB round-trips, e.g. "return batch aggregations only if the batch is ready", perhaps.
* The retry delay for retries of collection jobs which are not ready for collection is fixed. We may well want to turn this into exponential backoff, eventually.

tgeoghegan (commented):
First set of comments. I still want to spend more time understanding the lifecycle of aggregations jobs, especially how batch closing works now, and will review further.

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -699,92 +515,81 @@ where
...
-        )?;
-
-        let batches: HashMap<_, _> = batches
-            .into_iter()
-            .flat_map(|batch: Option<Batch<SEED_SIZE, Q, A>>| {
-                batch.map(|b| (b.batch_identifier().clone(), (Operation::Update, b)))
-            })
-            .collect();
+        // Read all relevant batch aggregations from the datastore.
+        let batch_aggregation_ord = thread_rng().gen_range(0..writer.batch_aggregation_shard_count);
```
tgeoghegan, line 519: Dumb question: what if the relevant batch aggregation was previously written with a different `batch_aggregation_ord` than what is generated here?

More broadly, it'd help to have a high level explanation of how batch aggregation sharding works. What's the data model? For a given batch, can we have multiple batch aggregations in each shard, or just one?

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -809,129 +614,161 @@ where
...
+    fn update_batch_aggregations(&mut self) -> Result<(), Error> {
+        let aggregation_parameter = match self.writer.aggregation_parameter() {
+            Some(aggregation_parameter) => aggregation_parameter,
+            // None means there are no aggregation jobs to write, so we can safely short-circuit.
+            None => return Ok(()),
+        };
+
+        for (batch_identifier, by_aggregation_job_index) in &self.writer.by_batch_identifier_index {
+            // Grab the batch aggregation we read for this batch identifier, or create a new empty
+            // one to aggregate into.
```
tgeoghegan, line 635: I'm trying to reason through how we preserve the existing logic where we used a batch's `OPEN, CLOSING, CLOSED` state to prevent the leader from creating new aggregation jobs once a collection had been initiated. How is that achieved now?

db/00000000000001_initial_schema.up.sql
```diff
@@ -324,15 +296,20 @@ CREATE TYPE BATCH_AGGREGATION_STATE AS ENUM(
...
+    state                       BATCH_AGGREGATION_STATE NOT NULL,  -- the current state of this batch aggregation
+    aggregate_share             BYTEA,                             -- the (possibly-incremental) aggregate share; populated unless report_count is 0 or the batch aggregation has been scrubbed.
+    report_count                BIGINT,                            -- the (possibly-incremental) client report count; populated unless the batch aggregation has been scrubbed.
+    checksum                    BYTEA,                             -- the (possibly-incremental) checksum; populated unless the batch aggregation has been scrubbed.
+    aggregation_jobs_created    BIGINT,                            -- the number of aggregation jobs that have been created for this batch; populated unless the batch aggregation has been scrubbed.
+    aggregation_jobs_terminated BIGINT,                            -- the number of aggregation jobs that have been terminated (finished or abandoned) for this batch; populated unless the batch aggregation has been scrubbed.
```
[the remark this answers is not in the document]
branlwyd: I think these should be `NULL` to indicate they do not exist, rather than 0/an empty string, which is semantically different (and, at least for the counters, a valid value). This reduces the odds of incorrect interpretations of the data.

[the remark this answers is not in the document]
tgeoghegan: Good point, I hadn't thought through that 0 is a legitimate, non-null value.

aggregator_core/src/datastore.rs
```diff
@@ -3181,18 +3211,27 @@ impl<C: Clock> Transaction<'_, C> {
...
                 JOIN tasks ON tasks.id = batch_aggregations.task_id
-                JOIN batches ON batches.task_id = batch_aggregations.task_id
-                            AND batches.batch_identifier = batch_aggregations.batch_identifier
-                            AND batches.aggregation_param = batch_aggregations.aggregation_param
                 WHERE tasks.task_id = $1
                   AND batch_aggregations.batch_identifier = $2
                   AND batch_aggregations.aggregation_param = $3
                   AND ord = $4
-                  AND UPPER(COALESCE(batches.batch_interval, batches.client_timestamp_interval)) >= COALESCE($5::TIMESTAMP - tasks.report_expiry_age * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)",
+                  AND EXISTS(SELECT 1 FROM non_gc_batches WHERE task_id = $1 AND batch_identifier = $2 AND aggregation_param = $3)",
```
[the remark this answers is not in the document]
branlwyd: 1. I took a crack at a comment, let me know if this makes sense. [Somewhat off-topic, but it's unfortuante that we call a "batch" something identified by (task ID, batch identifier, aggregation param) -- ideally, a batch would be identified by (task ID, batch identifier). But we don't have a good name for the entity addressed by (task ID, batch identifier, aggregation param).]

2. Surprisingly, it appears that joining against `non_gc_batches` directly is *worse* -- see query plans at end of this comment. It appears that the join version does a sequential scan over `batch_aggregations` for every row with the desired `ord`, then filters that based on a scan through `non_gc_batches`. The original query, OTOH, does a direct index lookup based on the expected values, followed by a filter based on the subquery. This might be affected by the scale of the DB -- I was working against a DB that had run a single `batch_aggregations` unit test, so there were a single-digit number of rows. But for now, I'm OK moving forward with this query.

<details>
<summary>Current query & query plan</summary>

```
EXPLAIN WITH non_gc_batches AS (
                    SELECT
                        tasks.task_id, batch_identifier, aggregation_param
                    FROM batch_aggregations
                    JOIN tasks ON tasks.id = batch_aggregations.task_id
                    WHERE tasks.task_id = '\xaf6d8d803ec6d78832c8fcb420f85bbe7b24c5ae4a540d0525849df9127cfdc1'
                      AND batch_aggregations.batch_identifier = '\x00000000000003e80000000000000064'
                      AND batch_aggregations.aggregation_param = '\x0c'
                    GROUP BY tasks.task_id, batch_identifier, aggregation_param
                    HAVING MAX(UPPER(COALESCE(batch_interval, client_timestamp_interval))) >= COALESCE(now() - MAX(tasks.report_expiry_age) * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)
                )
                SELECT
                    client_timestamp_interval, batch_aggregations.state, aggregate_share,
                    report_count, checksum, aggregation_jobs_created, aggregation_jobs_terminated
                FROM batch_aggregations
                JOIN tasks ON tasks.id = batch_aggregations.task_id
                WHERE tasks.task_id = '\xaf6d8d803ec6d78832c8fcb420f85bbe7b24c5ae4a540d0525849df9127cfdc1'
                  AND batch_aggregations.batch_identifier = '\x00000000000003e80000000000000064'
                  AND batch_aggregations.aggregation_param = '\x0c'
                  AND ord = 4
                  AND EXISTS(SELECT 1 FROM non_gc_batches WHERE task_id = '\xaf6d8d803ec6d78832c8fcb420f85bbe7b24c5ae4a540d0525849df9127cfdc1' AND batch_identifier = '\x00000000000003e80000000000000064' AND aggregation_param = '\x0c')

                                                                                                                                                  QUERY PLAN                                                                                                                                                  
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Result  (cost=16.68..32.73 rows=1 width=124)
   One-Time Filter: $1
   InitPlan 1 (returns $1)
     ->  Subquery Scan on non_gc_batches  (cost=0.29..16.39 rows=1 width=0)
           ->  GroupAggregate  (cost=0.29..16.38 rows=1 width=96)
                 Group Key: tasks_1.task_id, batch_aggregations_1.batch_identifier, batch_aggregations_1.aggregation_param
                 Filter: (max(upper(COALESCE(batch_aggregations_1.batch_interval, batch_aggregations_1.client_timestamp_interval))) >= COALESCE((now() - ((max(tasks_1.report_expiry_age))::double precision * '00:00:01'::interval)), ('-infinity'::timestamp without time zone)::timestamp with time zone))
                 ->  Nested Loop  (cost=0.29..16.34 rows=1 width=168)
                       ->  Index Scan using task_id_index on tasks tasks_1  (cost=0.14..8.16 rows=1 width=48)
                             Index Cond: (task_id = '\xaf6d8d803ec6d78832c8fcb420f85bbe7b24c5ae4a540d0525849df9127cfdc1'::bytea)
                       ->  Index Scan using batch_aggregations_unique_task_id_batch_id_aggregation_param on batch_aggregations batch_aggregations_1  (cost=0.14..8.17 rows=1 width=136)
                             Index Cond: ((task_id = tasks_1.id) AND (batch_identifier = '\x00000000000003e80000000000000064'::bytea) AND (aggregation_param = '\x0c'::bytea))
   ->  Nested Loop  (cost=16.68..32.73 rows=1 width=124)
         ->  Index Scan using task_id_index on tasks  (cost=0.14..8.16 rows=1 width=8)
               Index Cond: (task_id = '\xaf6d8d803ec6d78832c8fcb420f85bbe7b24c5ae4a540d0525849df9127cfdc1'::bytea)
         ->  Index Scan using batch_aggregations_unique_task_id_batch_id_aggregation_param on batch_aggregations  (cost=0.14..8.17 rows=1 width=132)
               Index Cond: ((task_id = tasks.id) AND (batch_identifier = '\x00000000000003e80000000000000064'::bytea) AND (aggregation_param = '\x0c'::bytea) AND (ord = 4))
(17 rows)

```
</details>

<details>
<summary>Updated query & query plan</summary>

```
EXPLAIN WITH non_gc_batches AS (
                    SELECT
                        tasks.id AS task_id, batch_identifier, aggregation_param
                    FROM batch_aggregations
                    JOIN tasks ON tasks.id = batch_aggregations.task_id
                    WHERE tasks.task_id = '\xaf6d8d803ec6d78832c8fcb420f85bbe7b24c5ae4a540d0525849df9127cfdc1'
                      AND batch_aggregations.batch_identifier = '\x00000000000003e80000000000000064'
                      AND batch_aggregations.aggregation_param = '\x0c'
                    GROUP BY tasks.id, batch_identifier, aggregation_param
                    HAVING MAX(UPPER(COALESCE(batch_interval, client_timestamp_interval))) >= COALESCE(now() - MAX(tasks.report_expiry_age) * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)
                )
                SELECT
                    client_timestamp_interval, batch_aggregations.state, aggregate_share,
                    report_count, checksum, aggregation_jobs_created, aggregation_jobs_terminated
                FROM batch_aggregations
JOIN non_gc_batches
ON non_gc_batches.task_id = batch_aggregations.task_id
AND non_gc_batches.batch_identifier = batch_aggregations.batch_identifier
AND non_gc_batches.aggregation_param = batch_aggregations.aggregation_param
AND ord = 4;

                                                                                                                                             QUERY PLAN                                                                                                                                             
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=16.35..29.55 rows=1 width=124)
   Join Filter: ((batch_aggregations.task_id = tasks.id) AND (batch_aggregations.batch_identifier = batch_aggregations_1.batch_identifier) AND (batch_aggregations.aggregation_param = batch_aggregations_1.aggregation_param))
   ->  Seq Scan on batch_aggregations  (cost=0.00..13.12 rows=1 width=196)
         Filter: (ord = 4)
   ->  GroupAggregate  (cost=16.35..16.40 rows=1 width=72)
         Group Key: tasks.id, batch_aggregations_1.batch_identifier, batch_aggregations_1.aggregation_param
         Filter: (max(upper(COALESCE(batch_aggregations_1.batch_interval, batch_aggregations_1.client_timestamp_interval))) >= COALESCE((now() - ((max(tasks.report_expiry_age))::double precision * '00:00:01'::interval)), ('-infinity'::timestamp without time zone)::timestamp with time zone))
         ->  Sort  (cost=16.35..16.36 rows=1 width=144)
               Sort Key: tasks.id
               ->  Nested Loop  (cost=0.29..16.34 rows=1 width=144)
                     ->  Index Scan using task_id_index on tasks  (cost=0.14..8.16 rows=1 width=16)
                           Index Cond: (task_id = '\xaf6d8d803ec6d78832c8fcb420f85bbe7b24c5ae4a540d0525849df9127cfdc1'::bytea)
                     ->  Index Scan using batch_aggregations_unique_task_id_batch_id_aggregation_param on batch_aggregations batch_aggregations_1  (cost=0.14..8.17 rows=1 width=136)
                           Index Cond: ((task_id = tasks.id) AND (batch_identifier = '\x00000000000003e80000000000000064'::bytea) AND (aggregation_param = '\x0c'::bytea))
(14 rows)
```
</details>

3. The purpose of `non_gc_batches` is to evaluate the GC condition, which requires evaluating all shards of `batch_aggregations`; thus, grouping (across shards) is necessary in that query. But this means we can't return `batch_aggregations.id`, since we aren't grouping (& can't group) over it; a query with a `GROUP BY` clause can only return group-by'ed columns or aggregate columns: https://www.postgresql.org/docs/current/sql-select.html#SQL-GROUPBY

aggregator/src/aggregator/aggregation_job_writer.rs
tgeoghegan: Here's what I'm putting together. Let's suppose that the aggregation job creator has found some unaggregated reports and decided to create a new aggregation job.

- `AggregationJobWriter::write` will choose a random shard to consult each time it's called
- If it finds a relevant batch aggregation in the chosen shard, then either that shard gets updated with the new aggregation job and report aggregations, or `AggregationJobWriter` will refuse to do that for a shard that's in the `Collected` or `Scrubbed` states.
- If it does not find a relevant batch aggregation in the chosen shard, `AggregationJobWriter` creates a new BA in state `Aggregating`.
- As a result, for any given `BatchIdentifier`, we can have at most one `BatchAggregation` per shard. 

But suppose this happens:

- `batch_aggregation_shard_count` is 2, and we are running a time interval task.
- A bunch of reports arrive, timestamped between `[t0, t1)`.
- The aggregation job creator runs, and `AggregationJobWriter::write` winds up creating 1 batch aggregation in shard 0.
- The aggregation jobs run to completion, and the BA in shard 0 is updated into the `Collected` state.
- More reports arrive, also timestamped between `[t0, t1)`.
- The aggregation job creator runs again, finds those unaggregated reports, and creates a new aggregation job.
- This time, `AggregationJobWriter::write` chooses shard 1, and finds no relevant existing batch aggregation.

At this point, I'm pretty sure `WriteState::update_batch_aggregations` will create a brand new `BatchAggregation` in state `Aggregating`, and then the new reports will be aggregated, when they should get failed with `BatchCollected`.

Does the logic that checks if the batch is already collected need to check whether a relevant batch aggregation exists in _any_ shard, not just the one randomly chosen on this run?

Or maybe the fix here is for `AggregationJobWriter`/`WriteState` to eagerly create a `BatchAggregation` in all shards, even though some of them may never have aggregation jobs or report aggregations assigned to them, so that others parts of the machinery can update _all_ the `BatchAggregations` into `Collected` or `Scrubbed` in lockstep?

n.b. I still have to carefully study the code that drives aggregation jobs and creates/drives collection jobs, so this may well be accounted for elsewhere that I haven't seen yet.

aggregator_core/src/datastore.rs
```diff
@@ -3890,12 +3947,22 @@ impl<C: Clock> Transaction<'_, C> {
     ) -> Result<(), Error> {
         let stmt = self
             .prepare_cached(
-                "INSERT INTO outstanding_batches (
+                "WITH non_gc_batches AS (
+                    SELECT
+                        tasks.task_id, batch_identifier
```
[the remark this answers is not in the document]
branlwyd: It ignores the aggregation parameter because `outstanding_batches` is not aggregation-parameter-aware; `outstanding_batches` is not aggregation-parameter-aware since it is used for `current_batch` queries which only work with batches that have not yet been aggregated (or, more precisely, associated with a collection job).

Added a reference to #225.

aggregator_core/src/datastore.rs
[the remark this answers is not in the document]
tgeoghegan: Thanks for taking the trouble of trying out the JOIN. That result is indeed surprising, but good enough justification for structuring the query as you already did. I think an explanatory comment will suffice.

aggregator/src/aggregator/aggregation_job_writer.rs
branlwyd: > More broadly, it'd help to have a high level explanation of how batch aggregation sharding works. What's the data model? For a given batch, can we have multiple batch aggregations in each shard, or just one?

The data model is: a batch aggregation is identified by a batch (i.e. `(task_id, batch_id, aggregation_param`)) and a shard number (i.e. `ord`). A given batch can have many batch aggregations, but each one has a different `ord`. The `ord` defines the shard number of the particular batch aggregation. So, each shard (or `ord`) can hold exactly one `BatchAggregation`. This was true before this PR, and has not changed with this PR.

> At this point, I'm pretty sure WriteState::update_batch_aggregations will create a brand new BatchAggregation in state Aggregating, and then the new reports will be aggregated, when they should get failed with BatchCollected.

This won't happen: first, the step where the first aggregation job runs to completion will not update the BA to `Collected` (the aggregation job writer does not update batch aggregation states). This will only happen once a collection job is received from the collector, and the collection job is stepped and notices that it can complete the collection job (i.e. there are no outstanding aggregation jobs & no outstanding reports in the batch). At that point, the collection job stepper will update _all_ batch aggregation shards to the `Collected` state, to avoid the problem you describe.

That is,

> Or maybe the fix here is for AggregationJobWriter/WriteState to eagerly create a BatchAggregation in all shards, even though some of them may never have aggregation jobs or report aggregations assigned to them, so that others parts of the machinery can update all the BatchAggregations into Collected or Scrubbed in lockstep?

This is what is already implemented, except that it happens in the collection job driver (for the Leader) or the aggregate share request handler (for the Helper) instead of anywhere an aggregation job is written. See e.g. https://github.com/divviup/janus/blob/79597db649cc01cb18680bfda17b4e95a626aed8/aggregator/src/aggregator/collection_job_driver.rs#L192-L233.

I'm going to write up a description of state handling in the `docs/` directory, as I think this is pretty complicated.

(FWIW, I think we should rename `batch_aggregations` to `batches` as a next step, since `batch_aggregations` stores all of the per-batch information now. But let's not confuse the current discussion!)

tgeoghegan: Those explanations make sense, thank you!

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -609,68 +609,37 @@ where
...
-                            warn!(
-                                report_id = %stepped_aggregation.report_aggregation.report_id(),
-                                ?err,
-                                "Could not update batch aggregation",
-                            );
-                            self.aggregate_step_failure_counter
-                                .add(1, &[KeyValue::new("type", "accumulate_failure")]);
-                            ReportAggregationState::<SEED_SIZE, A>::Failed {
-                                prepare_error: PrepareError::VdafPrepError,
-                            }
```
tgeoghegan, line 660: I understand that this failure path goes away because we don't use `Accumulator` anymore. But we also are no longer incrementing metrics with the `accumulate_failure` label. Should we plumb that into `AggregationJobWriter`?

aggregator/src/aggregator.rs
```diff
@@ -2127,217 +2039,118 @@ impl VdafOps {
                 agg_param,
                 req.batch_selector().batch_identifier().clone(),
                 client_timestamp_interval,
-                if saw_continue {
-                    AggregationJobState::InProgress
-                } else {
-                    AggregationJobState::Finished
-                },
+                AggregationJobState::InProgress,
```
tgeoghegan, line 2042:
```suggestion
                // For one-round VDAFs, the aggregation job will actually be finished,
                // but AggregationJobWriter will handle updating its state.
                AggregationJobState::InProgress,
```

tgeoghegan (approved):
TY for your patience with my questions. Everything here makes sense to me and I think my remaining comments aren't blocking.

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -699,92 +515,81 @@ where
...
-                != Some(BatchState::Closed)
-            {
+        // batch which is not still accepting aggregations (i.e. not in the Aggregating state)
+        // instead fail with a BatchCollected error (unless they were already in an failed state).
+        for (batch_identifier, by_aggregation_job_index) in &self.writer.by_batch_identifier_index {
+            if !matches!(
+                self.batch_aggregations
+                    .get(batch_identifier)
+                    .map(|(_, b)| b.state()),
+                Some(BatchAggregationState::Collected { .. } | BatchAggregationState::Scrubbed)
```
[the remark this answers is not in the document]
branlwyd: SG, though note this isn't as big a win as it might seem at first, since we need to handle the case that the batch is nonexistent.

branlwyd:
I added a basic data model doc to try to capture some of the more complex/less obvious bits of how we interact with our DB. Thoughts on additional content welcome; the Helper section mostly refers back to the Leader section, but IMO that's a good thing (it indicates our Helper is acting identically to the Leader, which keeps things simpler).

aggregator/src/aggregator/aggregation_job_driver.rs
branlwyd: Good point -- I stared at this for a while and things would have to go very wrong for this merge to fail, but that means that if something does go wrong we'll probably appreciate having the metric.

branlwyd:
Closes #1392.
Closes #2751.

branlwyd:
I'm going to go ahead and merge this to avoid head-of-line blocking. If anyone else is reviewing, please feel free to leave comments & I will address them in a followup PR.

PR #2809, opened 2024-03-13
Add limit to transaction retries

inahga:
Closes https://github.com/divviup/janus/issues/2124.

Breaks out of the transaction loop if it's tried too many times. This is governed by the config option `max_transaction_retries`. We could hard code this limit, but it's not a big deal to plumb it through to configuration in the off chance we need to raise it.

### DIFF
--- a/aggregator/src/config.rs
+++ b/aggregator/src/config.rs
@@ -43,10 +43,20 @@ pub struct CommonConfig {
     /// Address to serve HTTP health check requests on.
     #[serde(default = "default_health_check_listen_address")]
     pub health_check_listen_address: SocketAddr,
+
+    /// The maximum number of times a transaction can be retried. The intent is to guard against bugs
+    /// that induce infinite retries. It should be set to a reasonably high limit to prevent legitimate
+    /// work from being cancelled.
+    #[serde(default = "default_max_transaction_retries")]
+    pub max_transaction_retries: u64,
 }
 
 fn default_health_check_listen_address() -> SocketAddr {
     SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 9001)
+}
+
+pub fn default_max_transaction_retries() -> u64 {
+    1000
 }
 
 /// Trait describing configuration structures for various Janus binaries.
@@ -277,6 +287,7 @@ pub mod test_util {
 mod tests {
     use crate::{
         config::{
+            default_max_transaction_retries,
             test_util::{generate_db_config, generate_metrics_config, generate_trace_config},
             CommonConfig, DbConfig, JobDriverConfig,
         },
@@ -307,6 +318,7 @@ mod tests {
             logging_config: generate_trace_config(),
             metrics_config: generate_metrics_config(),
             health_check_listen_address: SocketAddr::from((Ipv4Addr::UNSPECIFIED, 8080)),
+            max_transaction_retries: default_max_transaction_retries(),
         })
     }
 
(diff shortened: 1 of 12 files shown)


aggregator_core/src/datastore/tests.rs
```diff
@@ -99,6 +106,38 @@ async fn down_migrations(
...
+async fn retry_limit() {
+    install_test_trace_subscriber();
+    for max_transaction_retries in [0, 1, 1000] {
+        let ephemeral_datastore = EphemeralDatastoreBuilder::new().build().await;
+        let datastore = ephemeral_datastore
+            .datastore_with_max_transaction_retries(MockClock::default(), max_transaction_retries)
+            .await;
+
+        // The number of times the transaction was actually run.
+        let num_runs = Arc::new(AtomicU64::new(0));
```
tgeoghegan, line 119: Honest question: if it's an `AtomicU64`, does it need to be further wrapped in an atomic reference cell?

aggregator_core/src/datastore.rs
```diff
@@ -128,9 +129,17 @@ impl<C: Clock> Datastore<C> {
         crypter: Crypter,
         clock: C,
         meter: &Meter,
+        max_transaction_retries: u64,
     ) -> Result<Datastore<C>, Error> {
-        Self::new_with_supported_versions(pool, crypter, clock, meter, SUPPORTED_SCHEMA_VERSIONS)
-            .await
+        Self::new_with_supported_versions(
```
tgeoghegan, line 134: At some point we'll want to use the builder pattern here instead of tacking on more constructor arguments.

aggregator/src/config.rs
```diff
@@ -43,12 +43,22 @@ pub struct CommonConfig {
...
+    #[serde(default = "default_max_transaction_retries")]
+    pub max_transaction_retries: u64,
 }
 
 fn default_health_check_listen_address() -> SocketAddr {
     SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 9001)
 }
 
+pub fn default_max_transaction_retries() -> u64 {
+    1000
```
tgeoghegan, line 59: Why 1000? Is this informed by retry counts we've observed in production?

tgeoghegan (approved):

aggregator_core/src/datastore/tests.rs
inahga: Yes. If it's AtomicU64, it's still an owned type, so it will be moved into the `async move` block. So we need `Arc` for the pattern of sharing owned values across multiple threads/futures.

I also pondered it being a `Arc<u64>`, but that won't work because the value in an `Arc` is immutable--we some integer data structure that allows interior mutability.

It does feel a bit odd to me for just a counter, but perhaps there's a simpler set of data structures I've overlooked.

aggregator_core/src/datastore.rs
inahga: Indeed, I think we've reached the limit--any more we'll want to move to builder pattern.

aggregator/src/config.rs
inahga: Yes, insofar as I've not seen retry counts that approach this number--the maximum retry count I've seen in production is about 20, corresponding to normal serialization retries.

I target this number so as not to abort transactions that are retrying due to concurrency errors, rather only abort in case of application-driven retry bugs, so I leave considerable safety margin.

aggregator_core/src/datastore/tests.rs
tgeoghegan: Oh, right, the DB transaction stuff is awkward that way. Good enough!

PR #2813, opened 2024-03-14
aggregator api: Properly report we're DAP-09

inahga:


inahga:
We should make another prerelease version before merging this because it will break interop with divviup-api (AFAICT). Need some time to work on https://github.com/divviup/divviup-api/issues/878.

divergentdave (approved):

tgeoghegan (approved):

branlwyd (approved):

Issue #2562, opened 2024-01-29
Proposed style change for embedded SQL queries


divergentdave:
Currently, our convention for SQL queries is to use multiline `""` literals, start the query at the beginning of the literal, and indent following lines of the query either to the same column as the opening `"` or further, guided by the syntactic position within the SQL. I suggest that we change our conventions to instead start with a single-line `--` SQL comment, containing the name of the `Transaction` method in which it appears, and then begin the SQL query proper on the following line, left justified to column 0 (and indenting from there based on SQL syntax). This would have the following benefits:

- No need for manual style fixes when making unrelated changes
  - Currently, if rustfmt indents or dedents a `prepare_cached()` call, it won't change the number of spaces within the string literal, since they are semantically meaningful. However, this results in query bodies no longer lining up with the start of the string literal. We have a few cases like this present in the code now.
- Easier to unambiguously identify queries from query logs
- Less pressure on line length in `datastore.rs` and `datastore/tests.rs`
  - Less cases where rustfmt can't format a chain
- Less whitespace in query logs
  - Maybe more headroom before hitting Cloud SQL Query Insights's query length limit (it is unclear if the limit is on actual query length or normalized query length)

This is inspired by a similar trick I picked up somewhere for embedded YAML, wherein string literals start with a `---` document separator, and the indentation-sensitive YAML then follows against the left gutter, just as it would in a separate file.

Alternately, if we don't want to maintain the method names within queries, we could instead start query strings with an empty `--` single-line comment, then make the same indentation change. This would get us all but one of the above benefits.

inahga:
The other idea I had idling in my head was to relocate queries into their own files, then use [`include_bytes!`](https://doc.rust-lang.org/std/macro.include_bytes.html) to substitute the text at compile-time.

The major drawback to this is that we lose locality of query and code. I can accept it since my workflow enables side-by-side editing, but I can't speak for others. I think this satisfies all of the stated problems, and it also lets us more easily pipe queries through a [SQL formatter](https://github.com/darold/pgFormatter).

branlwyd:
Sure, no objections from me.

PR #2818, opened 2024-03-15
Datastore: treat writes to logically GC'ed rows as new writes.

branlwyd:
Previously, such writes would lead to a MutationTargetAlreadyExists error. However, the same error would happen if two transactions concurrently wrote the same entity (regardless of GC), which would have to be manually retried (by `tx.retry()`) by application code if it was sure the row shouldn't exist. This could lead to infinitely-repeated retries if the MutationTargetAlreadyExists error was due to the row being GC'ed rather than a concurrent write.

Now, we perform an upsert if the existing row is logically GC'ed. This allows overwriting logically-nonexistent rows with a put, but more importantly, this should allow concurrent writes to receive a serialization failure, rather than a
MutationTargetAlreadyExists error. This removes the need for application-level retries entirely, which should remove the possibility of further infinite-retry bugs.

Also:
* Remove delete-after-put for all put operations. We will wait for the next GC run to delete these rows. This is a tradeoff in complexity for the new upserts. I think writing GC'ed rows should be rare enough that this delay will not matter in practice.
* put_report_share renamed to put_scrubbed_report, and it now has the same semantics as all other put datastore methods. Note that this required some changes to the application code using this functionality, please review carefully.
* put_report_aggregation & similar now properly reports MutationTargetAlreadyExists.
* put_client_report: fix the expiry check.
* get_report_aggregation renamed to get_report_aggregation_for_report_id, now test-only. This was already only used for tests.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -24,7 +24,7 @@ use fixed::{
     types::extra::{U15, U31},
     FixedI16, FixedI32,
 };
-use futures::future::try_join_all;
+use futures::{future::try_join_all, TryFutureExt as _};
 use http::{header::CONTENT_TYPE, Method};
 use itertools::iproduct;
 use janus_aggregator_core::{
@@ -93,7 +93,7 @@ use std::{
     sync::{Arc, Mutex as SyncMutex},
     time::{Duration as StdDuration, Instant},
 };
-use tokio::{join, sync::Mutex, try_join};
+use tokio::{sync::Mutex, try_join};
 use tracing::{debug, info, trace_span, warn, Level};
 use url::Url;
 
@@ -2098,6 +2098,7 @@ impl VdafOps {
                     }
 
                     // Write report shares, and ensure this isn't a repeated report aggregation.
+                    // TODO(#225): on repeated aggregation, verify input share matches previously-received input share
                     try_join_all(report_share_data.iter_mut().map(|rsd| {
                         let task = Arc::clone(&task);
                         let aggregation_job = Arc::clone(&aggregation_job);
@@ -2105,8 +2106,14 @@ impl VdafOps {
                         async move {
                             // Verify that we haven't seen this report ID and aggregation parameter
                             // before in another aggregation job.
-                            let put_report_share_fut =
-                                tx.put_report_share(task.id(), &rsd.report_share);
+                            let put_report_share_fut = tx
+                                .put_scrubbed_report(task.id(), &rsd.report_share)
+                                .or_else(|err| async move {
+                                    match err {
+                                        datastore::Error::MutationTargetAlreadyExists => Ok(()),
+                                        _ => Err(err),
+                                    }
+                                });
                             let report_aggregation_exists_fut = tx
                                 .check_other_report_aggregation_exists::<SEED_SIZE, A>(
                                     task.id(),
@@ -2114,17 +2121,10 @@ impl VdafOps {
                                     aggregation_job.aggregation_parameter(),
                                     aggregation_job.id(),
                                 );
-                            let (put_report_share_rslt, report_aggregation_exists_rslt) =
-                                join!(put_report_share_fut, report_aggregation_exists_fut);
+                            let (_, report_aggregation_exists) =
+                                try_join!(put_report_share_fut, report_aggregation_exists_fut)?;
 
-                            let report_share_exists = match put_report_share_rslt {
-                                Ok(()) => false,
-                                Err(datastore::Error::MutationTargetAlreadyExists) => true,
-                                Err(err) => return Err(err),
-                            };
-                            let report_aggregation_exists = report_aggregation_exists_rslt?;
-
-                            if report_share_exists || report_aggregation_exists {
+                            if report_aggregation_exists {
                                 rsd.report_aggregation = rsd
                                     .report_aggregation
                                     .clone()
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -2104,8 +2104,6 @@ impl VdafOps {
                         let aggregation_job = Arc::clone(&aggregation_job);
 
                         async move {
-                            // Verify that we haven't seen this report ID and aggregation parameter
-                            // before in another aggregation job.
                             let put_report_share_fut = tx
                                 .put_scrubbed_report(task.id(), &rsd.report_share)
                                 .or_else(|err| async move {
@@ -2114,6 +2112,8 @@ impl VdafOps {
                                         _ => Err(err),
                                     }
                                 });
+                            // Verify that we haven't seen this report ID and aggregation parameter
+                            // before in another aggregation job.
                             let report_aggregation_exists_fut = tx
                                 .check_other_report_aggregation_exists::<SEED_SIZE, A>(
                                     task.id(),
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -559,16 +559,6 @@ impl<C: Clock> Transaction<'_, C> {
         self.run_op(self.raw_tx.query_opt(statement, params)).await
     }
 
-    /// Calling this method will force the transaction to eventually be rolled back and retried; all
-    /// datastore writes in this try will be lost. Calling this method does not interrupt or
-    /// otherwise directly affect the transaction-processing callback; the caller may wish to e.g.
-    /// use an error or some other signalling method to cause the callback to terminate.
-    ///
-    /// There is no upper limit on the number of retries a single transaction may incur.
-    pub fn retry(&self) {
-        self.retry.store(true, Ordering::Relaxed);
-    }
-
     /// Returns the current schema version of the datastore and the description of the migration
     /// script that applied it.
     async fn get_current_schema_migration_version(&self) -> Result<(i64, String), Error> {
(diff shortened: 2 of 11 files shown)


aggregator_core/src/datastore.rs
```diff
@@ -2036,59 +1967,52 @@ impl<C: Clock> Transaction<'_, C> {
...
+                    error_code
                 FROM report_aggregations
                 JOIN aggregation_jobs ON aggregation_jobs.id = report_aggregations.aggregation_job_id
                 JOIN tasks ON tasks.id = aggregation_jobs.task_id AND tasks.id = report_aggregations.task_id
                 WHERE tasks.task_id = $1
                   AND aggregation_jobs.aggregation_job_id = $2
-                  AND report_aggregations.client_report_id = $3
-                  AND UPPER(aggregation_jobs.client_timestamp_interval) >= COALESCE($4::TIMESTAMP - tasks.report_expiry_age * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)",
+                  AND UPPER(client_timestamp_interval) >= COALESCE($3::TIMESTAMP - tasks.report_expiry_age * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)
+                ORDER BY ord ASC",
```
inahga, line 1987: I wonder if for `get_report_aggregations_for_aggregation_job()` we should take the similar approach to splitting the query as https://github.com/divviup/janus/pull/2807.

On one hand, without a query plan against a large enough table this could be premature optimization, OTOH we know postgres has a hard time planning this COALESCE statement.

aggregator_core/src/datastore.rs
```diff
@@ -559,16 +559,6 @@ impl<C: Clock> Transaction<'_, C> {
...
 
-    /// Calling this method will force the transaction to eventually be rolled back and retried; all
-    /// datastore writes in this try will be lost. Calling this method does not interrupt or
-    /// otherwise directly affect the transaction-processing callback; the caller may wish to e.g.
-    /// use an error or some other signalling method to cause the callback to terminate.
-    ///
-    /// There is no upper limit on the number of retries a single transaction may incur.
-    pub fn retry(&self) {
-        self.retry.store(true, Ordering::Relaxed);
-    }
```
inahga, line 570: I suppose we can revert https://github.com/divviup/janus/pull/2809 (main branch only), since now there's no way for the application to trigger unbounded retries.

aggregator_core/src/datastore.rs
```diff
@@ -3561,12 +3505,22 @@ impl<C: Clock> Transaction<'_, C> {
...
+                        aggregation_jobs_created, aggregation_jobs_terminated, created_at,
+                        updated_at, updated_by
+                    ) = (
+                        excluded.client_timestamp_interval, excluded.state,
+                        excluded.aggregate_share, excluded.report_count, excluded.checksum,
+                        excluded.aggregation_jobs_created, excluded.aggregation_jobs_terminated,
+                        excluded.created_at, excluded.updated_at, excluded.updated_by
+                    )
+                    WHERE GREATEST(UPPER(COALESCE(batch_aggregations.batch_interval, batch_aggregations.client_timestamp_interval)), (SELECT MAX(UPPER(COALESCE(ba.batch_interval, ba.client_timestamp_interval))) FROM batch_aggregations ba WHERE ba.task_id = batch_aggregations.task_id AND ba.batch_identifier = batch_aggregations.batch_identifier AND ba.aggregation_param = batch_aggregations.aggregation_param)) < COALESCE($16::TIMESTAMP - (SELECT report_expiry_age FROM tasks WHERE id = batch_aggregations.task_id) * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)",
```
inahga, line 3519: Not to rehash older discussions on this, but I do really find this tough to read when it's a single line, to such an extent that I have to break it out in my editor. No changes necessary right now I think, I'll just make a note in https://github.com/divviup/janus/issues/2562 for us to decide on a style and pass through the GC queries.

aggregator_core/src/datastore/tests.rs
```diff
@@ -126,7 +126,7 @@ async fn retry_limit() {
                 let num_runs = Arc::clone(&num_runs);
                 Box::pin(async move {
                     num_runs.fetch_add(1, Ordering::Relaxed);
-                    tx.retry();
+                    tx.retry.store(true, Ordering::Relaxed); // artificially force a retry
```
inahga, line 129: More ammunition for reverting the retry change. If in test we have to induce something that can't actually be done in user code, since tx.retry is private, we might as well ditch that code.

aggregator/src/aggregator.rs
```diff
@@ -2098,33 +2098,33 @@ impl VdafOps {
...
 
                     // Write report shares, and ensure this isn't a repeated report aggregation.
+                    // TODO(#225): on repeated aggregation, verify input share matches previously-received input share
                     try_join_all(report_share_data.iter_mut().map(|rsd| {
                         let task = Arc::clone(&task);
                         let aggregation_job = Arc::clone(&aggregation_job);
 
                         async move {
                             // Verify that we haven't seen this report ID and aggregation parameter
                             // before in another aggregation job.
```
inahga, line 2108: nit: this comment doesn't reflect how `put_report_share_fut` works, it belongs next to `report_aggregation_exists_fut`.

aggregator_core/src/datastore.rs
```diff
@@ -1440,26 +1430,27 @@ impl<C: Clock> Transaction<'_, C> {
...
+                ON CONFLICT(task_id, report_id) DO UPDATE
+                    SET (
+                        client_timestamp, extensions, public_share, leader_input_share,
+                        helper_encrypted_input_share, created_at, updated_at, updated_by
+                    ) = (
+                        excluded.client_timestamp, excluded.extensions, excluded.public_share,
+                        excluded.leader_input_share, excluded.helper_encrypted_input_share,
+                        excluded.created_at, excluded.updated_at, excluded.updated_by
+                    )
+                    WHERE client_reports.client_timestamp < COALESCE($11::TIMESTAMP - (SELECT report_expiry_age FROM tasks WHERE id = client_reports.task_id) * '1 second'::INTERVAL, '-infinity'::TIMESTAMP)",
```
tgeoghegan, line 1449: What prevents a client from changing an existing report's input shares, public shares, etc? i.e.:

- report gets uploaded, gets written via put_client_report
- report gets aggregated, scrubbed
- evil client uploads report with same report ID + task ID, but changes input shares

aggregator_core/src/datastore.rs
branlwyd: I'd like to keep it -- while I can't think of a plausible way to hit a retry loop with the current code, IMO we may as well keep the mitigation in place in case we invent a new way.

aggregator_core/src/datastore/tests.rs
branlwyd: We can still retry based on errors received from Postgres; think of this as a very cheap way to emulate fault-injection into our connection to Postgres.

db/00000000000001_initial_schema.up.sql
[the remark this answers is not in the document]
branlwyd: I'm maintaining the convention that all fields of the effective primary key [in this case, `(task_id, aggregation_job_id, ord)` -- i.e. the fields added to the `UNIQUE` constraint] are grouped together, directly after `id`. I can revert if you like, but as you note this change is harmless.

aggregator_core/src/datastore.rs
branlwyd: Note that, due to the `WHERE` clause on the update, we will only update a row that is already (logically) GC'ed. So the answer, in the datastore, is: nothing, just like with other rows that have been deleted. But there was no such protection before, either -- once the (physical) GC ran, the row would be gone, and nothing in the datastore would the report ID from being reused. The change here is to treat logical GC the same as physical GC.

The protection against this lives in the upload handler, which will reject reports which would be immediately expired:

https://github.com/divviup/janus/blob/5db944a7952c5e49756868cb14a45e9854a72d0e/aggregator/src/aggregator.rs#L1564-L1574

db/00000000000001_initial_schema.up.sql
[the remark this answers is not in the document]
branlwyd: N.B. there are two(?) other tables that don't follow this convention, though arguably one doesn't do so because it puts the derived `batch_interval` next to the `batch_identifier` the interval is derived from.

aggregator_core/src/datastore.rs
branlwyd: I agree. I think I'm going to extend the approach in #2807 to every relevant query (and maybe add caching), but in a separate PR.

aggregator_core/src/datastore.rs
branlwyd: Yeah, this is fair; these lines started pretty long and are now enormous. I think I'm going to address this in the same follow-on PR that takes the approach in #2807 -- another benefit of query-splitting is that it will make these queries somewhat more tractable.

aggregator_core/src/datastore.rs
divergentdave: I'm uneasy about allowing mutations to columns that were previously write-once as well. Note that there would now be a race condition where a process that started a query with an older value of `clock.now()` may see the `ON CONFLICT DO UPDATE` write from another process.

tgeoghegan (approved):

Issue #2819, opened 2024-03-18
`tasks` is missing `updated_at` column


inahga:

inahga:
@tgeoghegan was this intentional in https://github.com/divviup/janus/pull/2079? I was thinking maybe we didn't have `updated_at` because we intended tasks to be fully immutable--but if that were the case I figured we would omit `updated_by` too.

tgeoghegan:
My thinking was that `tasks` rows are immutable, so `updated_at` is unnecessary. The `updated_by` column is still present because we want to record who was responsible for creating a row, and I didn't bother to create distinct `updated_by` and `created_by` columns in any tables.

In any case, since `tasks` is now mutable, we should add an `updated_at` column to `tasks`, and default its value to the creation time.

inahga:
Makes sense, thanks for clarifying!

PR #2824, opened 2024-03-18
aggregator api: Allow patching a task's expiration time

inahga:
Lets us issue `PATCH /tasks/:task_id` with the following bodies and results
- `{}` -> no changes, returns the task as-is.
- `{"task_expiration": null}` -> unsets the task expiration. There currently isn't a use case for this, but I did it for consistency with how PUT of a task works, i.e. you can make a task with no expiration.
- `{"task_expiration": 1000}` -> works as you'd expect, content is seconds since epoch.

Supports https://github.com/divviup/divviup-api/issues/862

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -729,7 +729,7 @@ impl<C: Clock> Transaction<'_, C> {
 
     /// Deletes a task from the datastore, along with all related data (client reports,
     /// aggregations, etc).
-    #[tracing::instrument(skip(self))]
+    #[tracing::instrument(skip(self), err(level = Level::DEBUG))]
     pub async fn delete_task(&self, task_id: &TaskId) -> Result<(), Error> {
         // Deletion of other data implemented via ON DELETE CASCADE.
         let stmt = self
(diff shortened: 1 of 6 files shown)


aggregator_core/src/datastore.rs
```diff
@@ -710,6 +710,34 @@ impl<C: Clock> Transaction<'_, C> {
...
+                "UPDATE tasks SET task_expiration = $1, updated_by = $2 WHERE task_id = $3",
+            )
+            .await?;
+        check_single_row_mutation(
+            self.execute(
+                &stmt,
+                &[
+                    /* task_expiration */
+                    &task_expiration.map(Time::as_naive_date_time).transpose()?,
+                    // TODO(#2819): set updated_at
```
inahga, line 731: https://github.com/divviup/janus/issues/2819. I suspect omitting `updated_at` from `tasks` was intentional, as we'd assume it would always be immutable. I will change this in a separate PR, since for backporting we'll need a database migration.

branlwyd (approved):

aggregator_core/src/datastore.rs
```diff
@@ -710,6 +710,34 @@ impl<C: Clock> Transaction<'_, C> {
         Ok(())
     }
 
+    /// Sets or unsets the expiration date of a task.
+    #[tracing::instrument(skip(self))]
```
divergentdave, line 714: nit: this should have the lower event level set too
```suggestion
    #[tracing::instrument(skip(self, err(level = Level::DEBUG)))]
```

divergentdave (approved):

inahga:
@jbr do you think this API is sufficient? Just checking in before I merge it.

jbr (approved):
LGTM! Keeping track of updated_at at some point seems useful in subsequent PR.

Are there concerns about what happens if expiry is backdated before reports have already been processed? Or the only thing that matters is if at the time of ingest whether the timestamp exists and is in the past?

inahga:
> Are there concerns about what happens if expiry is backdated before reports have already been processed? Or the only thing that matters is if at the time of ingest whether the timestamp is "none or in the future?"
 
It should only matter at ingest time, yes. Currently the only place we check task expiration are in the upload handler. So the effect of setting task expiration should just be to effectively turn off report uploads.

(Well, we also check expiry for taskprov opt-in/out on all handlers, but taskprov is a bit more strict about task expiration).

PR #2858, opened 2024-03-20
Split messages/src/lib.rs

divergentdave:
This splits in-line modules out of `messages/src/lib.rs` to cut down its size.


tgeoghegan (approved):
Nice! I'd be OK with chopping up `messages/src/tests.rs` even further, maybe grouping items by the upload, aggregation and collect interactions. However I think the change as-is accomplishes the main ergonomic improvement of being able to edit the code and the tests side by side in two buffers.

PR #2861, opened 2024-03-20
datastore: Add updated_at to tasks table

inahga:
Supports https://github.com/divviup/janus/issues/2819, follow up to https://github.com/divviup/janus/pull/2824 since task rows are no longer immutable.


tgeoghegan (approved):

tgeoghegan:
Flagging this for @divergentdave because I think you'll need to re-deploy sisyphus and salmoneus to deal with the schema change.

branlwyd (approved):

PR #2863, opened 2024-03-20
datastore: Migration for adding updated_at to tasks table

inahga:
Supports https://github.com/divviup/janus/issues/2819.

This is just the migration for this change.

Notice the new column is nullable. I think this is the only way to make this migration work with zero downtime--if we apply the migration, older versions of Janus would fail to insert tasks due to the NOT NULL constraint. It also means we don't have to backfill for existing tasks.

Since this column is only for debugging, I think it's fine to have null values for `updated_at`. We'll just have to note that on this version of Janus, a null `updated_at` means the task was never updated.

This migration will acquire an exclusive table lock for the duration of the migration. This will block every query that needs to join against `tasks`... which is almost all of them. However, the tasks table is relatively small and this migration doesn't force a table rewrite, so it should complete quickly.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -100,7 +100,7 @@ macro_rules! supported_schema_versions {
 // version is seen, [`Datastore::new`] fails.
 //
 // Note that the latest supported version must be first in the list.
-supported_schema_versions!(4, 3);
+supported_schema_versions!(5, 4, 3);
 
 /// Datastore represents a datastore for Janus, with support for transactional reads and writes.
 /// In practice, Datastore instances are currently backed by a PostgreSQL database.
--- a/db/00000000000005_tasks_updated_at.down.sql
+++ b/db/00000000000005_tasks_updated_at.down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE tasks DROP COLUMN updated_at;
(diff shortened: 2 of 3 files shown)


tgeoghegan (approved):
I'm guessing that it's not possible to have `updated_at` default to the current value of `created_at`?

Past that, we might consider just not taking this migration in `release/0.6`, since we hope to drop support for that version soon anyway.

All that said, this change accomplishes what it says it does, and I agree that locking the `tasks` table should be low-risk.

inahga:
> I'm guessing that it's not possible to have updated_at default to the current value of created_at?

Not easily, I think. https://www.postgresql.org/docs/current/sql-createtable.html#SQL-CREATETABLE-PARMS-DEFAULT

> The DEFAULT clause assigns a default data value for the column whose column definition it appears within. The value is any variable-free expression (in particular, cross-references to other columns in the current table are not allowed).

I think we could work around this with a trigger, but that's a lot of work for something we only care about for debugging.

> Past that, we might consider just not taking this migration in release/0.6, since we hope to drop support for that version soon anyway.

This is fair. I'm inclined to do this, since it's a moderately risky change due to the exclusive lock on an important table.

inahga:
> This is fair. I'm inclined to do this, since it's a moderately risky change due to the exclusive lock on an important table.

Yep I think I'll avoid making this change on release/0.6. It's of marginal benefit for a branch we're going to drop support for soon.

PR #2865, opened 2024-03-20
Add conditions on aggregation_jobs.task_id back to queries

divergentdave:
This is a follow-up to #2857 to add conditions on `aggregation_jobs.task_id` back to some queries. The importance of these predicates varies: some are in `test-util` methods, some may be fine as it is if other joined tables can be searched over first, and some don't have any such compensating factors. These conditions will help when looking up aggregation jobs by their `aggregation_job_id` column, because there is only one index covering this column, and its first column is `task_id`. We know that all `report_aggregations` rows linked to an aggregation job will share the same `task_id` as the `aggregation_jobs` row, but the database doesn't, so supplying this redundant constraint when joining across both opens up new opportunities for efficient index scans. Without this, my concern is that doing a full table scan of aggregation jobs would be the best way to look up a job by its DAP AggregationJobId only.

I also fixed the indentation of some queries where the enclosing method call got reformatted. Enable "Hide whitespace" to just view the substantive changes.

FWIW, I left `delete_expired_aggregation_artifacts()` alone, where a `report_aggregations.task_id` predicate was removed from a subquery, because in that case the lookup on `report_aggregations` is being done by aggregation job foreign key, which there is an index for, and the aggregation job primary keys were looked up in an earlier CTE.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -2137,6 +2137,7 @@ impl<C: Clock> Transaction<'_, C> {
                 "SELECT 1 FROM report_aggregations
                 JOIN aggregation_jobs ON aggregation_jobs.id = report_aggregations.aggregation_job_id
                 WHERE report_aggregations.task_id = $1
+                  AND aggregation_jobs.task_id = $1
                   AND report_aggregations.client_report_id = $2
                   AND aggregation_jobs.aggregation_param = $3
                   AND aggregation_jobs.aggregation_job_id != $4
@@ -2190,6 +2191,7 @@ impl<C: Clock> Transaction<'_, C> {
                 FROM report_aggregations
                 JOIN aggregation_jobs ON aggregation_jobs.id = report_aggregations.aggregation_job_id
                 WHERE report_aggregations.task_id = $1
+                  AND aggregation_jobs.task_id = $1
                   AND aggregation_jobs.aggregation_job_id = $2
                   AND UPPER(client_timestamp_interval) >= $3
                 ORDER BY ord ASC",
@@ -2241,20 +2243,22 @@ impl<C: Clock> Transaction<'_, C> {
         };
 
         let stmt = self
-                .prepare_cached(
-                    "SELECT
-                        ord, client_timestamp, last_prep_resp, report_aggregations.state,
-                        public_share, leader_extensions, leader_input_share,
-                        helper_encrypted_input_share, leader_prep_transition, helper_prep_state,
-                        error_code
-                    FROM report_aggregations
-                    JOIN aggregation_jobs ON aggregation_jobs.id = report_aggregations.aggregation_job_id
-                    WHERE report_aggregations.task_id = $1
-                      AND aggregation_jobs.aggregation_job_id = $2
-                      AND report_aggregations.client_report_id = $3
-                      AND UPPER(client_timestamp_interval) >= $4",
-                )
-                .await?;
+            .prepare_cached(
+                "SELECT
+                    ord, client_timestamp, last_prep_resp, report_aggregations.state,
+                    public_share, leader_extensions, leader_input_share,
+                    helper_encrypted_input_share, leader_prep_transition, helper_prep_state,
+                    error_code
+                FROM report_aggregations
+                JOIN aggregation_jobs
+                    ON aggregation_jobs.id = report_aggregations.aggregation_job_id
+                WHERE report_aggregations.task_id = $1
+                    AND aggregation_jobs.task_id = $1
+                    AND aggregation_jobs.aggregation_job_id = $2
+                    AND report_aggregations.client_report_id = $3
+                    AND UPPER(client_timestamp_interval) >= $4",
+            )
+            .await?;
         self.query_opt(
             &stmt,
             &[
@@ -2309,6 +2313,7 @@ impl<C: Clock> Transaction<'_, C> {
                 FROM report_aggregations
                 JOIN aggregation_jobs ON aggregation_jobs.id = report_aggregations.aggregation_job_id
                 WHERE report_aggregations.task_id = $1
+                  AND aggregation_jobs.task_id = $1
                   AND UPPER(aggregation_jobs.client_timestamp_interval) >= $2",
             )
             .await?;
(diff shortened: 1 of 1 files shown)


branlwyd (commented):
I'm confused why this is necessary, apologies if I'm missing something obvious: in all queries touched by this PR, the primary join condition is on `aggregation_jobs.id = report_aggregations.aggregation_job_id`, i.e. a primary-key lookup via `aggregation_jobs`' primary key index based on the relevant report aggregation's `aggregation_job_id`. I would not expect lookup of aggregation job by (DAP task ID, DAP aggregation job ID) to be necessary at all.

divergentdave:
Ah, the change in `check_other_report_aggregation_exists()` may be moot because of that, and also the condition on `aggregation_jobs.aggregation_job_id` is `!=`, not `=`.

In the case of `get_report_aggregations_for_aggregation_job()`, the database would need to look up the aggregation job first by its DAP IDs, and then join from that to `report_aggregations`. Here's a comparison of query plans, note they have the same topology, and scan the same indices, but one has to scan more of the index.

```
(before PR)
postgres=> EXPLAIN SELECT ord, client_report_id, client_timestamp, last_prep_resp, report_aggregations.state, public_share, leader_extensions, leader_input_share, helper_encrypted_input_share, leader_prep_transition, helper_prep_state, error_code FROM report_aggregations JOIN aggregation_jobs ON aggregation_jobs.id = report_aggregations.aggregation_job_id WHERE report_aggregations.task_id = 20 AND aggregation_jobs.aggregation_job_id = '\x00' AND UPPER(client_timestamp_interval) >= '2024-03-06Z' ORDER BY ord ASC;
                                                          QUERY PLAN                                                          
------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=191679.52..191679.65 rows=54 width=263)
   Sort Key: report_aggregations.ord
   ->  Nested Loop  (cost=1.25..191677.96 rows=54 width=263)
         ->  Index Scan using aggregation_jobs_unique_id on aggregation_jobs  (cost=0.55..186415.80 rows=1 width=8)
               Index Cond: (aggregation_job_id = '\x00'::bytea)
               Filter: (upper(client_timestamp_interval) >= '2024-03-06 00:00:00'::timestamp without time zone)
         ->  Index Scan using report_aggregations_unique_ord on report_aggregations  (cost=0.70..5216.09 rows=4608 width=271)
               Index Cond: ((task_id = 20) AND (aggregation_job_id = aggregation_jobs.id))
(8 rows)

(after PR)
postgres=> EXPLAIN SELECT ord, client_report_id, client_timestamp, last_prep_resp, report_aggregations.state, public_share, leader_extensions, leader_input_share, helper_encrypted_input_share, leader_prep_transition, helper_prep_state, error_code FROM report_aggregations JOIN aggregation_jobs ON aggregation_jobs.id = report_aggregations.aggregation_job_id WHERE report_aggregations.task_id = 20 AND aggregation_jobs.task_id = 20 AND aggregation_jobs.aggregation_job_id = '\x00' AND UPPER(client_timestamp_interval) >= '2024-03-06Z' ORDER BY ord ASC;
                                                          QUERY PLAN                                                          
------------------------------------------------------------------------------------------------------------------------------
 Sort  (cost=5266.50..5266.63 rows=54 width=263)
   Sort Key: report_aggregations.ord
   ->  Nested Loop  (cost=1.25..5264.94 rows=54 width=263)
         ->  Index Scan using aggregation_jobs_unique_id on aggregation_jobs  (cost=0.55..2.78 rows=1 width=8)
               Index Cond: ((task_id = 20) AND (aggregation_job_id = '\x00'::bytea))
               Filter: (upper(client_timestamp_interval) >= '2024-03-06 00:00:00'::timestamp without time zone)
         ->  Index Scan using report_aggregations_unique_ord on report_aggregations  (cost=0.70..5216.09 rows=4608 width=271)
               Index Cond: ((task_id = 20) AND (aggregation_job_id = aggregation_jobs.id))
(8 rows)
```

Regarding `update_report_aggregation()`, in that case we need to identify the `aggregation_jobs` row before we can identify the `report_aggregations` row as well. The predicates and the `ReportAggregation` struct over-determine the tuple to be modified with the DAP TaskId, DAP AggregationJobId, the `ord` within the aggregation job, and the client ReportId and timestamp. (i.e. we could uniquely determine the tuple with certain subsets of these five attributes) When it comes to the physical plan, we can't look up the report aggregation first, and then use `report_aggregations.aggregation_job_id` to look up the aggregation job by primary key, because we start the query with only the task's primary key, and the full set of DAP IDs. I think the best way to look up the `report_aggregations` tuple would be via the `UNIQUE` index covering the task primary key, the aggregation job primary key, and the `ord`, presupposing that we look up the aggregation job first.

The changes to the other two methods we can take or leave because they are `test-util`-only.

branlwyd (approved):
Ah, that's fair, thank you for clarifying. If you don't mind, please clean up the unnecessary `report_aggregation.task_id` conditions in the case where we want to check `aggregation_jobs.task_id` for efficiency -- it would be good to have a minimal set of conditions, so that well-meaning folks don't introduce inefficient joins in future changes.

divergentdave:
I'd actually prefer to leave both predicates in, under the rubric of giving the query planner as much information as we can.

Taking `get_report_aggregations_for_aggregation_job()` as an example, the report aggregations could be looked up using either the `report_aggregations_unique_ord` index or the `report_aggregations_aggregation_job_id_index` index. Just as the `aggregation_jobs.task_id` predicate enables the use of `aggregation_jobs_unique_id` to look up the aggregation job with a single btree traversal, the `report_aggregations.task_id` predicate enables the use of `report_aggregations_unique_ord` to fetch the report aggregations with an index scan, and no sorting. If we didn't have the predicate on `report_aggregations.task_id`, then using `report_aggregations_aggregation_job_id_index` would likely be the next best option. But since this index only covers the `aggregation_job_id` column, the tuples would be retrieved in an arbitrary order, and they would have to be sorted by `ord` separately.

`ORDER BY` aside, I would be hard pressed to determine whether using `report_aggregations_unique_ord` or `report_aggregations_aggregation_job_id_index` would be categorically more efficient, across all ranges of table cardinalities, etc. By keeping both options on the table, we can let the planner adaptively choose, based on the current table and index statistics. The two possible downsides I see are evaluating extra filter conditions and bad plans due to bad row count estimates. Whenever some predicate that is not strictly necessary doesn't get used in an index condition, it will get checked as a filter condition later. This is cheap enough that I don't think it is of concern. As for bad row count estimates, my hope is that keeping both predicates will help more than hurt, because the `task_id` can put a good ceiling on the selectivity over each table.

tgeoghegan (approved):

PR #2867, opened 2024-03-21
Reverse proxy inside `janus_interop_aggregator`

### LINKED ISSUE #446: Interop-test containers should use the same binaries as a production deployment.

divergentdave:
This closes #446. The aggregator interop container is changed so that it runs the aggregator component as a separate process, managed by supervisord, and the `janus_interop_aggregator` binary reverse proxies DAP requests it receives to the aggregator. I also moved schema migration out of `janus_interop_aggregator`, so we can sequence it before starting other processes. This was necessary because the aggregator component doesn't tolerate missing tables well when it loads global HPKE keys at startup. With this change, the `janus_interop_aggregator` binary is much smaller, since it doesn't include the entire aggregator.

I also cleaned up old `ENV CARGO_REGISTRIES_CRATES_IO_PROTOCOL=sparse` lines, because this option is now the default, and has been for a while.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -710,6 +710,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "cidr"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8d18b093eba54c9aaa1e3784d4361eb2ba944cf7d0a932a830132238f483e8d8"
+
+[[package]]
 name = "cipher"
 version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


inahga (approved):
Nice!

Issue #2312, opened 2023-12-04
Produce single Janus binary, container image


tgeoghegan:
Currently, we build separate binaries and containers for `janus_aggregator`, `janus_aggregation_job_creator`, `janus_aggregation_job_driver` and `janus_collection_job_driver`. Besides serving HTTP requests, `janus_aggregator` also optionally can do garbage collection.

We should instead build a single binary, named `janus_aggregator`, that can be configured to do any of these five things (serve HTTP, periodically schedule aggregation jobs, consume aggregation jobs, consume collection jobs or collect garbage) based on a subcommand. Correspondingly we would build and distribute a single container image.

To make this easy for existing usage of `janus_aggregator` as a helper, we could rig up the container's entrypoint so that by default, it acts as an HTTP handler (which implements everything needed to act as a DAP helper).

tgeoghegan:
We should go do this soon, while we're in a breaking changes window for Janus 0.7.

inahga:
I think we could take this as a non-breaking change, for the sake of backporting to 0.6.

We can do this by backporting the single binary functionality, but also have the binary parse its own binary name, `argv[0]`, to determine which subcommand to run.

```rust
let bin = std::env::args.collect::<Vec<_>>()[0];
match bin {
     "janus" => { /* behave normally */ },
     "janus_aggregration_job_driver" => { /* behave as if we're aggregation_job_driver */ }
    ...
}
```

We maintain backwards compatibility by having the containers be renamed copies of the main binary that determine functionality by their binary name. This is a similar theory as to how busybox works.

This should simplify our deployment code since we can use the same code for 0.6 and 0.7. Not sure if it's worth the effort though, just throwing the idea out there.

inahga:
> We should instead build a single binary, named janus_aggregator, that can be configured to do any of these five things (serve HTTP, periodically schedule aggregation jobs, consume aggregation jobs, consume collection jobs or collect garbage) based on a subcommand. Correspondingly we would build and distribute a single container image.

FWIW I think we should roll the `janus_cli` into it as well, so we truly only have one binary.

divergentdave:
We can make use of clap's `multicall` flag, and add a subcommand enum around our existing command line parsers.

Bonus: we could build one image with the binary hardlinked under several names, and tag it with all five image names.

PR #2873, opened 2024-03-22
Unify interop test binary targets

divergentdave:
This combines the three binary targets under `janus_interop_binaries` into one. It can either be used via symlinks with the original names of the three binaries, dispatching on argv[0], or with its default name and a subcommand. Dockerfiles are updated so that they work the same way as before. Moreover, we can now share builder layers between all three interop test containers. Overall, this should slightly improve build time and total image size. This is a warm-up for #2312. Note that I had to move `Options::parse()` out of `janus_main()` here; this will be doubly necessary when implementing #2312.

### DIFF
--- a/Dockerfile.interop
+++ b/Dockerfile.interop
@@ -45,9 +45,9 @@ FROM alpine:3.19.1 AS final
 ARG BINARY
 ARG PROFILE
 RUN mkdir /logs
-COPY --from=builder /src/target/$PROFILE/janus_interop_binaries /janus_interop_binaries
-RUN ln -s /janus_interop_binaries /janus_interop_client
-RUN ln -s /janus_interop_binaries /janus_interop_collector
+COPY --from=builder /src/target/$PROFILE/janus_interop /janus_interop
+RUN ln -s /janus_interop /janus_interop_client
+RUN ln -s /janus_interop /janus_interop_collector
 EXPOSE 8080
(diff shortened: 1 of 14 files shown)


interop_binaries/src/main.rs
```diff
@@ -0,0 +1,44 @@
+use clap::{Parser, Subcommand};
+use janus_interop_binaries::commands::{
+    janus_interop_aggregator, janus_interop_client, janus_interop_collector,
+};
+
+#[derive(Debug, Parser)]
+#[clap(multicall = true)]
```
inahga, line 7: Awesome!

Dockerfile.interop
```diff
@@ -39,15 +39,15 @@ COPY interop_binaries /src/interop_binaries
 COPY messages /src/messages
 COPY tools /src/tools
 COPY xtask /src/xtask
-ARG BINARY
-RUN cargo build --features fpvec_bounded_l2 --profile $PROFILE -p janus_interop_binaries \
-    --bin $BINARY
+RUN cargo build --features fpvec_bounded_l2 --profile $PROFILE -p janus_interop_binaries
```
inahga, line 42: nit: `janus_interop_binaries` is a bit of an awkward name for a binary, I'd just call it `janus_interop` if it's not too much work to plumb through.

inahga (approved):

PR #2915, opened 2024-03-26
Run `cargo update`.

branlwyd:
Many of our dependencies are out-of-date in terms of minor version. I noticed this because some of Janus' dependents failed to take an update to the Janus crates due to using a too-old version of the chrono crate which was missing a function introduced in a later minor version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
This looks fine, but we will need to update `Cargo.toml` to address issues in downstream crates automatically, without requiring them to run `cargo update` too. This kind of issue could be caught by [`cargo-minimal-versions`](https://github.com/taiki-e/cargo-minimal-versions), but our practice of using `0.x` requirements for intra-workspace dependencies would also be flagged by this (once we add a new item in one crate and use it in another in a patch release)

PR #2918, opened 2024-03-27
Add metric for aggregation success.

branlwyd:
The new "janus_aggregation_success_counter" tracks the number of successfully-aggregated reports. This metric is not keyed by task ID or anything else, so it will be most useful (a) in single-teneant deployments or (b) as an overall metric of aggregation throughput.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -6,7 +6,7 @@ use crate::{
         aggregate_share::compute_aggregate_share,
         aggregation_job_writer::{
             AggregationJobWriter, InitialWrite, ReportAggregationUpdate as _,
-            WritableReportAggregation,
+            WritableReportAggregation, AggregationJobWriterMetrics,
         },
         error::{handle_ping_pong_error, ReportRejection, ReportRejectionReason},
         error::{BatchMismatch, OptOutReason},
@@ -117,6 +117,16 @@ pub mod report_writer;
 #[cfg(test)]
 mod taskprov_tests;
 
+pub(crate) fn aggregation_success_counter(meter: &Meter) -> Counter<u64> {
+    let aggregation_success_counter = meter
+        .u64_counter("janus_aggregation_success_counter")
+        .with_description("Number of successfully-aggregated report shares")
+        .with_unit(Unit::new("{report}"))
+        .init();
+    aggregation_success_counter.add(0, &[]);
+    aggregation_success_counter
+}
+
 pub(crate) fn aggregate_step_failure_counter(meter: &Meter) -> Counter<u64> {
     let aggregate_step_failure_counter = meter
         .u64_counter("janus_step_failures")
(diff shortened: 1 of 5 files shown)


aggregator/src/aggregator.rs
```diff
@@ -117,6 +117,16 @@ pub mod report_writer;
 #[cfg(test)]
 mod taskprov_tests;
 
+pub(crate) fn aggregation_success_counter(meter: &Meter) -> Counter<u64> {
```
tgeoghegan, line 120: nit/suggestion: we should avoid growing this file if we can. `janus_aggregator::metrics` seems like a more natural place for this item, if it's going to be `pub(crate)` anyway. `aggregate_step_failure_counter` could be moved there too, but perhaps not in this PR.

tgeoghegan, line 120: separate nit/suggestion: would be nice to align the name of this item and of the metric with `aggregate_step_failure_counter`. That said, the other metric's name is bad and needs work (`janus_step_failure` doesn't tell me what was stepped), and we don't want to fix that here since changing metric names involves gymnastics in alerts and Grafana charts.

So concretely: how about `janus_report_aggregation_success_counter`, unless we feel that the `report` unit makes this clear enough?

aggregator/src/aggregator.rs
```diff
@@ -180,6 +190,8 @@ pub struct Aggregator<C: Clock> {
     /// Counter tracking the number of failed message decodes while handling the
     /// `tasks/{task-id}/reports` endpoint.
     upload_decode_failure_counter: Counter<u64>,
+    /// Counter tracking the number of successfully-aggregated reports.
+    aggregation_success_counter: Counter<u64>,
```
tgeoghegan, line 194: We might want to put all these `Counter<T>`s behind some kind of `struct AggregatorMetrics`, so that we'd have fewer functions to update in the future when we add new metrics. This would also give us more runway before we hit `too_many_arguments` lints.

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -657,12 +664,16 @@ where
 
                     match ra_batch_aggregation.merged_with(batch_aggregation) {
                         Ok(merged_batch_aggregation) => {
+                            if let Some(metrics) = self.writer.metrics.as_ref() {
+                                metrics.aggregation_success_counter.add(1, &[])
+                            }
```
tgeoghegan, line 669: related to a previous suggestion: if we defined a `struct AggregatorMetrics`, then you could hide the `if let Some(metrics)` logic in there, and you'd have less metrics noise in application logic.

tgeoghegan (approved):
Nothing but nits that I'm happy to defer. This does what it says it will do and I agree we should do that thing.

aggregator/src/aggregator.rs
branlwyd: I'm fine with `janus_report_aggregation_success_counter`; I avoided matching the name of the other metric too closely as this tracks success of entire aggregations rather than individual aggregation steps.

aggregator/src/aggregator/aggregation_job_writer.rs
branlwyd: Sure, but note that I maintain separation between `AggregatorMetrics` and `AggregationJobWriterMetrics`, as these two types are & should remain separate (unless/until we do a change to use a single `AggregatorMetrics` struct for all metrics storage across the entire `janus_aggregator` crate, which IMO is definitely for another PR).

PR #2920, opened 2024-03-29
Janus CLI commands to configure taskprov aggregators.

branlwyd:
These don't allow complete management, but they are sufficient to allow initial configuration.


branlwyd:
(N.B. this is because we're not planning to pair the taskprov-enabled environment with our API.)

aggregator/src/binaries/janus_cli.rs
```diff
@@ -54,6 +65,86 @@ pub async fn run(command_line_options: CommandLineOptions) -> Result<()> {
...
+        /// The peer's endpoint, as a URL.
+        #[arg(long)]
+        peer_endpoint: Url,
+
+        /// This aggregator's role.
+        #[arg(long)]
+        role: Role,
+
+        /// The taskprov verify_key_init value, in unpadded base64url.
+        #[arg(long, env = "VERIFY_KEY_INIT")]
```
inahga, line 122:
```suggestion
        #[arg(long, env = "VERIFY_KEY_INIT", hide_env_values = true)]
```
(ouch, there's other secrets here that we don't hide_env_values for. I can run through these in another PR, or you can if you want)

aggregator/src/binaries/janus_cli.rs
```diff
@@ -54,6 +65,86 @@ pub async fn run(command_line_options: CommandLineOptions) -> Result<()> {
...
+        /// the datastore, in seconds.
+        #[arg(long)]
+        report_expiry_age_secs: Option<u64>,
+
+        /// The amount of clock skew that the system will accept, in seconds.
+        #[arg(long)]
+        tolerable_clock_skew_secs: u64,
+
+        /// The aggregator auth token, which must be in the format `bearer:value` or `dap:value`.
+        #[arg(long, env = "AGGREGATOR_AUTH_TOKEN")]
```
inahga, line 140:
```suggestion
        #[arg(long, env = "AGGREGATOR_AUTH_TOKEN", hide_env_values = true)]
```

aggregator/src/binaries/janus_cli.rs
```diff
@@ -54,6 +65,86 @@ pub async fn run(command_line_options: CommandLineOptions) -> Result<()> {
...
+        aggregator_auth_token: String,
+
+        /// The collector auth token, which must be in the format `bearer:value` or `dap:value`.
+        #[arg(long, env = "COLLECTOR_AUTH_TOKEN")]
```
inahga, line 144:
```suggestion
        #[arg(long, env = "COLLECTOR_AUTH_TOKEN", hide_env_values = true)]
```

aggregator/src/binaries/janus_cli.rs
```diff
@@ -91,6 +182,101 @@ impl Command {
...
+                    kubernetes_secret_options,
+                    command_line_options,
+                    config_file,
+                    &kube_client,
+                )
+                .await?;
+
+                // Parse flags into proper types.
+                let verify_key_init =
+                    VerifyKeyInit::try_from(URL_SAFE_NO_PAD.decode(verify_key_init)?.as_ref())?;
```
inahga, line 256: nit: we can make an `impl FromStr for VerifyKeyInit` with this logic, similar to what we have in messages, then clap can directly parse a VerifyKeyInit.

I generally like to coerce clap into using a ValueParser https://docs.rs/clap/latest/clap/macro.value_parser.html, which in practice has usually involved implementing `FromStr`. This makes clap print pretty error messages, and generally `FromStr` is nice to have around elsewhere.

Same idea with the bearer tokens below. Can take it or leave it though, it's a minor refactor.

aggregator/src/binaries/janus_cli.rs
```diff
@@ -155,6 +341,95 @@ async fn install_tracing_and_metrics_handlers(
...
+    kem: HpkeKemId,
+    kdf: HpkeKdfId,
+    aead: HpkeAeadId,
+    hpke_config_out_file: Option<&Path>,
+) -> Result<()> {
+    let hpke_keypair = Arc::new(generate_hpke_config_and_private_key(id, kem, kdf, aead)?);
+
+    if !dry_run {
+        datastore
+            .run_tx("generate-global-hpke-key", |tx| {
```
inahga, line 357:
```suggestion
            .run_tx("generate_global_hpke_key", |tx| {
```
nit: aligns with our other tx names.

aggregator/src/binaries/janus_cli.rs
```diff
@@ -155,6 +341,95 @@ async fn install_tracing_and_metrics_handlers(
...
+
+async fn set_global_hpke_key_state<C: Clock>(
+    datastore: &Datastore<C>,
+    dry_run: bool,
+    id: HpkeConfigId,
+    state: HpkeKeyState,
+) -> Result<()> {
+    if !dry_run {
+        datastore
+            .run_tx("set-global-hpke-key-state", |tx| {
```
inahga, line 380:
```suggestion
            .run_tx("set_global_hpke_key_state", |tx| {
```

aggregator/src/binaries/janus_cli.rs
```diff
@@ -155,6 +341,95 @@ async fn install_tracing_and_metrics_handlers(
...
+        collector_hpke_config,
+        report_expiry_age,
+        tolerable_clock_skew,
+        Vec::from([aggregator_auth_token.clone()]),
+        collector_auth_tokens,
+    ));
+
+    if !dry_run {
+        datastore
+            .run_tx("add-taskprov-peer-aggregator", |tx| {
```
inahga, line 422:
```suggestion
            .run_tx("add_taskprov_peer_aggregator", |tx| {
```

aggregator/src/binaries/janus_cli.rs
```diff
@@ -155,6 +341,95 @@ async fn install_tracing_and_metrics_handlers(
...
+                let peer_aggregator = Arc::clone(&peer_aggregator);
+
+                Box::pin(async move { tx.put_taskprov_peer_aggregator(&peer_aggregator).await })
+            })
+            .await?;
+    }
+
+    Ok(())
+}
+
```
inahga, line 432: We may also want to plumb through `delete_taskprov_peer_aggregator`, since if we make a mistake in the configuration we'll otherwise have to go into the DB to drop it.

inahga (approved):
Looks good, I have only nits.

tools/Cargo.toml
```diff
@@ -17,6 +17,7 @@ base64.workspace = true
 clap.workspace = true
 derivative.workspace = true
 fixed = { version = "1.26", optional = true }
+janus_aggregator_core.workspace = true
```
divergentdave, line 20: This import appears to be unused now. Plus, it would prevent us from publishing the hpke_keygen/dap_decode/collect tools to crates.io in the future.

aggregator/src/binaries/janus_cli.rs
branlwyd: ah, nice spot; there are a few test tx's with names that use dashes rather than underscores, but I'll leave them be for now.

aggregator/src/binaries/janus_cli.rs
branlwyd: Agreed, but I'm going to leave this for a later PR. If we make a mistake during configuration, we really can hop into the DB without much risk as the relevant deployment will not be used until after configuration is complete.

aggregator/src/binaries/janus_cli.rs
branlwyd: Hmm, I like this idea. Implementing on `VerifyKeyInit` seems like an obvious win as there is already serde code that uses the same format. `AuthenticationToken` was less obvious; I considered wrapping a newtype, but I decided I'm OK adding a `FromStr` as there was no string format for an `AuthenticationToken` yet, and prefixed-value is flexible enough to handle any new token types that may come to exist.

PR #2921, opened 2024-04-01
Hoist all dependencies up to top-level Cargo.toml

tgeoghegan:
Use `cargo-autoinherit` to rewrite all the workspace `Cargo.toml`s to pull in dependencies from the top-level using `workspace = true`. I then hand-sorted the top-level `Cargo.toml` because I couldn't get `cargo-sort` to work.

### DIFF
--- a/aggregator_api/Cargo.toml
+++ b/aggregator_api/Cargo.toml
@@ -10,20 +10,20 @@ version.workspace = true
 
 [dependencies]
 anyhow.workspace = true
-async-trait = "0.1"
+async-trait = { workspace = true }
 base64.workspace = true
 derivative.workspace = true
 janus_aggregator_core.workspace = true
 janus_core.workspace = true
 janus_messages.workspace = true
 opentelemetry.workspace = true
-querystring = "1.1.0"
+querystring = { workspace = true }
 rand = { workspace = true, features = ["min_const_gen"] }
-ring = "0.17.8"
+ring = { workspace = true }
 serde.workspace = true
 serde_json.workspace = true
 thiserror.workspace = true
-tracing = "0.1.40"
+tracing = { workspace = true }
 trillium.workspace = true
 trillium-api.workspace = true
 trillium-opentelemetry.workspace = true
@@ -32,7 +32,7 @@ url.workspace = true
 
 [dev-dependencies]
 assert_matches.workspace = true
-futures = "0.3.30"
+futures = { workspace = true }
 janus_aggregator_core = { workspace = true, features = ["test-util"] }
 rstest.workspace = true
 serde_test.workspace = true
--- a/aggregator_core/Cargo.toml
+++ b/aggregator_core/Cargo.toml
@@ -22,16 +22,16 @@ test-util = [
 
 [dependencies]
 anyhow.workspace = true
-async-trait = "0.1"
-backoff = { version = "0.4.0", features = ["tokio"] }
+async-trait = { workspace = true }
+backoff = { workspace = true, features = ["tokio"] }
 base64.workspace = true
 bytes.workspace = true
 chrono.workspace = true
-deadpool = { version = "0.11.0", features = ["rt_tokio_1"] }
-deadpool-postgres = "0.13.0"
+deadpool = { workspace = true, features = ["rt_tokio_1"] }
+deadpool-postgres = { workspace = true }
 derivative.workspace = true
-futures = "0.3.30"
-hex = { version = "0.4.3", features = ["serde"], optional = true }
+futures = { workspace = true }
+hex = { workspace = true, features = ["serde"], optional = true }
 http.workspace = true
 http-api-problem.workspace = true
 itertools = { workspace = true, optional = true }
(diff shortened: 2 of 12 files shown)


aggregator_core/Cargo.toml
```diff
@@ -40,25 +40,25 @@ janus_messages.workspace = true
 k8s-openapi.workspace = true
 kube.workspace = true
 opentelemetry.workspace = true
-postgres-protocol = "0.6.6"
-postgres-types = { version = "0.2.6", features = ["derive", "array-impls"] }
+postgres-protocol = { workspace = true }
+postgres-types = { workspace = true, features = ["derive", "array-impls"] }
```
inahga, line 44: Darn, it doesn't look like `cargo-autoinherit` factors out common features--both usages of `postgres-types` use the same featureset.

inahga (commented):
Could we also factor out common features?

tgeoghegan:
> Could we also factor out common features?

We could, but I did this because `cargo autoinherit` made it fast and low-risk, whereas factoring out features is a chore I'd have to do by hand.

inahga (approved):

messages/Cargo.toml
```diff
@@ -16,15 +16,15 @@ test-util = []
...
 base64.workspace = true
 derivative.workspace = true
-hex = "0.4"
-num_enum = "0.7.2"
+hex = { workspace = true }
+num_enum = { workspace = true }
 # We can't pull prio in from the workspace because that would enable default features, and we do not
 # want prio/crypto-dependencies
-prio = { version = "0.16.2", default-features = false, features = ["multithreaded", "experimental"] }
+prio = { workspace = true, features = ["multithreaded", "experimental"] }
```
divergentdave, line 23: We should move `default-features = false` up to the workspace level (or roll this line back) to avoid requiring the crypto-dependencies feature from here when not necessary. Mozilla isn't currently using `janus_messages`, but this change would block them from doing so.

divergentdave (approved):

aggregator_api/Cargo.toml
```diff
@@ -10,20 +10,20 @@ version.workspace = true
...
-querystring = "1.1.0"
+querystring = { workspace = true }
 rand = { workspace = true, features = ["min_const_gen"] }
-ring = "0.17.8"
+ring = { workspace = true }
 serde.workspace = true
 serde_json.workspace = true
 thiserror.workspace = true
-tracing = "0.1.40"
+tracing = { workspace = true }
```
branlwyd, line 26: Ideally, rather than having two formats (`foo.workspace = true`, `foo = { workspace = true }`) we'd settle on just one. I'm fine with either.

branlwyd (approved):

aggregator_api/Cargo.toml
tgeoghegan: I agree, but I don't think `cargo autoinherit` supports choosing a style like that.

messages/Cargo.toml
divergentdave: I think that's insufficient. I tried this patch:

```patch
diff --git a/messages/Cargo.toml b/messages/Cargo.toml
index 8c8515a1..d0c9bdfc 100644
--- a/messages/Cargo.toml
+++ b/messages/Cargo.toml
@@ -20,7 +20,7 @@ hex = { workspace = true }
 num_enum = { workspace = true }
-prio = { workspace = true, features = ["multithreaded", "experimental"] }
+prio = { workspace = true, default-features = false, features = ["multithreaded", "experimental"] }
 rand.workspace = true
 serde.workspace = true
 thiserror.workspace = true
```

Running `cargo tree -p janus_messages` with this applied, I see the following.

```
...
├── prio v0.16.2
│   ├── aes v0.8.4
...
```

Using this patch instead, we get what we want. The same `cargo tree -p janus_messages` command shows that `prio` is only pulling in `sha3`, not `aes` etc.

```patch
diff --git a/Cargo.toml b/Cargo.toml
index 42c9ff83..986b3f30 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,7 @@ opentelemetry_sdk = { version = "0.22", features = ["metrics"] }
 opentelemetry-semantic-conventions = "0.14"
 postgres-protocol = "0.6.6"
 postgres-types = "0.2.6"
-prio = { version = "0.16.2", features = ["multithreaded", "experimental"] }
+prio = { version = "0.16.2", default-features = false, features = ["multithreaded", "experimental"] }
 prometheus = "0.13.3"
 querystring = "1.1.0"
 rand = "0.8"
diff --git a/core/Cargo.toml b/core/Cargo.toml
index 6e822dce..d3e37301 100644
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -48,7 +48,7 @@ http-api-problem.workspace = true
 janus_messages.workspace = true
 kube = { workspace = true, optional = true, features = ["rustls-tls"] }
 k8s-openapi = { workspace = true, optional = true }
-prio.workspace = true
+prio = { workspace = true, default-features = true }
 rand.workspace = true
 regex = { workspace = true }
 reqwest = { workspace = true, features = ["json"] }
diff --git a/messages/Cargo.toml b/messages/Cargo.toml
index 8c8515a1..2186ecd9 100644
--- a/messages/Cargo.toml
+++ b/messages/Cargo.toml
@@ -18,8 +18,6 @@ base64.workspace = true
 derivative.workspace = true
 hex = { workspace = true }
 num_enum = { workspace = true }
-# We can't pull prio in from the workspace because that would enable default features, and we do not
-# want prio/crypto-dependencies
 prio = { workspace = true, features = ["multithreaded", "experimental"] }
 rand.workspace = true
 serde.workspace = true
```

This is because features selected in package files are additive on top of those selected in the workspace file. If we don't include `default-features = false` at the workspace level, there's no way to go back.

aggregator_api/Cargo.toml
tgeoghegan: https://github.com/mainmatter/cargo-autoinherit/issues/15

PR #2922, opened 2024-04-01
Update minimum version of chrono

divergentdave:
This sets the minimum version of chrono to 0.4.33, which is when `Duration::try_seconds()` was introduced.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -696,9 +696,9 @@ dependencies = [
 
 [[package]]
 name = "chrono"
-version = "0.4.36"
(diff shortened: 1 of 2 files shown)


tgeoghegan (approved):
No objection to updating `chrono`, but I'm curious whether adopting `Duration::try_seconds` unblocks something we want?

divergentdave:
`Duration::seconds()` was deprecated in 0.4.35, (then un-deprecated in 0.4.37) so we started using it to fix warnings.

PR #2925, opened 2024-04-01
Rephrase description of signing key option

divergentdave:
This applies a late suggestion I made on #2880.

While trying to propagate this to the trycmd tests, I discovered that #2879 made trycmd ignore relevant tests, since binary names changed. I'll fix that separately.

### DIFF
--- a/aggregator/src/binaries/aggregator.rs
+++ b/aggregator/src/binaries/aggregator.rs
@@ -215,8 +215,10 @@ pub struct Options {
     )]
     pub aggregator_api_auth_tokens: Vec<String>,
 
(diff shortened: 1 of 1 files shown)


branlwyd:
PEM-encoding can in general be used on any binary data; PEM itself does not imply DER-encoding (or any other encoding): https://datatracker.ietf.org/doc/html/rfc7468

The relevant RFC suggests use of DER as a `SHOULD`, so I think mentioning the format is useful: https://datatracker.ietf.org/doc/html/rfc5915#section-4

branlwyd (approved):
(that said, rephrasing the doccomment while keeping the DER-encoding note makes sense IMO, so I'm going to LGTM)

PR #2954, opened 2024-04-02
interop: Wait for subprocesses to be ready before reporting ready

inahga:
See https://github.com/divviup/divviup-android/issues/101#issuecomment-2032813271, if the subprocesses aren't ready, we shouldn't report that the interop API is ready.

### DIFF
--- a/interop_binaries/src/commands/janus_interop_aggregator.rs
+++ b/interop_binaries/src/commands/janus_interop_aggregator.rs
@@ -140,17 +140,18 @@ async fn make_handler(
     let upstream = format!("http://{aggregator_address}/").into_upstream();
     let proxy_handler = Proxy::new(
         Client::new(ClientConfig::default()).with_default_pool(),
-        upstream.clone(),
+        upstream,
     );
+    let health_check_client = Client::new(ClientConfig::default()).with_default_pool();
 
     let handler = Router::new()
         .post("internal/test/ready", move |conn: Conn| {
             let health_check_peers = health_check_peers.clone();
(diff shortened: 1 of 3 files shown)


interop_binaries/src/commands/janus_interop_aggregator.rs
```diff
@@ -130,17 +133,36 @@ async fn make_handler(
...
     aggregator_address: SocketAddr,
+    health_check_peers: Vec<Url>,
 ) -> anyhow::Result<impl Handler> {
     let keyring = Keyring::new();
 
     let upstream = format!("http://{aggregator_address}/").into_upstream();
     let proxy_handler = Proxy::new(
         Client::new(ClientConfig::default()).with_default_pool(),
-        upstream,
+        upstream.clone(),
```
divergentdave, line 143: minor nit: this change isn't necessary anymore

interop_binaries/src/commands/janus_interop_aggregator.rs
```diff
@@ -130,17 +133,36 @@ async fn make_handler(
...
     );
 
     let handler = Router::new()
-        .post("internal/test/ready", Json(serde_json::Map::new()))
+        .post("internal/test/ready", move |conn: Conn| {
+            let health_check_peers = health_check_peers.clone();
+            let client = Client::new(ClientConfig::default()).with_default_pool();
```
divergentdave, line 149: We could save a bit of TCP connection setup if we constructed one client, then used a clone in the proxy, and a clone for the health check handler. It's not a big deal, since this is all test-only.

divergentdave (approved):

Issue #2812, opened 2024-03-14
Improve collection job polling


branlwyd:
With #2782, we now poll collection jobs to determine if they can be progressed to a final state (e.g. ready for retrieval by the Collector). Currently, we poll on a fixed schedule. There are several improvements we can make to make this more efficient:

* Polling collection jobs for readiness can be expensive if the batch aggregations are numerous/large. We can make this cheaper if we implement a DB method that sums the relevant aggregation_jobs_{created,terminated} counters and tells us if they are equal, before we go to the effort of reading the aggregate shares & other batch-aggregation data. I think if we are clever we can even do so in a way that doesn't require additional DB round-trips, e.g. "return batch aggregations only if the collection job is ready", perhaps.
* The retry delay for retries of collection jobs which are not ready for collection is fixed. We may well want to turn this into exponential backoff, eventually.
* Perhaps we can come up with a scheme where we only retry collection jobs if at least one relevant aggregation job has terminated since the last try, though I'd need to think through whether this could pull us back into unsharded write-contention on the collection jobs.

branlwyd:
Closing this:
* The first & second ideas above are implemented in the linked PRs.
* The third idea above is, on further thought, unsound in corner-ish cases: if we only retry collection jobs if at least one relevant aggregation job has terminated, we won't retry if an unaggregated client report in the batch interval ends up being GC'ed. We might decide we don't care about such cases, but for now, I don't plan to pursue this idea -- I think the above two ideas will make collection job polling cheap enough that we don't need to pursue "edge-triggering" of state changes.

PR #2957, opened 2024-04-03
Collection job driver: reduce work required to detect step readiness.

branlwyd:
Specifically, we now check if there are unaggregated reports, or unfinished aggregation jobs, in the same pipelined round-trip as reading the task and collection job. The outstanding aggregation job count is determined without reading the set of batch aggregations, which can be quite large in some cases as they include partial aggregate shares.

We also now check for a matching, already-completed collection job before reading the batch aggregations.


branlwyd:
Part of #2812.

tgeoghegan (approved):

PR #2963, opened 2024-04-04
Exponential backoff for collection job polling.

branlwyd:
This is controlled by a couple of new collection job driver config knobs, which determine the maximum delay between polling attempts & the exponential factor to use. No randomization factor is included as these delays are the _minimum_ delays possible before retry; thundering herd is not a risk because the collection job driver already controls how many collection jobs it tries to step at once.

Also, make min_collection_job_retry_delay optional (along with the new collection job driver configuration options).

### DIFF
--- a/aggregator/src/aggregator/collection_job_driver.rs
+++ b/aggregator/src/aggregator/collection_job_driver.rs
@@ -1807,7 +1807,7 @@ mod tests {
             assert_eq!(
                 want_delay_s,
                 got_delay_s,
-                "compute_reacquire_delay({step_attempts}, {min_delay_s}, {max_delay_s}, {exponential_factor})"
+                "RetryDelay({min_delay_s}, {max_delay_s}, {exponential_factor}).compute_retry_delay({step_attempts})"
             );
         }
 
@@ -1824,7 +1824,7 @@ mod tests {
                 StdDuration::from_secs(max_delay_s),
                 exponential_factor,
             )
-            .unwrap_err();
+            .expect_err(&format!("RetryDelay({min_delay_s}, {max_delay_s}, {exponential_factor})"));
         }
     }
 }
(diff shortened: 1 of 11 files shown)


branlwyd:
Part of #2812. Stacked on #2961. DO NOT MERGE until #2961 is released.

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -48,9 +48,15 @@ pub struct CollectionJobDriver<B> {
...
-    // not ready yet because not all included reports had finished aggregation.
+    /// The minimum duration to wait before retrying a collection job that has been stepped but was
+    /// not ready yet because of pending aggregation.
     min_collection_job_retry_delay: Duration,
+    /// The maximum duration to wait before retrying a collection job that has been stepped but was
+    /// not ready yet because of pending aggregation.
+    max_collection_job_retry_delay: Duration,
+    /// The exponential factor used when computing the next retry delay for a collection job that
+    /// has been stepped but was not ready yet because of pending aggregation.
+    collection_job_retry_delay_exponential_factor: f64,
```
tgeoghegan, line 59: nit: you could factor these three values into a `struct JobRetryStrategy` or something, and then `compute_reacquire_delay` becomes a method on that. It would then be incrementally easier to adopt this exponential backoff logic in the aggregation job driver.

tgeoghegan (approved):

tgeoghegan:
We will need a writeup in operator notes discussing the sequencing of this and #2961.

aggregator/src/aggregator/collection_job_driver.rs
branlwyd: Sure, YAGNI is a risk but we'll plausibly use this again -- I factored into a simple `RetryStrategy` type.

PR #2969, opened 2024-04-06
Schema changes for "aggregation_job_creator: mark filled outstanding …

branlwyd:
…batches."

### DIFF
--- a/db/00000000000003_outstanding_batch_state.down.sql
+++ b/db/00000000000003_outstanding_batch_state.down.sql
@@ -1,4 +1,4 @@
 CREATE INDEX outstanding_batches_task_and_time_bucket_index ON outstanding_batches(task_id, time_bucket_start);
-DROP INDEX outstanding_batches_task_id_and_time_bucket_start CASCADE;
+DROP INDEX outstanding_batches_task_id_and_time_bucket_start;
 ALTER TABLE outstanding_batches DROP COLUMN state;
-DROP TYPE OUTSTANDING_BATCH_STATE CASCADE;
+DROP TYPE OUTSTANDING_BATCH_STATE;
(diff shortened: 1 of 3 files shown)


db/00000000000003_outstanding_batch_state.up.sql
```diff
@@ -0,0 +1,8 @@
+-- Specifies the possible state of an outstanding batch.
+CREATE TYPE OUTSTANDING_BATCH_STATE AS ENUM(
+    'FILLING',  -- this outstanding batch is still being considered for additional reports
+    'FILLED'    -- this outstanding batch has received enough reports, no more are necessary
+);
+ALTER TABLE outstanding_batches ADD COLUMN state OUTSTANDING_BATCH_STATE NOT NULL DEFAULT 'FILLING';
+CREATE INDEX outstanding_batches_task_id_and_time_bucket_start ON outstanding_batches(task_id, time_bucket_start) WHERE state = 'FILLING';
```
inahga, line 7: Hmm. My first thought is that we may want to CREATE INDEX CONCURRENTLY to persuade postgres to take more lenient locks, but then again aggregation job processing is non-critical.

db/00000000000003_outstanding_batch_state.up.sql
```diff
@@ -0,0 +1,8 @@
...
+DROP INDEX outstanding_batches_task_and_time_bucket_index CASCADE;
```
inahga, line 8: Did the CASCADE turn out strictly necessary? It makes me a bit uncomfortable to CASCADE delete things on a production database so I'm curious what are the dependents of this index.

db/00000000000003_outstanding_batch_state.down.sql
```diff
@@ -0,0 +1,4 @@
+CREATE INDEX outstanding_batches_task_and_time_bucket_index ON outstanding_batches(task_id, time_bucket_start);
+DROP INDEX outstanding_batches_task_id_and_time_bucket_start CASCADE;
+ALTER TABLE outstanding_batches DROP COLUMN state;
+DROP TYPE OUTSTANDING_BATCH_STATE CASCADE;
```
inahga, line 4: Similar arguments could apply with the down migration. Although I don't care that much about cascading deletes on a down migration.

inahga (commented):
Looks good, just some mild deployment concerns.

inahga (approved):

PR #2970, opened 2024-04-06
aggregation_job_creator: mark filled outstanding batches.

branlwyd:
These marked batches will no longer be considered when generating aggregation jobs in the future; this is a performance optimization which will be especially impactful when there are many outstanding_batches, which might occur if the collector falls behind in collection.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_creator.rs
+++ b/aggregator/src/aggregator/aggregation_job_creator.rs
@@ -1871,7 +1871,9 @@ mod tests {
 
                     Box::pin(async move {
                         Ok((
-                            tx.get_outstanding_batches(task.id(), &None).await.unwrap(),
+                            tx.get_unfilled_outstanding_batches(task.id(), &None)
+                                .await
+                                .unwrap(),
                             read_and_verify_aggregate_info_for_task::<
                                 VERIFY_KEY_LENGTH,
                                 FixedSize,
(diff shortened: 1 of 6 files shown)


branlwyd:
Stacked on #2969. DO NOT MERGE until #2969 is merged.

inahga (approved):
Makes sense to me.

tgeoghegan (commented):
IIUC, a batch can be filled but still outstanding, which would mean it cannot have new reports added to it but can be collected.

Should `Datastore::acquire_filled_outstanding_batch` be updated to consider `outstanidng_batches.state`? It seems to only matter in the fixed size case.

branlwyd:
> IIUC, a batch can be filled but still outstanding, which would mean it cannot have new reports added to it but can be collected.
> 
> Should `Datastore::acquire_filled_outstanding_batch` be updated to consider `outstanidng_batches.state`?

Good observation: I'd prefer not to use `state` when acquiring a filled batch, because `state` is only updated by the `aggregation_job_creator` (the other point we might try to update `state` is on completion of an aggregation job; but if we tried to update `state` when completing an aggregation job, we would run into write contention similarly to what we saw with the `batches` table).

If we relied on the `state` for acquiring `outstanding_batches`, we would potentially never collect an `outstanding_batch` which has passed the minimum batch size if we never received another report that could be added to the batch. (This matters more when using time-bucketed fixed-size, where the previous condition would require a report to arrive in the time window associated with the `outstanding_batch`.)

That said, this is awkward. Maybe the answer is additional documentation or better naming?


> It seems to only matter in the fixed size case.

`outstanding_batches` in general is only used for fixed-size.

tgeoghegan (approved):
Yeah, I think we might have some batch taxonomy issues. However I don't want to block this performance improvement on the refactor that we might want to do. So let's go ahead and do this, and do no more than add some explanatory comments.

branlwyd:
I renamed the existing `acquire_filled_outstanding_batches` method to `acquire_outstanding_batches_with_report_count`. This is more accurate, and doesn't imply that the functionality will be based on the explicit `FILLED` state.

PR #2994, opened 2024-04-08
[0.6] Schema changes for "aggregation_job_creator: mark filled outsta…

branlwyd:
…nding batches."

PR #2995, opened 2024-04-08
[0.6] aggregation_job_creator: mark filled outstanding batches.

branlwyd:
These marked batches will no longer be considered when generating aggregation jobs in the future; this is a performance optimization which will be especially impactful when there are many outstanding_batches, which might occur if the collector falls behind in collection.


branlwyd:
Stacked on #2994. DO NOT MERGE until #2994 is merged.

inahga (approved):

PR #2997, opened 2024-04-09
Bump version to 0.7.3

divergentdave:
This bumps the version on the main branch. I plan to cut a release manually, in order to include a note about #2971.


inahga (approved):
We can edit the release notes after they've been created, that's how I've been adding additional details when using `make-release`.

Up to you, though.

Issue #2959, opened 2024-04-04
Where possible, don't let an issue with one report share cause an aggregation job to fail


divergentdave:
I noticed that we have some places in aggregation job handling where we propagate errors when not strictly necessary, such as `public_share.get_encoded()?` (though this ought to be unreachable from a message we just decoded off the wire). These could be turned into prepare errors, which would let the rest of the aggregation job proceed.

divergentdave:
Fixed in #3002.

PR #3002, opened 2024-04-10
Fixup aggregator per-report-share error handling.

branlwyd:
There were a few places where a per-report-share failure would result in failure of processing of the entire aggregation job. Now, these locations will fail only the report share in question.

Also:
 * Reformat a few comments that had crept past 100 columns.
 * Replace `panic_any` with `resume_unwind` when we are resuming a panic.
 * Remove implicit conversion from CodecError to Error::MessageDecode.


branlwyd:
Closes #2959.

aggregator/src/aggregator.rs
```diff
@@ -2086,7 +2111,7 @@ impl VdafOps {
                 Error::Internal("threadpool failed to send VDAF preparation result".into())
             })?
             .unwrap_or_else(|panic_cause: Box<dyn Any + Send>| {
-                panic_any(panic_cause);
+                resume_unwind(panic_cause);
```
branlwyd, line 2114: A note on this change: calling `panic_any` will result in a second panic, with all associated logging; `resume_unwind` will continue the "existing" panic without additional logic.

For example, the following program:

```rust
fn main() {
    println!("Before panic");
    let result = panic::catch_unwind(|| { panic!("uh oh") });
    println!("Before panic handling");
    if let Err(ongoing_panic) = result {
        panic::panic_any(ongoing_panic);
    }
    println!("After panic handling");
}
```

will print:

```
Before panic
thread 'main' panicked at src/main.rs:5:43:
uh oh
Before panic handling
thread 'main' panicked at src/main.rs:8:9:
Box<dyn Any>
```

... while the following program:

```rust
fn main() {
    println!("Before panic");
    let result = panic::catch_unwind(|| { panic!("uh oh") });
    println!("Before panic handling");
    if let Err(ongoing_panic) = result {
        panic::resume_unwind(ongoing_panic);
    }
    println!("After panic handling");
}
```

will print:

```
Before panic
thread 'main' panicked at src/main.rs:5:43:
uh oh
Before panic handling
```

divergentdave: Thanks for figuring this out! The duplicate messages had bothered me previously, but I didn't get to the bottom of it.

aggregator/src/aggregator/error.rs
```diff
@@ -25,12 +25,12 @@ pub enum Error {
     /// An invalid configuration was passed.
     #[error("invalid configuration: {0}")]
     InvalidConfiguration(&'static str),
-    /// Error decoding an incoming message.
+    /// Error decoding a message.
     #[error("message decoding failed: {0}")]
     MessageDecode(#[from] prio::codec::CodecError),
-    /// Error encoding a response to a request.
+    /// Error encoding a message.
     #[error("response encoding failed: {0}")]
```
tgeoghegan, line 32:
```suggestion
    #[error("message encoding failed: {0}")]
```

aggregator/src/aggregator/error.rs
```diff
@@ -273,7 +273,7 @@ impl Error {
         match self {
             Error::InvalidConfiguration(_) => "invalid_configuration",
             Error::MessageDecode(_) => "message_decode",
-            Error::ResponseEncode(_) => "response_encode",
+            Error::MessageEncode(_) => "response_encode",
```
tgeoghegan, line 276:
```suggestion
            Error::MessageEncode(_) => "message_encode",
```

tgeoghegan (approved):

divergentdave (approved):

divergentdave (approved):

PR #3004, opened 2024-04-12
Remove unused parameter from `Transaction` method

tgeoghegan:
`Transaction::get_unaggregated_client_reports_for_task` took a `vdaf: &A` parameter that isn't actually used. Deleting that lets us remove some generics and trait bounds from the method, which probably emits less code. Curiously, this didn't trip any unused parameter warnings. I think this is because we put a `tracing::instrument` annotation on the method, which must generate code that uses that parameter.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3811,7 +3811,7 @@ dependencies = [
  "percent-encoding",
  "pin-project-lite",
  "rustls 0.22.3",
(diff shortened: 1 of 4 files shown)


tgeoghegan:
Known issue in `tracing`: https://github.com/tokio-rs/tracing/issues/2209

PR #3005, opened 2024-04-12
Remove unused VDAF method parameters

divergentdave:
This is a follow-up to #3004 that removes a couple more unused VDAF parameters. I found these by patching in a local modified copy of `tracing`, with an `instrument` attribute macro that just returned the provided item as-is.


inahga (approved):
> I found these by patching in a local modified copy of tracing, with an instrument attribute macro that just returned the provided item as-is.

Nice trick!

tgeoghegan (approved):

PR #3006, opened 2024-04-12
daphne interop: Set log source and correct port

cjpatton:


integration_tests/src/daphne.rs
```diff
@@ -50,7 +50,7 @@ impl<'a> Daphne<'a> {
             let daphne_container = ContainerLogsDropGuard::new(
                 test_name,
                 container_client.run(runnable_image),
-                ContainerLogsSource::Docker,
+                ContainerLogsSource::Path("/logs".into()),
```
inahga, line 53: Nice, we can probably eliminate this `ContainerLogsSource` abstraction entirely, since Daphne was the only use case for the `Docker` variant. (You don't have to do this, I'll follow up a PR :grin:).

integration_tests/src/daphne.rs
```diff
@@ -108,6 +108,53 @@ impl<'a> Daphne<'a> {
             );
         }
 
+        // Reset Daphne's state.
```
inahga, line 111: Unless I'm reading the diff wrong, it looks like the `// Reset Daphne's state` and `// Add an HPKE receiver config to Daphne` blocks are duplicated above this line. i.e. it looks like we're already initializing, or at least attempting to.

cjpatton: 😮‍💨 oh gosh, how dumb. Fixed.

PR #3026, opened 2024-04-16
Time-bucketed fixed size support (release/0.6 backport)

inahga:
Backports these changes:
- https://github.com/divviup/janus/pull/3025
- https://github.com/divviup/janus/pull/3001

I don't really think we'll end up using this on 0.6, but it's low effort to backport these changes anyway. Supports https://github.com/divviup/divviup-api/issues/291

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1221,13 +1221,13 @@ dependencies = [
 
 [[package]]
 name = "divviup-client"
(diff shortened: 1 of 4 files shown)


inahga:
> I don't really think we'll end up using this on 0.6.

Asked around, we definitely won't be using this on 0.6.

PR #3027, opened 2024-04-17
Add GC subcommand

divergentdave:
This adds a new subcommand that just does garbage collection, as an alternative to enabling garbage collection in the aggregator subcommand. Since we are now building one multicall-enabled binary, this is effectively free. This new subcommand and Docker image is intended to make our deployment simpler by removing the need for providing aggregator configuration to a deployment that is only responsible for garbage collection. However, I figure we'll likely wait on making that change until we decommission all older versions of Janus. I updated the deployment documentation to mention that both the aggregator and garbage collector components can do garbage collection.

I also noticed that our garbage collector task does not shut down upon SIGTERM, so I updated the aggregator graceful termination test to cover this, and added the `Stopper` into the loop.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -158,6 +158,7 @@ jobs:
         targets: janus
         load: true
     - run: docker run --rm janus_aggregator --help
+    - run: docker run --rm janus_garbage_collector --help
     - run: docker run --rm janus_aggregation_job_creator --help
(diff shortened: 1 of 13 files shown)


tgeoghegan (approved):
LGTM. We'll need to coordinate this with a janus-ops change to modify the GC deployment and add something to deploy notes for that week.

branlwyd (approved):

Issue #238, opened 2022-06-13
Improve task caching strategy


branlwyd:
Currently, the Janus aggregator binary will read tasks the first time they are referenced by a request, and then store the read task in memory forever. This is not ideal for a few reasons: first, if there are enough tasks this would constitute an arbitrary growth of memory usage; second, we will eventually want to change tasks in some way (e.g. verify parameter rotation) and it would be very nice not to have to restart all aggregator tasks to do so.

We should think through a caching strategy that does not require storing all tasks forever in memory, and allows rereading tasks to discover changes.

It would be very nice to also cache the _lack_ of a task at a given task ID, too, to avoid the case of a misconfigured client causing an undue number of DB reads.

inahga:
I think I'm going to pick this up in support of https://github.com/divviup/divviup-api/pull/981, hopefully it's a quick change. I'm thinking a cache invalidation timeout is a simple enough approach.

PR #3028, opened 2024-04-17
Task aggregator cache: remove redundant reads on concurrent requests.

branlwyd:
Previously, if multiple requestors concurrently retrieved the same task from the task aggregator cache, we would query the DB for the task for each requestor.

Now, we use an additional layer of per-task locking so that, in the concurrent-requestor scenario, one requestor will read the task from the DB, and the remaining requestors will wait on that requestor's value.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -140,7 +140,7 @@ pub struct Aggregator<C: Clock> {
     /// Report writer, with support for batching.
     report_writer: Arc<ReportWriteBatcher<C>>,
     /// Cache of task aggregators.
-    task_aggregators: Mutex<HashMap<TaskId, Arc<TaskAggregator<C>>>>,
+    task_aggregators: TaskAggregatorCache<C>,
     /// Metrics.
     metrics: AggregatorMetrics,
 
@@ -150,6 +150,9 @@ pub struct Aggregator<C: Clock> {
     /// Cache of taskprov peer aggregators.
     peer_aggregators: PeerAggregatorCache,
 }
+
+type TaskAggregatorCache<C> =
+    SyncMutex<HashMap<TaskId, Arc<Mutex<Option<Arc<TaskAggregator<C>>>>>>>;
 
 #[derive(Clone)]
 struct AggregatorMetrics {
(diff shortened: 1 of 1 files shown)


aggregator/src/aggregator.rs
```diff
@@ -271,7 +274,7 @@ impl<C: Clock> Aggregator<C> {
             clock,
             cfg,
             report_writer,
-            task_aggregators: Mutex::new(HashMap::new()),
+            task_aggregators: SyncMutex::new(HashMap::new()),
```
inahga, line 277: nit: we only use the std::sync::Mutex so we don't need to alias this type to SyncMutex.

aggregator/src/aggregator.rs
```diff
@@ -676,45 +679,53 @@ impl<C: Clock> Aggregator<C> {
...
-            // None, we must provide some mechanism for taskprov tasks to force a cache refresh.
-            None => Ok(None),
+            cache_entry.as_ref().map(Arc::clone)
+        };
+
+        // If the task doesn't exist, remove the task entry from the cache to avoid caching a
+        // negative result. Then return the result.
+        //
+        // TODO(#238): once cache eviction is implemented, we can likely remove this step. We only
+        // need to do this to avoid trivial DoS via a requestor spraying many nonexistent task IDs.
```
inahga, line 722: I believe we can never drop this step, because of the taskprov case--any aggregator at any time can insert a task _and_ expect it to be immediately serviceable, so we can never cache None. I think it'll ultimately be up to rate limiting to mitigate this.

inahga (approved):
Nice use of data structures, I thought I had exhausted all possibilities.

aggregator/src/aggregator.rs
branlwyd: Not so -- the inner mutexes are Tokio's `Mutex` implementation, and they need to be since they are held across await points.

inahga: Ah, oops!

aggregator/src/aggregator.rs
branlwyd: Good point, though we might decide to e.g. cache `None` in the common case & include some configuration knob to allow us to configure taskprov environments to _not_ cache `None` (`nonexistent_task_cache_duration` or something?). I updated the comment to account for taskprov's needs.

inahga: I'll consider it for #238. I imagine it's a simple enough tweak, since we already have a flag that signals when taskprov is enabled.

PR #3041, opened 2024-04-22
Split unit tests from aggregation job driver file

divergentdave:
This splits another module worth of unit tests out into a separate file.


inahga (approved):

branlwyd (commented):
I don't object to breaking things out if individual files are getting too large to be manageable, but with an IDE (which is basically necessary for Rust) it's very easy to navigate the items in a file. Also, this loses `git blame` since we're not using [a trick to preserve history](https://stackoverflow.com/questions/16937359/git-copy-file-preserving-history), which I often use when tracking down the reasoning behind a change.

Is file-size/manageability the only reason to split these out? Either way, if we do split, can we make effort to preserve `blame`?

inahga:
Yes, last time I did this I turned off squashed commits temporarily to merge. https://github.com/divviup/janus/pull/1656

PR #3058, opened 2024-04-24
client: add client example in the docs

philips:
add a code example similar to the collector package

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -1,4 +1,41 @@
-//! DAP protocol client
+//! A [DAP](https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/) client
+//!
+//! This library implements the client role of the DAP-PPM protocol. It uploads measurements to two
+//! DAP aggregator servers which in turn compute a statistical aggregate over data from many
+//! clients, while preserving the privacy of each client's data.
+//!
+//! # Examples
+//!
+//! ```no_run
+//! use url::Url;
+//! use prio::vdaf::prio3::Prio3Histogram;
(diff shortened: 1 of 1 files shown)


client/src/lib.rs
```diff
@@ -1,4 +1,41 @@
-//! DAP protocol client
+//! A [DAP-PPM](https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/) client
```
tgeoghegan, line 1: I would just say "DAP". PPM is the working group that is writing DAP, not the protocol itself.
```suggestion
//! A [DAP](https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/) client
```

client/src/lib.rs
```diff
@@ -1,4 +1,41 @@
...
+//! ```no_run
+//! use url::Url;
+//! use prio::vdaf::prio3::Prio3Histogram;
+//! use janus_messages::{Duration, TaskId};
+//! use std::str::FromStr;
+//!
+//! #[tokio::main]
+//! async fn main() {
+//!     let leader_url = Url::parse("https://staging-dap-09-2.api.divviup.org/").unwrap();
+//!     let helper_url = Url::parse("https://staging-dap-09-1.api.divviup.org/").unwrap();
```
tgeoghegan, line 18: Let's steer docs.rs readers away from our staging environment:
```suggestion
//!     let leader_url = Url::parse("https://leader.example.org").unwrap();
//!     let helper_url = Url::parse("https://helper.example.org").unwrap();
```

branlwyd (approved):

Issue #2919, opened 2024-03-29
Migrate off serde_yaml


divergentdave:
The `serde_yaml` crate is unmaintained for now. We should keep an eye on the ecosystem over the next few weeks, and switch to whatever alternative makes the most sense. Meanwhile, I'm going to check the test coverage of our related serialization and deserialization routines.

PR #3062, opened 2024-04-25
Additional YAML serialization format tests

divergentdave:
This adds a few test cases to enforce backwards compatibility of YAML serialization of a few value-carrying enums. If we end up changing YAML libraries (see #2919), I figure these are the most likely sources of compatibility issues, both because use of tags is uncommon, and because we opt out of YAML tags in certain places with a "singleton map" adapter.

I also fixed a dev-dependency so that `cargo test -p janus_aggregator` works again.


aggregator/Cargo.toml
```diff
@@ -115,6 +115,7 @@ janus_aggregator = { path = ".", features = ["fpvec_bounded_l2", "test-util"] }
 janus_aggregator_core = { workspace = true, features = ["test-util"] }
 mockito = { workspace = true }
 opentelemetry_sdk = { workspace = true, features = ["testing"] }
+prio = { workspace = true, features = ["multithreaded"] }
```
inahga, line 118: This change seems unrelated?

inahga (approved):

aggregator/Cargo.toml
inahga: Oh oops I should read the PR description.

branlwyd (approved):

Issue #999, opened 2023-02-10
Reuse database server for ephemeral datastores


divergentdave:
We could probably save a lot of per-test setup and teardown time if we reused the same Docker container for multiple parallel tests, and instead isolated the tests from each other with different databases. Extensions and enum types both get created per-database, so our whole schema file should go through fine. We don't read from the system catalog or qualify any relation names with databases or schemas, so everything else should work transparently.

The container needs to be torn down explicitly by `testcontainers` on drop, so we can't just leak the container. Holding a static `RwLock<Weak<Container>>` and sticking corresponding `Arc<Container>`s inside the ephemeral datastore handle would resolve this.

branlwyd:
Sharing the DB containers is a good idea.

Unfortunately, `Container` isn't `Send` or `Sync`, so the `RwLock` idea won't work directly (since static variables must be `Send` & `Sync`). However, that doesn't sink the idea: users of the ephemeral database don't actually need a reference to the `Container`, only (a) the connection information and (b) to ensure that the `Container` isn't dropped while it is in use.

So an approach that would work would be:
* When we need to start a container, start a thread which starts the container then waits to be signaled for shutdown.
* Create a type called something like `EphemeralDatabase`. On creation, this type starts a container-management thread. It keeps the required information to generate connection strings, as well as enough to signal the container-management thread for shutdown. (this will be slightly tricky: we'll need to do the appropriate synchronization to make sure the container actually starts before creation of `EphemeralDatabase` returns) Arrange that dropping an `EphemeralDatabase` signals the container-management thread to halt (then, optionally, waits for container shutdown to be complete via additional synchronization).
* Keep a `Mutex<Weak<EphemeralDatabase>>` (or `RwLock`, but IMO test code doesn't need this level of optimization) around as per your original suggestion.

FWIW, completely unscientifically, the pre-change stats on my machine from one (1) run:
```
$ cargo test -p janus_aggregator --lib -- 'datastore::'
test result: ok. 37 passed; 0 failed; 0 ignored; 0 measured; 71 filtered out; finished in 26.75s
```

branlwyd:
Nerd-sniped by a concurrency/synchronization problem.

branlwyd:
Initial results look promising:
```
$ cargo test -p janus_aggregator --lib -- 'datastore::'
test result: ok. 37 passed; 0 failed; 0 ignored; 0 measured; 71 filtered out; finished in 12.57s
```

(PR still in-progress, resource cleanups/documentation needed)

PR #3063, opened 2024-04-29
Upgrade testcontainers to 0.16

divergentdave:
Testcontainers-rs published a big refactor over the weekend, this PR updates to adapt to the API changes. The biggest changes are that testcontainer clients are no longer part of the API surface, and the library now takes care of lazily instantiating them, plus the non-async client now wraps the HTTP-based client in a new Tokio runtime, rather than shelling out to Docker's CLI. We have to switch to the async API in several places where we are already running inside a Tokio runtime.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,7 +98,7 @@ signal-hook-tokio = "0.3.1"
 sqlx = "0.7.4"
 stopper = "0.2.7"
 tempfile = "3.10.1"
-testcontainers = "0.15.0"
+testcontainers = { version = "0.16.3", features = ["blocking"] }
 thiserror = "1.0"
 tracing = "0.1.40"
 tracing-chrome = "0.7.2"
--- a/aggregator_core/src/datastore/test_util.rs
+++ b/aggregator_core/src/datastore/test_util.rs
@@ -23,7 +23,7 @@ use std::{
     thread::{self, JoinHandle},
     time::Duration,
 };
-use testcontainers::RunnableImage;
+use testcontainers::{runners::SyncRunner, RunnableImage};
 use tokio::sync::{oneshot, Mutex};
 use tokio_postgres::{connect, Config, NoTls};
 use tracing::trace;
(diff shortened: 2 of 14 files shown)


Cargo.toml
```diff
@@ -98,7 +98,7 @@ signal-hook-tokio = "0.3.1"
 sqlx = "0.7.4"
 stopper = "0.2.7"
 tempfile = "3.10.1"
-testcontainers = "0.15.0"
+testcontainers = { version = "0.16.3", features = ["blocking"] }
```
branlwyd, line 101: Is there a good reason not to use the `AsyncRunner` everywhere -- it looks like you already used it everywhere except for `EphemeralDatabase`? Dropping the use of `SyncRunner` would let us drop the `blocking` feature.

divergentdave:
The next release of testcontainers should fix the cargo-deny issues by pushing the serde-java-properties behind a Cargo feature.

aggregator_core/src/datastore/test_util.rs
```diff
@@ -57,8 +57,7 @@ impl EphemeralDatabase {
             let shutdown_barrier = Arc::clone(&shutdown_barrier);
             move || {
                 // Start an instance of Postgres running in a container.
-                let container_client = testcontainers::clients::Cli::default();
-                let db_container = container_client.run(RunnableImage::from(Postgres::default()));
+                let db_container = RunnableImage::from(Postgres::default()).start();
```
branlwyd, line 60: If the reason we're still using the sync interface in this one place is because of the use of a (synchronous) thread, I think this would be pretty easy to replace with an async task -- other than the testcontainers interface & the async channel (which are already async), we use a `Barrier` to communicate shutdown. This Barrier could pretty easily be replaced by another `oneshot` channel, for example, to allow us to async-ify all of the blocking operations.

Cargo.toml
divergentdave: No, I was thinking of doing that as a follow-up once everything was working

branlwyd (approved):
Very nice, I like dropping the thread/task entirely. (I don't remember why we used a separate thread in the first place; hopefully that functionality wasn't unnecessary from the start.)

divergentdave:
Looks like `Container` wasn't `Send + Sync` at the time, so that's why it had to be owned by its own thread. This changed in version 0.15.0. See https://github.com/divviup/janus/issues/999#issuecomment-1426613988.

PR #3075, opened 2024-05-01
Apply SQL formatting w/ SQLTools VS Code extension

divergentdave:
For the sake of argument, here's what I get when I apply "format selected query" command from the [SQLTools extension for VS Code](https://marketplace.visualstudio.com/items?itemName=mtxr.sqltools) to each inline query. Downsides: this is still not fully automated, it requires VS Code, and always putting columns/values on their own line uses up a fair amount of vertical space.


divergentdave:
I'm going to close this with no action. I don't think it's worth chasing this sort of formatting until extensibility of Rust static analysis tooling is better, so that this can be done in an automated way easily.

PR #3076, opened 2024-05-01
Use nightly rustfmt features

### LINKED ISSUE #3056: Use nightly rustfmt
We could automate and standardize a few more style choices if we used a nightly build of rustfmt, and set up a configuration file with the following options.

- `format_strings = true` (unstable): I made some one-off changes using this previously, to wrap very long string literals. At the time this un-stuck about half of the chains that couldn't be formatted. (the rest were due to long turbofishes)
- `group_imports` (unstable): Our initial convention in this repository was to group all imports together, which maps to "One". Personally I would prefer "StdExternalCrate", which creates three groups of imports separated by blank lines in that order. This is closer to what rust-analyzer's [`auto-import`](https://rust-analyzer.github.io/manual.html#auto-import) assist does, though that creates up to five groups, splitting out `self` and `super`.
- `imports_granularity = "Crate"` (unstable): Combines multiple import statements from the same crate into one.

divergentdave:
This resolves #3056.


divergentdave:
I was surprised to see that [`comment_width`](https://rust-lang.github.io/rustfmt/?version=master&search=#comment_width) defaults to 80, not 100, so I overrode that when I enabled `wrap_comments`.

PR #3078, opened 2024-05-02
Fix Clippy lints

divergentdave:
This fixes some new warnings with a 1.78.0 toolchain.

### DIFF
--- a/aggregator_core/src/datastore/tests.rs
+++ b/aggregator_core/src/datastore/tests.rs
@@ -3433,10 +3433,6 @@ trait TestQueryTypeExt: CollectableQueryType {
 
     fn batch_identifier_for_client_timestamps(client_timestamps: &[Time]) -> Self::BatchIdentifier;
 
-    fn shortened_batch_identifier(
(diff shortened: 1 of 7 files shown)


aggregator_core/src/datastore/tests.rs
```diff
@@ -3433,10 +3433,6 @@ trait TestQueryTypeExt: CollectableQueryType {
 
     fn batch_identifier_for_client_timestamps(client_timestamps: &[Time]) -> Self::BatchIdentifier;
 
-    fn shortened_batch_identifier(
-        batch_identifier: &Self::BatchIdentifier,
-    ) -> Self::BatchIdentifier;
```
tgeoghegan, line 3438: I'm assuming this was never used?

tgeoghegan (approved):

aggregator_core/src/datastore/tests.rs
divergentdave: Correct, and it appears catching unused associated trait methods is a new feature

PR #3081, opened 2024-05-02
Re-enable divviup-ts integration tests

divergentdave:
The integration tests need a small fix in how they handle Prio3Count. Due to how the generics work out, we are currently relying on `<bool As Display>`, when we should be sending `"0"` or `"1"`. See https://divergentdave.github.io/draft-dcook-ppm-dap-interop-test-design/draft-dcook-ppm-dap-interop-test-design.html#table-3.

> Error: measurement is not a valid base 10 integer (was 'true')

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -34,7 +34,7 @@ jobs:
       id: default-input-values
       run: |
         DIVVIUP_TS_INTEROP_CONTAINER= ${{ inputs.divviup_ts_interop_container }}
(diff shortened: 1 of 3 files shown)


divergentdave:
This bug in the `InteropClientEncoding` implementation for `Prio3Count` had gone unnoticed since we upgraded the `prio` crate because we only ever use it during the divviup-ts integration tests. All other integration tests just use `janus_client` in-process instead of using a client in a container. Note that the `e2e_prio3_count` test under the `janus_interop_binaries` crate checks that the `janus_interop_client` container can handle correctly-encoded requests.

tgeoghegan (approved):

PR #3083, opened 2024-05-02
Disable incremental compilation in small profile

divergentdave:
I noticed that the "ci" profile disables incremental compilation, while the "small" profile does not. Since we only use "small" from inside `docker build`, disabling incremental compilation should be an easy win in terms of layer cache usage.

After this change, the two profiles only differ in their `opt-level` and `strip` settings. It may make sense to unify these profiles, but I'd want to measure build times and layer sizes first.


divergentdave:
Actually this doesn't make a difference, as we already disable incremental compilation in the Dockerfile here https://github.com/divviup/janus/blob/321dbeea3d56f6a3dea79e2332d869d1cb4a1056/Dockerfile.interop#L5

PR #3084, opened 2024-05-03
Use ci profile in interop containers

divergentdave:
This changes the profile we use to compile interop containers in tests, from `small` to `ci`. This results in an increase in layer sizes, and thus cache usage, and a huge speedup in compile time. I think this effect is primarily attributable to changing `opt-level` from `"z"` to `0`.

The total size of all Docker stages increased from 5397MiB to 5724MiB. The two chief contributors are the `builder-aggregator` and `final` stages from `Dockerfile.interop_aggregator`, which grew by 158MiB and 136MiB. The relative increase in the `final` stages of both Dockerfiles is large, at 41% and 141%, but note that the size of the final stage doesn't have a big impact anymore, since we are no longer embedding it in later compilations.

Locally, compilation time for the various layers improved by 2x to 6x, depending. The time to compile the aggregator, including both `cargo chef cook` and `cargo build`, went from 18 minutes to 3 minutes. I'm thus expecting a big improvement in CI times.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -181,7 +181,7 @@ jobs:
       with:
         files: docker-bake.hcl
         workdir: .
-        targets: interop_binaries_small
+        targets: interop_binaries_ci
 
   rustsec_advisories:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


divergentdave:
This saved nine minutes on a fresh build of the `janus_interop_docker` job, from 22 minutes to 13 minutes. I tried to time a rebuild, but it didn't work out, maybe we would have to merge to main first.

tgeoghegan (approved):

Issue #3107, opened 2024-05-07
Authentication scheme in `Authorization` header should not be case sensitive


tgeoghegan:
When authenticating leader -> helper requests, Janus requires a bearer token presented in an [`Authorization` header](https://datatracker.ietf.org/doc/html/rfc9110#name-authorization). The header name is not case sensitive (per HTTP), but we require [exactly the string prefix "Bearer " on the value](https://github.com/divviup/janus/blob/main/core/src/http.rs#L110).

Per RFC 9110, the `Authorization` header has the form `Authorization = credentials`. [`credentials`](https://datatracker.ietf.org/doc/html/rfc9110#name-credentials) in turn is `credentials = auth-scheme [ 1*SP ( token68 / #auth-param ) ]` and finally [`auth-scheme`](https://datatracker.ietf.org/doc/html/rfc9110#name-authentication-scheme) is: "a case-insensitive token to identify the authentication scheme". Janus should normalize the case of the auth scheme before checking it.

tgeoghegan:
Just for thoroughness, an authorization header will look like:

```
Authorization: Bearer some-arbitrary-string
^^^^^^^^^^^^^  ^^^^^^ ^^^^^^^^^^^^^^^^^^^^^
     |           |              |
HTTP header     auth          token
name           scheme
```

As discussed above, the HTTP header name and auth scheme are case *insensitive*. However, [RFC 6750](https://datatracker.ietf.org/doc/html/rfc6750), which we lean on for the `Bearer` auth scheme, says in section 1.1:

> Unless otherwise noted, all the protocol parameter names and values are *case sensitive*.

(emphasis mine) 

Section 2.1, describing the "Authorization Request Header Field" does not otherwise note. So we should continue treating the actual token as case sensitive.

PR #3111, opened 2024-05-08
Relax constraints on bearer token format

inahga:
Closes https://github.com/divviup/janus/issues/3107. The `bearer` part is now case insensitive. Add some basic tests.

I also allow whitespace between `bearer` and the token. The `1*SP` part of the grammar, IIUC, indices one or more whitespace:
```
Authorization = credentials
credentials = auth-scheme [ 1*SP ( token68 / #auth-param ) ]
```

### DIFF
--- a/core/src/http.rs
+++ b/core/src/http.rs
@@ -111,12 +111,12 @@ pub fn extract_bearer_token(conn: &Conn) -> Result<Option<AuthenticationToken>, 
         .get("authorization")
         .map(HeaderValue::to_string)
     {
-        let (bearer, token) = authorization
+        let (auth_scheme, token) = authorization
             .split_once(char::is_whitespace)
-            .ok_or_else(|| anyhow!("invalid bearer token"))?;
+            .ok_or_else(|| anyhow!("invalid authorization header"))?;
 
-        if bearer.to_lowercase() != "bearer" {
-            return Err(anyhow!("authorization header value is not a bearer token"));
+        if auth_scheme.to_lowercase() != "bearer" {
+            return Err(anyhow!("authorization scheme is not bearer"));
         }
 
         return Ok(Some(AuthenticationToken::new_bearer_token_from_string(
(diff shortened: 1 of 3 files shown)


divergentdave (approved):

branlwyd (approved):

core/src/http.rs
```diff
@@ -106,16 +106,78 @@ impl Display for HttpErrorResponse {
...
+    if let Some(authorization) = conn
+        .request_headers()
+        .get("authorization")
+        .map(HeaderValue::to_string)
+    {
+        let (bearer, token) = authorization
+            .split_once(char::is_whitespace)
+            .ok_or_else(|| anyhow!("invalid bearer token"))?;
+
+        if bearer.to_lowercase() != "bearer" {
```
tgeoghegan, line 118: nit:
```suggestion
        let (auth_scheme, token) = authorization
            .split_once(char::is_whitespace)
            .ok_or_else(|| anyhow!("invalid scheme in authorization header"))?;

        if auth_scheme.to_lowercase() != "bearer" {
```


To line up with the `credentials = auth-scheme [ 1*SP ( token68 / #auth-param ) ]` in the Backaus-whatever notation.

core/src/http.rs
```diff
@@ -106,16 +106,78 @@ impl Display for HttpErrorResponse {
...
+                    .with_request_header("authorization", "bearer gVfRUu9krhxrUgFsEo-P5w"),
+            ),
+            Ok(Some(_))
+        );
+        assert_matches!(
+            extract_bearer_token(
+                &TestConn::build("get", "/", "body")
+                    .with_request_header("authorization", "BeArEr     gVfRUu9krhxrUgFsEo-P5w"),
+            ),
+            Ok(Some(_))
```
tgeoghegan, line 151: I think it'd be nice to check the contents of the `AuthenticationToken` to make sure nothing bad happens when we call `String::trim_start`.

tgeoghegan (changes requested):

tgeoghegan (approved):

Issue #3117, opened 2024-05-10
Parallelize report processing


divergentdave:
We could improve latency of aggregation job requests if we parallelized report share preparation using Rayon. We'd have to do some refactoring first, to do duplicate report detection in a separate first pass, and to return results from an iterator rather than adding them to different local variables.

Issue #3035, opened 2024-04-19
Abort work in threadpool if initiating future is cancelled


divergentdave:
Once we start cancelling futures based on disconnections and/or timeouts, (see #3033 and #3034) we may want to plumb through a signal to either skip running a task on a threadpool if the initiating future has already been cancelled, or check inside the loop over report shares, and short-circuit in such cases. If the future has been cancelled, then the oneshot channel's sender has been dropped, so any further work on that threadpool task will ultimately be wasted.

Issue #3033, opened 2024-04-19
Cancel on disconnect


divergentdave:
This is a tracking issue for cancelling request handling upon disconnection. See #2644.

branlwyd:
Implemented by PR #3119 above.

PR #3119, opened 2024-05-10
Implement cancellation on disconnect, & parallelize Helper aggregation computations.

branlwyd:
This will be very helpful when we have requests which timeout, which is presented to Janus as a client disconnect. Otherwise, we would continue processing the request until it is complete.

Previously, we used a rayon threadpool for aggregation computations, but the computations within the handling of a single aggregation job were still serialized. Now, we use a parallel iterator for handling each report share in turn, allowing the VDAF evaluations to be performed in parallel. The Helper methods are also updated such that the parallel computations respect cancellation. Specifically, the `receiver` will be dropped on cancellation, causing the producer threads' `sender` to return a `SendError` when they attempt to send, which will cause `try_for_each_with` to stop processing, which will cause the producer thread to terminate.

Also, remove the (unstated!) requirement in aggregation_job_writer than report aggregations be provided in order. This eases parallelization of the aggregation job continuation logic.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -90,6 +90,7 @@ use prio::{
         xof::XofTurboShake128,
     },
 };
+use rayon::iter::{IndexedParallelIterator as _, IntoParallelRefIterator as _, ParallelIterator};
 use reqwest::Client;
 use ring::{
     digest::{digest, SHA256},
@@ -97,20 +98,15 @@ use ring::{
     signature::{EcdsaKeyPair, Signature},
 };
 use std::{
-    any::Any,
     collections::{HashMap, HashSet},
     fmt::Debug,
     hash::Hash,
-    panic::{catch_unwind, resume_unwind, AssertUnwindSafe},
     path::PathBuf,
     sync::{Arc, Mutex as SyncMutex},
     time::{Duration as StdDuration, Instant},
 };
-use tokio::{
-    sync::oneshot::{self, error::RecvError},
-    try_join,
-};
-use tracing::{debug, info, info_span, trace_span, warn, Level, Span};
+use tokio::try_join;
+use tracing::{debug, info, trace_span, warn, Level};
 use url::Url;
 
 #[cfg(test)]
@@ -1900,10 +1896,8 @@ impl VdafOps {
             return Ok(response);
         }
 
-        let agg_param = Arc::new(
-            A::AggregationParam::get_decoded(req.aggregation_parameter())
-                .map_err(Error::MessageDecode)?,
-        );
+        let agg_param = A::AggregationParam::get_decoded(req.aggregation_parameter())
+            .map_err(Error::MessageDecode)?;
 
         let report_deadline = clock
             .now()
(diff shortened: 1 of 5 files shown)


branlwyd:
A couple of things to pay attention to in review:
* This is mostly "free" parallelism outside of whatever `rayon`'s inherent synchronization costs are, but there are a couple of places that aren't free. First, in aggregation job continuation we now serially build up a new vector associating prepare steps to report aggregations; I think this is required since otherwise we'd need to mutate the report aggregation iterator from multiple callbacks in parallel. Second, I added a sort over report aggregations to `aggregation_job_writer`, to remove the requirement that report aggregations be in order. If we can drop either of these, we'd see some efficiency gains.
* I'm not totally sure traces/spans are being performed properly, please pay close attention here.

branlwyd:
Closes #3117.

aggregator/src/aggregator.rs
```diff
@@ -1761,358 +1755,318 @@ impl VdafOps {
...
-                    "handle_aggregate_init_generic threadpool task"
+        // We validate that each prepare_init can be represented by a `u64` ord value here, so that
+        // inside the parallel iterator we can unwrap. A conversion failure here will fail the
+        // entire aggregation. However, this is desirable: this can only happen if we receive too
+        // many report shares in an aggregation job for us to store, which is a whole-aggregation
+        // problem rather than a per-report problem. (separately, this would require more than
+        // u64::MAX report shares in a single aggregation job, which is practically impossible.)
+        u64::try_from(req.prepare_inits().len())?;
+
+        let mut report_share_data = Vec::new();
```
divergentdave, line 1766: nit: We can pre-allocate this vector

aggregator/src/aggregator.rs
```diff
@@ -1761,358 +1755,318 @@ impl VdafOps {
...
-                Error::Internal("threadpool failed to send VDAF preparation result".into())
-            })?
-            .unwrap_or_else(|panic_cause: Box<dyn Any + Send>| {
-                resume_unwind(panic_cause);
-            })?;
-
-        // TODO: Use Arc::unwrap_or_clone() once the MSRV is at least 1.76.0.
-        let agg_param = Arc::try_unwrap(agg_param).unwrap_or_else(|arc| arc.as_ref().clone());
+            })
+            .collect_into_vec(&mut report_share_data);
```
divergentdave, line 2069: Dropping the `rayon::spawn()` and making this call from an async task results in regressing on poll times. I saw one poll between 8 and 17 seconds per aggregation job when testing with a large number of report shares.

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -120,10 +120,12 @@ where
     pub fn put(
         &mut self,
         aggregation_job: AggregationJob<SEED_SIZE, Q, A>,
-        report_aggregations: Vec<RA>,
+        mut report_aggregations: Vec<RA>,
     ) -> Result<(), Error> {
         self.update_aggregation_parameter(aggregation_job.aggregation_parameter());
 
+        report_aggregations.sort_unstable_by_key(RA::ord);
```
divergentdave, line 127: I think this sort may not be necessary since we're just using order-preserving rayon operations, but I haven't checked all the docs closely yet. `par_bridge()`/`ParallelBridge` does not preserve the order, but others can, since they start with full knowledge of the items to be processed, and subdivide both the input iterator and output destination.

divergentdave (commented):
I haven't read the changes closely yet, but I have some early feedback based on testing.

As for tracing spans, we will want to create a span, make it a child of the current span, and pass it to the closure currently used in `ParallelIterator::map()`. (similar to before) There, we should enter the span at the top of the closure. This will ensure our events logged inside the closure have an enclosing span. Note that the span documentation says "it is entirely valid for multiple threads to enter the same span concurrently", so it's okay to share one child span. We can pass it as the initialization argument of `map_with()`, so it gets cloned once per worker thread.

branlwyd:
I merged #3131 into this, as it addresses all current comments. I will update Leader aggregation methods in a follow-on PR, to parallelize the computations & respect cancellation.

Stacked on #3119.
Part of #3033.
Part of #3035.

aggregator/src/aggregator/aggregation_job_writer.rs
branlwyd: This is needed (at least) because aggregation job continuation's preprocessing step to match prep steps to report aggs will now place any failed report aggs at the front of the list, so they will no longer be in `ord` order.

aggregator/src/aggregator.rs
branlwyd: Fixed by merging the follow-on PR (lol).

aggregator/src/aggregator.rs
branlwyd: Fixed by merging the follow-on PR (lol).

aggregator/src/aggregator.rs
divergentdave: Looking much better, the worst-case poll time is now between a quarter second and a half second.

inahga (approved):

aggregator/src/aggregator.rs
```diff
@@ -1771,345 +1779,329 @@ impl VdafOps {
...
 
-                    let task_hpke_keypair = task.hpke_keys().get(
-                        prepare_init
-                            .report_share()
-                            .encrypted_input_share()
-                            .config_id(),
-                    );
+            move || {
+                let span = info_span!(parent: parent_span, "handle_aggregate_init_generic threadpool task");
+                let _entered = span.enter();
```
divergentdave, line 1946: This is going to happen on the Tokio blocking thread, but we don't do much here, just kick off Rayon tasks and wait for them. The relevant spans and events will be happening on Rayon threads. Thus, we should pass the span in a tuple along with the `sender` to be cloned onto Rayon threads, and enter it at the top of the closure passed to `try_for_each_with()`. The impact of this issue is that the "VDAF preparation (helper initialization)" span below would be a root span, and various events would have no parent span.

branlwyd:
(Rebased on latest `main`.)

aggregator/src/aggregator.rs
```diff
@@ -1932,345 +1940,329 @@ impl VdafOps {
...
-                                // which is a whole-aggregation problem rather than a per-report
-                                // problem. (separately, this would require more than u64::MAX
-                                // report shares in a single aggregation job, which is practically
-                                // impossible.)
-                                ord.try_into()?,
-                                Some(PrepareResp::new(
+                                ),
+                            };
+
+                        sender.send(ReportShareData {
```
divergentdave, line 2229: I see that a SendError returned from here will cause `try_for_each_with()` to short-circuit, and stop other tasks as it can. This, combined with the fact that `send()` fails when the corresponding `UnboundedReceiver` is dropped, means that work being done in Rayon tasks will be stopped promptly when the initiating future is cancelled. We should describe this in a comment, maybe just above the Rayon parallel iterator calls.

aggregator/src/aggregator/aggregation_job_writer.rs
divergentdave: We definitely need this sort now that data is flowing back over a channel.

aggregator/src/aggregator.rs
```diff
@@ -1932,345 +1940,329 @@ impl VdafOps {
...
+        // Await the producer task to resume any panics that may have occurred, and to ensure we can
+        // unwrap the aggregation parameter's Arc in a few lines. The only other errors that can
+        // occur are: a `JoinError` indicating cancellation, which is impossible because we do not
+        // cancel the task; and a `SendError`, which can only happen if this future is cancelled (in
+        // which case we will not run this code at all).
+        let _ = producer_task.await.map_err(|join_error| {
+            if let Ok(reason) = join_error.try_into_panic() {
+                panic::resume_unwind(reason);
             }
-        };
```
divergentdave, line 2249: I wonder if we should add a defensive check that report_share_data is the right length, if the blocking task didn't panic.

aggregator/src/aggregator/aggregation_job_continue.rs
```diff
@@ -59,201 +56,216 @@ impl VdafOps {
...
+        let producer_task = tokio::task::spawn_blocking({
+            let parent_span = Span::current();
+            let metrics = metrics.clone();
+            let task = Arc::clone(&task);
+            let vdaf = Arc::clone(&vdaf);
+            let aggregation_job = Arc::clone(&aggregation_job);
+
+            move || {
+                let span = info_span!(parent: parent_span, "step_aggregation_job threadpool task");
+                let _entered = span.enter();
```
divergentdave, line 154: As with aggregation init, this span should be entered in the closure passed to `try_for_each_with()`.

aggregator/src/aggregator.rs
branlwyd: Sure. I used an `assert_eq!` as it would be programmer error if this condition did not hold.

divergentdave (approved):

PR #3129, opened 2024-05-13
Log agg job init forbidden mutations.

tgeoghegan:
When we detect that an aggregation job ID is being re-used, we check that the incoming request matches whatever request we previously saw, and reject it with `Error::ForbiddenMutation`, which is rendered as HTTP 409 Conflict.

We now record an event when this occurs for aggregation job initialization. The event contains the expected request hash, the incoming request's hash and the list of `ReportMetadata`s for both.

Because the resulting event object is quite large, it gets written to a file, created under the directory specified in the `log_forbidden_mutations` option.

~The event is logged in a new span `aggregation_job_init_forbidden_mutation_span` at the DEBUG level. This event will be logged if a suitable filter is set up. For example, running Janus with `RUST_LOG=debug` or a `RUST_LOG` environment variable containing this string will do the trick:~

~`janus_aggregator::aggregator[aggregation_job_init_forbidden_mutation_span]=debug`~

Logging the forbidden mutation is expensive because it requires fetching all the `report_aggregations` rows for the existing aggregation job, so this is gated on the `log_forbidden_mutations` option.

Note that there are also cases in aggregation job continuation and collection job creation where we check for resource mutation and return 409 Conflict. This commit only logs on the aggregation job init path because that's what we're trying to debug right now.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -18,6 +18,7 @@ use crate::{
         TASK_AGGREGATOR_CACHE_DEFAULT_CAPACITY, TASK_AGGREGATOR_CACHE_DEFAULT_TTL,
     },
     config::TaskprovConfig,
+    diagnostic::AggregationJobInitForbiddenMutationEvent,
     metrics::{aggregate_step_failure_counter, report_aggregation_success_counter},
 };
 use backoff::{backoff::Backoff, Notify};
@@ -101,6 +102,7 @@ use std::{
     fmt::Debug,
     hash::Hash,
     panic::{catch_unwind, resume_unwind, AssertUnwindSafe},
+    path::PathBuf,
     sync::{Arc, Mutex as SyncMutex},
     time::{Duration as StdDuration, Instant},
 };
@@ -108,7 +110,7 @@ use tokio::{
     sync::oneshot::{self, error::RecvError},
     try_join,
 };
-use tracing::{debug, debug_span, info, info_span, trace_span, warn, Level, Span};
+use tracing::{debug, info, info_span, trace_span, warn, Level, Span};
 use url::Url;
 
 #[cfg(test)]
@@ -215,9 +217,9 @@ pub struct Config {
     /// Configuration for the taskprov extension.
     pub taskprov_config: TaskprovConfig,
 
-    /// Whether forbidden mutations of resources (e.g., re-using the same aggregation job ID but
-    /// with different reports in it) should be logged when detected.
-    pub log_forbidden_mutations: bool,
+    /// If set, forbidden mutations of resources (e.g., re-using the same aggregation job ID but
+    /// with different reports in it) will be logged to files under this path when detected.
+    pub log_forbidden_mutations: Option<PathBuf>,
 }
 
 impl Default for Config {
@@ -232,7 +234,7 @@ impl Default for Config {
             taskprov_config: TaskprovConfig::default(),
             task_cache_ttl: TASK_AGGREGATOR_CACHE_DEFAULT_TTL,
             task_cache_capacity: TASK_AGGREGATOR_CACHE_DEFAULT_CAPACITY,
-            log_forbidden_mutations: false,
+            log_forbidden_mutations: None,
         }
     }
 }
@@ -467,7 +469,7 @@ impl<C: Clock> Aggregator<C> {
                 self.cfg.batch_aggregation_shard_count,
                 aggregation_job_id,
                 taskprov_task_config.is_some(),
-                self.cfg.log_forbidden_mutations,
+                self.cfg.log_forbidden_mutations.clone(),
                 req_bytes,
             )
             .await
@@ -1017,8 +1019,8 @@ impl<C: Clock> TaskAggregator<C> {
         metrics: &AggregatorMetrics,
         batch_aggregation_shard_count: u64,
         aggregation_job_id: &AggregationJobId,
-        log_forbidden_mutations: bool,
         require_taskprov_extension: bool,
+        log_forbidden_mutations: Option<PathBuf>,
         req_bytes: &[u8],
     ) -> Result<AggregationJobResp, Error> {
         self.vdaf_ops
@@ -1030,8 +1032,8 @@ impl<C: Clock> TaskAggregator<C> {
                 Arc::clone(&self.task),
                 batch_aggregation_shard_count,
                 aggregation_job_id,
-                log_forbidden_mutations,
                 require_taskprov_extension,
+                log_forbidden_mutations,
                 req_bytes,
             )
             .await
@@ -1396,8 +1398,8 @@ impl VdafOps {
         task: Arc<AggregatorTask>,
         batch_aggregation_shard_count: u64,
         aggregation_job_id: &AggregationJobId,
-        log_forbidden_mutations: bool,
         require_taskprov_extension: bool,
+        log_forbidden_mutations: Option<PathBuf>,
         req_bytes: &[u8],
     ) -> Result<AggregationJobResp, Error> {
         match task.query_type() {
@@ -1413,8 +1415,8 @@ impl VdafOps {
                         batch_aggregation_shard_count,
                         aggregation_job_id,
                         verify_key,
-                        log_forbidden_mutations,
                         require_taskprov_extension,
+                        log_forbidden_mutations,
                         req_bytes,
                     )
                     .await
@@ -1432,8 +1434,8 @@ impl VdafOps {
                         batch_aggregation_shard_count,
                         aggregation_job_id,
                         verify_key,
-                        log_forbidden_mutations,
                         require_taskprov_extension,
+                        log_forbidden_mutations,
                         req_bytes,
                     )
                     .await
@@ -1729,7 +1731,7 @@ impl VdafOps {
         aggregation_job_id: &AggregationJobId,
         verify_key: &VerifyKey<SEED_SIZE>,
         require_taskprov_extension: bool,
-        log_forbidden_mutations: bool,
+        log_forbidden_mutations: Option<PathBuf>,
         req_bytes: &[u8],
     ) -> Result<AggregationJobResp, Error>
     where
@@ -2168,6 +2170,7 @@ impl VdafOps {
                 let aggregation_job_writer_metrics = metrics.for_aggregation_job_writer();
                 let aggregation_job = Arc::clone(&aggregation_job);
                 let mut report_share_data = report_share_data.clone();
+                let log_forbidden_mutations = log_forbidden_mutations.clone();
 
                 Box::pin(async move {
                     // Check if this is a repeated request, and if it is the same as before, send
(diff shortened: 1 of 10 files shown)


aggregator/src/aggregator.rs
```diff
@@ -2173,6 +2186,39 @@ impl VdafOps {
...
+                                    "aggregation_job_init_forbidden_mutation_span",
+                                );
+                                let _span_entered = span.enter();
+                                let existing_report_metadatas: Vec<_> = tx
+                                    .get_report_aggregations_for_aggregation_job(
+                                        vdaf.as_ref(),
+                                        &Role::Helper,
+                                        task.id(),
+                                        aggregation_job.id(),
+                                    )
```
tgeoghegan, line 2201: I'm a little concerned that the order in which we'll get `report_aggregations` rows from this query won't be guaranteed to match the order in which they appeared in an `AggregationJobInitReq`, but I think they _should_.

branlwyd: `get_report_aggregations_for_aggregation_job` retrieves report aggregations ordered by `ord`, which is the correct (required) order for the `AggregationJobInitReq`'s `prepare_inits`.

aggregator/src/aggregator.rs
```diff
@@ -2173,6 +2186,39 @@ impl VdafOps {
...
+                                    .await?
+                                    .iter()
+                                    .map(|ra| ra.report_metadata())
+                                    .collect();
+                                let incoming_request_report_metadatas: Vec<_> = report_share_data
+                                    .iter()
+                                    .map(|rsd| rsd.report_share.metadata())
+                                    .collect();
+                                tracing::event!(
+                                    Level::DEBUG,
```
branlwyd, line 2211: If this is a config option, we might as well use an `INFO` log IMO.

branlwyd (dismissed):

aggregator/src/aggregator.rs
tgeoghegan: Yes, I suppose that's simpler.

branlwyd:
Maybe we should also log: the agg param (though it should always be `()` in this case), and the `BatchId` contained in the partial batch identifier?

tgeoghegan:
We discussed this offline and are concerned that in the case we are trying to learn more about, the log messages this generates will be too big for ingestion into Google Cloud Logging (https://cloud.google.com/logging/quotas). I'm going to rework this so that instead of logging these events to stdout, we'll log each such event to a file, which we can then copy off production nodes and analyze. Additionally, as Bran suggests, we'll try to log as much as we can about the requests to look for other mismatches that could cause hashes to not line up.

tgeoghegan:
I iterated on this so that we now have a config option which is a path under which to write event descriptions. This means we could do something like mount a volume into Janus containers rather than use the underlying node's tiny storage. The result is a file like `/tmp/events/agg-job-illegal-mutation-1715647260.935990888s-bdb4a376-26ce-4447-8b8c-022214855413-0.7.9` (the filename interpolates event type, UNIX time, a unique ID, and Janus version) whose contents are like:

```
AggregationJobInitForbiddenMutationEvent {
    aggregation_job_id: AggregationJobId(55pm_TTfjJqi773oLDOguQ),
    original_request_hash: Some("85968b369375c34bd0e8554433e693839a387adff4a8a43a8794bb6d2677ea56"),
    original_report_metadatas: [
        ReportMetadata {
            report_id: ReportId(JUNUjNs44OVIyvldUj171Q),
            time: Time(
                999993600,
            ),
        },
        ReportMetadata {
            report_id: ReportId(8oC9qWIGfW3v8h63lbVjZw),
            time: Time(
                999993600,
            ),
        },
    ],
    original_batch_id: "()",
    original_aggregation_parameter: [
        0,
    ],
    mutating_request_hash: Some("97e9c2181163cdc1923126eb1892afba77780e7dabb432d4802189d3516e7436"),
    mutating_request_report_metadatas: [
        ReportMetadata {
            report_id: ReportId(JUNUjNs44OVIyvldUj171Q),
            time: Time(
                999993600,
            ),
        },
        ReportMetadata {
            report_id: ReportId(8oC9qWIGfW3v8h63lbVjZw),
            time: Time(
                999993600,
            ),
        },
    ],
    mutating_request_batch_id: "()",
    mutating_request_aggregation_parameter: [
        1,
    ],
}
```

Though a real one of these will have many thousands of `original_report_metadatas` and `mutating_request_report_metadatas`. This needs more testing in Kind before we take it out to prod but we can review.

aggregator/src/aggregator.rs
```diff
@@ -212,7 +214,12 @@ pub struct Config {
     /// The key used to sign HPKE configurations.
     pub hpke_config_signing_key: Option<EcdsaKeyPair>,
 
+    /// Configuration for the taskprov extension.
     pub taskprov_config: TaskprovConfig,
+
+    /// If set, forbidden mutations of resources (e.g., re-using the same aggregation job ID but
+    /// with different reports in it) will be logged to files under this path when detected.
+    pub log_forbidden_mutations: Option<PathBuf>,
```
inahga, line 222: Do we want to keep this option around long-term, or are we planning to back this change out once diagnosis is done? If the former, we may want to note that this option is not covered by our stability guarantees.

aggregator/src/aggregator.rs
```diff
@@ -2173,6 +2189,62 @@ impl VdafOps {
...
+                                        .aggregation_parameter()
+                                        .get_encoded()
+                                        .map_err(|e| datastore::Error::User(e.into()))?,
+                                };
+                                let event_id = crate::diagnostic::write_event(
+                                    log_forbidden_mutations,
+                                    "agg-job-illegal-mutation",
+                                    Box::new(event),
+                                )
+                                .map_err(|e| datastore::Error::User(e.into()))?;
```
inahga, line 2237: I think I'm fine with this, but on the edge. If the target directory is unwritable, e.g. out of disk space, then we'll be trickling up 500 for what amounts to an error encountered while logging. But this flag and situation have a narrow enough use case that we'll be watching it closely anyway.

inahga: Also note that I don't think we log errors in transactions like this. So if something went wrong here the only indication will be at DEBUG level and a 500 error in the metrics. Can consider adding an `error!()` in the `.map_err()`.

inahga (approved):

aggregator/src/diagnostic.rs
```diff
@@ -0,0 +1,123 @@
...
+use janus_messages::{AggregationJobId, ReportMetadata};
+use std::{fmt::Debug, fs::File, io::Write, path::Path, time::SystemTime};
+use uuid::Uuid;
+
+/// Represents an illegal attempt to mutate an aggregation job.
+#[derive(Derivative, Clone)]
+#[derivative(Debug)]
+pub struct AggregationJobInitForbiddenMutationEvent {
+    /// The ID of the aggregation job.
+    pub aggregation_job_id: AggregationJobId,
```
divergentdave, line 14: It would be useful to stash the TaskId as well.

aggregator/src/diagnostic.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// The SHA-256 of the request that created the aggregation job.
+    #[derivative(Debug(format_with = "fmt_hash_option"))]
+    pub original_request_hash: Option<[u8; 32]>,
+
+    /// The ordered report metadatas from the request that created the aggregation job.
+    pub original_report_metadatas: Vec<ReportMetadata>,
+
+    /// The batch ID in the original request.
+    pub original_batch_id: String,
```
divergentdave, line 24: nit: We could skip some extra quote marks, here and on the other batch ID, since this is already Debug formatted before the struct is constructed.
```suggestion
    #[derivative(Debug(format_with = "std::fmt::Display::fmt"))]
    pub original_batch_id: String,
```

aggregator/src/diagnostic.rs
```diff
@@ -0,0 +1,123 @@
...
+    let version = env!("CARGO_PKG_VERSION");
+
+    let mut f = File::create(
+        event_storage
+            .as_ref()
+            .join(format!("{event_type}-{now:?}-{event_id}-{version}")),
+    )
+    .context("failed to open file to write event")?;
+
+    write!(&mut f, "{event:#?}").context("failed to write event")?;
```
divergentdave, line 74: It might be good to call `sync_all()` to catch possible errors when syncing and closing as well.

Separately, since we expect to be writing at least 10MiB of formatted text, this call could take a while. I don't know what sort of latency we would expect from file creation in the FUSE driver, for example. Thus, I think we should make `write_event()` an async method, and then use `spawn_blocking()` around creating the file, writing to it, and closing it. (The only good stories for heavy-duty filesystem work in Tokio are spawn_blocking with standard syscalls or an experimental io_uring crate)

aggregator/src/aggregator.rs
tgeoghegan: Yeah, I think we should document that this isn't stable. If we do decide to keep it, we can always promote it to stable.

aggregator/src/aggregator.rs
tgeoghegan: I think you're right. What I did is change the call to `write_event` so that if it fails, we log the error it returns but still move on in this function, so the client will still get HTTP 409. But I only did this for that call, as I think the other fallible calls in this block failing would be so unusual and rare that an HTTP 500 would be fine.

aggregator/src/diagnostic.rs
tgeoghegan: I think this is moot now because I switched the event serialization format to JSON. The `*_batch_id` fields end up looking like `"original_batch_id": "()",`

aggregator/src/diagnostic.rs
tgeoghegan: Done!

tgeoghegan (commented):
Here's another take on this where we serialize the events to file as JSON, to facilitate later analysis.

aggregator/src/diagnostic.rs
tgeoghegan: Done. I considered using `tokio::fs::File` for this, but it's not obvious to me how to use an async writer with `serde_json`, so I stuck with the synchronous `std::fs::File`.

docs/samples/advanced_config/aggregator.yaml
```diff
@@ -130,6 +130,22 @@ taskprov_config:
...
+# metadatas in both, into a file created under the path provided here.
+#
+# Janus does not unconditionally log this information because retrieving the report metadata for the
+# existing request is expensive and the resulting event is large.
+#
+# This option is not stable and not subject to Janus' typical API and configuration stability
+# promises.
+#
+# Optional; the default is not to log events.
+log_forbidden_mutations:
```
branlwyd, line 147: nit: specify an example path?

branlwyd (approved):

aggregator/src/aggregator.rs
branlwyd: I agree with _not_ failing out if we fail to write diagnostic information, i.e. we should log but continue on a failure to write diagnostic information.

docs/samples/advanced_config/aggregator.yaml
tgeoghegan: Sure!

divergentdave (approved):

aggregator/src/diagnostic.rs
```diff
@@ -0,0 +1,252 @@
...
+/// Represents an illegal attempt to mutate an aggregation job.
+#[derive(Derivative, Clone, Serialize, Deserialize)]
+#[derivative(Debug)]
+pub struct AggregationJobInitForbiddenMutationEvent {
+    /// The ID of the task.
+    #[serde(with = "serialize_task_id")]
+    pub task_id: TaskId,
+
+    /// The ID of the aggregation job.
+    #[serde(with = "serialize_job_id")]
```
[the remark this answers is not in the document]
tgeoghegan: Yeah, and it's worth my while to derive `Deserialize` so I can use it in the unit test to check the contents of the written event, since I use a different type as the inner event there. Additionally, I suspect we will soon add real implementations of `Deserialize` on the illegal mutation event so we can write tools that ingest the JSON events and analyze them, so I don't feel _too_ bad about the TODOs.

PR #3131, opened 2024-05-13
Implement cancellation on disconnect.

branlwyd:
This will be very helpful when we have requests which timeout, which is presented to Janus as a client disconnect. Otherwise, we would continue processing the request until it is complete.

Update Helper aggregation methods, which use rayon to parallelize processing, to respect cancellation. Specifically, the `receiver` will be dropped, causing the producer's `sender` to return a SendError, which will cause `try_for_each_with` to stop processing.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -100,10 +100,11 @@ use std::{
     collections::{HashMap, HashSet},
     fmt::Debug,
     hash::Hash,
+    panic,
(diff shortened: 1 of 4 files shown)


branlwyd:
I will update Leader aggregation methods in a follow-on PR which also implements parallelization.

Stacked on #3119.
Part of #3033.
Part of #3035.

branlwyd:
I'm going to merge this into the parent PR (#3119), as it addresses all current comments on that PR.

Issue #3132, opened 2024-05-14
Optimistically check for repeated aggregation job requests


divergentdave:
Currently the aggregation job init handler performs all VDAF preparation work first, before opening a database transaction. It might be operationally beneficial to add a second database transaction, and duplicate the check of whether there is an existing aggregation job in the database already. (we'd want to keep the existing check in the transaction that records everything, to catch races) In the case where a past job initialization successfully completed, but the leader didn't get the response for whatever reason, this would drastically speed up retried requests. In the case where requests don't match, this would let us return a 409 Conflict response without performing wasted VDAF preparation work. The obvious downsides of this would be increased complexity, and additional database transactions and round trips.

branlwyd:
Retitling & keeping open: the scope of this issue is now performing early repeated-request checks for all aggregation job operations (Leader/Helper, initialization/continuation). The above PR handles the Helper initialization case only.

tgeoghegan:
When the transaction that checks for illegal mutations returns an error, it causes our metrics for database transaction errors to be incremented. However that isn't really a database error but a normal bit of application logic being evaluated successfully. We should change the illegal mutation check so that it doesn't get treated as an error [here](https://github.com/divviup/janus/blob/c9487bfad6a5b8d025e234d88941a02fe29ecf77/aggregator_core/src/datastore.rs#L268).

branlwyd:
Oh, this is done:
* It doesn't apply to Leader aggregation operations, as the Leader uses a different mechanism (leasing) to ensure we don't repeatedly step the same aggregation job.
* Helper aggregation initialization was handled by PR #3135.
* Helper aggregation continuation already checks for a repeated request before performing VDAF compuations.

PR #3135, opened 2024-05-15
Helper agg init: check for repeated request before VDAF computations.

branlwyd:
This will considerably reduce the computational cost of repeated requests, at the cost of an additional (read-only) DB transaction per aggregation initialization.


branlwyd:
Part of #3132.

inahga (approved):

tgeoghegan (approved):

Issue #2147, opened 2023-10-18
Drop support for per-task HPKE configs


divergentdave:
We are considering dropping support for per-task HPKE configs to simplify. Note that this will have an impact on a number of unit tests. Global HPKE configs currently can only be set up via the aggregator API, so we may need to make the aggregator API required instead of optional, provide another means to set up a global keypair, or do so automatically on first boot.

inahga:
Dupe of https://github.com/divviup/janus/issues/1641, I'll let this issue supersede that one.

inahga:
Some notes: We already have support for global keys for the taskprov use case. Functionality should already be present for using these keys. Janus also refreshes its internal cache of these keys every 30 minutes.

We'll need some tweaks to remove `if taskprov_enabled`, and change the behavior of `/hpke_configs` to always present global keys. We will need to retain the ability for Janus to trial decrypt using a task-specific key, until the next breaking windows change. Task provisioning will need to be modified to stop creating task-specific keys. Several tests will need to be changed to stop using task-specific keys.

The main missing functionality is that keys currently need to be manually provisioned. This is a non-starter for normal deployments of Janus, of which we run several. We also need to consider that BYOH deployments probably don't want to have to worry about key management.

We need a subprocess `key-rotator` (to borrow terminology from ENPA) that ensures that an active HPKE key is present, and fully handles the lifecycle of keys. This can run independently for Janus leader use cases, and in the same binary for BYOH use cases. There is a first-start edge case that requires consideration, where an `aggregator` process starts running before the `key-rotator` does. A few options here, perhaps the aggregator's cache is always invalid until it has at least one active key.

inahga:
> There is a first-start edge case that requires consideration, where an aggregator process starts running before the key-rotator does. A few options here, perhaps the aggregator's cache is always invalid until it has at least one active key.

This can be accomplished by having `GlobalHpkeKeypairCache::new()` backoff retry fetching keypairs until it finds at least one active one. This will block the process from fully starting, until the key rotator service inserts an active keypair.

For BYOH use cases where we're supplying an all-in-one binary, the key rotator task should be started before the HPKE cache task. The key rotator should run a sweep immediately on creation.

inahga:
I've sketched out this logic for the key rotator (may have errors, let me know if anything stands out as wrong).

Configuration:
```rust
struct Options {
    /// How often to run the key rotator, in seconds
    pub frequency_s: u64,
    pub hpke: HpkeOptions,
}

[...]

PR #3136, opened 2024-05-15
Key rotator

inahga:

Introduces the `key_rotator`, which is a utility for managing the lifecycle of global HPKE keys. It will bootstrap keys, and rotate them according to the rotation policy in the config.

Keys are unique by their HPKE ciphersuite. For each ciphersuite, a key is run through the pending->active->expired->deleted lifecycle. It is permissive of some manual operator changes, e.g. if a manual key rotation needs to be executed through janus_cli or the aggregator API.

In this iteration, it runs as a standalone binary for use in a cronjob. It is suitable for deployment in our environments, including taskprov ones. A future PR will add support for BYOH deployments by letting the `aggregator` process run the key rotator.

### DIFF
--- a/aggregator/src/aggregator/key_rotator.rs
+++ b/aggregator/src/aggregator/key_rotator.rs
@@ -1,7 +1,8 @@
-use std::sync::Arc;
+use std::{collections::HashSet, sync::Arc};
 
 #[allow(unused_imports)]
-use crate::aggregator::Config as AggregatorConfig; // used in doccomment.
+use crate::aggregator::Config as AggregatorConfig;
+use crate::cache::GlobalHpkeKeypairCache; // used in doccomment.
 use anyhow::{anyhow, Error};
 use janus_aggregator_core::datastore::{
     models::{GlobalHpkeKeypair, HpkeKeyState},
@@ -11,8 +12,8 @@ use janus_core::{
     hpke::{generate_hpke_config_and_private_key, HpkeCiphersuite},
     time::{Clock, TimeExt},
 };
-use janus_messages::{Duration, HpkeConfigId, Time};
-use serde::{Deserialize, Serialize};
+use janus_messages::{Duration, HpkeAeadId, HpkeConfigId, HpkeKdfId, HpkeKemId, Time};
+use serde::{de, Deserialize, Deserializer, Serialize};
 use tracing::{debug, info};
 
 /// Handles key rotation for Janus, according to policies defined in the configuration.
@@ -31,6 +32,7 @@ use tracing::{debug, info};
 ///   - Promote pending keypairs too early, unless all Janus replicas have been rebooted since
 ///     their insertion.
 ///   - Directly insert active keypairs.
+///
 /// Keypairs that are manually inserted are adopted by the key rotator and will have their lifecycle
 /// managed. The number of keypairs per ciphersuite will trend towards 1, i.e. 2 keys with the same
 /// ciphersuite will eventually be replaced with 1 key.
@@ -46,17 +48,21 @@ pub struct HpkeKeyRotatorConfig {
     /// How long key remains in [`HpkeKeyState::Pending`] before being moved to
     /// [`HpkeKeyState::Active`]. This should be greater than
     /// [`AggregatorConfig::global_hpke_configs_refresh_interval`].
+    #[serde(rename = "pending_duration_secs", default = "default_pending_duration")]
     pub pending_duration: Duration,
 
     /// The time-to-live of the key. Once this is exceeded, the key is moved to
     /// [`HpkeKeyState::Expired`]. It is at operator discretion as to how long this should be.
+    #[serde(rename = "active_duration_secs", default = "default_active_duration")]
     pub active_duration: Duration,
 
     /// How long the key remains in [`HpkeKeyState::Expired`] before being deleted. This should
     /// be greater than the clients' HPKE key cache maximum age.
+    #[serde(rename = "expired_duration_secs", default = "default_expired_duration")]
     pub expired_duration: Duration,
 
     /// The ciphersuite of the key.
+    #[serde(default = "default_hpke_ciphersuite")]
     pub ciphersuite: HpkeCiphersuite,
 
     /// Safely phase out this ciphersuite. This will immediately delete any pending keys and
@@ -68,7 +74,32 @@ pub struct HpkeKeyRotatorConfig {
     ///
     /// There must be at least one non-retired config present with its key in the active state. If
     /// this is the only config and it is marked retired, the key rotator will refuse to run.
+    #[serde(default)]
     pub retire: bool,
+}
+
+/// Returns [`GlobalHpkeKeypairCache::DEFAULT_REFRESH_INTERVAL`] times 2, for safety margin.
+fn default_pending_duration() -> Duration {
+    Duration::from_seconds(GlobalHpkeKeypairCache::DEFAULT_REFRESH_INTERVAL.as_secs() * 2)
+}
+
+/// 4 weeks. This is long enough not to be unnecessary churn, but short enough that misbehaving
+/// clients reveal themselves somewhat imminently.
+fn default_active_duration() -> Duration {
+    Duration::from_seconds(60 * 60 * 24 * 7 * 4)
+}
+
+/// 1 week.
+fn default_expired_duration() -> Duration {
+    Duration::from_seconds(60 * 60 * 24 * 7)
+}
+
+fn default_hpke_ciphersuite() -> HpkeCiphersuite {
+    HpkeCiphersuite::new(
+        HpkeKemId::X25519HkdfSha256,
+        HpkeKdfId::HkdfSha256,
+        HpkeAeadId::Aes128Gcm,
+    )
 }
 
 impl<C: Clock> KeyRotator<C> {
@@ -85,7 +116,7 @@ impl<C: Clock> KeyRotator<C> {
     /// This is not permissive of [`HpkeConfigId`] space exhaustion, i.e. if there are more rows
     /// than what fit into a `u8`, this process will error. To avoid this, keep the list of managed
     /// ciphersuites small, and the expiration duration less than the key age.
-    #[tracing::instrument]
+    #[tracing::instrument(err)]
     pub async fn run(&self) -> Result<(), Error> {
         self.datastore
             .run_tx("global_hpke_key_rotator", |tx| {
@@ -96,7 +127,7 @@ impl<C: Clock> KeyRotator<C> {
             .map_err(|err| err.into())
     }
 
-    #[tracing::instrument(skip(tx))]
+    #[tracing::instrument(err, skip(tx))]
     async fn run_hpke(
         tx: &Transaction<'_, C>,
         configs: Vec<HpkeKeyRotatorConfig>,
@@ -135,7 +166,7 @@ impl<C: Clock> KeyRotator<C> {
         }
     }
 
-    #[tracing::instrument(skip(tx, current_keypairs, available_ids))]
+    #[tracing::instrument(err, skip(tx, current_keypairs, available_ids))]
     async fn run_hpke_for_config(
         tx: &Transaction<'_, C>,
         current_keypairs: &[GlobalHpkeKeypair],
@@ -308,6 +339,29 @@ struct PartitionedKeypairs<'a> {
     pending: Vec<&'a GlobalHpkeKeypair>,
     active: Vec<&'a GlobalHpkeKeypair>,
     expired: Vec<&'a GlobalHpkeKeypair>,
+}
+
+/// Enforces that there's at least one [`HpkeKeyRotatorConfig`].
+pub fn deserialize_hpke_key_rotator_configs<'de, D>(
+    deserializer: D,
+) -> Result<Vec<HpkeKeyRotatorConfig>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let configs: Vec<HpkeKeyRotatorConfig> = Deserialize::deserialize(deserializer)?;
+    let mut unique = HashSet::new();
+    if configs.is_empty() {
+        Err(de::Error::custom(
+            "must provide at least one hpke key rotator config",
+        ))
+    } else if !configs
+        .iter()
+        .all(|config| unique.insert(config.ciphersuite))
+    {
+        Err(de::Error::custom("each ciphersuite must be unique"))
+    } else {
+        Ok(configs)
+    }
 }
 
 #[cfg(test)]
--- a/aggregator/src/aggregator/key_rotator.rs
+++ b/aggregator/src/aggregator/key_rotator.rs
@@ -23,9 +23,6 @@ use tracing::{debug, info};
 /// The key rotator can handle key rotation for global HPKE keys. The key rotator is tolerant of
 /// some manual operator changes to the `global_hpke_keys` table.
 ///
-/// Operators _must not_ manually change the `created_at` or `updated_at` columns, as these are
-/// used for rotation decision-making.
-///
 /// It is strongly discouraged to:
 ///   - Delete active or expired keypairs too early, as that would leave Janus unable to decrypt
 ///     report shares using the keypair.
@@ -143,7 +140,7 @@ impl<C: Clock> KeyRotator<C> {
         tx.lock_global_hpke_keypairs().await?;
 
         let current_keypairs = tx.get_global_hpke_keypairs().await?;
-        debug!(?current_keypairs, "before");
+        debug!(?current_keypairs, "table state before running key rotator");
         let mut available_ids = Self::available_ids(&current_keypairs);
         for config in &configs {
             Self::run_hpke_for_config(tx, &current_keypairs, &mut available_ids, config).await?;
@@ -154,7 +151,7 @@ impl<C: Clock> KeyRotator<C> {
         // so we should rollback. This captures any bugs, and if the operator is trying to retire
         // the last available key.
         let current_keypairs = tx.get_global_hpke_keypairs().await?;
-        debug!(?current_keypairs, "after");
+        debug!(?current_keypairs, "table state after running key rotator");
         if !current_keypairs
             .iter()
             .any(|keypair| keypair.state() == &HpkeKeyState::Active)
@@ -232,7 +229,8 @@ impl<C: Clock> KeyRotator<C> {
             let to_be_expired_keypairs: Vec<_> = active_keypairs
                 .iter()
                 .filter(|keypair| {
-                    Self::duration_since(clock, keypair.updated_at()) > config.active_duration
+                    Self::duration_since(clock, keypair.last_state_change_at())
+                        > config.active_duration
                 })
                 .cloned()
                 .collect();
@@ -249,7 +247,8 @@ impl<C: Clock> KeyRotator<C> {
                 tx.put_global_hpke_keypair(&next).await?;
             } else {
                 let pending_key_is_ready = pending_keypairs.iter().any(|keypair| {
-                    Self::duration_since(clock, keypair.updated_at()) > config.pending_duration
+                    Self::duration_since(clock, keypair.last_state_change_at())
+                        > config.pending_duration
                 });
                 if pending_key_is_ready {
                     for to_be_expired_keypair in to_be_expired_keypairs {
@@ -267,7 +266,7 @@ impl<C: Clock> KeyRotator<C> {
             // as active keypairs, because it's possible for there to be a pending keypair
             // without an active one.
             for pending_keypair in pending_keypairs {
-                if Self::duration_since(clock, pending_keypair.updated_at())
+                if Self::duration_since(clock, pending_keypair.last_state_change_at())
                     > config.pending_duration
                 {
                     info!(id = ?pending_keypair.id(), "pending key is ready, moving it to active");
@@ -278,7 +277,9 @@ impl<C: Clock> KeyRotator<C> {
         }
 
         for expired_keypair in &expired_keypairs {
-            if Self::duration_since(clock, expired_keypair.updated_at()) > config.expired_duration {
+            if Self::duration_since(clock, expired_keypair.last_state_change_at())
+                > config.expired_duration
+            {
                 info!(id = ?expired_keypair.id(), "deleting expired keypair");
                 tx.delete_global_hpke_keypair(expired_keypair.id()).await?;
             }
@@ -321,7 +322,7 @@ impl<C: Clock> KeyRotator<C> {
         // config error, rather than attempting to decrypt them with the wrong key.
         let newest_id = current_keypairs
             .iter()
-            .max_by(|a, b| a.created_at().cmp(b.created_at()))
+            .max_by_key(|keypair| keypair.id())
             .map(|keypair| (*keypair.id()).into())
             .unwrap_or(0);
         (newest_id..=u8::MAX)
--- a/aggregator/src/aggregator/key_rotator.rs
+++ b/aggregator/src/aggregator/key_rotator.rs
@@ -242,21 +242,19 @@ impl<C: Clock> KeyRotator<C> {
                     tx.set_global_hpke_keypair_state(keypair.id(), &HpkeKeyState::Expired)
                         .await?;
                 }
+            } else if pending_keypairs.is_empty() {
+                let next = next_key()?;
+                info!(id = ?next.config().id(), "inserting new pending key to replace expired one(s)");
+                tx.put_global_hpke_keypair(&next).await?;
             } else {
-                if pending_keypairs.is_empty() {
-                    let next = next_key()?;
-                    info!(id = ?next.config().id(), "inserting new pending key to replace expired one(s)");
-                    tx.put_global_hpke_keypair(&next).await?;
-                } else {
-                    let pending_key_is_ready = pending_keypairs.iter().any(|keypair| {
-                        Self::duration_since(clock, keypair.updated_at()) > config.pending_duration
-                    });
-                    if pending_key_is_ready {
-                        for keypair in to_be_expired_keypairs {
-                            info!(id = ?keypair.id(), "a pending key is ready, marking key expired");
-                            tx.set_global_hpke_keypair_state(keypair.id(), &HpkeKeyState::Expired)
-                                .await?;
-                        }
+                let pending_key_is_ready = pending_keypairs.iter().any(|keypair| {
+                    Self::duration_since(clock, keypair.updated_at()) > config.pending_duration
+                });
+                if pending_key_is_ready {
+                    for keypair in to_be_expired_keypairs {
+                        info!(id = ?keypair.id(), "a pending key is ready, marking key expired");
+                        tx.set_global_hpke_keypair_state(keypair.id(), &HpkeKeyState::Expired)
+                            .await?;
                     }
                 }
             }
@@ -416,14 +414,14 @@ mod tests {
                     pending_duration,
                     active_duration,
                     expired_duration,
-                    ciphersuite: ciphersuite_0.clone(),
+                    ciphersuite: ciphersuite_0,
                     retire: false,
                 },
                 HpkeKeyRotatorConfig {
                     pending_duration,
                     active_duration,
                     expired_duration,
-                    ciphersuite: ciphersuite_1.clone(),
+                    ciphersuite: ciphersuite_1,
                     retire: false,
                 },
             ]),
@@ -432,10 +430,11 @@ mod tests {
         // Checks that there's a keypair with the given state for each ciphersuite.
         let ciphersuites = Vec::from([ciphersuite_0, ciphersuite_1]);
         let assert_state = |keypairs: &[GlobalHpkeKeypair], state: HpkeKeyState| {
-            assert!(ciphersuites.iter().all(|ciphersuite| keypairs
-                .iter()
-                .find(|keypair| keypair.state() == &state && &keypair.ciphersuite() == ciphersuite)
-                .is_some()));
+            assert!(ciphersuites.iter().all(|ciphersuite| {
+                keypairs.iter().any(|keypair| {
+                    keypair.state() == &state && &keypair.ciphersuite() == ciphersuite
+                })
+            }));
         };
 
         // First iteration: We should create active keys for each ciphersuite.
@@ -517,7 +516,7 @@ mod tests {
                 pending_duration,
                 active_duration,
                 expired_duration,
-                ciphersuite: ciphersuite_0.clone(),
+                ciphersuite: ciphersuite_0,
                 retire: false,
             }]),
         );
@@ -541,14 +540,14 @@ mod tests {
                     pending_duration,
                     active_duration,
                     expired_duration,
-                    ciphersuite: ciphersuite_0.clone(),
+                    ciphersuite: ciphersuite_0,
                     retire: false,
                 },
                 HpkeKeyRotatorConfig {
                     pending_duration,
                     active_duration,
                     expired_duration,
-                    ciphersuite: ciphersuite_1.clone(),
+                    ciphersuite: ciphersuite_1,
                     retire: false,
                 },
             ]),
@@ -612,7 +611,7 @@ mod tests {
                 pending_duration,
                 active_duration,
                 expired_duration,
-                ciphersuite: ciphersuite_0.clone(),
+                ciphersuite: ciphersuite_0,
                 retire: false,
             }]),
         );
@@ -628,7 +627,6 @@ mod tests {
         // Operator inserts a new key in the pending state.
         let id = HpkeConfigId::from(255);
         ds.run_unnamed_tx(|tx| {
-            let ciphersuite_0 = ciphersuite_0.clone();
             Box::pin(async move {
                 tx.put_global_hpke_keypair(
                     &generate_hpke_config_and_private_key(
@@ -650,8 +648,7 @@ mod tests {
         assert_eq!(keypairs.len(), 3);
         assert!(keypairs
             .iter()
-            .find(|keypair| keypair.state() == &HpkeKeyState::Active)
-            .is_some());
+            .any(|keypair| keypair.state() == &HpkeKeyState::Active));
         let pending_keypairs: Vec<_> = keypairs
             .iter()
             .filter(|keypair| keypair.state() == &HpkeKeyState::Pending)
@@ -667,10 +664,7 @@ mod tests {
             .filter(|keypair| keypair.state() == &HpkeKeyState::Active)
             .collect();
         assert_eq!(active_keypairs.len(), 2);
-        assert!(active_keypairs
-            .iter()
-            .find(|keypair| *keypair.id() == id)
-            .is_some());
+        assert!(active_keypairs.iter().any(|keypair| *keypair.id() == id));
 
         // Step into the future, we should eventually replace both active keypairs with only one.
         for _ in 0..30 {
@@ -708,7 +702,7 @@ mod tests {
                 pending_duration,
                 active_duration,
                 expired_duration,
-                ciphersuite: ciphersuite_0.clone(),
+                ciphersuite: ciphersuite_0,
                 retire: false,
             }]),
         );
@@ -722,7 +716,6 @@ mod tests {
         // Operator inserts a new key in the active state.
         let id = HpkeConfigId::from(255);
         ds.run_unnamed_tx(|tx| {
-            let ciphersuite_0 = ciphersuite_0.clone();
             Box::pin(async move {
                 tx.put_global_hpke_keypair(
                     &generate_hpke_config_and_private_key(
@@ -757,7 +750,7 @@ mod tests {
             key_rotator.run().await.unwrap();
             clock.advance(&Duration::from_seconds(30));
         }
-        let keypairs = dbg!(get_global_hpke_keypairs(&ds).await);
+        let keypairs = get_global_hpke_keypairs(&ds).await;
         let active_keypairs: Vec<_> = keypairs
             .iter()
             .filter(|keypair| keypair.state() == &HpkeKeyState::Active)
@@ -787,7 +780,7 @@ mod tests {
                 pending_duration,
                 active_duration,
                 expired_duration,
-                ciphersuite: ciphersuite_0.clone(),
+                ciphersuite: ciphersuite_0,
                 retire: true,
             }]),
         );
@@ -800,7 +793,7 @@ mod tests {
                 pending_duration,
                 active_duration,
                 expired_duration,
-                ciphersuite: ciphersuite_0.clone(),
+                ciphersuite: ciphersuite_0,
                 retire: false,
             }]),
         );
@@ -816,7 +809,7 @@ mod tests {
                 pending_duration,
                 active_duration,
                 expired_duration,
-                ciphersuite: ciphersuite_0.clone(),
+                ciphersuite: ciphersuite_0,
                 retire: true,
             }]),
         );
@@ -852,14 +845,14 @@ mod tests {
                     pending_duration,
                     active_duration,
                     expired_duration,
-                    ciphersuite: ciphersuite_0.clone(),
+                    ciphersuite: ciphersuite_0,
                     retire: false,
                 },
                 HpkeKeyRotatorConfig {
                     pending_duration,
                     active_duration,
                     expired_duration,
-                    ciphersuite: ciphersuite_1.clone(),
+                    ciphersuite: ciphersuite_1,
                     retire: false,
                 },
             ]),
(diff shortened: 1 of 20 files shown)


aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,266 @@
...
+                                        }
+                                    }
+                                }
+                            }
+                            None => {
+                                // Bootstrapping case: there is no active keypair with the given
+                                // ciphersuite. This means one of the following cases:
+                                //   - Fresh instance of Janus coming up, and a new key needs to be
+                                //     bootstrapped.
+                                //   - The operator has added a new ciphersuite to the config.
```
divergentdave, line 156: Technically in this case, we would want to go through the usual delay with the key in a pending state. That would either take more aggressive checks in the upload path, which may not be wise, or Tink-style feedback of which keys are in use where, which we don't have appropriate infrastructure for in our deployment model. We could just document that you have to create a pending global keypair with a given ciphersuite, and wait, before you can add that ciphersuite to the key rotator's configuration.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,266 @@
...
+#[derive(Debug)]
+pub struct KeyRotator<C: Clock> {
+    clock: C,
+    datastore: Arc<Datastore<C>>,
+    hpke: HpkeKeyRotatorConfig,
+}
+
+/// Handles key rotation for global HPKE keys. It is assumed that the [`KeyRotator`] is the sole
+/// entity responsible for the `global_hpke_keys` table. Do not make manual changes to the
+/// `global_hpke_keys` table when the key rotator is used.
```
divergentdave, line 26: I think we'll need to be able to support manually-initiated key rotations, (informed by our experience with ENPA) even if it's a multi-step process. If we make the key rotator robust to multiple pending keys and multiple active keys, that will help out a lot. We will still have to tell the user not to delete active keys, and not to promote pending keys to active early unless they've rebooted aggregator processes, and write up a playbook with janus_cli commands for a manual rotation.

Previously, with `key-rotator`, we had to emulate manual rotations by changing times in `key-rotator`'s configuration. That led to some surprises when we did so, and when we restored the original configuration again. I previously editorialized about this here: https://github.com/divviup/janus-ops/issues/142#issuecomment-1629571706

One nice thing about HPKE keys for DAP, as opposed to other secrets we need to rotate, is that we can get away with having multiple "active" keys with the same ciphersuites at once, as that just means we'll have a longer HpkeConfigList. Thus, it's okay for a manual rotation playbook to update key states in separate steps, not transactionally, and we don't necessarily need a new janus_cli command.

aggregator/src/aggregator/key_rotator.rs
divergentdave: Maybe we could make the step where we immediately set the state to active conditional on `current_keypairs` being empty. Then, it would only come into play in the bootstrapping scenario or in the case of outside interference.

aggregator/src/aggregator/key_rotator.rs
inahga: Thanks for the context, I'll add support for manual key rotations.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,266 @@
...
+                    // caches without being blocked.
+                    tx.lock_global_hpke_keypairs().await?;
+
+                    let current_keypairs = tx.get_global_hpke_keypairs().await?;
+
+                    let ids: Vec<_> = current_keypairs
+                        .iter()
+                        .map(|keypair| u8::from(*keypair.hpke_keypair().config().id()))
+                        .collect();
+                    let mut available_ids = (0..=u8::MAX).filter_map(|id| {
```
divergentdave, line 77: This is less important, but we could ensure that our IDs cycle through the whole u8 space if we identified which keypair was created most recently, and then replaced `(0..=u8::MAX)` with `(newest_id..=u8::MAX).chain(0..newest_id)`. That way, weird clients with very old cached keys will have a much bigger window during which we will send them urn:ietf:params:ppm:dap:error:outdatedConfig, instead of trying to decrypt it with a new key with the same ID.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,673 @@
...
+        // in the table. If one is absent, committing the transaction would leave Janus unstartable
+        // so we should rollback.
+        let current_keypairs = tx.get_global_hpke_keypairs().await?;
+        debug!(?current_keypairs, "after");
+        if !current_keypairs
+            .iter()
+            .any(|keypair| keypair.state() == &HpkeKeyState::Active)
+        {
+            Err(DatastoreError::User(
+                anyhow!("unexpected state: no keypairs are active").into(),
```
divergentdave, line 119: nit: This could fire if the database is empty and the `hpke` list in the config is empty. We could check if there are no algorithms chosen first, log a warning, and exit early to provide a nicer message in that case.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,673 @@
...
+            if Self::duration_since(tx, keypair.updated_at()) > config.expired_duration {
+                info!(id = ?keypair.id(), "deleting expired keypair");
+                tx.delete_global_hpke_keypair(keypair.id()).await?;
+            }
+        }
+
+        Ok(())
+    }
+
+    fn duration_since(tx: &Transaction<C>, time: &Time) -> Duration {
```
divergentdave, line 227: nit: I think making this take a `&C` argument instead would be a bit clearer, as it would be obvious there's no database manipulation happening. You could also add a `let clock = tx.clock();` binding at the top of `run_hpke_for_config()` to eliminate some repetition.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,673 @@
...
+                .iter()
+                .filter(|keypair| {
+                    Self::duration_since(tx, keypair.updated_at()) > config.active_duration
+                })
+                .cloned()
+                .collect();
+
+            if to_be_expired_keypairs != active_keypairs {
+                for keypair in to_be_expired_keypairs {
+                    info!(id = ?keypair.id(), "allowing already active keypair to replace this one");
```
divergentdave, line 180: IMO this message should be a bit more specific that the relevant key's state is being set to expired.
```suggestion
                    info!(id = ?keypair.id(), "multiple active keys are present, marking key expired");
```

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,673 @@
...
+
+    fn available_ids(current_keypairs: &[GlobalHpkeKeypair]) -> impl Iterator<Item = HpkeConfigId> {
+        let ids: Vec<_> = current_keypairs
+            .iter()
+            .map(|keypair| u8::from(*keypair.id()))
+            .collect();
+
+        // Try to cycle through the entire u8 space before going back to zero. This allows us a
+        // bigger window to quicky reject faulty old clients with an outdated HPKE config error,
+        // rather than attempting to decrypt them with the wrong key.
```
divergentdave, line 262: nit
```suggestion
        // bigger window to quicky reject reports from faulty old clients with an outdated HPKE
        // config error, rather than attempting to decrypt them with the wrong key.
```

divergentdave (commented):
Looking good!

aggregator/src/aggregator/key_rotator.rs
inahga: My (eventual) intent is to provide a sane default for `hpke`, such that the `hpke` list is never empty.

inahga:
One edge case currently unhandled is when a ciphersuite is removed from the config. The key is then no longer managed at all by the key rotator.

The use case for this is to retire old/undesired algorithms. I think we ought to add a parameter `retire: true` which automatically phases out a key, i.e. discontinues renewing it and quickly moves it to the expired state. This logic should be simple enough, and takes out manual toil if we need to retire a ciphersuite.

aggregator/src/aggregator/key_rotator.rs
inahga: On further thought I don't think it's wise to provide a default.

If we provide a default, if the operator wants to introduce a non-default key, they must provide the default key in their `hpke:` stanza, otherwise the key rotator will forget about the default key.

I think instead I'll just make providing at least one ciphersuite mandatory. (The goal of #2147 is to make global keys a requirement, so this approach makes sense to me).

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,911 @@
...
+            .map_err(|e| DatastoreError::User(e.into()))
+        };
+
+        if config.retire {
+            for pending_keypair in pending_keypairs {
+                // Janus replicas should have never advertised this keypair, so it should be safe
+                // to delete outright.
+                info!(id = ?pending_keypair.id(), "deleting pending keypair for retired ciphersuite");
+                tx.delete_global_hpke_keypair(pending_keypair.id()).await?;
+            }
```
inahga, line 202: n.b. we could `try_join` a lot of calls like this, but since this code is not performance sensitive, I prioritized readability and straightforwardness.

tgeoghegan (commented):
It makes sense to have a key rotation component in Janus to handle those keys that Janus itself is aware of, like HPKE configurations or maybe aggregator API tokens. With some work, this could be extended to handle database encryption keys, too.

But in Divvi Up, we have several secrets that we want to manage outside of Janus' scope, such as seeds for OHTTP key configurations. Does this key-rotator component fit into a bigger plan for managing all of Divvi Up's secrets? For example, could we write a higher-level key rotator that execs `janus_key_rotator` alongside other things?

divergentdave (approved):

aggregator/src/binaries/key_rotator.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    fn common_config_mut(&mut self) -> &mut CommonConfig {
+        &mut self.common_config
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct KeyRotatorConfig {
+    #[serde(default, deserialize_with = "deserialize_hpke_key_rotator_configs")]
```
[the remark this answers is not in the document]
inahga: Nice catch!

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,911 @@
+use std::{collections::HashSet, sync::Arc};
+
+#[allow(unused_imports)]
+use crate::aggregator::Config as AggregatorConfig;
+use crate::cache::GlobalHpkeKeypairCache; // used in doccomment.
```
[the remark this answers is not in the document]
inahga: Oops, I guess `rustfmt` mangled this at some point.

inahga:
> Does this key-rotator component fit into a bigger plan for managing all of Divvi Up's secrets? For example, could we write a higher-level key rotator that execs janus_key_rotator alongside other things?

I don't see why not. Since this process runs one sweep at a time, we have flexibility in what context the process is run, e.g. cronjob, as part of another process, etc.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,911 @@
...
+            .iter()
+            .map(|keypair| u8::from(*keypair.id()))
+            .collect();
+
+        // Try to cycle through the entire u8 space before going back to zero. This allows us a
+        // bigger window to quicky reject reports from faulty old clients with an outdated HPKE
+        // config error, rather than attempting to decrypt them with the wrong key.
+        let newest_id = current_keypairs
+            .iter()
+            .max_by(|a, b| a.created_at().cmp(b.created_at()))
```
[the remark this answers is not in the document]
divergentdave: Note that treating the highest config ID as the most recent will eventually result in cycling through 253, 254, and 255, instead of 0, 1, and 2 if we picked the lowest unused ID.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,911 @@
...
+use tracing::{debug, info};
+
+/// Handles key rotation for Janus, according to policies defined in the configuration.
+///
+/// # Global HPKE Keys
+///
+/// The key rotator can handle key rotation for global HPKE keys. The key rotator is tolerant of
+/// some manual operator changes to the `global_hpke_keys` table.
+///
+/// Operators _must not_ manually change the `created_at` or `updated_at` columns, as these are
```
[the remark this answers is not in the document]
inahga: > WDYT about introducing a dedicated column for tracking rotation information, e.g. last_rotated_at?

That sounds fine. I was similarly on the fence for the same points you raised.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,911 @@
...
+    #[serde(default)]
+    pub retire: bool,
+}
+
+/// Returns [`GlobalHpkeKeypairCache::DEFAULT_REFRESH_INTERVAL`] times 2, for safety margin.
+fn default_pending_duration() -> Duration {
+    Duration::from_seconds(GlobalHpkeKeypairCache::DEFAULT_REFRESH_INTERVAL.as_secs() * 2)
+}
+
+/// 4 weeks. This is long enough not to be unnecessary churn, but short enough that misbehaving
```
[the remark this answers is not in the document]
inahga: I was hoping that a short rotation would force clients to have proper key fetching and caching strategy, i.e. the surprises would happen in the 1 month timespan rather than the 1 year timespan, and we would discourage any permanent retention of the key.

Short rotation also validates whether all this stuff works in production, so that we're not shocked 1 year later when we've completely forgotten how this tool works.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,911 @@
...
+    ) -> Result<(), DatastoreError> {
+        let clock = tx.clock();
+
+        let PartitionedKeypairs {
+            pending_keypairs,
+            active_keypairs,
+            expired_keypairs,
+        } = Self::partition_keypairs(current_keypairs, &config.ciphersuite);
+
+        let mut next_key = || {
```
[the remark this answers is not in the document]
inahga: Worth a try!

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,911 @@
...
+/// ciphersuite will eventually be replaced with 1 key.
+#[derive(Debug)]
+pub struct KeyRotator<C: Clock> {
+    datastore: Arc<Datastore<C>>,
+    hpke: Vec<HpkeKeyRotatorConfig>,
+}
+
+/// Defines the ciphersuite and rotation policy of a global HPKE key.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct HpkeKeyRotatorConfig {
```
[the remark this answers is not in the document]
inahga: I'll take this suggestion to enable the `retired` fix below.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,911 @@
...
+    /// expire any active keys. Expired keys are retained for at least [`Self::expired_duration`]
+    /// before being deleted.
+    ///
+    /// After [`Self::expired_duration`], this config can be removed entirely. It is not recommended
+    /// to remove key configurations without running the key rotator with `retire` first.
+    ///
+    /// There must be at least one non-retired config present with its key in the active state. If
+    /// this is the only config and it is marked retired, the key rotator will refuse to run.
+    #[serde(default)]
+    pub retire: bool,
```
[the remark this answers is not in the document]
inahga: That SGTM.

I had introduced `retire` to work around the fact that we would need some "global" set of durations to handle unaccounted for keys, but yeah it's just as simple to have global durations for every key.

aggregator/src/aggregator/key_rotator.rs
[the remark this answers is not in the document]
inahga: Ah, the reason I had done this that the iterator is shared across different runs of `run_hpke_for_config`, so we need a config/ciphersuite specific function that consumes the iterator for each `run_hpke_for_config`.

I suppose the iterator could be something like `Iterator<Item = FnOnce(HpkeCiphersuite) -> Result<HpkeKeypair, DatastoreError>>` but maybe that's too clever. Either way I'll play around with this more.

aggregator/src/aggregator/key_rotator.rs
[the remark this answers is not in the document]
inahga: Good suggestion, this should be operationally simpler.

I'm treating "unknown" keys, i.e. ones that we don't have a ciphersuite in the config for, as ones that should be imminently retired. I figure this case is most often the operator removing a ciphersuite entry from the config.

It does mean that any keys that are manually inserted by the operator and aren't in the list of known ciphersuites are unceremoniously expired/deleted though. I'd prefer this so that we retain the predictable behavior of config removal resulting in key retirement.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,858 @@
...
+            //
+            // Otherwise, the ciphersuite was likely added to the configuration and its key
+            // needs to go through the normal pending->active state change.
+            if current_keypairs.is_empty() {
+                info!(id = ?new_keypair.config().id(), "bootstrapping: moving new key to active state");
+                tx.set_global_hpke_keypair_state(new_keypair.config().id(), &HpkeKeyState::Active)
+                    .await?;
+            }
+        } else {
+            let to_be_expired_keypairs: Vec<_> = active_keypairs
```
branlwyd, line 238: I think I follow the logic here, but it's pretty complicated & procedural. I might just be missing an obvious policy, but it also seems somewhat ad-hoc. It is also challenging to be completely confident in this logic, as the logic switches between working with in-memory representations & actually touching the DB -- the in-memory representations will be out of sync with the state of the DB after we start touching the DB, so it's somewhat challenging to reason about this code.

I would suggest writing an explicit human-language policy for rotation, then implementing that policy. If there's no simple policy that matches the current code, a per-ciphersuite policy that might work would be (cribbing from [the `prio-server` key rotator](https://github.com/divviup/prio-server/blob/5aeb486a0f0a0d58a028cedcb9f6baa1f2bbac44/key-rotator/key/key.go#L214)):

1. For each pending key, if the pending key is at least age `active_min_age`, promote it to active.
1. If the ciphersuite is configured, and there are no pending/active keys or the newest pending/active key is at least age `creation_min_age`, create a new pending key.
1. While there are more than `maximum_active_keys` active keys, expire the oldest active key. [I think `maximum_active_keys` would typically be set to 1, and would be treated 0 for unconfigured ciphersuites. Maybe this doesn't need to be a config knob at all & can always be treated as 1 for configured ciphersuites, 0 for unconfigured?]
1. For each expired key, if the expired key is at least age `deletion_min_age`, delete it.

(It would be good to write this policy down in comments, near the code which implements each step. This policy is just an example that I think works, but maybe we can come up with a simpler one? Especially the special-casing around unconfigured ciphersuites might permit a simpler policy.)

To make the code clearer/more obviously correct, I suggest structuring it into three phases:
1) Read the existing DB state into an in-memory structure.
2) Manipulate the in-memory structure to perform rotation (in-memory). Ideally this would remove the "we have a bunch of in-memory stuff we make decisions on, but which is out of sync with changes we've already written to the DB" property the current code has, as decisions would be made at each step based on the current state of the in-memory structure, which would reflect the operations already performed on the in-memory structure.
3) Write the in-memory structure back to the DB, replacing the existing contents of the DB. This could be done by keeping track of a diff in the in-memory structure, generating a diff based on the in-memory structure, or even wiping the table & rewriting it with each rotation.

I would also like it if the handling for configured vs unconfigured ciphersuites was unified, as long as that doesn't introduce too much complexity. The suggested policy above does an OK job of this IMO, but maybe we can do better?

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,858 @@
...
+                        tx.set_global_hpke_keypair_state(keypair.id(), &HpkeKeyState::Expired)
+                            .await?;
+                    }
+                    HpkeKeyState::Pending => {
+                        // Janus replicas should have never advertised this keypair, so it should
+                        // be safe to delete outright.
+                        info!(id = ?keypair.id(), "deleting pending keypair for unknown ciphersuite");
+                        tx.delete_global_hpke_keypair(keypair.id()).await?;
+                    }
+                    HpkeKeyState::Expired => {}
```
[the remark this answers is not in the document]
inahga: No, the expired key will eventually hit this arm above:
```
            if keypair.state() == &HpkeKeyState::Expired
                && Self::duration_since(tx.clock(), keypair.last_state_change_at())
                    > config.expired_duration
```
So we don't need to take any action in the `!config.ciphersuites.contains(&keypair.ciphersuite())` arm.

aggregator/src/aggregator/key_rotator.rs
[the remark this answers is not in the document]
inahga: That's fair, but I do feel strongly that 1 year is too long, for at least the other reasons. Can split the difference though, 3 months? 6 months?

> Still, if that's not convincing, let's make sure we communicate with all deployed folks before we enable key rotation in their respective environments.

:+1:

aggregator/src/aggregator/key_rotator.rs
inahga: > I would suggest writing an explicit human-language policy for rotation, then implementing that policy.

:+1:

> Read the existing DB state into an in-memory structure.
> ...
> Write the in-memory structure back to the DB, replacing the existing contents of the DB.

I had considered this early but chose not to since I felt the logic was simple enough, though borderline, that we could avoid the extra code.

In any case, I'm amenable to this idea. I was careful to not rely on fields that would have been outdated, but perhaps it's unnecessary mental overhead to have to consider the current table state versus in-memory state for each operation.

> I would also like it if the handling for configured vs unconfigured ciphersuites was unified, as long as that doesn't introduce too much complexity.

I disagree that we should handle unconfigured ciphersuites like we do configured ones. IMO if we encounter an unconfigured ciphersuite we should swiftly phase it out. This strikes me as the least surprising behavior.

That is, as an operator, the least surprising result of removing a ciphersuite from the configuration would be that the respective keys are removed from the system. We, as developers, balance that out by retaining the expired key for as long as we have to, but still immediately remove the key from advertisement.

The alternative approaches, i.e. continue rotating the key or let the key run out its lifespan, are surprising results, because now the operator has to go in and make manual changes to _actually_ remove the key from the system.

Or do you mean more mechanically, i.e. if we can sweep unconfigured keys in the same set of logic as configured keys?

branlwyd: > I disagree that we should handle unconfigured ciphersuites like we do configured ones. IMO if we encounter an unconfigured ciphersuite we should swiftly phase it out. This strikes me as the least surprising behavior.
>
> That is, as an operator, the least surprising result of removing a ciphersuite from the configuration would be that the respective keys are removed from the system. We, as developers, balance that out by retaining the expired key for as long as we have to, but still immediately remove the key from advertisement.

OK -- I don't object to expiring keys for quickly, I think that can be done without making the policy too complicated. (N.B. the suggested policy would immediately expire any active keys for an unconfigured policy.)

I do think we want to let expired keys for unconfigured ciphersuites time out before deletion, but I think we agree on that point.

> Or do you mean more mechanically, i.e. if we can sweep unconfigured keys in the same set of logic as configured keys?

Yes, this would be nice. The current logic has many special cases; while we do need (effectively) a `ciphersuite_configured` signal, I think it should basically feed into the same set of steps used by configured ciphersuites. Basically I want to avoid having two different codepaths for configured vs unconfigured ciphersuites.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,911 @@
...
+    /// Errors on general datastore errors.
+    ///
+    /// This is not permissive of [`HpkeConfigId`] space exhaustion, i.e. if there are more rows
+    /// than what fit into a `u8`, this process will error. To avoid this, keep the list of managed
+    /// ciphersuites small, and the expiration duration less than the key age.
+    #[tracing::instrument(err)]
+    pub async fn run(&self) -> Result<(), Error> {
+        self.datastore
+            .run_tx("global_hpke_key_rotator", |tx| {
+                let configs = self.hpke.clone();
```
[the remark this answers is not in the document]
inahga: Can you elaborate?
```
let configs = Arc::new(self.hpke.clone())
```
doesn't strike me as cheaper.

inahga:
I still need to finagle migrations and test this in kind, but it's ready once again for a deep review. @divergentdave this underwent significant changes since your review, so it might be worth a re-review.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,951 @@
...
+                match keypair.state() {
+                    HpkeKeyState::Active => match ciphersuite {
+                        Some(ciphersuite) => {
+                            let keypairs: Vec<_> = self
+                                .keypairs
+                                .values()
+                                .filter(|keypair| keypair.ciphersuite() == *ciphersuite)
+                                .collect();
+
+                            let pending_key_ready = keypairs.iter().any(|keypair| {
```
branlwyd, line 244: This is pretty fragile: it needs to match the logic in the `Pending` case. You got it right (I think), but later changes could easily break this, leading to us expiring all active keys (or, less seriously, advertising two active keys for a given ciphersuite at a time).

In general, the operation for each keypair is computed entirely independently of each other keypair, so "lockstep" changes (such as promoting a pending keypair to active & simultaneously demoting the old active keypair to expired) have to "agree" in the logic of their processing. [1]

I suggest an "imperative" implementation strategy like the following. First, split the stored keypairs by ciphersuite so that you have a list of keypairs per ciphersuite, and process each ciphersuite's keypairs indepdendently. For each ciphersuite's keypairs:
1) Clone the list of keypairs (or perhaps map it into `(id, state, last_state_change_at)`).
2) Apply each policy rule iteratively, implemented by actually updating the in-memory state of the keypairs from the previous step. (An order that would work would be: promote old-enough pending to active -> demote old active to expired -> delete old expired; if implemented properly, I think we'd get the correct behavior for non-configured ciphersuites without needing special steps for that case.) This way, each update would be apparent to the later steps.
3) Diff the in-memory state against the original state to determine which operations to perform against the DB.
4) Finally, perform the DB operations.

I think this implementation strategy, other than being more obviously correct, would also be easier to follow & simpler to write down. WDYT? (If you disagree, definitely push back -- you've been staring at this problem longer than I have at this point, so if you think this is a bad idea, maybe it is.)

[1] Well, they could look at the operation computed for the other keypairs, but IMO this is more confusing than the suggested imperative update method; and this would also depend on the order than the keypairs are processed in.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,951 @@
...
+    }
+
+    /// Rotation policy:
+    ///   - If the key has been in [`HpkeKeyState::Expired`] for at least
+    ///     [`HpkeKeyRotatorConfig::expired_duration`], delete it.
+    ///   - If the key has a ciphersuite that is not known, safely phase it out. If the key is
+    ///     in [`HpkeKeyState::Pending`], delete it immediately. If it is in
+    ///     [`HpkeKeyState::Active`], move it to expired.
+    ///   - For each configured ciphersuite, move the latest pending key to active if it has been
+    ///     pending for at least [`HpkeKeyRotatorConfig::pending_duration`]. Delete other pending
```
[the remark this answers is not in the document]
inahga: > though this isn't guaranteed if an operator manually adds one.

Indeed that's the case I'm trying to cover. If the operator is adding a new pending key, and there is already one present, both could be promoted at the same time, and one of them will be immediately expired. Which one is expired is indeterminate, since they'll have the same update timestamp--or rather there's no guarantee that it'll be the one that the operator didn't insert.

Note the window for the "both promoted at the same time" case is not gated by `pending_duration`, but by how often the key rotator is run. If it's run once per day/week/etc., the window for this case during a manual key rotation is fairly large.

I figured this was an easy enough logic add-on to simplify manual key rotation.

aggregator/src/aggregator/key_rotator.rs
inahga: It's a more-than-plausible idea, but I'm not sure if there's enough benefit to justify additional work to iterate on this again, since the changes proposed seem rather sweeping.

Is the only problem with this code the duplicated logic between Pending and Active? If so, I think it can be factored out into a small function.

> but later changes could easily break this, leading to us expiring all active keys (or, less seriously, advertising two active keys for a given ciphersuite at a time).

Unit tests ought to catch any serious regressions like this--I think they should be covering this arm pretty well. LMK if they don't, though.

(I'd like if the defensive assertion could validate one active keypair per ciphersuite, but the bootstrap case confounds that since there could be a pending key but no active key.)

> I think we'd get the correct behavior for non-configured ciphersuites without needing special steps for that case.

I don't see how, since the policy for unconfigured ciphersuites is explicitly different than configured ones, i.e. immediate phase out without replacement. If we were treating the two the same, then yes I think this solution would be good for that.

branlwyd (approved):

aggregator_core/src/datastore.rs
```diff
@@ -4958,6 +4967,61 @@ DELETE FROM global_hpke_keys WHERE config_id = $1",
...
+                self.execute(
+                    &stmt,
+                    &[
+                        /* config_id */ &hpke_config_id,
+                        /* config */ &hpke_config,
+                        /* private_key */ &encrypted_hpke_private_key,
+                        /* state */ keypair.state(),
+                        /* last_state_change_at */
+                        &keypair.last_state_change_at().as_naive_date_time()?,
+                        /* created_at */ &now,
```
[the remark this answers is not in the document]
inahga: Good point.

I went with changing `rewrite_global_hpke_keypairs` to upserting rows, so that `created_at` touches as little application code as possible.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,948 @@
...
+                                    None
+                                }
+                            } else {
+                                info!(
+                                    id = ?keypair.id(),
+                                    ?ciphersuite,
+                                    "deleting extraneous pending keypair"
+                                );
+                                Some(Op::Delete(id))
+                            }
```
divergentdave, line 338: The following pathological case occurs to me. Suppose we have, all with the same ciphersuite, one active keypair that is old enough to be expired, and two pending keypairs, one that is old enough to be promoted, and one that is not. The logic under the `Active` arm will see that there's a pending keypair that's ready, and thus expire the active one. The logic under the `Pending` arm, however, will not promote either pending keypair to active. The older pending keypair, which is ready for promotion, will be deleted because it is not the latest pending keypair. The newer pending keypair, which is not ready for promotion, will have no action taken on it.

I think this bug arose when transforming the logic to consider one keypair at a time. Ideally the expiration and promotion should either both happen or not, but considering them separately created a hole because the preconditions weren't quite the same.

There are a few ways we could fix this: 
- Changing the criteria for which pending keypair we delete
- Writing more complicated conditions for promotion and expiration, so that either both happens or neither happens
- Making processing two-pass, so that we can choose whether to expire an active key based on whether we actually have promoted a pending key
- Replacing the `keypairs.iter.filter_map()` loop, and pushing multiple operations onto the `Vec` together, so we don't get out of sync.

We could also improve tests to catch this sort of thing by using `quickcheck` or similar to generate the database's initial state and the key rotator's configuration, then asserting various correctness properties about the initial and final states.

docs/samples/advanced_config/key_rotator.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+  # Path to a PEM file with root certificates to trust for TLS database
+  # connections. If present, TLS may be used depending on the "sslmode"
+  # connection string parameter, and the server's support for TLS. If absent,
+  # TLS will never be used. (optional)
+  tls_trust_store_path: /path/to/file.pem
+
+key_rotator:
+  # Rotation policy for global HPKE keys. At least one is required. Each entry
+  # represents a key with a particular ciphersuite.
```
divergentdave, line 23: I think the second and third sentences here now belong on the `ciphersuite` key below

aggregator/src/aggregator/key_rotator.rs
inahga: Nice find. That lends a lot of credence to @branlwyd point about precondition mismatch being a possible source of bugs.

inahga: I've solved this by more or less adopting @branlwyd's strategy of handling each ciphersuite independently. This allows us to make the state changes in lockstep and based on the same preconditions.

inahga: I'm still looking into the testing strategy you suggested, but the logic is ready for another pass.

aggregator/src/aggregator/key_rotator.rs
inahga: I've adopted your strategy after all, following @divergentdave noticing a bug related to preconditions not matching. Let me know if this is closer to what you're looking for.

Note that this implementation never writes to the in-memory state, instead we derive a list of `Op`s from the current table state, and apply them to the database. I think if we _can_ do it this way, we should, since I think the logic is easier to follow without having to consider intermediate state, i.e. we can think of `sweep()` as the "operations necessary to get the table to the desired state, given the initial state".

This implementation also turns out to be less indentation-riffic.

inahga: > Note that this implementation never writes to the in-memory state, instead we derive a list of Ops from the current table state, and apply them to the database.

OK, well I might have to back out that change if adopting property based testing suggested by @divergentdave. To have tests complete in a reasonable amount of time, we'll have to check in-memory state rather than taking obnoxious amounts of database roundtrips.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,922 @@
...
+
+                        if let Some(latest_active_key) = latest_active_key {
+                            ops.push(HpkeOp::Update(
+                                *latest_active_key.id(),
+                                HpkeKeyState::Expired,
+                                "pending key ready",
+                            ));
+                        }
+                    }
+                } else if latest_pending_key.is_none()
```
branlwyd, line 285: nit: the `latest_pending_key.is_none()` check is extraneous, as we can only reach this block if the `if let Some(latest_pending_key) = latest_pending_key` above fails.

branlwyd (approved):

aggregator/src/aggregator/key_rotator.rs
branlwyd: (If it's written this way for clarity, replace with a comment?)

divergentdave (approved):

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,922 @@
...
+                                entry.partition_point(|&candidate| {
+                                    candidate.last_state_change_at()
+                                        > keypair.last_state_change_at()
+                                }),
+                                keypair,
+                            )
+                        })
+                        .or_insert_with(|| VecDeque::from([keypair]));
+                })
+                .or_insert_with(|| HashMap::from([(*keypair.state(), VecDeque::from([keypair]))]));
```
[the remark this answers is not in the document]
inahga: Nice, that's a massive improvement.
```rust
        for keypair in self.keypairs.values() {
            let by_state = by_ciphersuite.entry(keypair.ciphersuite()).or_default();
            let keypairs = by_state.entry(*keypair.state()).or_default();
            keypairs.insert(
                keypairs.partition_point(|&candidate| {
                    candidate.last_state_change_at() > keypair.last_state_change_at()
                }),
                keypair,
            );
        }
```

aggregator/src/lib.rs
```diff
@@ -12,6 +12,12 @@ pub mod diagnostic;
 pub mod metrics;
 pub mod trace;
 
+#[cfg(test)]
+extern crate quickcheck;
```
branlwyd, line 16: `extern crate` is pretty archaic; can we `use quickcheck` instead?

docs/samples/advanced_config/key_rotator.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+key_rotator:
+  # Rotation policy for global HPKE keys.
+  hpke:
+    # How long keys remains pending before they're promoted to active. Should
+    # be greater than the global HPKE keypair cache refresh rate. Defaults to
+    # 1 hour.
+    pending_duration_s: 3600
+
+    # The TTL of keys. Defaults to 4 weeks.
```
branlwyd, line 29:
```suggestion
    # The TTL of keys. Defaults to 12 weeks.
```

branlwyd: (nit, I believe this matches what is written in code)

docs/samples/basic_config/key_rotator.yaml
```diff
@@ -0,0 +1,7 @@
+database:
+  # Database URL. (required)
+  url: "postgres://postgres:postgres@localhost:5432/postgres"
+
+key_rotator:
+  # Rotation policy for global HPKE keys
+  hpke: {}
```
branlwyd, line 7: nit: I think this config is so simple because it is attempting to be as minimal as possible, but I think this would effectively just remove any existing keys.

Maybe specify only `ciphersuite`, with a common choice of configuration values? (`P256HkdfSha256`, `HkdfSha256`, `Aes128Gcm`, perhaps?)

branlwyd (approved):

docs/samples/basic_config/key_rotator.yaml
inahga: An empty `ciphersuite` value should default to one with a reasonable ciphersuite.
```rust
fn default_hpke_ciphersuites() -> HashSet<HpkeCiphersuite> {
    HashSet::from([HpkeCiphersuite::new(
        HpkeKemId::X25519HkdfSha256,
        HpkeKdfId::HkdfSha256,
        HpkeAeadId::Aes128Gcm,
    )])
}
```

At another look I'm not happy with the test coverage on this though, so I'll at minimum add a test.

aggregator/src/lib.rs
inahga: Ah, we totally can. I copypasta'd from the examples without being critical about it.

divergentdave (approved):

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -0,0 +1,919 @@
...
+            .unwrap();
+
+        for ciphersuite in ciphersuites_to_insert {
+            let inserted: Vec<_> = key_rotator
+                .keypairs
+                .iter()
+                .filter(|(_, keypair)| &keypair.ciphersuite() == ciphersuite)
+                .collect();
+            if inserted.len() != 1 {
+                return TestResult::error("more than one key inserted for new ciphersuite");
```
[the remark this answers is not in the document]
inahga: Clarify why? `TestResult::error()` looks the same as `TestResult::failed()` except it sets a friendly error message https://docs.rs/quickcheck/latest/src/quickcheck/tester.rs.html#216. That way if a test fails it points us directly to the failing line, rather than having to run with RUST_LOG.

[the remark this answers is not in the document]
inahga: In both `error()` and `failed()` cases, we're presented with the input variables, only in `error()` we're given an additional line `Error: expired key was not deleted`.

e.g.
```
# cargo test hpke_key_rotator_expired_keys
   Compiling janus_core v0.7.19 (/home/inahga/Projects/janus/core)
   Compiling janus_aggregator_core v0.7.19 (/home/inahga/Projects/janus/aggregator_core)
   Compiling janus_aggregator_api v0.7.19 (/home/inahga/Projects/janus/aggregator_api)
   Compiling janus_aggregator v0.7.19 (/home/inahga/Projects/janus/aggregator)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 58.37s
     Running unittests src/lib.rs (/home/inahga/Projects/janus/target/debug/deps/janus_aggregator-4c9e31edbd742197)

running 1 test
test aggregator::key_rotator::tests::hpke_key_rotator_expired_keys ... FAILED

failures:

---- aggregator::key_rotator::tests::hpke_key_rotator_expired_keys stdout ----
thread 'aggregator::key_rotator::tests::hpke_key_rotator_expired_keys' panicked at /home/inahga/.cargo/registry/src/index.crates.io-6f17d22bba15001f/quickcheck-1.0.3/src/tester.rs:165:28:
[quickcheck] TEST FAILED (runtime error). Arguments: [the arguments, really long, somewhat justifying fixing `Arbitrary::shrink`.]
Error: expired key was not deleted


failures:
    aggregator::key_rotator::tests::hpke_key_rotator_expired_keys

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 193 filtered out; finished in 0.05s
```

PR #3138, opened 2024-05-16
Parallelize Leader aggregation computations.

branlwyd:
This implements for the Leader what #3119 did for the Helper. Like with the Helper, a rayon threadpool is used to parallelize aggregation computations. These computations respect cancellation (though I'm not sure if anything short of process death will currently cancel the aggregation_job_driver's computations).


branlwyd:
Closes #3117.
Closes #3035.

branlwyd:
(This isn't time-critical as we are the Helper in the integration that led to us parallelizing these computations, but it might be nice to parallelize all aggregation computations in a single release.)

divergentdave (approved):

tgeoghegan (approved):

Issue #3142, opened 2024-05-17
Helper: More than two aggregation jobs containing the same report cause 500 error.


inahga:
If two aggregation jobs come in that contain the same report, one will prepare the report successfully, and the second will reject the report, as expected.

If a third comes in, the result is a 500 error. The problem is in the function `check_other_report_aggregation_exists()` which does this query in a `query_opt()` call.
```
                "-- check_other_report_aggregation_exists()
SELECT 1 FROM report_aggregations
JOIN aggregation_jobs ON aggregation_jobs.id = report_aggregations.aggregation_job_id
WHERE report_aggregations.task_id = $1
  AND aggregation_jobs.task_id = $1
  AND report_aggregations.client_report_id = $2
  AND aggregation_jobs.aggregation_param = $3
  AND aggregation_jobs.aggregation_job_id != $4
  AND UPPER(aggregation_jobs.client_timestamp_interval) >= $5",
```
If there are multiple report aggregations, this query returns more than 1 row, which returns an error from `query_opt()` which propagates up to a 500 error.

branlwyd:
The OP is true; but more than that, if the aggregation jobs containing the replayed report arrived concurrently, Janus would have aggregated both. This is potentially a privacy bug. The fix for both of these issues is tracked in #3143.

PR #3143, opened 2024-05-17
Fix Helper aggregation-initialization report-replayed check.

branlwyd:
Specifically, we now check only if the corresponding row exists in the `client_reports` table, rather than ignoring that fact and looking for a corresponding `report_aggregations` row. This means we no longer respect whether the other aggregation used a different aggregation parameter, but that's fine -- Janus now supports only a single aggregation per report.

This fixes a few bugs:
* If the same report was repeatedly submitted in different aggregation jobs, the first instance would be aggregated; the second instance would be marked as a replay; and the third and further instances would result in an HTTP 500 error, as `check_other_report_aggregation_exists` mistakenly expected at most a single matching `report_aggregations` row to exist.
* If the same report was submitted in more than one aggregation job concurrently, Janus would have aggregated both. This was a privacy bug.


branlwyd:
Closes #3142.

aggregator/src/aggregator.rs
```diff
@@ -2336,33 +2336,18 @@ impl VdafOps {
...
-                                );
-                            let (_, report_aggregation_exists) =
-                                try_join!(put_report_share_fut, report_aggregation_exists_fut)?;
-
-                            if report_aggregation_exists {
+                            let report_already_aggregated =
+                                match tx.put_scrubbed_report(task.id(), &rsd.report_share).await {
+                                    Ok(()) => false,
+                                    Err(datastore::Error::MutationTargetAlreadyExists) => true,
+                                    Err(err) => Err(err)?,
```
divergentdave, line 2347: I think clippy prefers this
```suggestion
                                    Err(err) => return Err(err),
```

aggregator/src/aggregator/http_handlers/tests/aggregation_job_init.rs
```diff
@@ -354,29 +354,21 @@ async fn aggregate_init() {
...
             let task = helper_task.clone();
             let report_share_4 = prepare_init_4.report_share().clone();
-            let report_share_8 = prepare_init_8.report_share().clone();
 
             Box::pin(async move {
                 tx.put_aggregator_task(&task).await.unwrap();
 
-                // report_share_4 and report_share_8 are already in the datastore as they were
+                // report_share_4 are already in the datastore as they were
                 // referenced by existing aggregation jobs.
```
divergentdave, line 368: nit: grammar cleanup
```suggestion
                // report_share_4 is already in the datastore as it was
                // referenced by an existing aggregation job.
```

divergentdave (approved):

aggregator/src/aggregator.rs
branlwyd: Clippy is happy enough either way, but I think the explicit `return` makes it clearer no type-coercion magic is happening & (more obviously) that an early return is definitely happening.

tgeoghegan (approved):

PR #3158, opened 2024-05-21
client: move tests to distinct module

tgeoghegan:
Moves the `janus_client::tests` module into a separate file.

As we have found elsewhere, putting tests into separate files improves compilaton speed ([citation needed]), and makes it easier to navigate sources.


tgeoghegan:
I'd like to land this noisy change ahead of adding OHTTP support, which will make substantial changes to the tests.

divergentdave (approved):

PR #3161, opened 2024-05-23
`janus_client`: Uploads over OHTTP

tgeoghegan:
Adds support for uploading reports to the Leader via OHTTP.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -38,8 +38,12 @@
 //! ```
 
 use backoff::ExponentialBackoff;
+#[cfg(feature = "ohttp")]
+use bhttp::{ControlData, Message, Mode};
 use derivative::Derivative;
-use http::header::CONTENT_TYPE;
+#[cfg(feature = "ohttp")]
+use http::{header::ACCEPT, HeaderValue};
+use http::{header::CONTENT_TYPE, StatusCode};
 use itertools::Itertools;
 use janus_core::{
     hpke::{self, is_hpke_config_supported, HpkeApplicationInfo, Label},
@@ -52,13 +56,16 @@ use janus_messages::{
     Duration, HpkeConfig, HpkeConfigList, InputShareAad, PlaintextInputShare, Report, ReportId,
     ReportMetadata, Role, TaskId, Time,
 };
+#[cfg(feature = "ohttp")]
+use ohttp::{ClientRequest, KeyConfig};
 use prio::{
     codec::{Decode, Encode},
     vdaf,
 };
 use rand::random;
+#[cfg(feature = "ohttp")]
+use std::io::Cursor;
 use std::{convert::Infallible, fmt::Debug, time::SystemTimeError};
-use tokio::try_join;
 use url::Url;
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


tgeoghegan:
Still a draft; I want to add tests in `janus_client` that use `mockito` to simulate an OHTTP gateway.

inahga (commented):
Consider locking this behind a rust feature `ohttp`, so folks who use this crate but have no need for OHTTP can leave the feature off to keep their transitive dependency tree smaller.

I understand this might be trickier than it's immediately worth though, and probably complicates testing setup, so I'm fine taking it as a future improvement.

tgeoghegan:
I think that's a good idea. It shouldn't be too hard to do and in any case will only get harder the longer we wait to do it.

client/src/lib.rs
```diff
@@ -476,28 +662,127 @@ impl<V: vdaf::Client<16>> Client<V> {
...
+        &self,
+        upload_endpoint: &Url,
+        request_body: &[u8],
+    ) -> Result<StatusCode, Error> {
+        let (ohttp_config, key_configs) =
+            if let Some((ohttp_config, key_configs)) = &self.ohttp_config {
+                (ohttp_config, key_configs)
+            } else {
+                return self.put_report(upload_endpoint, request_body).await;
+            };
```
inahga, line 714: IIUC if `ohttp_config` is None, then we fallback to normal uploads? This strikes me as a footgun for an accidental privacy leak if the client is used improperly.

I understand this as allowing non-OHTTP uploads with the `ohttp` feature enabled`. The point above about having a separate suite of methods would resolve this.

client/src/lib.rs
```diff
@@ -258,11 +347,32 @@ impl<V: vdaf::Client<16>> ClientBuilder<V> {
...
+            Some(config) => config,
+            None => aggregator_hpke_config(&self.parameters, &Role::Leader, &http_client).await?,
+        };
+        let helper_hpke_config = match self.helper_hpke_config {
+            Some(config) => config,
+            None => aggregator_hpke_config(&self.parameters, &Role::Helper, &http_client).await?,
+        };
+
+        #[cfg(feature = "ohttp")]
+        let ohttp_config = if let Some(ohttp_config) = self.ohttp_config {
```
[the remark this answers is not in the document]
tgeoghegan: I'll give this a try.

client/src/lib.rs
tgeoghegan: I don't see it as a fallback. If the client didn't ask for OHTTP, we don't do OHTTP.

You're saying it should be impossible to do non-OHTTP uploads if feature `ohttp` is enabled? I don't agree. Case in point: we use `janus_client` in our own integration tests and we want to be able to exercise both OHTTP and non-OHTTP uploads without recompiling.

inahga: I see, I think my problem then is moreso code organization--it raises my alarm bells that we could do non-OHTTP upload in a function called `upload_with_ohttp`. Maybe hoisting this logic up to `upload_with_time()` would make it clearer.

Can take it or leave it, I don't feel strongly about it now that I know the intent.

> You're saying it should be impossible to do non-OHTTP uploads if feature ohttp is enabled?

Not really, I'm moreso looking for explicitness about whether an upload is being performed with OHTTP or not. Which, eh, I think `.with_ohttp_config()` in the client builder is sufficient, so no action necessary.

client/src/lib.rs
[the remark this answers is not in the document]
tgeoghegan: Done -- I tried adding the OHTTP key config future to the `try_join!`, but I can't work out how to add an extra future conditionally depending on a cargo feature, and I don't think it's worth the hassle of the gymnastics required to homogenize everything into a `Box<Future<Output = whatever>>` so it could go into `try_join_all`.

client/src/lib.rs
tgeoghegan: This is the least stupid way I found to do this. It's very awkward to work around the `ohttp` feature. I added a comment to this method explaining that it might not do OHTTP, but since it's not a public item, I'm less concerned about misuse.

inahga (approved):

Issue #3163, opened 2024-05-24
Accept repeated, duplicate `Authorization` headers as if they were one header


branlwyd:
In one of our deployments, we are observing repeated, duplicate `Authorization` headers. Due to the implementation details of Janus, this is treated as no `Authorization` header at all, leading to the request being treated as unauthorized.

In the short term, we should treat repeated, duplicate `Authorization` & `DAP-Auth-Token` headers as if only one instance of the header was sent. Repeated, non-duplicate authorization headers should cause the request to fail.

In the long term, we should remove the above change once we are no longer observing repeated headers. The HTTP spec does not allow the `Authorization` header to be duplicated. Per [RFC7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2):

```
A sender MUST NOT generate multiple header fields with the same field
name in a message unless either the entire field value for that
header field is defined as a comma-separated list [i.e., #(values)]
or the header field is a well-known exception (as noted below).
```

(and the `Authorization` header does not accept a comma-delimited list of values, nor is it a well-known exception)

tgeoghegan:
For the meager amount it's worth, the RFC 7230 reference is obsoleted by [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-order).

PR #3164, opened 2024-05-24
Accept repeated, duplicate Authorization headers as one.

branlwyd:
This is not to the RFC, but we are observing this behavior in a deployment.


branlwyd:
Part of #3163.

divergentdave (approved):

inahga (approved):

PR #3180, opened 2024-05-29
Improve efficiency of collection-artifact GC.

branlwyd:
Previously, we would need to scan the batch_aggregations table to fulfill batches_to_delete, which was especially painful if there were not enough batches to delete because we would have to scan the entire table to determine that.

Now, we do a quick first pass to determine candidate batches to delete, with the aid of a new index.

### DIFF
--- a/aggregator/src/aggregator/http_handlers/tests/aggregation_job_continue.rs
+++ b/aggregator/src/aggregator/http_handlers/tests/aggregation_job_continue.rs
@@ -390,7 +390,7 @@ async fn aggregate_continue_accumulate_batch_aggregation() {
     let hpke_key = helper_task.current_hpke_key();
     let measurement = IdpfInput::from_bools(&[true]);
     let aggregation_param =
-        Poplar1AggregationParam::try_from_prefixes(vec![measurement.clone()]).unwrap();
+        Poplar1AggregationParam::try_from_prefixes(Vec::from([measurement.clone()])).unwrap();
 
     // report_share_0 is a "happy path" report.
     let report_time_0 = first_batch_interval_clock
@@ -661,12 +661,10 @@ async fn aggregate_continue_accumulate_batch_aggregation() {
     // Map the batch aggregation ordinal value to 0, as it may vary due to sharding.
     let first_batch_got_batch_aggregations = datastore
         .run_unnamed_tx(|tx| {
-            let (task, vdaf, report_metadata_0, aggregation_param) = (
-                helper_task.clone(),
-                vdaf.clone(),
-                report_metadata_0.clone(),
-                aggregation_param.clone(),
-            );
+            let task = helper_task.clone();
+            let vdaf = vdaf.clone();
+            let aggregation_param = aggregation_param.clone();
+
             Box::pin(async move {
                 Ok(merge_batch_aggregations_by_batch(
                     TimeInterval::get_batch_aggregations_for_collection_identifier::<
@@ -678,14 +676,7 @@ async fn aggregate_continue_accumulate_batch_aggregation() {
                         task.id(),
                         task.time_precision(),
                         &vdaf,
-                        &Interval::new(
-                            report_metadata_0
-                                .time()
-                                .to_batch_interval_start(task.time_precision())
-                                .unwrap(),
-                            *task.time_precision(),
-                        )
-                        .unwrap(),
+                        &first_batch_identifier,
                         &aggregation_param,
                     )
                     .await
@@ -728,33 +719,28 @@ async fn aggregate_continue_accumulate_batch_aggregation() {
 
     let second_batch_got_batch_aggregations = datastore
         .run_unnamed_tx(|tx| {
-            let (task, vdaf, report_metadata_2, aggregation_param) = (
-                helper_task.clone(),
-                vdaf.clone(),
-                report_metadata_2.clone(),
-                aggregation_param.clone(),
-            );
+            let task = helper_task.clone();
+            let vdaf = vdaf.clone();
+            let aggregation_param = aggregation_param.clone();
+
             Box::pin(async move {
-                TimeInterval::get_batch_aggregations_for_collection_identifier::<
-                    VERIFY_KEY_LENGTH,
-                    Poplar1<XofTurboShake128, 16>,
-                    _,
-                >(
-                    tx,
-                    task.id(),
-                    task.time_precision(),
-                    &vdaf,
-                    &Interval::new(
-                        report_metadata_2
-                            .time()
-                            .to_batch_interval_start(task.time_precision())
-                            .unwrap(),
-                        Duration::from_seconds(task.time_precision().as_seconds()),
+                let mut got_batch_aggregations =
+                    TimeInterval::get_batch_aggregations_for_collection_identifier::<
+                        VERIFY_KEY_LENGTH,
+                        Poplar1<XofTurboShake128, 16>,
+                        _,
+                    >(
+                        tx,
+                        task.id(),
+                        task.time_precision(),
+                        &vdaf,
+                        &second_batch_identifier,
+                        &aggregation_param,
                     )
-                    .unwrap(),
-                    &aggregation_param,
-                )
-                .await
+                    .await
+                    .unwrap();
+                got_batch_aggregations.sort_unstable_by_key(|ba| ba.ord());
+                Ok(got_batch_aggregations)
             })
         })
         .await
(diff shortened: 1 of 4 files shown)


branlwyd:
Old:

```
EXPLAIN ANALYZE SELECT batch_identifier, aggregation_param
FROM batch_aggregations
WHERE task_id = 2
GROUP BY batch_identifier, aggregation_param
HAVING MAX(UPPER(COALESCE(batch_interval, client_timestamp_interval))) < NOW() - interval '1 month'
LIMIT 50;
```

```
                                                                                                      QUERY PLAN                                                                                                      
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.69..69734.84 rows=50 width=34) (actual time=352691.264..352691.266 rows=0 loops=1)
   ->  GroupAggregate  (cost=0.69..27988497.64 rows=20068 width=34) (actual time=352691.263..352691.264 rows=0 loops=1)
         Group Key: batch_identifier, aggregation_param
         Filter: (max(upper(COALESCE(batch_interval, client_timestamp_interval))) < (now() - '1 mon'::interval))
         Rows Removed by Filter: 252020
         ->  Index Scan using batch_aggregations_unique_task_id_batch_id_aggregation_param on batch_aggregations  (cost=0.69..27910230.67 rows=7721341 width=72) (actual time=1.668..346861.215 rows=8064609 loops=1)
               Index Cond: (task_id = 2)
 Planning Time: 0.204 ms
 Execution Time: 352691.392 ms
(9 rows)
```

New:

```
EXPLAIN ANALYZE WITH candidate_batches_to_delete AS (
    SELECT DISTINCT batch_identifier, aggregation_param
    FROM batch_aggregations
    WHERE task_id = 2
      AND UPPER(COALESCE(batch_interval, client_timestamp_interval)) < NOW() - interval '1 month'
)
SELECT ba.batch_identifier, ba.aggregation_param
FROM batch_aggregations AS ba
JOIN candidate_batches_to_delete AS candidate
  ON ba.batch_identifier = candidate.batch_identifier
 AND ba.aggregation_param = candidate.aggregation_param
WHERE ba.task_id = 2
GROUP BY ba.batch_identifier, ba.aggregation_param
HAVING MAX(UPPER(COALESCE(ba.batch_interval, ba.client_timestamp_interval))) < NOW() - interval '1 month'
LIMIT 50
```

```
                                                                                    QUERY PLAN                                                                                    
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=105.89..12961.81 rows=42 width=34) (actual time=1.901..1.905 rows=0 loops=1)
   ->  GroupAggregate  (cost=105.89..12961.81 rows=42 width=34) (actual time=1.899..1.902 rows=0 loops=1)
         Group Key: ba.batch_identifier, ba.aggregation_param
         Filter: (max(upper(COALESCE(ba.batch_interval, ba.client_timestamp_interval))) < (now() - '1 mon'::interval))
         ->  Nested Loop  (cost=105.89..12958.37 rows=125 width=72) (actual time=1.895..1.898 rows=0 loops=1)
               ->  Unique  (cost=105.20..105.38 rows=25 width=34) (actual time=1.894..1.895 rows=0 loops=1)
                     ->  Sort  (cost=105.20..105.26 rows=25 width=34) (actual time=1.890..1.891 rows=0 loops=1)
                           Sort Key: batch_aggregations.batch_identifier, batch_aggregations.aggregation_param
                           Sort Method: quicksort  Memory: 25kB
                           ->  Index Scan using batch_aggregations_gc_time on batch_aggregations  (cost=0.57..104.61 rows=25 width=34) (actual time=1.849..1.850 rows=0 loops=1)
                                 Index Cond: ((task_id = 2) AND (upper(COALESCE(batch_interval, client_timestamp_interval)) < (now() - '1 mon'::interval)))
               ->  Index Scan using batch_aggregations_unique_task_id_batch_id_aggregation_param on batch_aggregations ba  (cost=0.69..512.86 rows=125 width=72) (never executed)
                     Index Cond: ((task_id = 2) AND (batch_identifier = batch_aggregations.batch_identifier) AND (aggregation_param = batch_aggregations.aggregation_param))
 Planning Time: 7.280 ms
 Execution Time: 2.053 ms
```

branlwyd:
Unexpected test failure, back to draft.

divergentdave:
These changes look reasonable. How does the new query and index perform if GC is falling behind, and there are a lot of batches that are eligible according to their timestamp intervals? I think we could test this by replacing `interval '1 month'` with a smaller value. I'm curious if the sort node in the plan will go away if the index scan below it is expected to return many rows, or if its cost will blow up.

branlwyd:
Good point -- the new query does indeed increase in processing time based on the number of batches available. Looking to the same "old" & "new" queries, the plans for 2 weeks are as follows.

Old:

```                                                                                                                                                                                                       QUERY PLAN                                                                                                  
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.69..71802.24 rows=50 width=34) (actual time=2.857..4.045 rows=50 loops=1)
   ->  GroupAggregate  (cost=0.69..29615267.68 rows=20623 width=34) (actual time=2.856..4.036 rows=50 loops=1)
         Group Key: batch_identifier, aggregation_param
         Filter: (max(upper(COALESCE(batch_interval, client_timestamp_interval))) < (now() - '14 days'::interval))
         ->  Index Scan using batch_aggregations_unique_task_id_batch_id_aggregation_param on batch_aggregations  (cost=0.69..29532447.08 rows=8173791 width=72) (actual time=2.782..3.293 rows=1601 loops=1)
               Index Cond: (task_id = 2)
 Planning Time: 0.185 ms
 Execution Time: 4.083 ms
```

New:

```
QUERY PLAN                                                                                                  
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=141952.27..251292.29 rows=50 width=34) (actual time=124.738..126.623 rows=50 loops=1)
   ->  GroupAggregate  (cost=141952.27..14091552.06 rows=6379 width=34) (actual time=124.736..126.613 rows=50 loops=1)
         Group Key: ba.batch_identifier, ba.aggregation_param
         Filter: (max(upper(COALESCE(ba.batch_interval, ba.client_timestamp_interval))) < (now() - '14 days'::interval))
         ->  Nested Loop  (cost=141952.27..14091025.79 rows=19137 width=72) (actual time=124.671..126.080 rows=1601 loops=1)
               ->  Unique  (cost=141951.58..142244.57 rows=28970 width=34) (actual time=124.619..124.651 rows=51 loops=1)
                     ->  Sort  (cost=141951.58..142049.24 rows=39066 width=34) (actual time=124.616..124.628 rows=51 loops=1)
                           Sort Key: batch_aggregations.batch_identifier, batch_aggregations.aggregation_param
                           Sort Method: external merge  Disk: 2016kB
                           ->  Bitmap Heap Scan on batch_aggregations  (cost=537.00..138972.09 rows=39066 width=34) (actual time=16.738..91.727 rows=46571 loops=1)
                                 Recheck Cond: ((task_id = 2) AND (upper(COALESCE(batch_interval, client_timestamp_interval)) < (now() - '14 days'::interval)))
                                 Heap Blocks: exact=35742
                                 ->  Bitmap Index Scan on batch_aggregations_gc_time  (cost=0.00..527.23 rows=39066 width=0) (actual time=9.481..9.482 rows=46571 loops=1)
                                       Index Cond: ((task_id = 2) AND (upper(COALESCE(batch_interval, client_timestamp_interval)) < (now() - '14 days'::interval)))
               ->  Index Scan using batch_aggregations_unique_task_id_batch_id_aggregation_param on batch_aggregations ba  (cost=0.69..480.16 rows=132 width=72) (actual time=0.009..0.019 rows=31 loops=51)
                     Index Cond: ((task_id = 2) AND (batch_identifier = batch_aggregations.batch_identifier) AND (aggregation_param = batch_aggregations.aggregation_param))
 Planning Time: 0.463 ms
 Execution Time: 127.369 ms
```

This simulates being a few days behind. N.B. I ran each query twice; both sped up, but the new query sped up a bit more (the first attempt was around 30s). I think this is OK. The obvious improvement would be to add a `LIMIT` to `candidate_batches_to_delete`, but I don't want to do that unless we need to: the risk is that we might continually load the same candidate batches, find that they can't actually be deleted, and stop deleting anything at all.

divergentdave (approved):
I figure the first run took 30s because it was reading indexes and parts of the table into the buffer pool. Keeping more of batch_aggregations in the database's working set is not ideal, but so long as GC is able to keep up still, this problem will solve itself eventually.

I think starting with this as an incremental improvement makes sense. Adding a generous LIMIT to the SELECT DISTINCT query could be useful if we still need to optimize this. Note that this could only get stuck if many batches have batch aggregations with upper timestamps straddling both sides of the GC threshold. Even if we get into such a situation, the GC threshold keeps moving forward as time passes, which makes it harder to get stuck with pathological behavior. 

I wonder if there's a way to use window functions instead of a SELECT DISTINCT subquery/CTE to express what we want in a more efficient manner. I'm not confident enough in my knowledge of window functions to say if this is doable just yet.

PR #3184, opened 2024-05-29
`integration_tests`: OHTTP report uploads

tgeoghegan:
Add an integration test which configures its `janus_client::Client`s to upload over OHTTP. The OHTTP key configs and gateway are assumed to be adjacent to the leader's DAP API. The test only runs if feature `ohttp` is enabled.


tgeoghegan:
I successfully ran the janus-ops `cloud-integration-test` workflow with this change: https://github.com/divviup/janus-ops/actions/runs/9293234844, though actually I used a tweaked version of this change that enabled the OHTTP test, since I don't want to mess with `CLOUD_INTEGRATION_TEST_CONFIG` in staging just yet.

branlwyd (approved):

PR #3186, opened 2024-06-03
Stop accepting repeated, duplicate Authorization headers.

branlwyd:
The issue that led to us accepting this behavior has been fixed. Given that this behavior is not RFC-compliant, we remove it from Janus.

### DIFF
--- a/aggregator/src/aggregator/aggregate_init_tests.rs
+++ b/aggregator/src/aggregator/aggregate_init_tests.rs
@@ -309,32 +309,10 @@ pub(crate) async fn put_aggregation_job<Q: query_type::QueryType>(
     aggregation_job: &AggregationJobInitializeReq<Q>,
     handler: &impl Handler,
 ) -> TestConn {
-    put_aggregation_job_with_auth_header_count(
-        task,
-        aggregation_job_id,
-        aggregation_job,
-        handler,
-        1,
-    )
-    .await
-}
-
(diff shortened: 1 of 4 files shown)


branlwyd:
I might wait a few more days/another release before merging this. I'm not totally convinced we are out of the woods yet (though it is likely).

Closes #3163.

tgeoghegan (approved):
This change is sound and does what it's intended to. But WDYT about instead adding a configuration parameter for accepting duplicate headers, so that we could first verify that the issue is resolved, and then only yank support once we're confident? On the other hand, I suppose rolling Janus back/forward to/from a version that accepts duplicate headers is about as difficult as changing a configuration value.

inahga (approved):

branlwyd:
I don't think this issue is worth a config value: in the long run, we don't want to support this behavior at all, we only do so now because doing so allowed us to quickly work around a production issue. A config option effectively saying "do something RFC noncompliant", which no one activates or wants to activate, is not something I'd like to keep around long-term.

I do think it's a good idea to wait a bit (one more release, in practice?) before we yank support for this, for the reason you mention. I'm also not totally confident we are out of the woods.

divergentdave (approved):

inahga:
We're OK with yanking support now.

branlwyd:
Er, were we? I think we were waiting on seeing additional traffic in the affected deployment to be sure. That said, rolling this back (in the unlikely event it is needed to do so) is not so hard, so I think I'm OK letting this roll forward.

PR #3228, opened 2024-06-11
key rotator: allow running it as part of aggregator process

inahga:
Stacked on https://github.com/divviup/janus/pull/3136. Supports https://github.com/divviup/janus/issues/2147.

Lets the key rotator be run as part of a BYOH all-in-one aggregator deployment, like we support for the garbage collector.

This is not ideal for high-scale deployments, i.e. ours, because this effectively serializes the startup of Janus due to the exclusive lock taken by the key rotator.

### DIFF
--- a/aggregator/src/binaries/aggregator.rs
+++ b/aggregator/src/binaries/aggregator.rs
@@ -104,6 +104,9 @@ async fn run_aggregator(
                 info!("Running key rotator");
                 let key_rotator = KeyRotator::new(datastore, config.hpke);
                 let mut interval = interval(Duration::from_secs(config.frequency_s));
+                // Note that `interval` fires immediately at first, so the key rotator runs
+                // immediately on boot. This takes care of bootstrapping keys on the first run of
+                // Janus.
                 while stopper.stop_future(interval.tick()).await.is_some() {
                     if let Err(err) = key_rotator.run().await {
                         error!(?err, "key rotator error");
--- a/aggregator/src/binaries/aggregator.rs
+++ b/aggregator/src/binaries/aggregator.rs
@@ -89,6 +89,7 @@ async fn run_aggregator(
         let stopper = stopper.clone();
         async move {
             if let Some(gc_config) = gc_config {
+                info!("Running garbage collector");
                 run_garbage_collector(datastore, gc_config, meter, stopper).await;
             }
         }
@@ -100,6 +101,7 @@ async fn run_aggregator(
         let stopper = stopper.clone();
         async move {
             if let Some(config) = config {
+                info!("Running key rotator");
                 let key_rotator = KeyRotator::new(datastore, config.hpke);
                 let mut interval = interval(Duration::from_secs(config.frequency_s));
                 while stopper.stop_future(interval.tick()).await.is_some() {
--- a/aggregator/src/binaries/aggregator.rs
+++ b/aggregator/src/binaries/aggregator.rs
@@ -1,5 +1,9 @@
 use crate::{
-    aggregator::{self, http_handlers::aggregator_handler},
+    aggregator::{
+        self,
+        http_handlers::aggregator_handler,
+        key_rotator::{deserialize_hpke_key_rotator_config, HpkeKeyRotatorConfig, KeyRotator},
+    },
     binaries::garbage_collector::run_garbage_collector,
     binary_utils::{setup_server, BinaryContext, BinaryOptions, CommonBinaryOptions},
     cache::{
@@ -27,8 +31,8 @@ use std::{
     pin::Pin,
 };
 use std::{iter::Iterator, net::SocketAddr, sync::Arc, time::Duration};
-use tokio::{join, sync::watch};
-use tracing::info;
+use tokio::{join, sync::watch, time::interval};
+use tracing::{error, info};
 use trillium::Handler;
 use trillium_router::router;
 use url::Url;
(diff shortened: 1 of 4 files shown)


docs/samples/advanced_config/aggregator.yaml
```diff
@@ -162,3 +162,31 @@ garbage_collection:
...
   collection_limit: 50
+
+# Configuration for key rotator. Allows running the key rotator as part of the aggregator process.
+# If omitted, you should run the key rotator as a separate cronjob.
+key_rotator:
+  # How frequently to run the key rotator, in seconds. Required.
+  frequency_s: 3600
+
+  # Rotation policy for global HPKE keys. At least one is required. Each entry
+  # represents a key with a particular ciphersuite.
```
divergentdave, line 173: As in https://github.com/divviup/janus/pull/3136/files#r1635380069, I think the second and third sentences belong on the `ciphersuite` key now.

aggregator/src/binaries/aggregator.rs
```diff
@@ -134,6 +160,7 @@ async fn run_aggregator(
     join!(
         aggregator_server,
         garbage_collector_future,
+        key_rotator_future,
         aggregator_api_future
     );
```
divergentdave, line 165: This is an existing issue, but I think we should `tokio::spawn()` the four futures here, and `join!()` the resulting join handles instead. This would let us skip a bunch of bookkeeping in the join macro on each poll, and let the respective accept loops, timers, and other await points run concurrently with each other on separate threads.

inahga: Simple enough change. Note that I've wrapped these in a `try_join!` since the compiler complains about unused results, and it doesn't quite feel right to swallow the errors.

It does make this PR mildly risky though, because if any of these futures were quietly failing, they'll now crash the whole process (which IMO, is a more desirable outcome anyway).

aggregator/src/binaries/aggregator.rs
```diff
@@ -78,19 +81,41 @@ async fn run_aggregator(
...
+        spawn(async move {
             if let Some(gc_config) = gc_config {
+                info!("Running garbage collector");
                 run_garbage_collector(datastore, gc_config, meter, stopper).await;
             }
-        }
+        })
+    };
+
+    let key_rotator_handle = {
```
branlwyd, line 97: While we've moved to a single-binary model to improve our build times, we are also moving away from any given process serving multiple roles (e.g. we don't want a single process that is both an aggregator & a garbage collector).

Is there a reason to run the key rotator as part of the aggregator, rather than making it a separate subcommand that we'd run as a separate deployment? I don't see any synchronization between the key rotator's work & the other pieces that might be running in-process; and IMO we don't want every aggregator to run key rotation (it's wasted effort, even in the likely case there are no concurrency bugs).

docs/samples/advanced_config/aggregator.yaml
```diff
@@ -162,3 +162,31 @@ garbage_collection:
...
+  frequency_s: 3600
+
+  # Rotation policy for global HPKE keys.
+  hpke:
+    # How long keys remains pending before they're promoted to active. Should
+    # be greater than the global HPKE keypair cache refresh rate. Defaults to
+    # 1 hour.
+    pending_duration_s: 3600
+
+    # The TTL of keys. Defaults to 4 weeks.
```
branlwyd, line 179:
```suggestion
    # The TTL of keys. Defaults to 12 weeks.
```

(nit, I believe this expresses the current state of the code)

aggregator/src/binaries/aggregator.rs
branlwyd: Hmm. I guess it's because we think (small?) BYOH deployments might still want to run everything in a single process?

I guess this isn't the worst, but it will lead to pain when the BYOH deployment scales up & no longer wants to run GC/key rotation in the same process as the aggregators. For simplicity, I'd still somewhat prefer to factor things such that these are separate processes, but maybe that is controversial?

inahga: > I don't see any synchronization between the key rotator's work & the other pieces that might be running in-process;

I'm going to implement synchronization in a future PR, i.e. the one that makes global HPKE keys required. The global HPKE keys cache will block until at least one global HPKE becomes available.

> Hmm. I guess it's because we think (small?) BYOH deployments might still want to run everything in a single process?

This is the case, yes. Similar in spirit to GC where our BYOH deployments run it in the same process. It's indeed wasteful for when many aggregator replicas are involved, but this I think this is a somewhat reasonable tradeoff in the interest of ease of deployment.

I think if we had a more prescriptive deployment solution for BYOH, i.e. a helm chart or `kustomize`-esque spec, I would be more comfortable omitting this functionality entirely, because we would be able to ship a kubernetes `CronJob` that invokes the key rotator.

n.b. the key rotator as implemented in the previous PR can (and should) be run in a separate process as a cronjob, for our non-BYOH deployments. Higher scale-out BYOH deployments can remove `key_rotator` from the `aggregator.yaml` config, and move it to a separate `key_rotator.yaml` config for running the key rotator separately.

branlwyd: This sounds good; I guess it doesn't cost us much to _support_ running multiple logically-separate pieces in one process.

branlwyd (approved):

PR #3237, opened 2024-06-14
Ignore const_is_empty lint

inahga:
This lint does us no good, since `INTEROP_CLIENT_IMAGE_BYTES` et. al. are defined as
```rust
const INTEROP_AGGREGATOR_IMAGE_BYTES: &[u8] =
    include_bytes!(concat!(env!("OUT_DIR"), "/interop_aggregator.tar.zst"));
```
so if that file is absent at linting time, it will flag.

### DIFF
--- a/interop_binaries/src/lib.rs
+++ b/interop_binaries/src/lib.rs
@@ -35,6 +35,11 @@ use trillium_api::ApiConnExt;
 use url::Url;
 
 #[cfg(feature = "testcontainer")]
+// Consts in this module are derived from `include_bytes!()`. If the target file is empty at
+// lint-time, they'll be empty, so this lint does us no good. This lint is only on Rust >1.79, so
+// we have to ignore unknown lints for MSRVs.
+#[allow(unknown_lints)]
+#[allow(clippy::const_is_empty)]
 pub mod testcontainer;
 
 pub mod status {


interop_binaries/src/testcontainer.rs
```diff
@@ -30,6 +30,7 @@ impl Default for Client {
     fn default() -> Self {
         // One-time initialization step: load compiled image into docker, recording its image tag,
         // so that we can launch it later.
+        #[allow(clippy::const_is_empty)]
```
branlwyd, line 33: nit: add comments describing why this whitelisting is necessary (I would have had to think about it for a few minutes w/o the PR description)

branlwyd (approved):

interop_binaries/src/lib.rs
```diff
@@ -35,6 +35,11 @@ use trillium_api::ApiConnExt;
 use url::Url;
 
 #[cfg(feature = "testcontainer")]
+// Consts in this module are derived from `include_bytes!()`. If the target file doesn't exist at
+// lint-time, they'll be empty, so this lint does us no good. This lint is only on Rust >1.79, so
+// we have to ignore unknown lints for MSRVs.
```
divergentdave, line 40: nit: In the "skip" case, the file will still exist, but it will be empty. (I think `include_bytes!()` on a nonexistent file would error)

divergentdave (approved):

PR #3251, opened 2024-06-18
Update rustls from 0.23.9 to 0.23.10

inahga:
Addresses security advisory https://github.com/divviup/janus/actions/runs/9571918527/job/26390065093?pr=3228#step:4:9.


inahga:
Oh, I see, I should read the dependency tree more carefully instead of running that command--`hpke` needs to update its `x25519-dalek` dependency.

PR #3268, opened 2024-06-26
Feature flag for requiring global HPKE keys

inahga:

Introduces a feature flag `require_global_hpke_keys`. When this is enabled:
- The `GlobalHpkeKeypairCache` blocks and retries fetching keys until it gets a list containing at least one active key. This is a bootstrapping case--it allows for the key rotator to do initialization, whether it's running in-process or separately.
- The `/hpke_configs` endpoint unconditionally returns global keys, regardless of the `?task_id` parameter.

Draft, as it needs refactoring and more testing. Looking for feedback on whether the overall approach is sane.

### DIFF
--- a/aggregator/src/binaries/aggregator.rs
+++ b/aggregator/src/binaries/aggregator.rs
@@ -394,6 +394,11 @@ pub struct Config {
     /// [`TASK_AGGREGATOR_CACHE_DEFAULT_CAPACITY`]. You shouldn't normally have to specify this.
     #[serde(default)]
     pub task_cache_capacity: Option<u64>,
+
+    /// Experimental. Always advertise global HPKE keys instead of per-task HPKE keys. This will
+    /// become on by default in a future version of Janus.
+    #[serde(default)]
+    pub require_global_hpke_keys: Option<bool>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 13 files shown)


aggregator/src/binaries/aggregator.rs
```diff
@@ -394,6 +394,11 @@ pub struct Config {
     /// [`TASK_AGGREGATOR_CACHE_DEFAULT_CAPACITY`]. You shouldn't normally have to specify this.
     #[serde(default)]
     pub task_cache_capacity: Option<u64>,
+
+    /// Experimental. Always advertise global HPKE keys instead of per-task HPKE keys. This will
+    /// become on by default in a future version of Janus.
+    #[serde(default)]
+    pub require_global_hpke_keys: Option<bool>,
```
branlwyd, line 401: nit: is `None` the same as `Some(false)`? if so, for simplicity I think this should just be a `bool` which defaults to `false`, so that we only interpret a missing value once.

branlwyd (commented):
The approach SGTM. 👍🏻

inahga:
Eh, pulling this back to draft, I think some work on integration/interop tests is justified here.

inahga:
I've made integration and inteorp tests use the flag. I'll make unit tests use it in the next PR, to keep this PR relatively small.

tgeoghegan (approved):

branlwyd (approved):

PR #3284, opened 2024-07-09
tests: Use global HPKE keys in all tests

inahga:

Any tests that relied on the task-specific key now rely on a global hpke key, e.g. for report encryption. I accomplished this by having the test setup code insert a global key on initialization, then plumbed that through. I detected dependencies in tests by temporarily removing the `Task::current_hpke_key()` function, and temporarily removing task-specific key decryption code in the aggregate init and upload functions, then seeing what tests failed.

Plumbing was done by introducing `HttpHandlerTest` and `UploadTest` structs, rather than extending the returned tuple from `setup_*()` functions. This is a bit more future proof, because it's easier to modify fields of a struct rather than a tuple.

I have not converted all tests to use `HttpHandlerTest` and `UploadTest` structs. I'm going to do this in the next PR to keep this one relatively small.

I've added a couple of very basic tests that still provide coverage to task-specific keys. These can be removed during the next breaking windows change, where we'll completely remove all traces of task-specific keys.


aggregator/src/aggregator.rs
```diff
@@ -3526,80 +3536,88 @@ mod tests {
...
-        setup_upload_test_with_runtime(TestRuntime::default(), cfg).await
+    struct UploadTest {
+        vdaf: Prio3Count,
+        aggregator: Aggregator<MockClock>,
+        clock: MockClock,
+        task: Task,
+        datastore: Arc<Datastore<MockClock>>,
+        ephemeral_datastore: EphemeralDatastore,
+        hpke_keypair: HpkeKeypair,
     }
```
tgeoghegan, line 3555: This file has always been too big. If we're touching a lot of the test code anyway, then it could be wise to move `struct UploadTest` and supporting code to some dedicated test module like `aggregator/src/aggregator/aggregator_test.rs`. I think that makes using `struct UploadTest` from other test modules a little more graceful, too.

tgeoghegan (approved):

branlwyd (approved):

PR #3288, opened 2024-07-09
config: Standardize time suffixes

inahga:
In slack we agreed to standardize on `_s` and `_ms` for time suffixes. I've also used `_us` for microseconds, since I don't think requiring unicode characters in configuration is appropriate.

For backwards compatibility, I have used `serde(alias = "")`. During the next breaking windows change, we can remove these aliases. I have not altered any samples or doccomments in this PR, so that the CI can prove that the aliases work. I will fix docs in the next PR.

We could add unit tests that exercise backcompat, but that doesn't seem worth the extra test complexity--open to dissenting opinions on that.


branlwyd (approved):
The renaming LGTM. To ease cleanup, please add a TODO at each alias point that will eventually need to be cleaned up.

PR #3289, opened 2024-07-09
Upgrade testcontainers to 0.20.0

divergentdave:
There have been four testcontainers releases over the past few months, and Dependabot has been failing to propose upgrades. Plus, there were a number of breaking changes to improve the API. This PR upgrades to the latest version and fixes various errors along the way.


interop_binaries/src/lib.rs
```diff
@@ -687,6 +684,6 @@ pub mod test_util {
     }
 
     pub fn generate_unique_name(prefix: &str) -> String {
-        format!("{}_{}", prefix, hex::encode(random::<[u8; 4]>()))
+        format!("{}-{}", prefix, hex::encode(random::<[u8; 4]>()))
```
divergentdave, line 687: FWIW, this is necessary to fix issues in the `end_to_end` test. I was seeing error messages in container logs saying that requests couldn't be made because hostnames were invalid. I think this happened because `testcontainers` now turns on the `reqwest/hickory-dns` feature, and Hickory is more strict about underscores not being allowed in hostnames.

inahga (approved):

PR #3292, opened 2024-07-10
collector: Refresh example in doccomment

inahga:
Give a little bit of love to this doccomment. Suggest use of the `PrivateCollectorCredential` instead of bringing in the unstable `janus_core`. Also provide an example of using a fixed size query.

### DIFF
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -17,45 +17,45 @@
 //! use url::Url;
 //!
 //! # async fn run() {
-//!     let task_id = TaskId::from_str("[your DAP task ID here]").unwrap();
+//! let task_id = TaskId::from_str("[your DAP task ID here]").unwrap();
 //!
-//!     let collector_credential: PrivateCollectorCredential =
-//!         serde_json::from_reader(File::open("[path to JSON encoded collector credential]").unwrap())
-//!             .unwrap();
-//!
-//!     let leader_url =
-//!         Url::from_str("[absolute URI to the DAP leader, e.g. https://leader.dap.example.com/]")
-//!             .unwrap();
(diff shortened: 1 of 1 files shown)


collector/src/lib.rs
```diff
@@ -6,46 +6,56 @@
...
 //! # async fn run() {
-//! // Supply DAP task parameters.
-//! let task_id = random();
-//! let hpke_keypair = janus_core::hpke::generate_hpke_config_and_private_key(
-//!     HpkeConfigId::from(0),
-//!     HpkeKemId::X25519HkdfSha256,
-//!     HpkeKdfId::HkdfSha256,
-//!     HpkeAeadId::Aes128Gcm,
-//! ).unwrap();
+//!     let task_id = TaskId::from_str("[your DAP task ID here]").unwrap();
```
divergentdave, line 20: I'd recommend leaving this un-indented, since the function's opening and closing braces won't appear in the final output. (It's just present to allow us to use `.await` in this context)

divergentdave (approved):
LGTM

FWIW `janus_core` is published on crates.io, while `janus_aggregator_core` is not. (it is a dependency of `janus_collector` as well)

inahga:
> FWIW janus_core is published on crates.io, while janus_aggregator_core is not.

Indeed, but a while ago we decided ago to say "please don't use janus_core directly", so that we can make breaking changes to it. Granted, I don't think we really landed on hard consensus on it, so we can rehash the discussion if we need. https://github.com/divviup/janus/pull/2390#discussion_r1434609580 https://github.com/divviup/janus?tab=readme-ov-file#versioning-and-stability

PR #3294, opened 2024-07-10
Fix flaky tests

divergentdave:
This fixes some flaky tests I found. Common themes:

- Some more tests that assume an HPKE config ID is unused needed to be updated to take the initial global HPKE key into account.
- `put_global_hpke_keypair()` does not return `MutationTargetAlreadyExists`, so we can't try to insert a keypair and suppress errors, but we can compare config IDs up front in tests.
- Some of the checks in task cache tests are racy, and they are more likely to fail under high CPU load. I removed or changed these, so we no longer require that the cache reflect a stale value immediately after changing the database. (since sometimes "immediately" comes too late)

### DIFF
--- a/aggregator/src/aggregator/http_handlers/tests/aggregation_job_init.rs
+++ b/aggregator/src/aggregator/http_handlers/tests/aggregation_job_init.rs
@@ -265,6 +265,9 @@ async fn aggregate_init() {
         if helper_task.hpke_keys().contains_key(hpke_config.id()) {
             continue;
         }
+        if hpke_keypair.config().id() == hpke_config.id() {
+            continue;
+        }
         break hpke_config;
     };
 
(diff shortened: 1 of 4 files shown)


aggregator/src/aggregator/http_handlers/tests/aggregation_job_init.rs
```diff
@@ -734,28 +735,23 @@ async fn aggregate_init_with_reports_encrypted_by_task_specific_key() {
...
-                // Leave these in the PENDING state--they should still be decryptable.
-                match tx
-                    .put_global_hpke_keypair(&global_hpke_keypair_same_id)
-                    .await
-                {
-                    // Prevent test flakes in case a colliding key is already in the datastore.
-                    // The test context code randomly generates an ID so there's a chance for
-                    // collision.
-                    Ok(_) | Err(datastore::Error::MutationTargetAlreadyExists) => Ok(()),
-                    Err(err) => Err(err),
```
inahga, line 753: :facepalm: ah I see, there is no `ON CONFLICT DO NOTHING` in `put_global_hpke_keypair()` so if there's a conflict it'll fail with a harder-to-match postgres uniqueness constraint violation error.

inahga (approved):
Thanks!

branlwyd (approved):

PR #3295, opened 2024-07-10
Reject unknown fields in configuration files

divergentdave:
This changes all config-related deserialization routines to throw an error upon seeing unexpected fields, rather than ignoring them. This should be mentioned in the next release notes.

### DIFF
--- a/aggregator/src/aggregator/key_rotator.rs
+++ b/aggregator/src/aggregator/key_rotator.rs
@@ -60,6 +60,7 @@ pub struct KeyRotator<C: Clock> {
 
 /// Defines the ciphersuite and rotation policy of a global HPKE key.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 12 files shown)


inahga (approved):

tgeoghegan (approved):

branlwyd (approved):
I'd like it if we could apply this to every serde-serialized structure by default, but I can't find a way to do this.

divergentdave:
That would be possible if we rolled up our sleeves to write some proc macros, and then used our derive instead of serde's derive everywhere, but I don't know if that would be worth it.

Moreover, I am intentionally holding off on changing deserialization done on aggregator API requests for now. We may want to make that parsing strict as well at some point, but it will take a whole different set of testing, and affect a different set of API consumers.

branlwyd:
I definitely don't think writing proc macros ourselves is the right path. If we really wanted to address this, I'd try to upstream a feature into serde before doing that -- but I don't think _that_ is worth it, either. :)

PR #3302, opened 2024-07-11
Add discrete Laplace noise w/ SumVec and Histogram

### LINKED ISSUE #3301: Incorporate DP into Prio3SumVec tasks
Work is being done in libprio to add DP to Prio3SumVec: https://github.com/divviup/libprio-rs/issues/1068.

This issue is to track the work to incorporate that change into janus.

divergentdave:
This closes #3301. Depends on divviup/libprio-rs#1072.


tgeoghegan (approved):
Looks like straightforward plumbing changes.

integration_tests/tests/integration/janus.rs
```diff
@@ -443,3 +458,60 @@ async fn janus_in_process_one_round_with_agg_param_time_interval() {
...
+        before_timestamp,
+        &(),
+    )
+    .await;
+    assert_eq!(report_count, janus_pair.task_parameters.min_batch_size);
+
+    let mut un_noised_result = [0u128; HISTOGRAM_LENGTH];
+    un_noised_result[0] = report_count.into();
+    // Smoke test: Just confirm that some noise was added. Since epsilon is small, the noise will be
+    // large (drawn from Laplace_Z(20) + Laplace_Z(20)), and it is highly unlikely that all 100
```
inahga, line 514: Naive: what is `_Z`? (you can point me to a paper :D).

inahga (approved):

integration_tests/tests/integration/janus.rs
divergentdave: This is a lo-fi version of the notation $Lap_\mathbb{Z}(t)$ from section 4 of https://arxiv.org/pdf/2004.00010. (where blackboard Z represents the set of the integers)

inahga: Also naive: there is no bound on the noise that could be applied to the result, yes? I'm mildly uncomfortable that the set of all valid results for this test is "all integers except for `n`", but seeing as the test fails if I turn off DP, I'm fine with it.

divergentdave: Correct, the noise value could be any integer, thus we could be adding any field elements to the shares. (Though the probabilities become astronomically low once you get many multiples of the scale parameter out)

I also did a negative test where I cranked epsilon way up (to `u128::MAX`) and confirmed that the test failed then. Even though it was still adding noise, the noise was all zeros because of the absurd parameter choice.

We could improve this with some sort of statistical test that takes each dimension of the noise vector as one sample of the noise distribution, and then performs a statistical test to see how good of a fit it is for the desired distribution. (with a low enough false positive rate to not cause problems) I filed #3325 to track this as a follow-up. It'll be a bit tricky because the distribution of the sum of two numbers drawn from base distributions is the convolution of those distributions, plus I'll need to read up on statistical tests.

PR #3303, opened 2024-07-13
Helper agg init: remove copy of report share data.

### LINKED ISSUE #3285: Helper peak memory usage can be reduced
I set up a harness to run one big aggregation job under Valgrind, using the DHAT tool. For both the leader and the helper, the peak memory usage was similar, double the total measurement share/output share size, plus various smaller allocations. (in the case of the helper, the next largest allocation is the aggregation job initialization request body, in a `trillium_http::ReceivedBody`)

In the case of the leader, one copy of the measurement share is allocated when loading report aggregations from the database in the aggregation job driver. The other copy is made inside the VDAF's `prepare_init()`, because the Prio3 leader measurement share is cloned for storage in `Prio3PrepareState`. We could probably eliminate this copy with `Arc`s inside the Prio3 implementation internals, in order to share and then transfer ownership of this vector, but that's outside the scope of this repository.

In the case of the helper, peak memory usage occurs after VDAF preparation, when preparing to write results to the database. The largest allocations come from two copies of the output shares. One is allocated in `Flp::truncate()`, and the other is allocated when cloning a `Vec<ReportShareData>` (which contains in turn `WritableReportAggregation` and `A::OutputShare`). This vector is cloned for two reasons, to give an async block ownership while still allowing for transaction retries, and to allow the vector to be modified inside the transaction before the report aggregations are finally written out. One of these copies could be eliminated if the original `Vec<ReportShareData>` were moved into an `Arc`, with ownership shared between each retry of the transaction, and if the mid-transaction data structure updates were done on a `Vec<Cow<'_, _>>`, with borrow references pointing back to the original `Arc<Vec<ReportShareData>>`. This would require an involved refactor, to allow `AggregationJobWriter` to work with `Cow`s for both phases of operation, instead of only its second phase, but it could cut peak memory usage by almost half, giving us more headroom from OOM errors.

branlwyd:
Previously, we copied report share data into the transaction callback (re-copying it on transaction retry). Notably, this required us to keep two copies of the output shares in memory. This was done to allow modification of the report aggregations based on information retrieved from the datastore, i.e. if the report was replayed.

Now, we remove the copy by passing an Arc of the relevant data into the transaction callback. Modification of the report aggregation is implemented via a Cow, which will be borrowed from the data in the Arc in the common case of no modification. The aggregation job writer is augmented with the ability to accept report aggregations already stored in a Cow, and is taught to be smart enough not to double-wrap Cow around its already-Cow'ed internal state.

Closes #3285.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -103,6 +103,7 @@ use ring::{
     signature::{EcdsaKeyPair, Signature},
 };
 use std::{
+    borrow::Cow,
     collections::{HashMap, HashSet},
     fmt::Debug,
     hash::Hash,
@@ -2294,6 +2295,7 @@ impl VdafOps {
 
         let mut report_share_data = Vec::with_capacity(req.prepare_inits().len());
         while receiver.recv_many(&mut report_share_data, 10).await > 0 {}
+        let report_share_data = Arc::new(report_share_data);
 
         // Await the producer task to resume any panics that may have occurred, and to ensure we can
         // unwrap the aggregation parameter's Arc in a few lines. The only other errors that can
@@ -2346,11 +2348,11 @@ impl VdafOps {
 
         Ok(datastore
             .run_tx("aggregate_init", |tx| {
-                let vdaf = vdaf.clone();
+                let vdaf = Arc::clone(&vdaf);
                 let task = Arc::clone(&task);
                 let aggregation_job_writer_metrics = metrics.for_aggregation_job_writer();
                 let aggregation_job = Arc::clone(&aggregation_job);
-                let mut report_share_data = report_share_data.clone();
+                let report_share_data = Arc::clone(&report_share_data);
                 let req = Arc::clone(&req);
                 let log_forbidden_mutations = log_forbidden_mutations.clone();
 
@@ -2373,24 +2375,23 @@ impl VdafOps {
                     }
 
                     // Write report shares, and ensure this isn't a repeated report aggregation.
-                    try_join_all(report_share_data.iter_mut().map(|rsd| {
+                    let report_aggregations = try_join_all(report_share_data.iter().map(|rsd| {
                         let task = Arc::clone(&task);
 
                         async move {
-                            let report_already_aggregated =
-                                match tx.put_scrubbed_report(task.id(), &rsd.report_share).await {
-                                    Ok(()) => false,
-                                    Err(datastore::Error::MutationTargetAlreadyExists) => true,
-                                    Err(err) => return Err(err),
-                                };
-
-                            if report_already_aggregated {
-                                rsd.report_aggregation = rsd
-                                    .report_aggregation
-                                    .clone()
-                                    .with_failure(PrepareError::ReportReplayed);
-                            }
-                            Ok::<_, datastore::Error>(())
+                            let mut report_aggregation = Cow::Borrowed(&rsd.report_aggregation);
+                            match tx.put_scrubbed_report(task.id(), &rsd.report_share).await {
+                                Ok(()) => (),
+                                Err(datastore::Error::MutationTargetAlreadyExists) => {
+                                    report_aggregation = Cow::Owned(
+                                        report_aggregation
+                                            .into_owned()
+                                            .with_failure(PrepareError::ReportReplayed),
+                                    )
+                                }
+                                Err(err) => return Err(err),
+                            };
+                            Ok(report_aggregation)
                         }
                     }))
                     .await?;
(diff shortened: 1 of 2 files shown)


tgeoghegan (approved):

aggregator/src/aggregator/aggregation_job_writer.rs
```diff
@@ -507,7 +507,7 @@ where
                             aggregation_job: Cow::Borrowed(aggregation_job),
                             report_aggregations: report_aggregations
                                 .iter()
-                                .map(Cow::Borrowed)
+                                .map(RA::borrow)
```
inahga, line 510: Check my understanding: at this point, report_aggregations is (roughly) a `Vec<impl ReportAggregationUpdate>`, which, concretely, could be a `struct ReportAggregationMetadata` or a `Cow` of it?

In which case a naive `map(Cow::Borrowed)` would double `Cow` a `Cow<'_, ReportAggregationMetadata>`, so we need a trait method to avoid this.

aggregator/src/aggregator.rs
```diff
@@ -2373,24 +2375,23 @@ impl VdafOps {
                     }
 
                     // Write report shares, and ensure this isn't a repeated report aggregation.
-                    try_join_all(report_share_data.iter_mut().map(|rsd| {
+                    let report_aggregations = try_join_all(report_share_data.iter().map(|rsd| {
                         let task = Arc::clone(&task);
```
inahga, line 2379: Can we accomplish the same by doing `.iter_mut()` and modifying the elements of `ReportShareData` in place? We would need to at least wrap it in a `std::sync::Mutex` for this to work, but it would avoid copying in all cases and save a bunch of code.

(Just checking if you've considered it, it probably is more involved of a change than I'm thinking).

inahga (approved):
Changes LGTM and I _think_ they accomplish the goal.

Do you have memory profiles that show improvement? Since a few LGTMs is not necessarily evidence of it working. Broadly, it would be nice to document how we're profiling, if for no other reason than I'm unfamiliar with Rust profiling strategies (but I don't think that should hold up this PR).

aggregator/src/aggregator.rs
branlwyd: We can't modify `report_share_data`: it's in an `Arc` to avoid copying the `ReportShareData` inside the `Vec`, and `Arc` doesn't allow mutable references. Or, to look at it another way, we can't modify `report_share_data` because we might need the unmodified value if we retry the transaction callback.

divergentdave:
I re-tested this and it demonstrated the desired drop in peak memory usage. (something a bit below 50%) I merged https://github.com/divergentdave/janus/tree/david/experiment-dhat-harness-2 onto this branch, built it in release mode with debug symbols, populated test fixture files by running the different subcommands several times, and then ran `valgrind --tool=dhat --dhat-out-file=dhat.out.helper-arc-cow ./target/profiling/profiler_harness helper`. I popped the resulting file into the DHAT viewer static webapp, and looked at the "At t-gmax (bytes)" metric.

aggregator/src/aggregator/aggregation_job_writer.rs
branlwyd: That's correct. The implementation of `ReportAggregationUpdate` over `Cow` exists to allow a `Cow<WritableReportAggregation>` (or a `Cow` of anything else that implements `ReportAggregationUpdate`), _without_ needing to deal with `Cow<Cow<T>>`.

branlwyd:
Thank you for profiling this!

aggregator/src/aggregator.rs
inahga: FWIW I was able to coerce my suggestion into "working" by having `report_share_data` being in an `Arc<tokio::sync::Mutex<_>>`, but it then becomes difficult to take a `.report_aggregation` for use in the `AggregationJobWriter` without `.clone()`ing it--i.e. I landed on needing to pass around a `Cow` anyway. So I think your approach is super-correct :+1:.

branlwyd: Yeah -- N.B. the problem with `Arc<Mutex<Vec<T>>>` is that, without a Cow, ultimately we would be modifying the `report_share_data`, which would remain modified in the case of a transaction retry (and on retry, we want to start from the same data as in the previous attempt, to ensure correctness).

As you note, we could do `Arc<Mutex<Vec<Cow<T>>>>`, but at that point, I think we're better off just generating a new `Vec<Cow<T>>` based off of the `report_share_data` since we need to modify the vector anyway.

PR #3314, opened 2024-07-16
Add in-cluster test with DP

divergentdave:
This adds an in-cluster integration test exercising aggregator-based differential privacy noise.

This shouldn't be merged until #3302 and divviup/divviup-api#1173 have been deployed.


tgeoghegan (approved):

inahga (approved):

integration_tests/tests/integration/in_cluster.rs
```diff
@@ -897,3 +917,128 @@ mod rate_limits {
...
+    )
+    .await;
+    assert_eq!(report_count, janus_pair.task_parameters.min_batch_size);
+
+    let mut un_noised_result = [0u128; HISTOGRAM_LENGTH];
+    un_noised_result[0] = report_count.into();
+    // Smoke test: Just confirm that some noise was added. Since epsilon is small, the noise will be
+    // large (drawn from Laplace_Z(20) + Laplace_Z(20)), and it is highly unlikely that all 100
+    // noise values will be zero simultaneously.
+    assert_ne!(aggregate_result, un_noised_result);
```
branlwyd, line 979: Are there any useful bounds on the amount of noise added such that we could check that the actual result is close enough to the non-noised result to be plausibly the result of aggregation + DP noise? "Useful" in this case meaning the condition is unlikely to be satisfied if DP noise is not being added correctly.

branlwyd (approved):

integration_tests/tests/integration/in_cluster.rs
inahga: https://github.com/divviup/janus/pull/3302#discussion_r1688567299 :grin:

divergentdave: I could check that each coordinate is not in the middle of the field's range (p*1/4 to p*3/4). That'll be more than far enough out that false positives won't be a problem, while still catching errors in secret sharing/unsharding with a probability of 1/2 on each coordinate.

divergentdave:
This is now passing against staging. https://github.com/divviup/janus-ops/actions/runs/10097504244/job/27922510921

PR #3315, opened 2024-07-22
Schema changes for application-level aggregation metrics.

branlwyd:


aggregator_core/src/datastore.rs
```diff
@@ -102,7 +102,7 @@ macro_rules! supported_schema_versions {
 // version is seen, [`Datastore::new`] fails.
 //
 // Note that the latest supported version must be first in the list.
-supported_schema_versions!(6, 5);
+supported_schema_versions!(7, 6, 5);
```
tgeoghegan, line 105: If all our deployed Januses are up to schema version 6, I think we could drop 5 and claw back some CI time.

db/00000000000007_task_aggregation_counters.up.sql
```diff
@@ -0,0 +1,14 @@
+-- Per-task report aggregation counters, used for metrics.
+--
+-- Fillfactor is lowered to improve the likelihood of heap-only tuple optimizations. See the
+-- discussion around this setting for the task_upload_counters table.
+CREATE TABLE task_aggregation_counters(
```
tgeoghegan, line 5: nit: elsewhere we align types and comments column-wise.
```
CREATE TABLE task_upload_counters(
    id                     BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,  -- artificial ID, internal-only
    task_id                BIGINT NOT NULL,

    interval_collected     BIGINT NOT NULL DEFAULT 0, -- Reports submitted for an interval that was already collected.
    report_decode_failure  BIGINT NOT NULL DEFAULT 0, -- Reports which failed to decode.
    report_decrypt_failure BIGINT NOT NULL DEFAULT 0, -- Reports which failed to decrypt.
    report_expired         BIGINT NOT NULL DEFAULT 0, -- Reports that were older than the task's report_expiry_age.
    report_outdated_key    BIGINT NOT NULL DEFAULT 0, -- Reports that were encrypted with an unknown or outdated HPKE key.
    report_success         BIGINT NOT NULL DEFAULT 0, -- Reports that were successfully uploaded.
    report_too_early       BIGINT NOT NULL DEFAULT 0, -- Reports whose timestamp is too far in the future.
    task_expired           BIGINT NOT NULL DEFAULT 0, -- Reports sent to the task while it is expired.

    ord                    BIGINT NOT NULL,           -- Index of this task_upload_counters shard.

    CONSTRAINT fk_task_id FOREIGN KEY(task_id) REFERENCES tasks(id) ON DELETE CASCADE,
    CONSTRAINT task_upload_counters_unique UNIQUE(task_id, ord)
) WITH (fillfactor = 50);
```

db/00000000000007_task_aggregation_counters.up.sql
```diff
@@ -0,0 +1,14 @@
...
+    id BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,  -- artificial ID, internal only
+    task_id BIGINT NOT NULL,  -- task ID the counter is associated with
+    ord BIGINT NOT NULL,  -- the ordinal index of the task aggregation counter
+
+    success BIGINT NOT NULL DEFAULT 0,  -- reports successfully aggregated
```
tgeoghegan, line 10: Do we not want to track preparation failures by reason the way we do upload failures? If nothing else, keeping track of either total failures or total prep attempts would let us compute success rate.

db/00000000000007_task_aggregation_counters.up.sql
```diff
@@ -0,0 +1,14 @@
...
+
+    CONSTRAINT task_aggregation_counters_unique_id UNIQUE(task_id, ord),
+    CONSTRAINT fk_task_id FOREIGN KEY(task_id) REFERENCES tasks(id) ON DELETE CASCADE
+) WITH (fillfactor = 50);
```
tgeoghegan, line 14: Given that uploading these counter tables can be performance sensitive, I'm surprised we don't create an index on task ID. Maybe it's irrelevant because this table will be smallish anyway?

tgeoghegan (approved):

db/00000000000007_task_aggregation_counters.up.sql
inahga, line 10: I see we're not going to track the other failures that could come up, like task upload counters do. Rationale? I'm nominally OK with this, since if we want this functionality it should be fairly trivial to bolt it on later.

inahga (approved):

db/00000000000007_task_aggregation_counters.up.sql
inahga: `CONSTRAINT task_aggregation_counters_unique_id UNIQUE(task_id, ord)` implicitly creates an index on `(task_id, ord).` We only ever INSERT on `(task_id, ord)` or SELECT `task_id` and aggregate over `ord`.

I have some query plans from task upload counters https://github.com/divviup/janus/pull/2508#discussion_r1462320434.

divergentdave (approved):

db/00000000000007_task_aggregation_counters.up.sql
branlwyd: I'm implementing the least amount of functionality that gives us what we need; I'm OK leaning on operational metrics & logging for other issues, at least until we have a justification for splitting them out. (And yeah, I think it would be easy to add more to this later.)

db/00000000000007_task_aggregation_counters.up.sql
branlwyd: `UNIQUE` constraints generate a (unique) index on the relevant fields, so we have an index on `(task_id, ord)`.

See https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS:

> Adding a unique constraint will automatically create a unique B-tree index on the column or group of columns listed in the constraint.

db/00000000000007_task_aggregation_counters.up.sql
branlwyd: See response to inahga's comment:

> I'm implementing the least amount of functionality that gives us what we need; I'm OK leaning on operational metrics (e.g. prometheus/grafana) & logging for other issues, at least until we have a justification for splitting them out. (And yeah, I think it would be easy to add more to this later.)

Success rate, specifically, can be tracked via the operational metrics.

Issue #3323, opened 2024-07-23
Fixed size batch can get assigned too many reports if some get garbage collected


divergentdave:
We saw an issue in production where a collection job couldn't be stepped due to an "invalid number of reports" error, because the batch had more than `max_batch_size` reports. This can happen with tasks that have a slow trickle of reports (below max_batch_size/report_expiry_age). Note that `read_batch_size()` counts `report_aggregations` rows, which are subject to garbage collection, but this error message is triggered based on `batch_aggregations.report_count`, which still remembers reports until the entire batch is deleted. Thus, the aggregation job creator may overshoot `max_batch_size`, and the collection job driver will catch it.

PR #3324, opened 2024-07-23
Add discrete event simulation tests

divergentdave:
This implements a discrete event simulation framework, and then uses it as a declarative language to make a reproducer for #3323. See the big module doc comment for an explanation of how it works. I made two tweaks to non-test code: `AggregationJobCreator` now takes an `Arc<Datastore<C>>` instead of a `Datastore<C>`, to better align with other components. `AggregationJobCreator::create_aggregation_jobs_for_task()` is also public now, so I can run it once, instead of spawning a loop that runs it periodically.

This is all minimally functional now, but there are still some features I'd like to add. I also plan to copy this into an old branch, write a couple more reproducers for problems we've already fixed, and then port those tests forward as regression tests. I haven't spent much time looking at what quickcheck tests produce, but the whole input generation and shrinking machinery is working well. I already spotted one shallow configuration validation bug, but haven't addressed it yet (a minimum aggregation job size of zero breaks invariants and leads to an unwrap).

- [x] Inspect request and response bodies.
- [x] More regression tests.
- [x] `ByBatchId` fixed size queries.
- [x] Give `Arbitrary` control over more configuration parameters (i.e. time-bucketed fixed size)


integration_tests/tests/integration/simulation/reproduction.rs
```diff
@@ -0,0 +1,219 @@
...
+            },
+            Op::AggregationJobCreator,
+            Op::AggregationJobDriver,
+            Op::AdvanceTime {
+                amount: Duration::from_seconds(3600),
+            },
+            Op::LeaderGarbageCollector,
+            Op::Upload {
+                report_time: Time::from_seconds_since_epoch(1_700_003_600),
+            },
```
inahga, line 178: This is quite exciting, it kills tons of boilerplate when trying to reproduce production issues in tests.

inahga (commented):
Shallow review: it looks great! I can look more in depth later.

> This is all minimally functional now, but there are still some features I'd like to add.

Can we keep the changes relatively small, to aid review? Maybe we can take the minimum code to support the reproducer for https://github.com/divviup/janus/issues/3323 in one PR, then iterate on it with features in new PRs?

Because this is all test code we don't have to be as careful about completeness, compatibility, etc.

divergentdave:
I added a few more regression tests, I'll call this good for now.

Future TODOs:

- Iterate using the quickcheck tests.
- Give the Arbitrary implementation control over more configuration parameters.
- Try treating 5xx responses (excluding those from fault injection) as test failures.
- Upload different kinds of invalid reports that won't pass preparation, etc.

integration_tests/tests/integration/simulation/model.rs
```diff
@@ -0,0 +1,62 @@
+use janus_messages::{CollectionJobId, Duration, Interval, Time};
+
+#[derive(Debug, Clone)]
+pub(super) struct Input {
+    pub(super) is_fixed_size: bool,
```
inahga, line 5: Can we eliminate this field? It appears to only be used for constructing a `janus_aggregator_core::task::QueryType`. Perhaps it can be a verbatim `pub(super) query_type: QueryType` either in `Input` or `Config`?

inahga: (Or perhaps there's some subtlety with Arbitrary that I'm missing--if so then a comment justifying the greater purpose of this field is justified).

integration_tests/tests/integration/simulation/proxy.rs
```diff
@@ -0,0 +1,168 @@
...
+use trillium::{Conn, Handler, Status};
+use trillium_macros::Handler;
+
+/// A [`Handler`] wrapper that can be configured to drop requests or responses.
+#[derive(Handler)]
+pub(super) struct FaultInjectorHandler<H> {
+    #[handler(except = [run, before_send, name])]
+    inner: H,
+    error_before: Arc<Mutex<bool>>,
+    error_after: Arc<Mutex<bool>>,
```
inahga, line 17: Comments may be justified here, i.e. before and after _what_ happens? AIUI `before` induces error before business-logic-containing handler happens, and `after` induces it, well, after?

integration_tests/tests/integration/simulation/run.rs
```diff
@@ -0,0 +1,662 @@
...
+                return ControlFlow::Break(TestResult::error(format!("{error:?}")));
+            }
+        };
+        debug!(count = leases.len(), "acquired aggregation jobs");
+        for lease in leases {
+            if let Err(error) = (self.components.aggregation_job_driver_stepper_cb)(lease).await {
+                if let aggregator::Error::Http(_) = error {
+                    warn!(?error, "aggregation job driver error");
+                    return ControlFlow::Continue(());
+                }
```
inahga, line 279: Hmm? Is this to support reusing this code for the fault injection case?

Not sure how I feel about it--if there's a legitimate bug that triggers an `::Http` error, then it won't be caught, but maybe such a bug is better caught by unit tests.

integration_tests/tests/integration/simulation/proxy.rs
```diff
@@ -0,0 +1,168 @@
...
+        self.inner.run(conn).await
+    }
+
+    async fn before_send(&self, conn: Conn) -> Conn {
+        let mut conn = self.inner.before_send(conn).await;
+        if conn.state::<InspectMarker>().is_some() {
+            if conn.status() == Some(Status::Conflict) {
+                error!("409 Conflict response");
+                *self.failure.lock().unwrap() = true;
+            }
```
inahga, line 127: Any 4xx error strikes me as fatal? Or does that break the case where report uploads are too far in the future?

integration_tests/tests/integration/simulation/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+//! simulation share a `MockClock`. None of the components should run any asynchronous tasks
+//! continuously throughout the simulation (except for tokio-postgres connection tasks).
+//! Initialization will be akin to `JanusInProcessPair`, but more low-level. The simulation is fed a
+//! list of `Op` values, and it executes the operations described one after another. The operations
+//! are as follows:
+//!
+//! - AdvanceTime: Advance the `MockClock`'s time by some amount.
+//! - Upload: Have the client shard a report at the given timestamp, with the next sequential
+//!   measurement, and send it to the leader aggregator. The leader will handle the request and
+//!   store the report to the database. Note that, as currently implemented, this will wait for the
```
inahga, line 36: These should be doccomments on the relevant enum variant. Otherwise there's a high chance we'll forget to ever upload this list.

inahga (approved):
LGTM, I only have clarifications and nitpicky comments.

PR #3328, opened 2024-07-24
Application-level aggregation success metrics.

branlwyd:


docs/samples/advanced_config/aggregation_job_driver.yaml
divergentdave, line 1: We should update the comment on `task_counter_shard_count` in the sample aggregator configuration file as well.

aggregator/src/aggregator.rs
```diff
@@ -3325,6 +3355,38 @@ impl VdafOps {
...
+) {
+    // We write task aggregation counters back in a separate tokio task & datastore transaction,
+    // so that any slowness induced by writing the counters (e.g. due to transaction retry) does
+    // not slow the main processing. The lack of transactionality between writing the updated
+    // aggregation job & updating the counters means that process failure may cause us to leave
+    // some counter updates unaccounted for, but that is an acceptable tradeoff.
+    let ord = thread_rng().gen_range(0..shard_count);
+    tokio::task::spawn(async move {
+        let rslt = datastore
+            .run_tx("aggregate_continue_counters", |tx| {
```
divergentdave, line 3372: This transaction name is misleading, since it is used in all three aggregation code paths.
```suggestion
            .run_tx("aggregate_counters", |tx| {
```

docs/samples/basic_config/aggregation_job_driver.yaml
```diff
@@ -29,3 +29,8 @@ maximum_attempts_before_failure: 10
 # Number of sharded database records per batch aggregation. Must not be greater
 # than the equivalent setting in the collection job driver. (required)
 batch_aggregation_shard_count: 32
+
+# Defines the number of shards to break report & aggregation metric counters into. Increasing this
+# value will reduce the amount of database contention during report uploads & aggregations, while
+# increasing the cost of getting task metrics.
+task_counter_shard_count: 32
```
divergentdave, line 36: I think we could leave this out of the basic file, since it has a default value. (we currently do the same with the upload counter shard count)

aggregator/src/aggregator/test_util.rs
```diff
@@ -154,3 +161,37 @@ pub fn generate_helper_report_share_for_plaintext(
...
+                        .unwrap())
+                })
+            })
+            .await
+            .unwrap();
+
+        if counters == expected_counters {
+            return;
+        }
+        if Instant::now() > end_instant {
```
divergentdave, line 190: I think we could make this a bit more resilient if we snapshotted `Instant::now()` at the top of the loop. That way, if fetching the counters the first time takes way too long due to resource exhaustion from other tests, or locks taken by database maintenance processes, then we will get a second chance to read from the database.

aggregator/src/binaries/aggregation_job_driver.rs
```diff
@@ -121,6 +123,12 @@ pub struct Config {
     /// will reduce the amount of database contention during leader aggregation, while increasing
     /// the cost of collection.
     pub batch_aggregation_shard_count: u64,
+
+    /// Defines the number of shards to break report & aggregation metric counters into. Increasing
+    /// this value will reduce the amount of database contention during report uploads &
+    /// aggregations, while increasing the cost of getting task metrics.
```
divergentdave, line 129: We should copy this comment over to `binaries/aggregator.rs`, since that field is now used for two kinds of counters.

divergentdave (approved):

aggregator/src/aggregator.rs
inahga: nit: I might even call it `update_task_aggregation_counters` to align with `update_task_upload_counters`.

aggregator/src/aggregator.rs
```diff
@@ -3325,6 +3355,38 @@ impl VdafOps {
...
+    task_id: TaskId,
+    counters: TaskAggregationCounter,
+) {
+    // We write task aggregation counters back in a separate tokio task & datastore transaction,
+    // so that any slowness induced by writing the counters (e.g. due to transaction retry) does
+    // not slow the main processing. The lack of transactionality between writing the updated
+    // aggregation job & updating the counters means that process failure may cause us to leave
+    // some counter updates unaccounted for, but that is an acceptable tradeoff.
+    let ord = thread_rng().gen_range(0..shard_count);
+    tokio::task::spawn(async move {
```
inahga, line 3370: I wonder if we should `spawn()` in the report writer too https://github.com/divviup/janus/blob/main/aggregator/src/aggregator/report_writer.rs#L201. (not in this PR, ofc)

inahga (approved):

tgeoghegan (approved):

aggregator/src/aggregator/test_util.rs
branlwyd: I suppose -- in my testing, this loop often terminated on the first attempt, and failing that, nearly always the second. If 10s is not enough time in a unit test, something has gone pretty wrong.

aggregator/src/aggregator.rs
branlwyd: I agree that `spawn` isn't warranted for the report-counters case, for the reason that David points out. (We may well want to swap the ordering, since one of the hoped-for purposes of writing the counters separately from the reports is that the counters' write latency won't be reflected onto the report upload.)

If anything, I think the task aggregation counters might eventually want a solution more like report writer's, e.g. a background task that receives updates, then batches & writes them as a unit. I'm mildly concerned that if the DB becomes unhealthy, this could lead to a bunch of unmanaged tasks hanging out & trying to write counters -- the background task would ensure that there is only one counter-writing task, I suppose.

PR #3345, opened 2024-07-29
fix: Count batch size via batch_aggregations rather than report_aggregations

### LINKED ISSUE #3323: Fixed size batch can get assigned too many reports if some get garbage collected

inahga:
Fixes #3323.

Track the number of FINISHED report_aggregations from `batch_aggregations.report_count`. This allows us to account for GC, since `batch_aggregations` are not sensitive to GC.

I don't think this is racy. The only possible race I can think of: If an aggregation job is in flight, there must be an accompanying report_aggregations row, as generated by the job creator. That row will be captured in `max_size`. Anyone else have thoughts?

This passes the regression test in https://github.com/divviup/janus/pull/3324.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -4444,7 +4444,8 @@ ON CONFLICT(task_id, batch_identifier, aggregation_param) DO UPDATE
         // Note that this ignores aggregation parameter, as `outstanding_batches` does not need to
         // worry about aggregation parameters.
         //
-        // TODO(#225): reevaluate whether we can ignore aggregation parameter here once we have experience with VDAFs requiring multiple aggregations per batch.
+        // TODO(#225): reevaluate whether we can ignore aggregation parameter here once we have
+        // experience with VDAFs requiring multiple aggregations per batch.
         let stmt = self
             .prepare_cached(
                 "-- put_outstanding_batch()
@@ -4570,7 +4571,7 @@ WHERE task_id = $1
 
         try_join_all(rows.into_iter().map(|row| async move {
             let batch_id = BatchId::get_decoded(row.get("batch_id"))?;
-            let size = self.read_batch_size(task_id, &batch_id).await?;
+            let size = self.read_batch_size(task_info.pkey, &batch_id).await?;
             Ok(OutstandingBatch::new(*task_id, batch_id, size))
         }))
         .await
@@ -4583,26 +4584,30 @@ WHERE task_id = $1
     //    aggregations in the batch which are in a non-failure state (START/WAITING/FINISHED).
     async fn read_batch_size(
         &self,
-        task_id: &TaskId,
+        task_pkey: i64,
         batch_id: &BatchId,
     ) -> Result<RangeInclusive<usize>, Error> {
-        // TODO(#1467): fix this to work in presence of GC.
         let stmt = self
             .prepare_cached(
                 "-- read_batch_size()
-WITH batch_report_aggregation_statuses AS
-    (SELECT report_aggregations.state, COUNT(*) AS count FROM report_aggregations
-     JOIN aggregation_jobs
+WITH report_aggregations_count AS (
+    SELECT COUNT(*) as count FROM report_aggregations
+    JOIN aggregation_jobs
         ON report_aggregations.aggregation_job_id = aggregation_jobs.id
-     WHERE aggregation_jobs.task_id = (SELECT id FROM tasks WHERE task_id = $1)
-     AND report_aggregations.task_id = aggregation_jobs.task_id
-     AND aggregation_jobs.batch_id = $2
-     GROUP BY report_aggregations.state)
+    WHERE aggregation_jobs.task_id = $1
+    AND report_aggregations.task_id = aggregation_jobs.task_id
+    AND aggregation_jobs.batch_id = $2
+    AND report_aggregations.state in ('START', 'WAITING')
+),
+batch_aggregation_count AS (
+    SELECT SUM(report_count) as count FROM batch_aggregations
+    WHERE batch_aggregations.task_id = $1
+    AND batch_aggregations.batch_identifier = $2
+)
 SELECT
-    (SELECT SUM(count)::BIGINT FROM batch_report_aggregation_statuses
-     WHERE state IN ('FINISHED')) AS min_size,
-    (SELECT SUM(count)::BIGINT FROM batch_report_aggregation_statuses
-     WHERE state IN ('START', 'WAITING', 'FINISHED')) AS max_size",
+    (SELECT count FROM batch_aggregation_count)::BIGINT AS min_size,
+    (SELECT count FROM report_aggregations_count)::BIGINT
+        + (SELECT count FROM batch_aggregation_count)::BIGINT AS max_size",
             )
             .await?;
 
@@ -4610,7 +4615,7 @@ SELECT
             .query_one(
                 &stmt,
                 &[
-                    /* task_id */ task_id.as_ref(),
+                    /* task_id */ &task_pkey,
                     /* batch_id */ batch_id.as_ref(),
                 ],
             )
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -4600,7 +4600,7 @@ WITH report_aggregations_count AS (
     AND report_aggregations.state in ('START', 'WAITING')
 ),
 batch_aggregation_count AS (
-    SELECT SUM(report_count) as count FROM batch_aggregations
+    SELECT SUM(report_count) AS count FROM batch_aggregations
     WHERE batch_aggregations.task_id = $1
     AND batch_aggregations.batch_identifier = $2
 )
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -4591,7 +4591,7 @@ WHERE task_id = $1
             .prepare_cached(
                 "-- read_batch_size()
 WITH report_aggregations_count AS (
-    SELECT COUNT(*) as count FROM report_aggregations
+    SELECT COUNT(*) AS count FROM report_aggregations
     JOIN aggregation_jobs
         ON report_aggregations.aggregation_job_id = aggregation_jobs.id
     WHERE aggregation_jobs.task_id = $1
--- a/aggregator_core/src/datastore/tests.rs
+++ b/aggregator_core/src/datastore/tests.rs
@@ -5176,7 +5176,10 @@ async fn roundtrip_outstanding_batch(ephemeral_datastore: EphemeralDatastore) {
                         .unwrap(),
                     BatchAggregationState::Aggregating {
                         aggregate_share: Some(dummy::AggregateShare(0)),
-                        report_count: 0,
+                        // Let report_count be 1 without an accompanying report_aggregation in a
+                        // terminal state. This captures the case where a FINISHED report_aggregation
+                        // was garbage collected and no longer exists in the database.
+                        report_count: 1,
                         checksum: ReportIdChecksum::default(),
                         aggregation_jobs_created: 4,
                         aggregation_jobs_terminated: 1,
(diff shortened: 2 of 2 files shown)


aggregator_core/src/datastore/tests.rs
```diff
@@ -5176,6 +5176,9 @@ async fn roundtrip_outstanding_batch(ephemeral_datastore: EphemeralDatastore) {
                         .unwrap(),
                     BatchAggregationState::Aggregating {
                         aggregate_share: Some(dummy::AggregateShare(0)),
+                        // Let report_count be 1 without an accompanying report_aggregation in a
+                        // terminal state. This captures the case where a FINISHED report_aggregation
+                        // was garbage collected and no longer exists in the database.
                         report_count: 1,
```
inahga, line 5182: I believe this test was actually somewhat reproducing the bug, except our assertion was wrong.

This batch doesn't have an accompanying report_aggregation in a FINISHED state to go along with the `report_count: 1`. We only have one in a START state. So in the test we were checking that the range was `[0, 1]` where it should have been `[1, 2]`.

inahga:
Draft for now while I cook up a query plan, in the mean time if anyone has thoughts on the approach...

inahga:
```
2024-07-31 19:17:39.503 UTC [57] LOG:  duration: 0.045 ms  plan:
        Query Text: -- read_batch_size()
        WITH report_aggregations_count AS (
            SELECT COUNT(*) as count FROM report_aggregations
            JOIN aggregation_jobs
                ON report_aggregations.aggregation_job_id = aggregation_jobs.id
            WHERE aggregation_jobs.task_id = $1
            AND report_aggregations.task_id = aggregation_jobs.task_id
            AND aggregation_jobs.batch_id = $2
            AND report_aggregations.state in ('START', 'WAITING')
        ),
        batch_aggregation_count AS (
            SELECT SUM(report_count) as count FROM batch_aggregations
            WHERE batch_aggregations.task_id = $1
            AND batch_aggregations.batch_identifier = $2
        )
        SELECT
            (SELECT count FROM batch_aggregation_count)::BIGINT AS min_size,
            (SELECT count FROM report_aggregations_count)::BIGINT
                + (SELECT count FROM batch_aggregation_count)::BIGINT AS max_size
        Result  (cost=24.58..24.60 rows=1 width=16) (actual time=0.044..0.044 rows=1 loops=1)
          CTE batch_aggregation_count
            ->  Aggregate  (cost=8.17..8.18 rows=1 width=32) (actual time=0.010..0.010 rows=1 loops=1)
                  ->  Index Scan using batch_aggregations_gc_time on batch_aggregations  (cost=0.14..8.16 rows=1 width=8) (actual time=0.007..0.008 rows=2 loops=1)
                        Index Cond: (task_id = '1'::bigint)
                        Filter: (batch_identifier = '\x040d48034f8e8b127e51d7dedfb3cc4b9087e4916f7bc303ac1dd34b57fab45b'::bytea)
          InitPlan 2 (returns $1)
            ->  CTE Scan on batch_aggregation_count  (cost=0.00..0.02 rows=1 width=32) (actual time=0.011..0.011 rows=1 loops=1)
          InitPlan 3 (returns $2)
            ->  Aggregate  (cost=16.34..16.35 rows=1 width=8) (actual time=0.030..0.030 rows=1 loops=1)
                  ->  Nested Loop  (cost=0.29..16.34 rows=1 width=0) (actual time=0.027..0.029 rows=2 loops=1)
                        Join Filter: (report_aggregations.aggregation_job_id = aggregation_jobs.id)
                        ->  Index Scan using report_aggregations_unique_ord on report_aggregations  (cost=0.14..8.16 rows=1 width=16) (actual time=0.007..0.009 rows=2 loops=1)
                              Index Cond: (task_id = '1'::bigint)
                              Filter: (state = ANY ('{START,WAITING}'::report_aggregation_state[]))
                              Rows Removed by Filter: 4
                        ->  Index Scan using aggregation_jobs_task_and_client_timestamp_interval on aggregation_jobs  (cost=0.14..8.16 rows=1 width=16) (actual time=0.010..0.010 rows=1 loops=2)
                              Index Cond: (task_id = '1'::bigint)
                              Filter: (batch_id = '\x040d48034f8e8b127e51d7dedfb3cc4b9087e4916f7bc303ac1dd34b57fab45b'::bytea)
          InitPlan 4 (returns $3)
            ->  CTE Scan on batch_aggregation_count batch_aggregation_count_1  (cost=0.00..0.02 rows=1 width=32) (actual time=0.000..0.000 rows=1 loops=1)
```

That's... a strange index to be using.
```
CREATE INDEX batch_aggregations_gc_time ON batch_aggregations(task_id, UPPER(COALESCE(batch_interval, client_timestamp_interval)));
```

divergentdave:
Since this plan was made using `enable_seqscan = false`, and it expected to select one row out of a very small table, I assume that all the candidate index scan query fragments had very similar costs. Each would have scanned one index page, and then done one random access heap page read. The choice of index used would thus be pretty arbitrary, and it may change from run to run depending on implementation details. I think we should avoid `enable_seqscan = false` in this situation; I don't think there's a good substitute for running with bigger tables.

inahga:
I think you're right. If I remove the `batch_aggregations_gc_time` index, it'll choose the:
```sql
CONSTRAINT batch_aggregations_unique_task_id_batch_id_aggregation_param UNIQUE(task_id, batch_identifier, aggregation_param, ord)
```
index, which is much more in line with what I expect.

(FWIW with enable_seqscan=true it also chooses the strange index, but yeah, you're right a large table is needed to get a true representation here).

divergentdave:
Hmm, interesting. In that case, the tuple width of the different indexes could be putting a thumb on the scale. Maybe the uniqueness index is a few pages longer than the GC index. I expect we'd only see pressure to choose the uniqueness index once we need both the selectivity of the task_id condition and the selectivity of the batch_identifier condition to push the number of heap tuples scanned down, enough that the savings due to better selectivity outweigh the index scan, etc.

inahga:
Testing against our load testing database, looks reasonable.
```
postgres=> EXPLAIN ANALYZE WITH report_aggregations_count AS (
    SELECT COUNT(*) as count FROM report_aggregations
    JOIN aggregation_jobs
        ON report_aggregations.aggregation_job_id = aggregation_jobs.id
    WHERE aggregation_jobs.task_id = 98
    AND report_aggregations.task_id = aggregation_jobs.task_id
    AND aggregation_jobs.batch_id = '\xcc586a332114e6966de8e8d59d903e5e2b4bacb27f1c3ae987ce613779c6ce8a'::BYTEA
    AND report_aggregations.state in ('START', 'WAITING')
),
batch_aggregation_count AS (
    SELECT SUM(report_count) as count FROM batch_aggregations
    WHERE batch_aggregations.task_id = 100
    AND batch_aggregations.batch_identifier = '\xcc586a332114e6966de8e8d59d903e5e2b4bacb27f1c3ae987ce613779c6ce8a'::BYTEA
)
SELECT
    (SELECT count FROM batch_aggregation_count)::BIGINT AS min_size,
    (SELECT count FROM report_aggregations_count)::BIGINT
        + (SELECT count FROM batch_aggregation_count)::BIGINT AS max_size;
                                                                                          QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Result  (cost=13.75..13.77 rows=1 width=16) (actual time=0.386..0.389 rows=1 loops=1)
   CTE batch_aggregation_count
     ->  Aggregate  (cost=3.77..3.78 rows=1 width=32) (actual time=0.046..0.047 rows=1 loops=1)
           ->  Index Scan using batch_aggregations_unique_task_id_batch_id_aggregation_param on batch_aggregations  (cost=0.42..3.76 rows=2 width=8) (actual time=0.043..0.043 rows=0 loops=1)
                 Index Cond: ((task_id = 100) AND (batch_identifier = '\xcc586a332114e6966de8e8d59d903e5e2b4bacb27f1c3ae987ce613779c6ce8a'::bytea))
   InitPlan 2 (returns $1)
     ->  CTE Scan on batch_aggregation_count  (cost=0.00..0.02 rows=1 width=32) (actual time=0.048..0.049 rows=1 loops=1)
   InitPlan 3 (returns $3)
     ->  Aggregate  (cost=9.91..9.92 rows=1 width=8) (actual time=0.333..0.333 rows=1 loops=1)
           ->  Nested Loop  (cost=0.98..9.91 rows=1 width=0) (actual time=0.332..0.332 rows=0 loops=1)
                 ->  Index Scan using aggregation_jobs_task_and_batch_id on aggregation_jobs  (cost=0.42..2.64 rows=1 width=16) (actual time=0.012..0.028 rows=12 loops=1)
                       Index Cond: ((task_id = 98) AND (batch_id = '\xcc586a332114e6966de8e8d59d903e5e2b4bacb27f1c3ae987ce613779c6ce8a'::bytea))
                 ->  Index Scan using report_aggregations_unique_ord on report_aggregations  (cost=0.56..7.26 rows=1 width=16) (actual time=0.025..0.025 rows=0 loops=12)
                       Index Cond: ((task_id = 98) AND (aggregation_job_id = aggregation_jobs.id))
                       Filter: (state = ANY ('{START,WAITING}'::report_aggregation_state[]))
                       Rows Removed by Filter: 40
   InitPlan 4 (returns $4)
     ->  CTE Scan on batch_aggregation_count batch_aggregation_count_1  (cost=0.00..0.02 rows=1 width=32) (actual time=0.000..0.001 rows=1 loops=1)
 Planning Time: 0.640 ms
 Execution Time: 0.472 ms
```

aggregator_core/src/datastore.rs
```diff
@@ -4583,34 +4584,38 @@ WHERE task_id = $1
...
     ) -> Result<RangeInclusive<usize>, Error> {
-        // TODO(#1467): fix this to work in presence of GC.
         let stmt = self
             .prepare_cached(
                 "-- read_batch_size()
-WITH batch_report_aggregation_statuses AS
-    (SELECT report_aggregations.state, COUNT(*) AS count FROM report_aggregations
-     JOIN aggregation_jobs
+WITH report_aggregations_count AS (
+    SELECT COUNT(*) as count FROM report_aggregations
```
branlwyd, line 4594:
```suggestion
    SELECT COUNT(*) AS count FROM report_aggregations
```

super-nit: our current standard is that SQL keywords are capitalized, let's stick to that for now.

aggregator_core/src/datastore.rs
```diff
@@ -4583,34 +4584,38 @@ WHERE task_id = $1
...
-     AND report_aggregations.task_id = aggregation_jobs.task_id
-     AND aggregation_jobs.batch_id = $2
-     GROUP BY report_aggregations.state)
+    WHERE aggregation_jobs.task_id = $1
+    AND report_aggregations.task_id = aggregation_jobs.task_id
+    AND aggregation_jobs.batch_id = $2
+    AND report_aggregations.state in ('START', 'WAITING')
+),
+batch_aggregation_count AS (
+    SELECT SUM(report_count) as count FROM batch_aggregations
```
branlwyd, line 4603:
```suggestion
    SELECT SUM(report_count) AS count FROM batch_aggregations
```

super-nit: same as above

branlwyd (approved):
I think this is not racy: the update to `report_aggregations` to move into the `FINISHED` state is transactional with the update to `batch_aggregations` to include those report aggregations in the batch aggregation. (another way to look at it: if these updates were not transactional, an inopportunely-timed crash could leave the batch in a state where the aggregation job is considered completed but not recorded in the batch aggregations, or vice-versa.)

aggregator_core/src/datastore.rs
inahga: Whoops!

PR #3347, opened 2024-07-31
aggregator_core: Allow dumping query plans to stdout

inahga:
Makes it easier to inspect query plans while working locally. These query plans do have to be taken with a grain of salt, because they're always running against an empty database and with `enable_seqscan=false`, so they're not entirely representative.

We dump to stdout because the output from tracing is fairly incomprehensible, since each node of the query plan is on its own tracing event.

### DIFF
--- a/aggregator_core/README.md
+++ b/aggregator_core/README.md
@@ -9,8 +9,7 @@ It is not published to crates.io and should not be depended on directly by users
 ## PostgreSQL Logs
 
 During tests, you can have PostgreSQL dump its logs to stdout by setting
-`JANUS_TEST_DUMP_POSTGRESQL_LOGS`. The test database is set to log all query plans. You should only
-run this option with one test at a time, otherwise you might not get all the logs.
+`JANUS_TEST_DUMP_POSTGRESQL_LOGS`. The test database is set to log all query plans.
 
 Example:
 ```
--- a/aggregator_core/README.md
+++ b/aggregator_core/README.md
@@ -1,0 +1,18 @@
+# `aggregator_core`
+
+`aggregator_core` contains helper code for the `aggregator` crate. It mainly consists of data
+structures that are only pertinent to a DAP aggregator, and subroutines for talking to a PostgreSQL
+database.
+
+It is not published to crates.io and should not be depended on directly by users of Janus.
+
+## PostgreSQL Logs
+
+During tests, you can have PostgreSQL dump its logs to stdout by setting
+`JANUS_TEST_DUMP_POSTGRESQL_LOGS`. The test database is set to log all query plans. You should only
+run this option with one test at a time, otherwise you might not get all the logs.
+
+Example:
+```
+JANUS_TEST_DUMP_POSTGRESQL_LOGS= RUST_LOG=debug cargo test datastore::tests::roundtrip_task::case_1 -- --exact --nocapture
+```
--- a/aggregator_core/src/datastore/test_util.rs
+++ b/aggregator_core/src/datastore/test_util.rs
@@ -70,36 +70,32 @@ impl EphemeralDatabase {
                 .build()
                 .unwrap()
                 .block_on(async move {
-                    // Start an instance of Postgres running in a container.
-                    let db_container = ContainerRequest::from(Postgres::default())
-                        .with_cmd(Self::postgres_configuration(&[
-                            // Many tests running concurrently can overwhelm the available,
-                            // connections, so bump the limit.
-                            "max_connections=200",
+                    let should_log = env::var_os("JANUS_TEST_DUMP_POSTGRESQL_LOGS").is_some();
+                    let mut configuration = Vec::from(["max_connections=200"]);
+                    if should_log {
+                        configuration.append(&mut Vec::from([
                             // Enable logging of query plans.
                             "shared_preload_libraries=auto_explain",
                             "log_min_messages=LOG",
                             "auto_explain.log_min_duration=0",
                             "auto_explain.log_analyze=true",
-                            // Discourage postgres from doing sequential scans, so we can analyze
-                            // whether we have appropriate indexes in unit tests. Because test
-                            // databases are not seeded with data, the query planner will often
-                            // choose a sequential scan because it would be faster than hitting an
-                            // index.
-                            "enable_seqscan=false",
                         ]))
+                    }
+
+                    // Start an instance of Postgres running in a container.
+                    let db_container = ContainerRequest::from(Postgres::default())
+                        .with_cmd(Self::postgres_configuration(&configuration))
                         .start()
                         .await
                         .unwrap();
 
                     let stdout = db_container.stdout(true);
                     let stderr = db_container.stderr(true);
-                    let log_consumer_handle =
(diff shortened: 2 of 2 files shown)


aggregator_core/README.md
```diff
@@ -0,0 +1,18 @@
...
+structures that are only pertinent to a DAP aggregator, and subroutines for talking to a PostgreSQL
+database.
+
+It is not published to crates.io and should not be depended on directly by users of Janus.
+
+## PostgreSQL Logs
+
+During tests, you can have PostgreSQL dump its logs to stdout by setting
+`JANUS_TEST_DUMP_POSTGRESQL_LOGS`. The test database is set to log all query plans. You should only
+run this option with one test at a time, otherwise you might not get all the logs.
```
inahga, line 13: I suspect this is a testcontainers bug--if running multiple tests at at time, the output is incomplete, and perhaps interspersed with other containers. IMO it's not a big deal, since we'd only care to run this against one test at a time, so I don't care to spend too much time on it.

inahga:
Eh, I think there's a more subtle synchronization error here, as sometimes I see incomplete results even when doing one test.

aggregator_core/README.md
divergentdave: Actually, we run multiple tests against the same container, to save on startup time, and separate by using different databases within the same process. Thus, database logs from queries in unrelated test aggregators will be commingled. It's also possible that we may shut down one database, and start another, if all the relevant Arcs go out of scope simultaneously. I wonder if there's some sort of race in the testcontainers library where deleting a container interrupts log forwarding, and if killing the container first, then deleting it once the log stream is done, would fix this.

inahga: Good point about sharing a container (I had thought it was 1:1, since I didn't read the code closely, and because I think the phenomena you describe about Arcs going out of scope applies, showing as multiple postgres containers in `docker ps`).

[It looks like the log consumer task is spawned via `tokio::spawn`](https://github.com/testcontainers/testcontainers-rs/blob/main/testcontainers/src/core/containers/async_container.rs#L71), and the corresponding handle is never joined, so testcontainers never waits for the log stream to be drained before the process is exited. So I think a bunch of log lines get queued up, but never get serviced because the process ends too quickly.

Indeed, if I tack a `sleep()` at the end of an affected test, I see a much more plausible volume of output.

inahga:
Yeesh, async drop is a hard problem. I've written an attempt at fixing synchronization without sleeps, done by spawning all testcontainers work in its own thread with its own tokio runtime.

I don't think it's possible for testcontainers to be in the same main thread and wait on logs, because blocking waiting for logs starves the logic that would terminate the log loop, i.e. a deadlock.

I think this hack is small enough to avoid pulling in something like the `async_drop` crate. I do think a bug report for testcontainers is justified over synchronization of logs, and I think [they have some logic to make it a (hopefully) easy enough change](https://github.com/testcontainers/testcontainers-rs/blob/main/testcontainers/src/core/async_drop.rs).

aggregator_core/src/datastore/test_util.rs
```diff
@@ -17,20 +17,30 @@ use sqlx::{
...
+};
 use tokio_postgres::{connect, Config, NoTls};
 use tracing::trace;
 
 use super::SUPPORTED_SCHEMA_VERSIONS;
 
 struct EphemeralDatabase {
-    _db_container: ContainerAsync<Postgres>,
+    db_thread: Option<JoinHandle<()>>,
+    db_thread_shutdown: Option<Sender<()>>,
```
inahga, line 43: Is there a better data structure for `.take()`'ing something in a `Drop`? Option works fine enough, but maybe there's a more pertinent structure I'm missing?

aggregator_core/src/datastore/test_util.rs
```diff
@@ -49,24 +59,73 @@ impl EphemeralDatabase {
...
+                        .await
+                        .unwrap();
+                    trace!("Postgres container is up with port {port_number}");
+
+                    // Send port information, which frees this function to continue.
+                    port_tx.send(port_number).unwrap();
+
+                    // Wait for shutdown to be signalled.
+                    shutdown_rx.await.unwrap();
+                    drop(db_container);
```
divergentdave, line 116: I'd recommend trying out calling `ContainerAsync::stop()` first, then waiting for the log consumers to finish, then dropping to delete the container. If the log consumer streams stop once the container has stopped, then that should work , and it will prevent race conditions leading to lost logs.

aggregator_core/src/datastore/test_util.rs
```diff
@@ -49,24 +59,73 @@ impl EphemeralDatabase {
...
-            .unwrap();
-        trace!("Postgres container is up with port {port_number}");
-
+        let (shutdown_tx, shutdown_rx) = oneshot::channel();
+        let (port_tx, port_rx) = oneshot::channel();
+
+        // Hack: run testcontainer logic under its own thread with its own tokio runtime, to avoid
+        // deadlocking the main runtime when waiting for logs to finish in the Drop implementation.
+        let db_thread = std::thread::spawn(move || {
+            tokio::runtime::Builder::new_current_thread()
```
divergentdave, line 68: This approach to resolving the deadlock LGTM. I suspect another way to solve it would be with `[tokio::test(flavor = "multi_threaded")]` everywhere, but that would be very intrusive.

aggregator_core/src/datastore/test_util.rs
inahga: Nice, that appears to work and is much cleaner, insofar as I still see the `database system is shut down` at the end of logs.

aggregator_core/src/datastore/test_util.rs
```diff
@@ -49,24 +59,73 @@ impl EphemeralDatabase {
...
+                            "shared_preload_libraries=auto_explain",
+                            "log_min_messages=LOG",
+                            "auto_explain.log_min_duration=0",
+                            "auto_explain.log_analyze=true",
+                            // Discourage postgres from doing sequential scans, so we can analyze
+                            // whether we have appropriate indexes in unit tests. Because test
+                            // databases are not seeded with data, the query planner will often
+                            // choose a sequential scan because it would be faster than hitting an
+                            // index.
+                            "enable_seqscan=false",
```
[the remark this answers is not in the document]
inahga: Surprisingly, spot checking a few tests, postgres still largely prefers index scans, so I think `enable_seqscan` was too presumptuous on my part.

No complaints from me, it points to these empty DB query plans still having a reasonable amount of signal.

aggregator_core/src/datastore/test_util.rs
inahga: I think you're right, if all tests run multithreaded, then maybe we can use `block_in_place()`. But yeah having to decorate every test would be irritating--we would probably be better off writing our own macro wrapper in that case. (I wonder why tokio tests default to the current thread runtime anyways.)

inahga:
https://github.com/testcontainers/testcontainers-rs/issues/719 for the testcontainers problem.

branlwyd (approved):

divergentdave (approved):

PR #3357, opened 2024-08-06
Add validation that min_batch_size is nonzero

divergentdave:
This adds a validation that `min_batch_size` is not zero. Without this, we can wind up taking this code path in the collection job driver: https://github.com/divviup/janus/blob/3258350fabbd6b1be8ad48637d7fbdd381893742/aggregator/src/aggregator/aggregate_share.rs#L98-L101

The following simulation test input reproduces this, based on an automatically-generated input.

```rust
Input {
    is_fixed_size: true,
    config: Config {
        time_precision: Duration::from_seconds(3600),
        min_batch_size: 0,
        max_batch_size: Some(255),
        batch_time_window_size: None,
        report_expiry_age: None,
        min_aggregation_job_size: 1,
        max_aggregation_job_size: 76,
    },
    ops: Vec::from([
        Op::Upload {
            report_time: Time::from_seconds_since_epoch(1700000903),
            count: 13,
        },
        Op::AggregationJobCreator,
        Op::AggregationJobDriverResponseError,
        Op::AdvanceTime {
            amount: Duration::from_seconds(27041),
        },
        Op::AggregationJobDriverResponseError,
        Op::AdvanceTime {
            amount: Duration::from_seconds(8541),
        },
        Op::CollectorStart {
            collection_job_id: random(),
            query: Query::FixedSizeCurrentBatch,
        },
        Op::AggregationJobDriverResponseError,
        Op::CollectionJobDriverResponseError,
    ]),
}
```

This looks like a variant of the test case showing that abandoning an aggregation job leads to a batch mismatch, but the collection job driver encounters this error before sending an aggregate share request. Instead, collection jobs with 0 aggregated reports can get picked up once the aggregation job is abandoned, and then they tick up their error counter.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_driver/tests.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver/tests.rs
@@ -54,6 +54,7 @@ use prio::{
 };
 use rand::random;
 use std::{sync::Arc, time::Duration as StdDuration};
+use tokio::time::timeout;
 use trillium_tokio::Stopper;
 
 #[tokio::test]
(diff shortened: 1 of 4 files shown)


divergentdave:
Changing the default min_batch_size in test tasks required some further changes, because various collection job tests were relying on the min_batch_size of zero. I added more fixture rows to the database where needed to set up batches that were collectable.

I also checked our staging and prod databases to confirm that no existing tasks have a min_batch_size of zero. (after this change, they would have become unusable due to validation errors upon reading from the database) Note that divviup-api's checks prevent a min_batch_size below 100, though there are other ways to create tasks without that check.

branlwyd (approved):

PR #3360, opened 2024-08-08
Use tmpfs in database container

### LINKED ISSUE #3359: Speed up database used in tests
I've seen that starting up the database process takes around four to five seconds locally, while setting up a new database and applying the schema takes a bit under one second. Cutting these times down could be a nice improvement to our test throughput.

Since the ephemeral databases used in tests don't need to be durable, we could LD_PRELOAD libeatmydata inside the database container, in order to make fsyncs no-ops. We'd need a Dockerfile that inherits from the postgres image, installs `libeatmydata-131-r1`, and sets the LD_PRELOAD environment variable.

Alternately, we could mount a tmpfs volume and point the image at it with the `PGDATA` environment variable. This would speed up all disk operations significantly, so long as we don't run out of memory.

divergentdave:
This mounts a tmpfs volume into test database containers, over where the Postgres data directory will be. Closes #3359. I saw the following runtime improvements locally.

Tests | Before | After
-------|-----------|-----------
All | 2m57s | 2m25s
`-p janus_aggregator --lib` | 82s | 68s
`-p janus_integration_tests --test integration` | 29s | 24s

### DIFF
--- a/aggregator_core/src/datastore/test_util.rs
+++ b/aggregator_core/src/datastore/test_util.rs
@@ -24,7 +24,7 @@ use std::{
     thread::JoinHandle,
     time::Duration,
 };
-use testcontainers::{runners::AsyncRunner, ContainerRequest, ImageExt};
+use testcontainers::{core::Mount, runners::AsyncRunner, ContainerRequest, ImageExt};
 use tokio::{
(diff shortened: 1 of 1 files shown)


inahga (approved):
Nice find.

branlwyd (approved):

PR #3361, opened 2024-08-09
Ephemeral database cleanup followup

divergentdave:
This is a follow-up to #3358, which was not actually deleting databases. First, we need to fix the SQL statement. The database name is an identifier, not a string, so we can't template it in as a parameter. This was producing the following error, which was being intentionally swallowed to handle container deletion races.

```
2024-08-09 19:41:26.520 UTC [30587] ERROR:  syntax error at or near "$1" at character 15
2024-08-09 19:41:26.520 UTC [30587] STATEMENT:  DROP DATABASE $1
```

Beyond that, we can't drop the database that the connection is using, so we have to spin up a fresh connection to the `postgres` database.

```
2024-08-09 19:52:39.758 UTC [843] ERROR:  cannot drop the currently open database
2024-08-09 19:52:39.758 UTC [843] STATEMENT:  DROP DATABASE janus_test_c71da38e6d7fe51b10f5c713d1130a0a
```

Lastly, dropping the database can take some time, and it will almost always happen near the end of a unit test. Thus, the Tokio runtime will be shutting down soon, which cancels the operation. This is fixed by spinning up a smaller runtime on its own thread, like we recently did for log forwarding. In this case, we don't need to block the `Drop` impl on the spawned operation, because we're just dropping the database for resource utilization reasons.

### DIFF
--- a/aggregator_core/src/datastore/test_util.rs
+++ b/aggregator_core/src/datastore/test_util.rs
@@ -27,7 +27,7 @@ use std::{
 use testcontainers::{core::Mount, runners::AsyncRunner, ContainerRequest, ImageExt};
 use tokio::{
     io::{AsyncBufRead, AsyncBufReadExt},
-    join, spawn,
+    join,
     sync::{
         oneshot::{self, Sender},
         Mutex,
@@ -263,17 +263,31 @@ impl Drop for EphemeralDatastore {
         // this moot.
         let db_name = self.db_name.clone();
         let connection_string = self.db.connection_string("postgres");
-        spawn(async move {
-            if let Ok((client, connection)) =
-                tokio_postgres::connect(&connection_string, NoTls).await
-            {
-                spawn(async move {
-                    let _ = connection.await;
-                });
-                let _ = client
-                    .execute(&format!("DROP DATABASE {db_name}"), &[])
-                    .await;
-            }
+
+        // Since each unit test has its own Tokio runtime, we need to run `DROP DATABASE` in a
+        // separate thread and Tokio runtime. If we just spawned a new task to do so in the same
+        // runtime, it would get cancelled when the runtime shuts down.
+        std::thread::spawn(move || {
+            let runtime = tokio::runtime::Builder::new_multi_thread()
+                .worker_threads(2)
+                .enable_all()
+                .build()
+                .unwrap();
+            runtime.block_on({
+                let handle = runtime.handle();
+                async move {
+                    if let Ok((client, connection)) =
+                        tokio_postgres::connect(&connection_string, NoTls).await
+                    {
+                        handle.spawn(async move {
+                            let _ = connection.await;
+                        });
+                        let _ = client
+                            .execute(&format!("DROP DATABASE {db_name}"), &[])
+                            .await;
+                    }
+                }
+            });
         });
     }
 }
--- a/aggregator_core/src/datastore/test_util.rs
+++ b/aggregator_core/src/datastore/test_util.rs
@@ -161,7 +161,7 @@ impl Drop for EphemeralDatabase {
 ///
 /// Dropping the EphemeralDatastore will cause it to be shut down & cleaned up.
 pub struct EphemeralDatastore {
-    _db: Arc<EphemeralDatabase>,
+    db: Arc<EphemeralDatabase>,
     connection_string: String,
     pool: Pool,
     datastore_key_bytes: Vec<u8>,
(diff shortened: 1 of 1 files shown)


aggregator_core/src/datastore/test_util.rs
```diff
@@ -261,12 +261,33 @@ impl Drop for EphemeralDatastore {
...
-                let _ = conn.execute("DROP DATABASE $1", &[&db_name]).await;
-            }
+        let connection_string = self.db.connection_string("postgres");
+
+        // Since each unit test has its own Tokio runtime, we need to run `DROP DATABASE` in a
+        // separate thread and Tokio runtime. If we just spawned a new task to do so in the same
+        // runtime, it would get cancelled when the runtime shuts down.
+        std::thread::spawn(move || {
+            let runtime = tokio::runtime::Builder::new_multi_thread()
+                .worker_threads(2)
```
inahga, line 272: It doesn't really matter, but any particular reason that a current_thread runtime wouldn't work here?

inahga (approved):

aggregator_core/src/datastore/test_util.rs
```diff
@@ -261,12 +261,33 @@ impl Drop for EphemeralDatastore {
...
         // this moot.
-        let pool = self.pool.clone();
         let db_name = self.db_name.clone();
-        spawn(async move {
-            if let Ok(conn) = pool.get().await {
-                let _ = conn.execute("DROP DATABASE $1", &[&db_name]).await;
-            }
+        let connection_string = self.db.connection_string("postgres");
+
+        // Since each unit test has its own Tokio runtime, we need to run `DROP DATABASE` in a
```
branlwyd, line 267: Starting a new tokio runtime & connecting to the database seems kinda heavyweight for a cleanup-on-drop operation.

Could `EphemeralDatabase` maintain a single background thread/task (running on its own runtime, if necessary), which holds a connection open to the default `postgres` DB, and receives DB names to drop over an MPSC channel? (admittedly, this is getting kind of complicated for a cleanup operation.)

branlwyd (approved):

aggregator_core/src/datastore/test_util.rs
divergentdave: A current thread runtime would have worked when using one runtime per database, but it doesn't work anymore after I rewrote it to spawn fewer threads. The current thread runtime only makes progress if at least one thread is calling `block_on()`, but there's no longer a natural place to do so. The long-lived cleanup thread instead blocks on reading from a `std::sync::mpsc` channel, then spawns async tasks. In turn, I'm using this channel instead of `tokio::sync::mpsc` because `drop()` is a synchronous function called within a Tokio context, so `tokio::sync::mpsc::Sender::blocking_send()` is off-limits.

aggregator_core/src/datastore/test_util.rs
```diff
@@ -120,10 +122,47 @@ impl EphemeralDatabase {
...
+                .build()
+                .unwrap();
+            let handle = runtime.handle();
+
+            let connection_string =
+                format!("postgres://postgres:postgres@127.0.0.1:{port_number}/postgres");
+
+            while let Ok(name) = cleanup_rx.recv() {
+                let connection_string = connection_string.clone();
+                handle.spawn({
```
branlwyd, line 143: optional: I would have connected to the DB once, outside of the recv loop, and kept only the `DROP DATABASE` inside the recv loop without spawning a separate task for each received DB name. We might end up buffering a few database names if dropping a database takes a while, but we'll already effectively do that in the tasks we generate if connecting takes a while. This would more effectively control concurrency, at the cost of the DB cleanup work sharing fate if e.g. the cleanup connection goes bad.

This doesn't need to be addressed (unless it nerd-snipes you) -- I don't want to take too much time on this cleanup, and I don't think a few extra tasks being generated during tests are going to meaningfully harm performance/resource usage.

branlwyd (approved):

aggregator_core/src/datastore/test_util.rs
divergentdave: I'd prefer to keep it this way, so we can pipeline `DROP DATABASE` commands inside the database. I tried dropping one database manually, while the container was loaded with other tests, and it took a few seconds. Unit tests can create several databases per second, and a simulation test loop can create a database about once per second. My concern is that serializing `DROP DATABASE` statements would lead to long-running simulation test runs falling behind in cleanup, and then hitting a resource limit. This can be painful, because a whole run can take a half hour, especially if a failing input is found and needs to be shrunk, and the database container becomes unusable by subsequent iterations.

PR #3381, opened 2024-08-28
Update sqlx to 0.8.1

inahga:
Fixes security advisory https://github.com/divviup/janus/security/dependabot/41. It's not really clear to me why dependabot doesn't want to take this update https://github.com/divviup/janus/actions/runs/10459997889/job/28965218103, but so be it.


inahga:
Ah, I see, the MSRV of `sqlx` was (mistakenly?) bumped to 1.80. https://github.com/launchbadge/sqlx/issues/3460#issuecomment-2308435214

We should wait for a release that contains https://github.com/launchbadge/sqlx/pull/3459, and I think dependabot should take it then.

PR #3406, opened 2024-09-11
Fix leak of channel sender

### LINKED ISSUE #3405: FD leak in queue tests
I'm running into test failures when running the test suite locally. I see "Too many open files" first, then a series of "metrics provider already shut down" messages. I bisected this to #3369, then bisected to 1bc432aaf540e313081f569cab8867e13548c0d7 within that. Interestingly, this is passing just fine on CI machines, so the number of threads and/or default ulimit settings must play a factor. Running one quickcheck test is not enough to trigger the issue, but running all four is.

I tried stracing these tests, and it looks like it fails after using 1024 file descriptors. I think about half are attributable to `fcntl(fd, F_DUPFD_CLOEXEC, 3)` calls, and about half are attributable to `eventfd2(0, EFD_CLOEXEC|EFD_NONBLOCK)`, though it's hard to follow FDs from opening to closing just based on logs. The `/proc/$PID/fd` listing shows a lot of `anon_inode:[eventfd]` and `anon_inode:[eventpoll]` plus some sockets.

divergentdave:
This fixes a leak of a channel sender, by removing a `forget()` call. We shouldn't leak this sender because the dispatcher task is supposed to run until all senders are dropped. Furthermore, some of our tests amplify this leak by creating and tearing down many queues, which causes file descriptor exhaustion for me locally. Closes #3405.

### DIFF
--- a/aggregator/src/aggregator/queue.rs
+++ b/aggregator/src/aggregator/queue.rs
@@ -1,6 +1,5 @@
 use std::{
     collections::BTreeMap,
-    mem::forget,
     sync::{
         atomic::{AtomicU64, Ordering},
         Arc,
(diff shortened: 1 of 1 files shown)


inahga (approved):
Nice find!

branlwyd (approved):
Nice indeed -- apologies for the bad suggestion (to use `mem::forget`).

PR #3414, opened 2024-09-18
docs: Avoid improperly using janus_messages::{Time, Duration}

### LINKED ISSUE #445: `janus_core::message::{Time,Duration}` should be used only for message serialization/deserialization

inahga:
Closes #445

### DIFF
--- a/docs/DEVELOPMENT.md
+++ b/docs/DEVELOPMENT.md
@@ -130,9 +130,10 @@ To create a new migration:
   data type (possibly expensive). This will also catch any mistakes where you
   intended to clone the Arc and not the underlying data.
 
-* Prefer to use `janus_messages::{Time, Duration}` _only_ for serialization and
-  deserialization of DAP messages. It's tempting to use them for representing
-  time for other use cases, but for any other use case we should prefer to use
-  `chrono`, or `std::time` where appropriate.
+* Prefer to use `janus_messages::{Time, Duration}` only for
+  serialization/deserialization of DAP messages, handling DAP client report timestamps,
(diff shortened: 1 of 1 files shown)


divergentdave (commented):
I think we should loosen the recommendation here, and say to only use these types when we are dealing with DAP report client timestamps, or times or durations related to those timestamps, and use `chrono` for all uses of wall clock time not directly related to report timestamps.

divergentdave (approved):

Issue #2810, opened 2024-03-13
Bug: helper-side panic when repeating request for GC-eligible aggregation job


inahga:
```
thread 'aggregator::http_handlers::tests::aggregate_init_with_expired_reports' panicked at aggregator/src/aggregator.rs:1746:17:
found no existing aggregation job for task ID CbZK5CdAfS7_hzp2L_4Y3vqlTgV1xhzMmv2pQH3M1qo and aggregation job ID Wj3h93hr6mqFyq_fqtRz-w
```
```rust
    #[tokio::test]
    #[allow(clippy::unit_arg)]
    async fn aggregate_init_repeat_garbage_collected_job() {
        let (clock, _ephemeral_datastore, datastore, handler) = setup_http_handler_test().await;

        let report_expiry_age = Duration::from_hours(24).unwrap();
        let task = TaskBuilder::new(QueryType::TimeInterval, VdafInstance::Fake)
            .with_report_expiry_age(Some(report_expiry_age))
            .build();

        let helper_task = task.helper_view().unwrap();

        let vdaf = dummy::Vdaf::new(1);
        // let verify_key: VerifyKey<0> = task.vdaf_verify_key().unwrap();
        // let hpke_key = helper_task.current_hpke_key();
        let measurement = 0;
        let prep_init_generator = PrepareInitGenerator::new(
            clock.clone(),
            helper_task.clone(),
            vdaf.clone(),
            dummy::AggregationParam(0),
        );

        // Create a happy path report.
        let (prepare_init_0, transcript_0) = prep_init_generator.next(&measurement);

        datastore
            .put_aggregator_task(&task.helper_view().unwrap())
            .await
            .unwrap();

        let aggregation_param = dummy::AggregationParam(0);
        let request = AggregationJobInitializeReq::new(
            aggregation_param.get_encoded().unwrap(),
            PartialBatchSelector::new_time_interval(),
            Vec::from([prepare_init_0.clone()]),
        );

        let aggregation_job_id: AggregationJobId = random();
        for _ in 0..2 {
            let mut test_conn =
                put_aggregation_job(&task, &aggregation_job_id, &request, &handler).await;
            assert_eq!(test_conn.status(), Some(Status::Ok));
            assert_headers!(
                &test_conn,
                "content-type" => (AggregationJobResp::MEDIA_TYPE)
            );
            let aggregate_resp: AggregationJobResp =
                dbg!(decode_response_body(&mut test_conn).await);

[...]

PR #3415, opened 2024-09-23
Fixup garbage-collection-check clause of put_report_aggregation.

branlwyd:
...and put_leader_report_aggregation. The condition here was reversed, only allowing a non-expired report aggregation row to be updated by the put operations. In practice, this wasn't a problem, as in application code put_report_aggregation was always paired with put_aggregation_job, and the condition on put_aggregation_job was correct.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -2604,7 +2604,7 @@ ON CONFLICT(task_id, aggregation_job_id, ord) DO UPDATE
     )
     WHERE (SELECT UPPER(client_timestamp_interval)
            FROM aggregation_jobs
-           WHERE id = report_aggregations.aggregation_job_id) >= $24",
+           WHERE id = report_aggregations.aggregation_job_id) < $24",
             )
             .await?;
         check_insert(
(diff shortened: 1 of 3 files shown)


branlwyd:
(see also https://github.com/divviup/janus/issues/2810#issuecomment-2359485245)

tgeoghegan:
The test in #2810 reliably fails, and is reliably fixed by this change. But it seems like we need an equivalent leader-side test, and I'd like to understand this better (maybe add some comments to SQL queries) before we merge this. I'll take care of that (thanks Bran!)

tgeoghegan:
Having reviewed #2818, #2857 and #2810, I now understand the guarantee that the `ON CONFLICT SET` clauses are meant to provide. The existing fix had missed the case in `put_leader_report_aggregation` here the report aggregation state is failed, so I fixed that too.

Rather than exercising this from the high level of driving aggregation jobs, I was able to write tests that reproduce this by directly exercising `Datastore` methods and advancing clocks to simulate GC. I also added tests for a few other database tables where we upsert on conflict. This is ready for review. @inahga and @branlwyd: if you're willing, I'd be grateful for your review!

divergentdave (approved):

branlwyd (commented):
I can't approve because I'm the author of this PR, but LGTM. 👍🏻

PR #3452, opened 2024-10-24
Replace gauges with observable gauges

divergentdave:
After deploying, I noticed that `janus_helper_max_outstanding_requests` only shows up as a single blip in one scrape, and it has no value after that. This PR fixes that by replacing the `Gauge`s with `ObservableGauge`s, so we can always record a value each time a callback is called.

See #3444.


inahga (approved):
Thanks!

PR #3469, opened 2024-11-08
Update HTTP verbs for DAP-13.

branlwyd:
Specifically:
* Use POST rather than PUT for report uploads.
* Use GET rather than POST for collection job polling.

### DIFF
--- a/aggregator/src/aggregator/http_handlers/tests/collection_job.rs
+++ b/aggregator/src/aggregator/http_handlers/tests/collection_job.rs
@@ -24,7 +24,7 @@ use serde_json::json;
 use trillium::{KnownHeaderName, Status};
 use trillium_testing::{
     assert_headers,
-    prelude::{delete, post, put},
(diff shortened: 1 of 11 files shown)


branlwyd:
I also disabled divviup-ts tests, as they're not going to work until both Janus & divviup-ts are updated to implement DAP-13. I filed #3479 to track re-enabling these tests.

aggregator/src/aggregator/http_handlers/tests/collection_job.rs
```diff
@@ -288,7 +288,7 @@ async fn collection_job_post_request_unauthenticated_collection_jobs() {
 
     // Incorrect authentication token.
     let mut test_conn = test_case
-        .post_collection_job_with_auth_token(&collection_job_id, Some(&random()))
```
divergentdave, line 291: nit: The name of this test could be updated as well

divergentdave (approved):

PR #3481, opened 2024-11-13
Remove multi-collection.

branlwyd:
DAP-11 removed the ability to collect a batch more than once (with differing aggregation parameters). This commit also includes related changes to a few messages, and removes the max_batch_query_count task parameter.


branlwyd:
(apologies, a couple of force-pushes to clean up formatting & a few merge issues with the rebase onto the renaming PR)

tools/src/bin/collect.rs
```diff
@@ -186,29 +184,14 @@ struct QueryOptions {
...
-    )]
-    batch_id: Option<BatchId>,
-    /// Have the aggregator select a batch that has not yet been collected
-    #[clap(
-        long,
-        action = ArgAction::SetTrue,
-        conflicts_with_all = ["batch_interval_start", "batch_interval_duration", "batch_id"],
-        help_heading = "Collect Request Parameters (Leader Selected)",
-    )]
-    current_batch: bool,
```
inahga, line 211: Hmm, so now without explicit `--batch-interval-{start,duration}`, we automatically assume we're operating on a fixed size task.

I suppose that's fine. Maybe we can think of a better way to clean up this CLI for this, but I don't think its important.

inahga (approved):

tools/src/bin/collect.rs
branlwyd: Yep, that's correct. I thought about this for a bit & came up with the same conclusion you did. We could introduce something like a `--leader-selected` flag to indicate a leader-selected query, but I'm not fully sure it's necessary.

Issue #3480, opened 2024-11-13
DAP-13: drop support for Poplar1


branlwyd:
With the removal of repeated collection of a batch, Poplar1 is no longer of practical value. We should stop supporting it in the DAP-13 / `0.8` release track of Janus.

Specifically:
* Application uses of Poplar1 should be dropped entirely.
* Uses of Poplar1 in unit tests should be replaced by `libprio-rs`' dummy VDAF, which can be configured to require multiple rounds for completion.
* Uses of Poplar1 in integration/end-to-end tests should be dropped entirely.

Issue #3436, opened 2024-10-11
Implement DAP-12


branlwyd:
DAP-12: https://www.ietf.org/archive/id/draft-ietf-ppm-dap-12.html

From DAP-09, this will require (roughly):

- [ ] Poll collection jobs with GET instead of POST. (DAP-10, [PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/544))
- [ ] Upload reports with POST rather than PUT. (DAP-10, [PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/543))
- [ ] Remove support for multi-collection of batches, removing `max_batch_query_count` & the leader-selected `by_batch_id` query. (DAP-11, [PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/554), [1])
- [ ] Remove per-task HPKE configurations. (DAP-12, [PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/567))
- [ ] Rename "query type" to "batch mode", "fixed-size" to "leader-selected", and remove `max_batch_size`. (DAP-12, [PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/563))
- [ ] Asynchronous aggregation. (DAP-12, [PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/564))
- [ ] Update collection-related messages to match aggregation messages. (DAP-12, [PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/596))
- [ ] Replay protection changes. (DAP-12, [PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/582), [2])
- [ ] Re-number `PrepareError` values. (DAP-12, [PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/611))
- [ ] Bump version tag. (DAP-12, [PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/601))

[1] Do not remove the `part_batch_selector` field, as it is restored in a later change.
[2] This may not require changes, but we should validate that Janus implements the specified behavior.

branlwyd:
> Replay protection changes

For this portion of the work, I think no changes are needed:
[...]

PR #3482, opened 2024-11-14
Remove support for Poplar1.

branlwyd:
DAP-11 removed multi-collection and, with it, the ability to usefully use Poplar1. This commit makes the following changes:

* Application-level support for Poplar1 is removed.
* Unit tests which used Poplar1 now use libprio-rs' dummy::Vdaf.
* Introduce a test-only ability to configure the fake dummy::Vdaf for taskprov, required to migrate some taskprov unit tests.


branlwyd:
Part of #3436. Closes #3480. Stacked on #3481.

inahga (approved):

Issue #3483, opened 2024-11-14
HTTP Status Code Divergence between DAP Specification and Janus Release 0.7


MaeMilc:
When perusing the [DAP specification](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-09.html) of the [Janus version under the release/0.7 branch](https://github.com/divviup/janus/tree/release/0.7), one comes across the following paragraph towards the end of [section 4.4.2](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-09.html#name-upload-request) regarding the possible responses a Leader can send to a Client that submitted a report earlier:

> The Leader responds to well-formed requests with HTTP status code 201 Created. Malformed requests are handled as described in Section 3.2. Clients SHOULD NOT upload the same measurement value in more than one report if the Leader responds with HTTP status code 201 Created.

Anyhow, the Leader seems to intentionally respond with __HTTP status code 200 OK__ in the [corresponding HTTP handler function](https://github.com/divviup/janus/blob/release/0.7/aggregator/src/aggregator/http_handlers.rs#L495) in Janus, despite this deviating from the aforementioned DAP specification on which the release/0.7 Janus branch is based. The code underlying the HTTP handler function referred to earlier is pasted below for reference:

```Rust
/// API handler for the "/tasks/.../reports" PUT endpoint.
async fn upload<C: Clock>(
    conn: &mut Conn,
    (State(aggregator), BodyBytes(body)): (State<Arc<Aggregator<C>>>, BodyBytes),
) -> Result<Status, ArcError> {
    validate_content_type(conn, Report::MEDIA_TYPE).map_err(Arc::new)?;

    let task_id = parse_task_id(conn).map_err(Arc::new)?;
    conn.cancel_on_disconnect(aggregator.handle_upload(&task_id, &body))
        .await
        .ok_or(Arc::new(Error::ClientDisconnected))??;

    // Handle CORS, if the request header is present.
    if let Some(origin) = conn.request_headers().get(KnownHeaderName::Origin) {
        // Unconditionally allow CORS requests from all origins.
        let origin = origin.clone();
        conn.response_headers_mut()
            .insert(KnownHeaderName::AccessControlAllowOrigin, origin);
    }

    Ok(Status::Ok)
}
```

[...]

PR #3485, opened 2024-11-14
Add DAP-09 errata for upload response code issue.

branlwyd:
We don't plan to fix this (in the short term) because Mozilla looks specifically for a 200 OK response from the Leader.


branlwyd:
Part of #3483.

divergentdave (approved):

PR #3486, opened 2024-11-15
Remove per-task aggregator HPKE keys.

branlwyd:
DAP-12 removed (specified) per-task HPKE key requests. We don't have a reason to want per-task HPKE keys, especially if they require custom functionality, so we remove per-task HPKE key support. With this change, all Janus deployments must use "global" HPKE keys.

Also, minorly refactor HpkeKeypairCache; mostly, to maintain all state under a single mutex and to remove an unnecessary "view" type.


branlwyd:
Part of #3436.

aggregator/src/aggregator.rs
```diff
@@ -269,9 +261,8 @@ impl<C: Clock> Aggregator<C> {
     ///
     /// Fails on general datastore errors.
     ///
-    /// If [`Self::require_global_hpke_keys`] is `true`, and there is not at least one
-    /// [`GlobalHpkeKeypair`] in the database in the [`HpkeKeyState::Active`] state then this
-    /// function will fail.
+    /// If  there is not at least one HPKE keypair in the database in the [`HpkeKeyState::Active`]
+    /// state then this function will fail.
```
inahga, line 265:
```suggestion
    /// If there is not at least one HPKE keypair in the database in the [`HpkeKeyState::Active`]
    /// state then this function will fail.
```

aggregator/src/aggregator.rs
```diff
@@ -1742,42 +1655,26 @@ impl VdafOps {
...
-            (Some(task_hpke_keypair), None) => try_hpke_open(task_hpke_keypair),
-            (Some(task_hpke_keypair), Some(global_hpke_keypair)) => {
-                try_hpke_open(task_hpke_keypair).or_else(|error| match error {
-                    // Only attempt second trial if _decryption_ fails, and not some
-                    // error in server-side HPKE configuration.
-                    hpke::Error::Hpke(_) => try_hpke_open(&global_hpke_keypair),
-                    error => Err(error),
-                })
-            }
-        };
```
inahga, line 1780: Phew, glad this nonsense is gone.

aggregator/src/aggregator/key_rotator.rs
```diff
@@ -421,10 +417,10 @@ impl<'a, C: Clock> HpkeKeyRotator<'a, C> {
...
         )?;
 
-        // Defensive assertion: Check our transaction snapshot for at least one active keypair in
-        // in the table. If one is absent, committing the transaction would leave Janus unstartable
-        // so we should rollback.
-        let keypairs = tx.get_global_hpke_keypairs().await?;
+        // Defensive assertion: Check our transaction snapshot for at least one active keypair in in
+        // the table. If one is absent, committing the transaction would leave Janus unstartable, so
+        // we should rollback.
+        let keypairs = tx.get_hpke_keypairs().await?;
```
inahga, line 423:
```suggestion
        // Defensive assertion: Check our transaction snapshot for at least one active keypair in
        // the table. If one is absent, committing the transaction would leave Janus unstartable, so
        // we should rollback.
        let keypairs = tx.get_hpke_keypairs().await?;
```

inahga (approved):

inahga:
Closes https://github.com/divviup/janus/issues/2147

aggregator/src/aggregator.rs
branlwyd: Yeah, this was some pretty hairy code, and in general I don't like trial decryption when it can be avoided. I'm glad to see it gone.

PR #3487, opened 2024-11-16
Remove max_batch_size parameter from leader-selected tasks.

branlwyd:
This parameter was removed in DAP-12, and we do not have any reason to keep it as functionality over-and-above what is specified in DAP, so we remove this parameter from Janus. The aggregation job creator will now always create batches which are exactly min_batch_size, matching previous functionality if max_batch_size was unspecified.


branlwyd:
Part of #3436.

inahga (approved):

PR #3488, opened 2024-11-18
Renumber ReportError values.

branlwyd:
Also, add a "reserved" value to match other similar enumerated values.


branlwyd:
Part of #3436.

inahga (approved):

PR #3501, opened 2024-11-19
Implement task start time.

branlwyd:
Note that DAP specifies the parameters as "task start time" and "task duration", rather than Janus' "task start time" and "task end time". Janus implements things this way so that either parameter can be left unspecified.

I leave taskprov configuration for a later commit, which will update taskprov functionality to match the latest specification.


branlwyd:
Part of #3436.

inahga (approved):
It's arguable whether we should 404 on report uploads to tasks that have not started. But I'm fine with giving expressive errors, since that'll aid in debugging.

inahga:
Nevermind, it's not arguable, behavior is in the spec https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#name-input-share-validation :smile:

branlwyd:
(force-push to rerun the `janus_migrations` CI check, now that the label is correctly applied.)

PR #3502, opened 2024-11-19
Update batch mode-specific messages to include a length prefix.

branlwyd:
As before, we will (correctly) generate an invalidMessage error if we receive a message with an unknown batch mode, starting from an error generated at the point of parsing the batch mode discriminator byte.


branlwyd:
Part of #3464.

inahga (approved):

PR #3503, opened 2024-11-21
Add support for public extensions.

branlwyd:
...and rename the existing (private) extensions to "private extensions". Per DAP, these extensions can conflict with one another.


branlwyd:
Part of #3436. See also https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/issues/94.

inahga (approved):

PR #3504, opened 2024-11-21
Switch to VDAF-13.

branlwyd:
I currently pin to a specific commit in libprio-rs tracking the VDAF-13 implementation, as the implementation is currently ongoing. I am fairly confident that the only breaking interface changes have already been implemented: there is a new "application context string", which is computed as the concatenation of hte current DAP draft in use & the task ID.

Also, bump the version tag while I'm at it, since I touched it anyway.

### DIFF
--- a/core/src/vdaf.rs
+++ b/core/src/vdaf.rs
@@ -1,5 +1,6 @@
+use crate::DAP_VERSION_IDENTIFIER;
 use derivative::Derivative;
-use janus_messages::taskprov;
+use janus_messages::{taskprov, TaskId};
 use prio::{
     field::Field64,
     flp::{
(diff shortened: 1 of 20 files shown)


branlwyd:
Part of #3436.

core/src/vdaf.rs
```diff
@@ -23,6 +24,18 @@ const ALGORITHM_ID_PRIO3_SUM_VEC_FIELD64_MULTIPROOF_HMACSHA256_AES128: u32 = 0xF
 /// of the VDAF specification.
 pub const VERIFY_KEY_LENGTH_HMACSHA256_AES128: usize = 32;
 
+/// Computes the VDAF "application context", which is an opaque byte string used by many VDAF
+/// operations for domain-separation purposes. In DAP, this is the DAP draft number concatenated
+/// with the task ID.
+pub fn vdaf_application_context(
+    task_id: &TaskId,
+) -> [u8; DAP_VERSION_IDENTIFIER.len() + TaskId::LEN] {
```
inahga, line 32: TIL you can do const operations in this spot.

inahga (approved):

Issue #3435, opened 2024-10-09
Replace `derivative` with `educe`


divergentdave:
The `derivative` crate has not been updated in years, and it is currently the only crate keeping syn v1 in our dependency tree. Following the lead of `kube-rs`, we could replace it with `educe`, which is actively maintained, and supports similar use cases.

PR #3506, opened 2024-11-22
deps: replace derivative with educe

inahga:
Closes https://github.com/divviup/janus/issues/3435.

Note this does not fully remove `derivative` from our dependency tree, because we depend on `divviup-client`, which depends on `janus_messages 0.7.42`. `divviup-client` needs to be updated to a new version of janus_messages, which seems non-trivial at the moment.


branlwyd (approved):

branlwyd:
ope, it looks like merging the last few PRs led to a merge conflict, investigating. (I don't think the build error we see here has anything to do with the changes in the PR)

inahga:
@branlwyd I caught them in https://github.com/divviup/janus/pull/3507.

PR #3508, opened 2024-11-22
cargo-deny: ignore instant and derivative unmaintained warnings

inahga:
`derivative` will naturally be fully phased out once we update divviup-client.

`instant` we will leave for the foreseeable future. If SHTF and this dependency is broken, we can use https://github.com/divviup/janus/commit/c4692719923009b0e7e1596b1850a98d02ce7a5d as a workaround.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -7,6 +7,15 @@ all-features = true
 
 [advisories]
 version = 2
+ignore = [
+    # `instant` is unmaintained, but it is feature-complete and small, so it is unlikely to have
+    # bugs or security vulnerabilities.
+    "RUSTSEC-2024-0384",
+
+    # Use of `derivative` has been removed from production code. It still exists as transitive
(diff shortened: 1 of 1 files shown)


inahga:
Whoops I had automerge on. I will file issue for at least `derivative`, since it does take action on our part to phase it out.

For `instant`, I don't foresee us really taking any action except for removing it, which we can do as needed.

> I suppose having CI go green is nice

IMO it's more than nice. If we leave it red, we're less inclined to check why it's failing, so an actual serious advisory could fly under our radar. An errored status check atm doesn't distinguish between "mindless ecosystem churn" and "critical vulnerability".

inahga:
Ah, actually, since David backported the `derivative` removal, we should be able to clear this once the updated janus_messages gets updated in divviup_client. I opened an issue and self-assigned to make sure this happens.

PR #3523, opened 2024-11-27
Bump MSRV to 1.80.

branlwyd:
This is, at time of writing, the "stable minus two releases" that Janus supports. Also, clean up a few uses of OnceLock -- mostly to replace with the simpler LazyLock stabilized in 1.80, in one place replacing it with an async mutex as neither OnceLock nor LazyLock were appropriate as the initializing function would be async.

### DIFF
--- a/aggregator/src/binaries/janus_cli.rs
+++ b/aggregator/src/binaries/janus_cli.rs
@@ -32,11 +32,12 @@ use serde::{Deserialize, Serialize};
 use std::{
     collections::BTreeMap,
     path::{Path, PathBuf},
-    sync::{Arc, OnceLock},
+    sync::Arc,
 };
 use tokio::{
     fs,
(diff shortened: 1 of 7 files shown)


aggregator/src/binaries/janus_cli.rs
```diff
@@ -700,40 +700,43 @@ impl BinaryConfig for ConfigFile {
...
-    fn from(value: kube::Client) -> Self {
+    fn from(kube_client: kube::Client) -> Self {
         Self {
-            lock: OnceLock::from(value),
+            lock: Mutex::new(Some(kube_client)),
         }
     }
 }
 
+
```
divergentdave, line 739: rustfmt doesn't like this
```suggestion
```

divergentdave (approved):

PR #3543, opened 2024-12-06
Use aws-lc-rs for hashing and encryption

### LINKED ISSUE #2719: Replace ring with RustCrypto crates
We currently use ring through rustls and directly for SHA-256 in authentication token hashes, batch checksums, hashes of requests for idempotency checking, and the aggregator API, for HKDF-SHA256 in Taskprov, and for AES-GCM in encryption of database records.

The forthcoming 0.23 release of rustls will swap ring out for aws-lc-rs by default (controlled by crate features). If we replaced our direct uses of ring for hashing and encryption with respective RustCrypto crates, we could eliminate ring from our dependency graph. This would be a win both because it is a very large crate, and because it has a complicated build script. Fully realizing these benefits would be blocked on related crates upgrading to rustls 0.23.

divergentdave:
This replaces direct uses of ring for hashing and encryption with aws-lc-rs. Closes #2719.

I was reminded recently that aws-lc-rs originally targeted being a drop-in replacement for ring's Rust API. Thus, this PR makes that swap. With this change, we will still benefit from the same hand-written assembly cores (more or less), but with a better maintenance story, and with the benefit of more high-assurance techniques.


Cargo.toml
```diff
@@ -26,6 +26,7 @@ version = "0.8.0-prerelease-1"
 anyhow = "1"
 assert_matches = "1"
 async-trait = "0.1"
+aws-lc-rs = "1.11.1"
```
inahga, line 29: I believe we can also drop `ring` from the workspace Cargo.toml

inahga (approved):

PR #3564, opened 2024-12-09
Implement Leader async aggregation.

branlwyd:
This change includes unit tests, but no integration tests -- those will need to come with the Helper async aggregation implementation, as without it we do not have anything to integration test against.

A few implementation notes:
 * I renamed the report aggregation states to better match their functionality (IMO).
 * If the Helper does not provide a retry-after header, the Leader will poll each "processing" aggregation job (at most) once per minute.
 * The retry-after header can specify either a number of seconds, or a specific date. Currently, we only support receiving a number of seconds.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_driver.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver.rs
@@ -53,7 +53,13 @@ use prio::{
 };
 use rayon::iter::{IndexedParallelIterator as _, IntoParallelIterator as _, ParallelIterator as _};
 use reqwest::Method;
-use std::{collections::HashSet, panic, str::FromStr, sync::Arc, time::Duration};
+use retry_after::RetryAfter;
+use std::{
+    collections::HashSet,
+    panic,
+    sync::Arc,
+    time::{Duration, UNIX_EPOCH},
+};
 use tokio::{join, sync::mpsc, try_join};
 use tracing::{debug, error, info, info_span, trace_span, warn, Span};
 
@@ -885,7 +891,7 @@ where
         aggregation_job: AggregationJob<SEED_SIZE, B, A>,
         stepped_aggregations: Vec<SteppedAggregation<SEED_SIZE, A>>,
         report_aggregations_to_write: Vec<WritableReportAggregation<SEED_SIZE, A>>,
-        retry_after: Option<&Duration>,
+        retry_after: Option<&RetryAfter>,
         helper_resp: AggregationJobResp,
     ) -> Result<(), Error>
     where
@@ -899,7 +905,6 @@ where
         A::PublicShare: Send + Sync,
     {
         match helper_resp {
-            // TODO(#3436): implement asynchronous aggregation
             AggregationJobResp::Processing => {
                 self.process_response_from_helper_pending(
                     datastore,
@@ -944,7 +949,7 @@ where
         aggregation_job: AggregationJob<SEED_SIZE, B, A>,
         stepped_aggregations: Vec<SteppedAggregation<SEED_SIZE, A>>,
         mut report_aggregations_to_write: Vec<WritableReportAggregation<SEED_SIZE, A>>,
-        retry_after: Option<&Duration>,
+        retry_after: Option<&RetryAfter>,
     ) -> Result<(), Error>
     where
         A::AggregationParam: Send + Sync + Eq + PartialEq,
@@ -993,7 +998,8 @@ where
         let aggregation_job_writer = Arc::new(aggregation_job_writer);
 
         let retry_after = retry_after
-            .copied()
+            .map(|ra| retry_after_to_duration(datastore.clock(), ra))
+            .transpose()?
             .or_else(|| Some(Duration::from_secs(60)));
         let counters = datastore
             .run_tx("process_response_from_helper_pending", |tx| {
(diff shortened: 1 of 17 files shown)


branlwyd:
Part of #3436.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -740,6 +780,91 @@ where
...
+                    ReportAggregationState::LeaderPoll { leader_state } => {
+                        Some(leader_state.clone())
+                    }
+                    _ => None,
+                };
+
+                leader_state.map(|leader_state| SteppedAggregation {
+                    report_aggregation,
+                    leader_state,
+                })
```
divergentdave, line 828: This is a potential candidate for future optimizations to reduce peak memory usage. Two copies of the leader state could be expensive for large instantiations of Prio3. Maybe we could put this in an Arc, or change SteppedAggregation to handle jobs in the poll state without this duplication.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -772,16 +898,155 @@ where
         A::PrepareState: Send + Sync + Encode,
         A::PublicShare: Send + Sync,
     {
-        let prepare_resps = match helper_resp {
+        match helper_resp {
             // TODO(#3436): implement asynchronous aggregation
```
divergentdave, line 902: nit: This can be deleted now
```suggestion
```

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -772,16 +898,155 @@ where
...
+        let mut aggregation_job_writer =
+            AggregationJobWriter::<SEED_SIZE, _, _, UpdateWrite, _>::new(
+                Arc::clone(&task),
+                self.batch_aggregation_shard_count,
+                Some(AggregationJobWriterMetrics {
+                    report_aggregation_success_counter: self.aggregation_success_counter.clone(),
+                    aggregate_step_failure_counter: self.aggregate_step_failure_counter.clone(),
+                    aggregated_report_share_dimension_histogram: self
+                        .aggregated_report_share_dimension_histogram
+                        .clone(),
```
divergentdave, line 989: We should rethink Prometheus metrics in the context of `AggregationJobResp::Processing` responses as well. I haven't looked at the metrics in detail yet, but we may want additional counters, or new labels, in order to distinguish forward progress versus mere polling.

divergentdave (approved):

aggregator_core/src/datastore/models.rs
```diff
@@ -1096,18 +1140,18 @@ where
...
                     transition: rhs_transition,
                 },
             ) => lhs_transition == rhs_transition,
             (
-                Self::WaitingHelper {
+                Self::HelperContinue {
                     prepare_state: lhs_state,
                 },
-                Self::WaitingHelper {
+                Self::HelperContinue {
```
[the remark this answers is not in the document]
branlwyd: Done -- this was slightly tricker than expected as `PingPongState`'s derived `PartialEq`/`Eq` implementations require `A: PartialEq` or `A: Eq`, respectively. I couldn't use them easily because, at least, `dummy::Vdaf` does not implement these traits.

I'm not sure why these bounds are required; maybe there is a good reason I am not yet thinking of. We might be able to get rid of these manual `PartialEq` implementations if we accept `A: PartialEq` and implement/derive it as-necessary on the aggregator types in use, which would be a good change for ensuring correctness. But I think that's beyond the bounds of this PR.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -1246,3 +1513,11 @@ struct SteppedAggregation<const SEED_SIZE: usize, A: vdaf::Aggregator<SEED_SIZE,
...
     leader_state: PingPongState<SEED_SIZE, 16, A>,
 }
+
+fn parse_retry_after(header_value: &HeaderValue) -> Result<Duration, Error> {
+    let val = header_value
+        .to_str()
+        .map_err(|err| Error::BadRequest(err.to_string()))?;
+    let val = u64::from_str(val).map_err(|err| Error::BadRequest(err.to_string()))?;
+    Ok(Duration::from_secs(val))
+}
```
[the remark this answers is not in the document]
branlwyd: Ah, thanks -- this is way better, and more feature-complete as well.

aggregator/src/aggregator/aggregation_job_driver.rs
branlwyd: Agreed -- using an `Arc`, or just moving the large value (where possible), would improve things here with little tradeoff.

aggregator/src/aggregator/aggregation_job_driver.rs
branlwyd: I think this is wise, but these metrics only concern themselves with completed report aggregations & failed steps, both of which are agnostic to the aggregation mode. I think we can get pretty far by looking to the request verb & the response code, though full these metrics won't give full clarity (i.e. we can differentiate between a Leader's attempt to poll vs an attempt to send a new step via the HTTP verb; but I don't think we can tell whether the Helper responded with a `processing` or `finished` response with these metrics). However, I strongly suspect that our current integrations will always use exactly one of asynchronous or synchronous responses, so this may be moot at the moment, at least for us.

PR #3589, opened 2025-01-13
Fix clippy lints

divergentdave:


jcjones (approved):
ref: https://doc.rust-lang.org/std/option/enum.Option.html#method.is_some_and

tgeoghegan (approved):

PR #3620, opened 2025-01-24
Upgrade to opentelemetry-rust 0.27

divergentdave:
This upgrades to the 0.27 release of opentelemetry-rust. This includes the following changes:

- Various OTel APIs were renamed.
- The `Meter` no longer allows supplying a multi-callback for several instruments. Instead, separate callbacks are registered on each instrument. Additionally, `observe()` can no longer be called on observable gauges, etc.; this method has moved to the argument provided to the instrument's callback.
- Support for the `MetricProducer` trait has been removed. As a result, the (unstable) Tokio poll time histogram has been removed, and the remaining Tokio runtime metrics have been reimplemented with SDK instruments.
- Histogram boundaries can now be set directly on instrument builders. Thus, I have done so and removed our custom `View`.
- `NoopMeterProvider` was removed in this release, so I vendored it for use in tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -286,31 +286,11 @@ checksum = "05b1b633a2115cd122d73b955eadd9916c18c8f510ec9cd1686404c60ad1c29c"
 dependencies = [
  "async-channel 2.2.0",
  "async-executor",
- "async-io 2.3.2",
+ "async-io",
  "async-lock 3.3.0",
  "blocking",
  "futures-lite 2.3.0",
  "once_cell",
-]
-
-[[package]]
-name = "async-io"
-version = "1.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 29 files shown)


tgeoghegan (approved):
The only thing I don't understand is why we drop all the poll time histogram config. Do we no longer track that metric, or do we no longer allow it to be configured?

divergentdave:
It's currently not possible to pass the poll time histogram through to the metrics exporters, because we only get already-aggregated bucket counts from Tokio, and the `MetricProducer` trait in the `opentelemetry-sdk` crate we previously used is no longer available. Now, our only entry point is the `Histogram` instrument. We can only observe one measurement at a time using that, but we don't have access to the original sequence of measurement values. Since there's no way to make use of the poll time histogram anymore, I removed the Janus configuration parameters for it, and stopped enabling it in Tokio runtime configuration as well.

I filed a [feature request](https://github.com/open-telemetry/opentelemetry-rust/issues/2505) to add `MetricProducer` back to the SDK. If that is implemented in the future, we can revisit this histogram.

PR #3621, opened 2025-01-25
Upgrade libprio to alpha prerelease

divergentdave:
This upgrades libprio, so we are now VDAF-interoperable. This includes a change to the field used by Prio3Sum and a change to the XofTurboShake128 seed size.

### DIFF
--- a/integration_tests/tests/integration/simulation/bad_client.rs
+++ b/integration_tests/tests/integration/simulation/bad_client.rs
@@ -12,7 +12,7 @@ use janus_core::{
     retries::retry_http_request,
     test_util::{install_test_trace_subscriber, runtime::TestRuntime},
     time::{Clock, RealClock, TimeExt},
-    vdaf::{vdaf_application_context, vdaf_dp_strategies, VdafInstance},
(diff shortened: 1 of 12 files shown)


jcjones (approved):
I see no mechanical issues, and obviously tests pass. 

It appears that the updates faithfully mirror the changes to the VDAF spec.

integration_tests/tests/integration/simulation/bad_client.rs
```diff
@@ -130,7 +130,7 @@ fn shard_encoded_measurement(
     task_id: &TaskId,
     encoded_measurement: Vec<Field128>,
     report_id: ReportId,
-) -> (Prio3PublicShare<16>, Vec<Prio3InputShare<Field128, 16>>) {
+) -> (Prio3PublicShare<32>, Vec<Prio3InputShare<Field128, 32>>) {
```
tgeoghegan, line 133: Can this/should this be `VERIFY_KEY_LENGTH` instead of `32`?

tgeoghegan (approved):

PR #3649, opened 2025-02-07
Use the mime crate to parse 'content-type' header into Media Types

### LINKED ISSUE #3609: Tolerate version parameter in media types
DAP draft-13 says that `Content-Type` header values may include a version parameter. (https://www.ietf.org/archive/id/draft-ietf-ppm-dap-13.html#section-9.1-4) We currently perform exact matches on the whole media type, which is too strict to interoperate with this.

jcjones:
Future work: 

-  Change the `validate_content_type` signature to include version constraints, and apply them using the values now parsed out and available by the `mime` crate. (https://github.com/divviup/janus/issues/3650)

Closes #3609.


aggregator/src/aggregator/http_handlers.rs
```diff
@@ -713,12 +713,22 @@ async fn aggregate_shares<C: Clock>(
...
+        )))?;
+
+        if let Ok(mime) = mime_str.parse::<mime::Mime>() {
+            if mime.essence_str() == expected_media_type {
+                Ok(())
+            } else {
+                Err(Error::BadRequest(format!(
+                    "unexpected Content-Type header: {mime}"
+                )))
+            }
```
tgeoghegan, line 727: I really like using `Option::ok_or` here when calling `HeaderValue::as_str`, but think we can go farther. Some observations:

- using [`Result::map_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err) on `mime_str.parse()` is tidier than `if let Ok(...) = ... { } else { }`, and also more consistent with what we do elsewhere in our Rust projects.
- `map_err` also de-indents the code somewhat, which makes it easier to read. We can move more code to the left by moving the `Ok(())` path to the end of the block and just checking for exception, error conditions above that.
- We shouldn't drop the error returned by `str::parse::<mime::Mime>` and instead incorporate it into the error we send up the stack.
- Where possible, we have a stylistic preference for type annotations on declared variables over use of the turbofish.
- We generally prefer importing modules rather than fully qualifying types, except where importing the specific type causes collisions or where the module name helps disambiguate what the type is. i.e., let's `use mime::Mime;` at the top of the file. 

Because of the way this diff is laid out, I can't neatly provide a suggestion here in the PR, but how about:

```rust
        let mime: Mime = mime_str
            .parse()
            .map_err(|e| Error::BadRequest(format!("unable to parse header {mime_str}: {e:?}")))?;

        if mime.essence_str() != expected_media_type {
            return Err(Error::BadRequest(format!(
                "unexpected Content-Type header: {mime}"
            )));
        }

        Ok(())
```

Interpolating `e`'s `std::debug::Debug` into the `Error::BadRequest` isn't ideal because we lose structure, but what's particularly valuable here is that the error will ultimately show up in logs and in the HTTP 400 presented to the client. Still, if you felt like it, you could add a variant to `janus_aggregator::aggregator::Error` that explicitly represents `FromStrError`, or more specifically a problem with content type. I think there's some other enums you'd have to fix up (e.g. where we construct HTTP Problem Documents from `Error`) but the compiler would walk you through it.

Once we've restructured the function to that extent, it's tempting to clean up the `if let Some(content_type) = conn.request_headers() {} else {}` that's already there:

```rust
    let content_type = conn
        .request_headers()
        .get(KnownHeaderName::ContentType)
        .ok_or_else(|| Error::BadRequest("no Content-Type header".to_owned()))?;

    let mime_str = content_type.as_str().ok_or(Error::BadRequest(format!(
        "invalid Content-Type header: {content_type}"
    )))?;

    let mime: Mime = mime_str
        .parse()
        .map_err(|e| Error::BadRequest(format!("unable to parse header {mime_str}: {e:?}")))?;

    if mime.essence_str() != expected_media_type {
        return Err(Error::BadRequest(format!(
            "unexpected Content-Type header: {mime}"
        )));
    }

    Ok(())
```

This code strikes me as much flatter, and it's way easier to figure out where the happy path is. Once again the questionable structure was there before your PR, but we have an opportunity clean up now.

core/src/retries.rs
```diff
@@ -485,4 +485,68 @@ mod tests {
...
+        mock_problem.assert_async().await;
+    }
+
+    #[tokio::test]
+    async fn http_retry_server_json_problem_versioned_doc() {
+        install_test_trace_subscriber();
+        let mut server = mockito::Server::new_async().await;
+
+        // Ensure that even with a complex media type, we parse a problem
+        // document and prefer its status code.
```
divergentdave, line 526: This part of the comment is incorrect, we happen to prefer the HTTP status code over the `status` member of the details. The latter is mostly advisory, see https://www.rfc-editor.org/rfc/rfc9457.html#name-status.
```suggestion
        // document.
```

collector/src/lib.rs
```diff
@@ -570,7 +570,11 @@ impl<V: vdaf::Collector> Collector<V> {
             .headers()
             .get(CONTENT_TYPE)
             .ok_or(Error::BadContentType(None))?;
-        if content_type != CollectionJobResp::<TimeInterval>::MEDIA_TYPE {
+        if let Ok(mime) = content_type.to_str()?.parse::<mime::Mime>() {
```
tgeoghegan, line 573: As argued in `http_handlers.rs`, prefer `Result::map_err` over `if let Ok(...) = foo.bar() {...} else {...}`

divergentdave (approved):

core/src/retries.rs
```diff
@@ -485,4 +485,68 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn http_retry_server_json_problem() {
+        install_test_trace_subscriber();
+        let mut server = mockito::Server::new_async().await;
+
+        let mock_problem = server
+            .mock("GET", "/")
+            .with_status(400)
+            .with_header("Content-Type", "application/problem+json")
```
tgeoghegan, line 497: nit: we have `static PROBLEM_DETAILS_JSON_MEDIA_TYPE: &str = "application/problem+json";` in `problem_details.rs` but I suppose it's not visible from here. Plus we reproduce this string literal all over our tests already, so this is no big deal.

core/src/http.rs
```diff
@@ -21,15 +22,20 @@ impl HttpErrorResponse {
...
         let status = response.status();
+
         if let Some(content_type) = response.headers().get(CONTENT_TYPE) {
-            if content_type == PROBLEM_JSON_MEDIA_TYPE {
-                match response.json::<HttpApiProblem>().await {
-                    Ok(mut problem) => {
-                        problem.status = Some(status);
-                        // Unwrap safety: the conversion always succeeds if the status is populated.
-                        return problem.try_into().unwrap();
+            if let Ok(content_type_str) = content_type.to_str() {
```
tgeoghegan, line 27: I want to make an observation about using things like `Result::map_err` or `Option::ok_or_else` to move some of this code left, but to be honest it's not immediately obvious to me how to do that since this method doesn't return `Result`. This function was a bit of a mess before you got here, so I think we can live with it.

tgeoghegan (approved):

aggregator/src/aggregator/http_handlers.rs
jcjones: Thanks for the help! I've updated this in 326436bf..2af7c099.

core/src/http.rs
jcjones: Also resolved in 326436bf..2af7c099.

core/src/retries.rs
jcjones: Updated to point to the `http_api_problem` crate's static.

collector/src/lib.rs
jcjones: Resolved in https://github.com/divviup/janus/pull/3649/commits/2ad6be874daf12c1c284c1606d5bf74cb2ff8a4c

PR #3708, opened 2025-03-04
Add verification of aggregation mode in PeerAggregator

### LINKED ISSUE #3666: Validate aggregation mode of taskprov peer aggregator against role
As a follow-up to #3630, we should add validation to the `PeerAggregator` constructor checking that an aggregation mode has been chosen if the peer aggregator's role is Leader (and our role is Helper).

divergentdave:
Closes #3666. This makes `PeerAggregator::new()` and `PeerAggregatorBuilder::build()` fallible, and adds a couple validation checks.

### DIFF
--- a/aggregator_api/src/routes.rs
+++ b/aggregator_api/src/routes.rs
@@ -456,7 +456,8 @@ pub(super) async fn post_taskprov_peer_aggregator<C: Clock>(
         req.tolerable_clock_skew,
         req.aggregator_auth_tokens,
         req.collector_auth_tokens,
-    );
+    )
+    .map_err(|e| Error::BadRequest(format!("Invalid request: {e}")))?;
 
     let inserted = ds
         .run_tx("post_taskprov_peer_aggregator", |tx| {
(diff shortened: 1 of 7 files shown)


jcjones (approved):
It felt like this would need some documentation updates, but I see the aggregator is marked as its Rust API is subject to arbitrary change - nice.

aggregator_api/src/routes.rs
```diff
@@ -456,7 +456,8 @@ pub(super) async fn post_taskprov_peer_aggregator<C: Clock>(
         req.tolerable_clock_skew,
         req.aggregator_auth_tokens,
         req.collector_auth_tokens,
-    );
+    )
+    .map_err(|e| Error::BadRequest(format!("Invalid request: {e}")))?;
```
tgeoghegan, line 460: super-nit: since this is `Error::BadRequest` anyway, I suspect the resulting error and/or HTTP problem doc will already say bad or invalid request and it's not necessary to repeat that in the format string.

tgeoghegan (approved):

aggregator_api/src/routes.rs
divergentdave: The aggregator API does not use HTTP problem documents, it just uses this string as the entirety of the response body.

PR #3755, opened 2025-04-03
Fix Clippy warnings

divergentdave:
This fixes miscellaneous Clippy warnings that appear with the 1.87 toolchain.

PR #3756, opened 2025-04-03
Fix Clippy warnings, 0.7 backport

divergentdave:
Backport of #3755


jcjones (approved):
also the protobuf/backoff rustsec advisories

Issue #3505, opened 2024-11-21
DAP-13: indicate unknown report extensions during upload.


branlwyd:
DAP-13 [specified a method](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/630) for aggregator implementations to indicates which report extensions where unknown when a report upload failed due to the inclusion of an unknown report extension type.

Indicating unknown extensions is optional; this issue tracks that work effort.

PR #3768, opened 2025-04-10
Harmonize errors with draft-ietf-ppm-dap-15

tgeoghegan:
draft-ietf-ppm-dap-15 (forthcoming) will make some changes to the fart problem types defined by DAP.

- `unauthorizedRequest` is gone. We still have `aggregator::Error::UnauthorizedRequest`, but now we construct a problem document with a URN pointing to Divvi Up docs (which will be updated in [133][1].
- `batchQueriedMultipleTimes` is gone, and now instead we are instructed to provide `invalidMessage` if someone tries to change a collection or aggregate share's aggregation parameter. Accordingly we remove `aggregator::Error::BatchQueriedMultipleTimes`.
- `aggregator::Error::MissingTaskId` was not used anywhere except tests.
- We no loner need `DapProblemTypeExt` because all DAP problem types can now be represented as HTTP 400 Bad Request, per DAP's guidance to use the most general possible status.
- We add a problem type `unsupportedExtension`. However we don't use it, since it's only prescribed when the Leader rejects a report, and our Leader doesn't currently handle any extensions (we just opaquely write them to the datastore).
- I re-ordered the `DapProblemType` enum to match the table in draft-ietf-ppm-dap-15.

[1]: https://github.com/divviup/public-docs/issues/133

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_driver/tests.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver/tests.rs
@@ -1610,7 +1610,7 @@ async fn leader_sync_leader_selected_aggregation_job_init_single_step() {
                 .unwrap()
                 .path(),
         )
-        .with_status(500)
+        .with_status(400)
         .with_header("Content-Type", "application/problem+json")
         .with_body("{\"type\": \"urn:ietf:params:ppm:dap:error:invalidMessage\"}")
         .create_async()
@@ -1660,7 +1660,7 @@ async fn leader_sync_leader_selected_aggregation_job_init_single_step() {
     assert_matches!(
         error,
         Error::Http(error_response) => {
-            assert_eq!(error_response.status(), StatusCode::INTERNAL_SERVER_ERROR);
+            assert_eq!(error_response.status(), StatusCode::BAD_REQUEST);
             assert_eq!(*error_response.dap_problem_type().unwrap(), DapProblemType::InvalidMessage);
         }
     );
--- a/aggregator/src/aggregator/aggregation_job_driver/tests.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver/tests.rs
@@ -1612,7 +1612,7 @@ async fn leader_sync_leader_selected_aggregation_job_init_single_step() {
         )
         .with_status(500)
         .with_header("Content-Type", "application/problem+json")
-        .with_body("{\"type\": \"urn:ietf:params:ppm:dap:error:unauthorizedRequest\"}")
+        .with_body("{\"type\": \"urn:ietf:params:ppm:dap:error:invalidMessage\"}")
         .create_async()
         .await;
     let (header, value) = agg_auth_token.request_authentication();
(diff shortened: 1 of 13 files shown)


tgeoghegan:
Some points for discussion that can feed back into the specification:
- We drop some information by downgrading `batchQueriedMultipleTimes` into `invalidMessage`. On the other hand, I don't think a more specific error is all that valuable, because a collector changing the aggregation param on a collection or the leader doing so on an aggregate share seems like a bug or misconfiguration rather than something a sound implementation could recover from automatically.
- It's odd that the leader specifically returns `unsupportedExtension`, but the helper would return a `PrepareResp` with `ReportError::invalid_message`. I think the idea is that a client could try again, dropping the unsupported extensions, but there's no way for a leader to delete the unsupported extensions from a helper's report share.

divergentdave:
> our Leader doesn't currently handle any extensions (we just opaquely write them to the datastore).

This is separately a problem because we're supposed to reject reports with unrecognized extensions.

tgeoghegan:
> This is separately a problem because we're supposed to reject reports with unrecognized extensions.

I agree. We could defer that to the aggregation phase, but doing it during upload seems easy enough (it's not like we have to do expensive storage txns to find out what extensions we support). We already have #3505 which seems relevant.

aggregator/src/aggregator/aggregation_job_driver/tests.rs
```diff
@@ -1612,7 +1612,7 @@ async fn leader_sync_leader_selected_aggregation_job_init_single_step() {
         )
         .with_status(500)
```
divergentdave, line 1613: nit: This should probably be 400 to make the mock a bit more realistic

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -1255,7 +1255,7 @@ mod tests {
             .match_body(leader_request.get_encoded().unwrap())
             .with_status(500)
```
divergentdave, line 1256: Same here, this mock should probably be using 400

aggregator/src/aggregator/http_handlers/tests/aggregate_share.rs
```diff
@@ -595,8 +595,8 @@ async fn aggregate_share_request() {
             take_problem_details(&mut test_conn).await,
             json!({
                 "status": Status::BadRequest as u16,
-                "type": "urn:ietf:params:ppm:dap:error:batchQueriedMultipleTimes",
-                "title": "The batch described by the query has been queried already.",
+                "type": "urn:ietf:params:ppm:dap:error:invalidMessage",
+                "title": "The message type for a response was incorrect or the payload was malformed.",
```
divergentdave, line 599: Seeing this makes me wonder if we should plumb the second field of `Error::InvalidMessage` to the detail field of the response.

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -128,7 +125,12 @@ async fn run_error_handler(error: &Error, mut conn: Conn) -> Conn {
         Error::UnrecognizedCollectionJob(_, _) => conn.with_status(Status::NotFound),
 
         Error::UnauthorizedRequest(task_id) => conn.with_problem_document(
-            &ProblemDocument::new_dap(DapProblemType::UnauthorizedRequest).with_task_id(task_id),
+            &ProblemDocument::new(
+                "https://docs.divviup.org/references/janus-errors#unauthorized-request",
+                "The request's authorization is not valid.",
+                Status::Forbidden,
+            )
+            .with_task_id(task_id),
```
divergentdave, line 133: I think just using a 403 status code would be sufficient for this case (similar to the UnrecognizedCollectionJob case above). If we want to provide more detail, we could still use a problem document with `about:blank`.

divergentdave (approved):

Issue #3725, opened 2025-03-12
Migrate from 'backoff' to 'backon'


jcjones:
cargo-deny has identified that `backoff` is unmaintained:

```
error[unmaintained]: `backoff` is unmainted.
   ┌─ /github/workspace/Cargo.lock:47:1
   │
47 │ backoff 0.4.0 registry+https://github.com/rust-lang/crates.io-index
   │ ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ unmaintained advisory detected
   │
   ├ ID: RUSTSEC-2025-0012
   ├ Advisory: https://rustsec.org/advisories/RUSTSEC-2025-0012
   ├ The [backoff](https://crates.io/crates/backoff) crate is no longer actively maintained. For exponential backoffs/retrying, you can use the [backon](https://crates.io/crates/backon) crate.
   ├ Announcement: https://github.com/ihrwein/backoff/issues/66
   ├ Solution: No safe upgrade is available!
```

`backon` is not a drop-in replacement, so it requires some re-tooling.

PR #3769, opened 2025-04-10
Migrate from 'backoff' to 'backon'

jcjones:
`cargo-deny` has identified that `backoff` is unmaintained. This changeset moves to the replacement, `backon`, which has different semantics.

A few things to orient reviewers about this change:
* [The Backon `Backoff` types aren't copy or clone](https://docs.rs/backon/1.5.0/backon/trait.Backoff.html), so the idomatic thing is to pass around the `BackoffBuilder` types which _are_. So everywhere we passed around a `Backoff` we're now passing around a `BackoffBuilder` of some kind.
* Since neither the `ExponentialBuilder` nor `ExponentialBackoff` types expose any of their internals back out, for the places where Janus wants to know what the configured maximum time is, I've introduced a composite type `ExponentialWithTotalDelayBuilder`.
* The semantics for [Backon's `with_total_delay`](https://docs.rs/backon/1.5.0/backon/struct.ExponentialBuilder.html#method.with_total_delay) are different than our semantics were for max_elapsed_time, which affects the `JobDriverConfig`: we go from elapsed, wall-clock time to the sum of the backoff intervals. That's fairly close, so I updated the docs and did _not_ change the config parameter this time, so job configs do not need an update. I am not sure that's the right decision!
* The `Backon` defaults for the `ExponentialBackoff` are to retry 3 times and stop, which isn't what we want in a bunch of places, so that's why some `.without_max_times()` are sprinkled about.
* Unlike `Backoff`, `Backon` has a useful `ConstantBackoff`, so I prefer it when that's what the tests were trying for anyway.
* I've worked to adopt idiomatic Rust everywhere I know how to. I'll call out the specifics that _I_ see where I didn't know what to do. Help obviously appreciated.

Fixes https://github.com/divviup/janus/issues/3725


aggregator/src/config.rs
```diff
@@ -224,8 +224,8 @@ pub struct JobDriverConfig {
         alias = "retry_max_interval_millis"
     )]
     pub retry_max_interval_ms: u64,
-    /// The maximum elapsed time, in milliseconds, to wait before giving up on retrying a retryable
-    /// HTTP request.
+    /// The maximum retry time, in milliseconds, to wait before giving up on retrying a retryable
+    /// HTTP request. Note this is the sum of the retry intervals, not wall-clock time.
     #[serde(
         default = "JobDriverConfig::default_retry_max_elapsed_time_ms",
         alias = "retry_max_elapsed_time_millis"
```
jcjones, line 231: As noted in the top-comment, I am of two minds whether we should rename this. 
```suggestion
        default = "JobDriverConfig::default_retry_max_delay_ms",
        alias = "retry_max_delay_millis"
```

collector/src/lib.rs
```diff
@@ -698,10 +707,16 @@ impl<V: vdaf::Collector> Collector<V> {
...
             };
 
-            let backoff_duration = if let Some(duration) = backoff.next_backoff() {
+            let backoff_duration = if let Some(duration) = backoff.next() {
                 duration
             } else {
-                // The maximum elapsed time has expired, so return a timeout error.
+                // Either the maximum retries or the maximum delay has been reached
+                // so return a timeout.
+                warn!(
```
jcjones, line 715: I'm fine taking these `warn!`s back out, but they were super valuable for debugging timeouts here.

integration_tests/tests/integration/common.rs
```diff
@@ -141,32 +143,26 @@ where
...
+        }
+    })
+    .retry(backoff)
+    .when(|e| {
+        if let janus_collector::Error::Http(error_response) = e {
+            if let Some(DapProblemType::InvalidBatchSize) = error_response.dap_problem_type() {
+                return true;
             }
         }
+        false
```
jcjones, line 165: This block screams to me that there must be some clever way to do this without indentation, but I haven't figured it out yet. Some kind of `match`? Clever use of `matches!`?

core/src/retries.rs
```diff
@@ -550,4 +641,49 @@ mod tests {
         assert_eq!(response.title(), Some("too many eels"));
         mock_problem.assert_async().await;
     }
+
+    #[tokio::test]
```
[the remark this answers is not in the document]
jcjones: done in fa485f2d

integration_tests/tests/integration/common.rs
jcjones: Maybe
```rust
if let janus_collector::Error::Http(error_response) = e {
            return error_response.dap_problem_type()
                     .is_some_and(|t| t==DapProblemType::InvalidBatchSize);
}
```

divergentdave: Rather than use a combinator, I'd recommend changing the outer if statement to let-else, and doing an early return in the else block.

tgeoghegan: Code golf: I believe this [match guard](https://doc.rust-lang.org/book/ch19-03-pattern-syntax.html#extra-conditionals-with-match-guards) works:

```rust
    .when(|e| {
        matches!(e, janus_collector::Error::Http(error_response) if error_response.dap_problem_type() == Some(&DapProblemType::InvalidBatchSize))
    })
```

It's not much prettier after the formatter gets done with it, but it does indent less.

Cargo.toml
```diff
@@ -27,7 +27,7 @@ anyhow = "1"
 assert_matches = "1"
 async-trait = "0.1"
 aws-lc-rs = "1.13.0"
-backoff = "0.4.0"
+backon = "1.5.0"
```
tgeoghegan, line 30: Do we ever not enable feature `tokio-sleep`? Is it easier to enable the feature up here and remove it from the other workspace `Cargo.toml`s?

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -104,7 +104,7 @@ pub struct AggregationJobDriver<B> {
 
 impl<R> AggregationJobDriver<R>
 where
-    R: Backoff + Clone + Send + Sync + 'static,
+    R: BackoffBuilder + Copy + Send + Sync + 'static,
```
tgeoghegan, line 107: nit: I think we could delete `Send + Sync` here because [`BackoffBuilder` already implies those](https://docs.rs/backon/1.5.0/backon/trait.BackoffBuilder.html).

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -60,7 +60,7 @@ pub struct CollectionJobDriver<B> {
 
 impl<R> CollectionJobDriver<R>
 where
-    R: Backoff + Clone + Send + Sync + 'static,
+    R: BackoffBuilder + Copy + Send + Sync + 'static,
```
tgeoghegan, line 63: As above, I think the `Send + Sync` is redundant.

aggregator/src/aggregator/queue.rs
```diff
@@ -436,30 +438,27 @@ mod tests {
...
         debug!("spawning requests");
         for _ in 0..(concurrency as usize + depth) {
             let handler = Arc::clone(&handler);
-            let backoff = backoff.clone();
+            let backoff = backoff.build();
             requests.push(tokio::spawn({
                 async move {
-                    retry(backoff, || {
+                    (|| async {
                         let handler = Arc::clone(&handler);
```
tgeoghegan, line 453: I think that since you remove the inner `async move` closure, we can remove this `clone`.

aggregator/src/binary_utils.rs
```diff
@@ -105,26 +105,25 @@ pub async fn database_pool(db_config: &DbConfig, db_password: Option<&str>) -> R
...
+
+    let _ = (|| async { pool.get().await })
+        .retry(backoff)
+        .when(|error| match error {
+            PoolError::Timeout(TimeoutType::Create) | PoolError::Backend(_) => {
+                debug!(?error, "transient error connecting to database");
+                true
+            }
+            _ => false,
+        })
```
tgeoghegan, line 124: I think using `matches!` would be a bit more compact, but maybe that wouldn't let you include the `debug!` call.

aggregator/src/config.rs
tgeoghegan: Renaming config options is kind of a chore, and I think the actual difference between the sum of retry intervals and the wall time wouldn't be a big deal anyway. I don't think we ever made particularly strong promises about the accuracy of the max elapsed time.

collector/src/lib.rs
tgeoghegan: If they're useful for debugging, let's move them to `debug!`.

core/src/retries.rs
```diff
@@ -22,22 +22,91 @@ fn find_io_error(original_error: &reqwest::Error) -> Option<&std::io::Error> {
     None
 }
 
-/// An [`ExponentialBackoff`] with parameters suitable for most HTTP requests. The parameters are
-/// copied from the parameters used in the GCP Go SDK[1].
+#[derive(Copy, Clone, Debug)]
```
tgeoghegan, line 25: nit: put derives below doccomment.

core/src/retries.rs
```diff
@@ -22,22 +22,91 @@ fn find_io_error(original_error: &reqwest::Error) -> Option<&std::io::Error> {
...
+/// Extends the Backon ExponentialBackoff and ExponentialBuilder to add
+/// a `total_delay` field, which we use for deadlining in the collector's
+/// `poll_until_complete` method, as the underlying Backoff implementation
+/// does not expose that delay after it's set.
```
tgeoghegan, line 29: nit: we wrap comments to 100 lines by convention.

core/src/retries.rs
```diff
@@ -22,22 +22,91 @@ fn find_io_error(original_error: &reqwest::Error) -> Option<&std::io::Error> {
...
+        self.builder = self.builder.without_max_times();
+        self
+    }
+
+    pub const fn with_jitter(mut self) -> Self {
+        self.builder = self.builder.with_jitter();
+        self
+    }
+
+    pub fn with_total_delay(mut self, d: Option<Duration>) -> Self {
```
tgeoghegan, line 79: Could this take `Duration` instead of `Option<Duration>`? Does anyone ever need to clear the total delay that's been set on one of these?

tgeoghegan (approved):
Some nits and code golf suggestions, but this looks sound to me. I appreciate your thoroughness on the deadline matter. I probably would have gotten lazy and broken the API for the collect crate.

core/src/retries.rs
jcjones: fixed in 601e8743d57f35a8aaad20cc8bad106226e1fc0f

integration_tests/tests/integration/common.rs
jcjones: I .. was not coming up with that syntax. Clever. Maybe I'll get it right next next time!

Applied in a7c4ff1968a31c3e7c09647eadf1dd6e3a30e3a9

collector/src/lib.rs
jcjones: Fixed in 9ecf65b784efbded5487e8d19ee56345c7b5e2ef

Cargo.toml
jcjones: Fixed in 6092c1d384690e438efc4b4021ec06f5bac063ed

aggregator/src/aggregator/collection_job_driver.rs
jcjones: Good catch, fixed in ee0aee9aef843d40c559cb6aca39006f4f819b4e

tgeoghegan (approved):
Re-approving to satisfy the requirement of >1 review from the committers group.

interop_binaries/src/lib.rs
```diff
@@ -608,16 +583,19 @@ pub mod test_util {
...
+        let _ = (|| async {
+            (|| async { http_client.post(url.clone()).send().await })
+                .retry(backoff.build())
+                .await?
                 .error_for_status()
-                .map_err(backoff::Error::transient)
         })
-        .await
+        .retry(backoff.build())
+        .await;
```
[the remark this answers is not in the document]
jcjones: Yeah, that trait bound is definitely why there's the additional closure. 

The two-stage retry is not-optimal, and honestly I meant to come back to this before posting, I forgot all about it. I was having test issues that this layering resolved, but now I bet those will be fixed by the rest of the fixes.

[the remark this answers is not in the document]
jcjones: I've reworked this in 071a97ac.

core/src/retries.rs
jcjones: I'd rather leave it as-is, to make these all consistent without having to use our composite type where it's unnecessary.

PR #3792, opened 2025-04-23
Enforce time_precision for Client Reports and Report Aggregations

jcjones:
This change enforces that timestamps for client reports and for report aggregations are divisible by the `time_precision` of a task.

NOTE: This does _not_ enforce that `Duration`s of `Interval`s be multiples of `time_precision`, that will be a follow-on PR. (This one is big enough!)

* Adds `validate_precision` to `TimeExt`, which yields `janus_messages::Error::InvalidParameter` if the timestamp of `self` is not divisible by the `time_precision`.
* Adds calls to this `validate_precision` to the datastore routines for `put_report_aggregation` and `put_client_report`, covering aggregations and reports.
* Adds trait `ClockExt` to aggregator's `test_util`, and a method `now_at_batch_interval_start`, which makes all the test copypasta more manageable. 
* Revisits every test that uses report timestamps, _generally_ setting TaskBuilder's `with_time_precision` to `100` seconds (though some I did `10`) and aligning the test as needed. For a couple of tests that had more significant timing logic, I used `time_precision` of `1` second, which made for no other test changes.

I was unsure about putting `ClockExt` into `test_util` while our other `...Ext` traits live elsewhere, but it's really test-only. I'm happy to refactor it elsewhere, let me know.


jcjones:
It's also worth pointing out that I put the validation of timestamps at the datastore layer, because the information needed was available there. It's feasible to move that up, it just increases the number of times we call the (admittedly already caching) task lookup datastore method.

aggregator/src/aggregator/test_util.rs
```diff
@@ -200,3 +203,15 @@ pub async fn assert_task_aggregation_counter(
         sleep(Duration::from_millis(100)).await;
     }
 }
+
+/// Extension methods on [`Clock`].
+pub trait ClockExt: Sized {
```
tgeoghegan, line 208: The reason we do extension traits on other time related types has to do with the layering of the crates. In particular, `janus_messages` gets used by Firefox's DAP client (I think; or at least we want that to be possible) so we want it to have minimal dependencies, and for that reason it doesn't pull in `chrono`. Instead we do that in `janus_core` and up.

However it seems like your objective is to have some functionality on `Clock`s that's only available in test configurations. Sticking this in `test_util` achieves that, but it could be simpler to stick this method in `janus_core::time::Clock` with `#[cfg(test)]` or gate it on feature `test_util`. Then at least all the clock logic is in one place. However I don't object to how you've done it here.

aggregator/src/aggregator/test_util.rs
```diff
@@ -200,3 +203,15 @@ pub async fn assert_task_aggregation_counter(
...
+    fn now_at_batch_interval_start(&self, time_precision: &janusDuration) -> Time;
```
tgeoghegan, line 209: I find the name of this method confusing, but I think that's due in part to the underlying `to_batch_interval_start` being badly named, too. That method is documented as "Compute the start of the batch interval containing this Time, given the task time precision" but I don't think that's right: what it does is round the time down to the provided precision. We must have changed what it does along the way without renaming it.

In any case, I think we could name this `ClockExt` method better. Is `ClockExt::now_aligned_to_precision` correct in your view?

tgeoghegan:
Thanks for slogging through all the time stuff in the tests. I know from experience that can be a real chore.

core/src/time.rs
```diff
@@ -184,6 +184,9 @@ pub trait TimeExt: Sized {
         time_precision: &Duration,
     ) -> Result<Self, janus_messages::Error>;
 
+    /// Confirm that the time is a multiple of the task time precision.
+    fn validate_precision(&self, time_precision: &Duration) -> Result<Self, janus_messages::Error>;
```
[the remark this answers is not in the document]
jcjones: That's a cool idea!

aggregator/src/aggregator/test_util.rs
jcjones: I think that's a great name.

jcjones: Updated in 693c69b2868b38242eca1584aa5c1e2ec56004d4

aggregator/src/aggregator/test_util.rs
jcjones: I think 23805016 works!

aggregator_core/src/datastore.rs
```diff
@@ -5722,6 +5733,9 @@ struct TaskInfo {
 
     /// The task's report expiry age, corresponding to the tasks.report_expiry_age column.
     report_expiry_age: Option<chrono::Duration>,
+
+    /// The task's time precision
+    time_precision: Duration,
```
tgeoghegan, line 5738: Hmm, mixing `chrono::Duration` and `janus_messages::Duration` in a struct is a little smelly. But I don't think it's worth fixing here. We have an issue backlogged for not using on-the-wire time types in business logic.

tgeoghegan (approved):
Most excellent

Issue #3785, opened 2025-04-21
Rework 'tolerable clock skew' in light of rounding reports to the task time precision


jcjones:
With the change from #3780, we need to rework the 'tolerable clock skew', minimally the default value in `aggregator_core/src/task.rs`, but probably as used elsewhere.

Since report timestamps [are rounded to the task time precision](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-4.5), clock skews less than the time precision are automatically tolerated due to report timestamps being rounded-down. 

We should:
* Change the default tolerable clock skew for tasks in task.rs,
* Add validation that the tolerable clock skew is not less than, or perhaps is a multiple of, the time precision,
* Add notes to documentation saying this.

jcjones:
https://github.com/divviup/janus/blob/8990a7dc565c93a4da19e4fc3d81cca35bc5f641/aggregator_core/src/task.rs#L1134-L1135

The default task tolerable clock skew is 10m with a time precision of 8h.

PR #3796, opened 2025-05-02
Use task time precisions for aggregation jobs Durations and Intervals.

### LINKED ISSUE #3780: Verify all request timestamps are truncated by time_precision
The dap-14 change to [timestamps requires they be multiples](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#timestamps) of `time_precision`.

The designated four places timestamps are used all need to check this and treat malformed timestamps as [errors](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-3.2).

jcjones:
This change means that Intervals must start on, and must have their duration be, multiples of the task time precision. This means that intervals in the tests are often changing from numbers such as `[100, 101)` to `[100, 200)`, due to a 100s time precision.

This appears significant, but is logically the same, as there's still no overlap with the next interval.

- Adds `validate_precision` to `Duration` and `Interval` and adds a test.
- Adds a new `datastore::Error::TimeUnaligned` and returns it from `validate_precision` failures.
  - Note: I don't actually return it directly because of scoping. I didn't want to add a one-off _type_ to `janus_messages` just to be able to use a `From`, as can't do that with an enum. Perhaps that's the wrong choice?
- Converts the new `TimeUnaligned` error into a 400 `BadRequest` in the `aggregator_api` and adds a test.
- Add a new Interval-creation method, `from_time_with_precision`, and use it for batches and batch tests.
- Because batches now have wider intervals, the garbage collection test now needs to proceed past the wider interval to operate.
- Because of the wider intervals, the `aggregator::http_handlers::tests::report::upload_handler` test had to adopt a wide tolerable clock skew. This will be re-addressed in https://github.com/divviup/janus/issues/3785, when clock skews get normalized to be multiples of the time precision.
- Note that _batch intervals_, like those produced by `CollectableBatchMode::to_batch_interval`, are still not checked for alignment. Creating them aligned is difficult. See https://github.com/divviup/janus/issues/3797.
- Catches a couple of places where timestamps should have been validated in #3792.

Fixes #3780.


aggregator_api/src/lib.rs
```diff
@@ -194,6 +194,7 @@ impl Handler for Error {
             Self::Db(err) => match err {
                 datastore::Error::MutationTargetNotFound => conn.with_status(Status::NotFound),
                 datastore::Error::MutationTargetAlreadyExists => conn.with_status(Status::Conflict),
+                datastore::Error::TimeUnaligned => conn.with_status(Status::BadRequest),
```
tgeoghegan, line 197: Module `aggregator_api` is the control plane API used by Divvi Up API to do things like create tasks or obtain metrics about how many reports are prepared. I would think that we want these `datastore::Error::TimeUnaligned` errors to get handled in the DAP API, implemented in the `aggregator` module. So maybe we need something like this case in [`run_error_handler`](https://github.com/divviup/janus/blob/087fbfd864811538bdf53b5f8c9f32a245d65a52/aggregator/src/aggregator/http_handlers.rs#L50)'s handling of `Error::Datastore`?

aggregator_core/src/datastore.rs
```diff
@@ -6025,6 +6089,9 @@ pub enum Error {
     /// An invalid parameter was provided.
     #[error("invalid parameter: {0}")]
     InvalidParameter(&'static str),
+    /// A time type (duration, interval, timestamp) was not aligned to the task precision.
+    #[error("time unaligned to task precision")]
+    TimeUnaligned,
```
tgeoghegan, line 6094: It's unfortunate to have to discard underlying errors everywhere. If you don't want to commit to doing `TimeUnaligned(janus_messages::Error)`, another option is `TimeUnaligned(Box<dyn Error>)`, which would still preserve error chains and let us print them even if we can't destructure them. `TimeUnaligned(String)` also works (then you'd do something like `Error::TimeUnaligned(format!("{:?}", underlying_error)`), but `Box<dyn Error>` is nicer and either way, it's a variable length heap allocation.

core/src/time.rs
```diff
@@ -1,6 +1,7 @@
 //! Utilities for timestamps and durations.
 
 use chrono::{DateTime, NaiveDateTime, Utc};
+
```
tgeoghegan, line 4:
```suggestion
```

Prefer to leave all the `use`s together so the linter can sort them for us.

core/src/time.rs
```diff
@@ -124,6 +125,9 @@ pub trait DurationExt: Sized {
     /// Return a duration representing this duration rounded up to the next largest multiple of
     /// `time_precision`, or the same duration if it already a multiple.
     fn round_up(&self, time_precision: &Duration) -> Result<Self, Error>;
+
+    /// Confirm that this duration is a multiple of the task time precision.
+    fn validate_precision(self, time_precision: &Duration) -> Result<Self, janus_messages::Error>;
```
tgeoghegan, line 130: I think this type is already imported as `Error`

```suggestion
    fn validate_precision(self, time_precision: &Duration) -> Result<Self, Error>;
```

aggregator_core/src/datastore.rs
jcjones: Fixed in 3369f41d

aggregator_api/src/lib.rs
jcjones: @tgeoghegan Are you envisioning something like this?
```rust
Error::Datastore(inner) => match inner {
    datastoreError::TimeUnaligned(error) => conn.with_problem_document(
        &ProblemDocument::new("about:blank", "Time Unaligned.", Status::BadRequest)
            .with_detail(&error.to_string()),
    ),
    _ => conn.with_status(Status::InternalServerError),
},
```

Specifically:
* Do you think this is a BadRequest status here?
* Should I plan to add a new error to https://docs.divviup.org/references/janus-error ?

jcjones: I think I have this worked out in https://github.com/divviup/janus/pull/3796/commits/144184f41bb099ba32a1f2a531b9a27c15167e92, but not for the helper yet.

tgeoghegan: Yeah, this is basically what I had in mind. Two notes:

You could also:
```
Error::Datastore(datastoreError::TimeUnaligned(error)) => conn.with_problem_document(...)
```

...and then the case where it's `Error::Datastore` wrapping anything else would still be handled by the `Error::Hpke(_)|Error::Datastore(_)|Error::Vdaf(_)|(etc)` case below. Should let you de-dent the code by one step.

As for the problem document itself: it seems like in Slack, you and @divergentdave concluded that either `reportRejected` or `invalidMessage` in the DAP error URN space is appropriate. Either sounds good to me. But since your question is a good one: if we _did_ define a new problem document type, then yeah, the thing to do would be to add an entry for it to https://docs.divviup.org/references/janus-errors, and then use `https://docs.divviup.org/references/janus-errors#the-particular-error-type` as the problem document URN.

tgeoghegan: Ah, you already updated the error type in the most recent commit. 👍🏻

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -141,8 +142,19 @@ async fn run_error_handler(error: &Error, mut conn: Conn) -> Conn {
                 .with_task_id(&inner.task_id)
                 .with_detail(&inner.to_string()),
         ),
+        Error::Datastore(inner) => match inner {
+            datastoreError::TimeUnaligned(error) => conn.with_problem_document(
```
tgeoghegan, line 146: as observed elsewhere, try:
```suggestion
        Error::Datastore(datastoreError::TimeUnaligned(error)) => conn.with_problem_document(
```
...and put the `Error::Datastore(_)` case back, below. I think it's prettier, but it's subjective.
(this suggestion won't lint or even compile unfortunately)

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -234,10 +234,16 @@ async fn upload_handler() {
         AggregationMode::Synchronous,
         VdafInstance::Prio3Count,
     )
+    // Since HttpHandlerTest's clock instance is a copy of ours, we can't simply
+    // advance it, and we have to instead tolerate skew.
```
tgeoghegan, line 238: one does not simply *advance* the clock instance

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -389,6 +400,28 @@ async fn upload_handler() {
         &test_conn,
         "access-control-allow-origin" => "https://example.com/"
     );
+
+    // Reports that aren't aligned to the task time precision must be rejected
+    clock.advance(&Duration::from_seconds(1));
+    let bad_leader_time_alignment_report = create_report(
+        &leader_task,
+        &hpke_keypair,
+        clock.now(), /* not aligned! */
```
tgeoghegan, line 409: super annoying nit: evidently this test passes so this is fine... but I feel like it'd be cleaner to do something like `clock.now().add(&Duration::from_seconds(1));`, because what if the `clock` is shared via `Arc<>` into somewhere else, and advancing it here messes with some other state?

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -389,6 +400,28 @@ async fn upload_handler() {
...
+        .with_request_body(bad_leader_time_alignment_report.get_encoded().unwrap())
+        .run_async(&handler)
+        .await;
+    check_response(
+        &mut test_conn,
+        Status::BadRequest,
+        "invalid_message",
+        "Time unaligned.",
+        leader_task.id(),
+        Some("time is unaligned (precision = 1000 seconds, inner error = timestamp is not a multiple of the time precision)"),
```
tgeoghegan, line 422: nit: wrap long line at 100
```suggestion
        Some("time is unaligned (precision = 1000 seconds, \
            inner error = timestamp is not a multiple of the time precision)"),
```

aggregator_core/src/datastore.rs
```diff
@@ -649,10 +649,34 @@ WHERE success = TRUE ORDER BY version DESC LIMIT(1)",
...
+
+        if let Some(start) = task.task_start() {
+            start
+                .validate_precision(task.time_precision())
+                .map_err(|e| Self::unaligned_time_error(task.id(), task.time_precision(), e))?;
+        }
+        if let Some(end) = task.task_end() {
+            end.validate_precision(task.time_precision())
+                .map_err(|e| Self::unaligned_time_error(task.id(), task.time_precision(), e))?;
+        }
```
tgeoghegan, line 678: Down below, we're already doing:

```
&task
                        .task_start()
                        .map(Time::as_naive_date_time)
                        .transpose()?,
                    /* task_end */
                    &task.task_end().map(Time::as_naive_date_time).transpose()?,
```

So you could stick the calls to `validate_precision` into that `map` call. Something like:

```
&task.task_start().map(|t| { // Map Option<Time> to Option<Result<NaiveDateTime, err>>
    t.validate_precision(task.time_precision())
        .map(Time::as_naive_date_time)
        .map_err(...)
}).transpose()? // Transpose Option<Result<T, E>> to Result<Option<T>, E>> and then ? away the Result<>
```

But that's code golf. Maybe even code figure skating. To be honest, I don't hate having the error handling being obvious and up front instead of buried in a loooooong list of SQL query parameters.

aggregator_core/src/datastore.rs
```diff
@@ -6045,3 +6145,21 @@ impl From<aws_lc_rs::error::Unspecified> for Error {
         Error::Crypt
     }
 }
+
+/// Details of a [`Error::TimeUnaligned`] error.
+#[derive(Debug)]
+pub struct TimeUnaligned {
```
tgeoghegan, line 6151: Another option, which would let you define this inline in the error enum and save you the `Display` impl, would be:

```
pub enum Error {
    ...
    #[error("time is unaligned (precision = {1}, inner error = {2}")
    TimeUnaligned {
        task_id: TaskId,
        time_precision: Duration,
        inner_error: Box<janus_messages::Error>,
    }
    ...
}
```

aggregator_core/src/datastore.rs
jcjones: I tried this first, actually, but I don't know how to get the same gating effects as the `if let` when I also need to provide a parameter and can't just call `map(validate_precision)` to get the `Some` value.

Thinking of it again, I guess I'd need to define a closure that captured `task.time_precision()` and then call the closure on `task.task_start()`. I'll give that a try, for figure skating purposes.

jcjones: ah, I misunderstood your code block's closure. Yeah, that seems like it'd work.

jcjones: Gonna come back to this. This diff:
```diff
diff --git i/aggregator_core/src/datastore.rs w/aggregator_core/src/datastore.rs
index 77a28fca..d63d18b6 100644
--- i/aggregator_core/src/datastore.rs
+++ w/aggregator_core/src/datastore.rs
@@ -677,6 +677,13 @@ WHERE success = TRUE ORDER BY version DESC LIMIT(1)",
                 .map_err(|e| Self::unaligned_time_error(task.id(), task.time_precision(), e))?;
         }

+        let validate_task_precision_and_convert = |time: &Time| {
+            // Time to Result<NaiveDateTime, err>
+            time.validate_precision(task.time_precision())
+                .map(|t| Time::as_naive_date_time(&t))
+                .map_err(|e| Self::unaligned_time_error(task.id(), task.time_precision(), e))
+        };
+
         // Main task insert.
         let stmt = self
             .prepare_cached(
@@ -710,10 +717,19 @@ ON CONFLICT DO NOTHING",
                     /* task_start */
                     &task
                         .task_start()
-                        .map(Time::as_naive_date_time)
+                        .map(|time| {
+                            time.validate_precision(task.time_precision())
+                                .map(|t| Time::as_naive_date_time(&t))
+                                .map_err(|e| {
+                                    Self::unaligned_time_error(task.id(), task.time_precision(), e)
+                                })
+                        })
                         .transpose()?,
                     /* task_end */
-                    &task.task_end().map(Time::as_naive_date_time).transpose()?,
+                    &task
+                        .task_end()
+                        .map(validate_task_precision_and_convert)
+                        .transpose()?,
                     /* report_expiry_age */
                     &task
                         .report_expiry_age()
```

is producing `std::result::Result<chrono::NaiveDateTime, janus_messages::Error>` which, uh, feels like I should be able to `?` down into a ToSql type, but it don't like it.

```
error[E0277]: the trait bound `std::result::Result<chrono::NaiveDateTime, janus_messages::Error>: postgres_types::ToSql` is not satisfied
   --> aggregator_core/src/datastore.rs:729:21
    |
729 | /                     &task
730 | |                         .task_end()
731 | |                         .map(validate_task_precision_and_convert)
732 | |                         .transpose()?,
    | |_____________________________________^ the trait `postgres_types::ToSql` is not implemented for `std::result::Result<chrono::NaiveDateTime, janus_messages::Error>`
    |
    = help: the following other types implement trait `postgres_types::ToSql`:
              &T
              &[T]
              &[u8]
              &str
              [T; N]
              [u8; N]
              bool
              chrono::DateTime<chrono::FixedOffset>
            and 38 others
    = note: required for `std::option::Option<std::result::Result<chrono::NaiveDateTime, janus_messages::Error>>` to implement `postgres_types::ToSql`
    = note: required for the cast from `&std::option::Option<std::result::Result<chrono::NaiveDateTime, janus_messages::Error>>` to `&dyn postgres_types::ToSql + std::marker::Sync`

```

tgeoghegan: What I meant to imply by "code figure skating" is that chaining maps and and_thens and so on like this might have aesthetic benefits, depending on subjective tastes, but I also wouldn't blame you for doing the obvious thing.

aggregator/src/aggregator/http_handlers.rs
jcjones: Matching syntax has so, so much to it. Very cool.

aggregator_core/src/datastore.rs
jcjones: Thanks for this. I first tried to do this, but I got caught up, believe it or not, interacting with the enum variant's internal fields during the matching in the `http_handlers.rs`, which you helpfully elucidated above. Very nice.

jcjones:
Everything but the datastore refactor are in https://github.com/divviup/janus/pull/3796/commits/4b9a702dc908553f6dd4ccc81a0d6fb3d856ab3e. 

The next follow-on to this change is https://github.com/divviup/janus/pull/3820 and has a lot more churn than this.

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -141,8 +142,23 @@ async fn run_error_handler(error: &Error, mut conn: Conn) -> Conn {
                 .with_task_id(&inner.task_id)
                 .with_detail(&inner.to_string()),
         ),
+        Error::Datastore(
+            error @ datastoreError::TimeUnaligned {
+                task_id,
+                time_precision: _,
+                inner_error: _,
```
tgeoghegan, line 149: nit: to more compactly elide the rest of the struct, use `..` as in `datastoreError::TimeUnaligned { task_id, .. }`

tgeoghegan (approved):
Just nits, please feel free to merge without re-review if you address them.

aggregator_api/src/lib.rs
```diff
@@ -194,6 +194,13 @@ impl Handler for Error {
             Self::Db(err) => match err {
                 datastore::Error::MutationTargetNotFound => conn.with_status(Status::NotFound),
                 datastore::Error::MutationTargetAlreadyExists => conn.with_status(Status::Conflict),
+                err @ datastore::Error::TimeUnaligned {
+                    task_id: _,
+                    time_precision: _,
+                    inner_error: _,
+                } => conn
```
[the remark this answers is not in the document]
jcjones: oh, yeah! 👍🏼

PR #3802, opened 2025-05-05
Upgrade testcontainers, use aws-lc-rs in more places

divergentdave:
This upgrades the testcontainers dependency, and changes feature flags on it and reqwest to use aws-lc-rs more. We now only depend on ring for tokio-postgres-rustls and rustls-platform-verifier. I had to explicitly initialize the rustls crypto provider in more tests, because reqwest only has special handling for there being no installed default crypto provider when its ring feature is enabled.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1548,6 +1548,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "etcetera"
+version = "0.10.0"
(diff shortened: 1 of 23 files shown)


tgeoghegan (approved):

divergentdave:
> Looks fine to me. The tests look like it could have been a spurious failure.

There was one more entry point that needed the same setup, in the interop test client/collector servers.

PR #3820, opened 2025-05-16
[WIP] Batch aggregation and collection job precision enforcement

jcjones:
NOTE: This is a follow-on to https://github.com/divviup/janus/pull/3796, and must be landed after that commit.

This adds calls to `validate_precision` to batch aggregation and collection job submissions. 

Note, due to the volume of test changes, this is not small.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_continue.rs
+++ b/aggregator/src/aggregator/aggregation_job_continue.rs
@@ -392,6 +392,7 @@ mod tests {
             VdafInstance::Fake { rounds: 2 },
         )
         .build();
+        let time_precision = task.time_precision().clone();
         let helper_task = task.helper_view().unwrap();
         let clock = MockClock::default();
         let ephemeral_datastore = ephemeral_datastore().await;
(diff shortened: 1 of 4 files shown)


aggregator_core/src/datastore/tests.rs
```diff
@@ -1690,7 +1690,7 @@ async fn roundtrip_aggregation_job(ephemeral_datastore: EphemeralDatastore) {
         random(),
         dummy::AggregationParam(23),
         batch_id,
-        Interval::new(OLDEST_ALLOWED_REPORT_TIMESTAMP, Duration::from_seconds(1)).unwrap(),
+        Interval::new(OLDEST_ALLOWED_REPORT_TIMESTAMP, *task.time_precision()).unwrap(),
```
jcjones, line 1693: not sure if it's cleaner to use `TIME_PRECISION` or the actual task value here. But it's a test...

tgeoghegan: Better to use `task.time_precision()` I think.

tgeoghegan (approved):

PR #3826, opened 2025-05-21
sync with libprio-rs changes for new VDAF spec

tgeoghegan:


Cargo.toml
```diff
@@ -76,7 +76,7 @@ postgres-types = "0.2.9"
 pretty_assertions = "1.4.1"
 # Disable default features so that individual workspace crates can choose to
 # re-enable them
-prio = { version = "0.18.1-alpha.0", default-features = false, features = ["experimental"] }
+prio = { version = "0.18.1-alpha.0", default-features = false, features = ["experimental"], git = "https://github.com/divviup/libprio-rs.git", rev = "2dc43855ada06f761a2d7a16aad53181d4357e00" }
```
tgeoghegan, line 79: We could release `prio 0.18.1-alpha.1` to crates.io, but since we don't yet ship this Janus version anywhere, I think pinning a git commit is good enough.

PR #3831, opened 2025-05-22
Adopt bulk upload API from DAP PR #686

tgeoghegan:
Plus my tweaks to the DAP PR in [this branch][1]

https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/686/files

[1]: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/tree/timg/bulk-upload

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -78,7 +78,8 @@ use janus_messages::{
     AggregationJobId, AggregationJobInitializeReq, AggregationJobResp, AggregationJobStep,
     BatchSelector, CollectionJobId, CollectionJobReq, CollectionJobResp, Duration, HpkeConfig,
     HpkeConfigList, InputShareAad, Interval, PartialBatchSelector, PlaintextInputShare,
-    PrepareResp, Report, ReportError, ReportIdChecksum, Role, TaskId,
+    PrepareResp, Report, ReportError, ReportIdChecksum, ReportUploadStatus, Role, TaskId,
+    UploadRequest, UploadResponse,
 };
 use opentelemetry::{
     metrics::{Counter, Histogram, Meter},
@@ -98,7 +99,7 @@ use prio::{
 use rand::{thread_rng, Rng};
 use reqwest::Client;
 use std::{
-    borrow::Cow,
+    borrow::{Borrow, Cow},
     collections::HashSet,
     fmt::Debug,
     panic,
@@ -351,9 +352,16 @@ impl<C: Clock> Aggregator<C> {
         Ok((encoded_hpke_config_list, signature))
     }
 
-    async fn handle_upload(&self, task_id: &TaskId, report_bytes: &[u8]) -> Result<(), Arc<Error>> {
-        let report =
-            Report::get_decoded(report_bytes).map_err(|err| Arc::new(Error::MessageDecode(err)))?;
+    async fn handle_upload(
+        &self,
+        task_id: &TaskId,
+        report_bytes: &[u8],
+    ) -> Result<UploadResponse, Arc<Error>> {
+        // Assume that report_bytes is the entire request body, and so its length is the HTTP
+        // Content-Length and can be used to decode the vector of reports.
+        let upload_request =
+            UploadRequest::get_decoded_with_param(&report_bytes.len(), report_bytes)
+                .map_err(|err| Arc::new(Error::MessageDecode(err)))?;
 
         let task_aggregator = self
             .task_aggregators
@@ -364,7 +372,12 @@ impl<C: Clock> Aggregator<C> {
             return Err(Arc::new(Error::UnrecognizedTask(*task_id)));
         }
         task_aggregator
-            .handle_upload(&self.clock, &self.hpke_keypairs, &self.metrics, report)
+            .handle_upload(
+                &self.clock,
+                &self.hpke_keypairs,
+                &self.metrics,
+                upload_request,
+            )
             .await
     }
 
@@ -983,8 +996,8 @@ impl<C: Clock> TaskAggregator<C> {
         clock: &C,
         hpke_keypairs: &HpkeKeypairCache,
         metrics: &AggregatorMetrics,
-        report: Report,
-    ) -> Result<(), Arc<Error>> {
+        upload_request: UploadRequest,
+    ) -> Result<UploadResponse, Arc<Error>> {
         self.vdaf_ops
             .handle_upload(
                 clock,
@@ -992,7 +1005,7 @@ impl<C: Clock> TaskAggregator<C> {
                 metrics,
                 &self.task,
                 &self.report_writer,
-                report,
+                upload_request,
             )
             .await
     }
@@ -1392,8 +1405,8 @@ impl VdafOps {
         metrics: &AggregatorMetrics,
         task: &AggregatorTask,
         report_writer: &ReportWriteBatcher<C>,
-        report: Report,
-    ) -> Result<(), Arc<Error>> {
+        upload_request: UploadRequest,
+    ) -> Result<UploadResponse, Arc<Error>> {
         match task.batch_mode() {
             task::BatchMode::TimeInterval => {
                 vdaf_ops_dispatch!(self, (vdaf, VdafType, VERIFY_KEY_LENGTH) => {
@@ -1404,7 +1417,7 @@ impl VdafOps {
                         metrics,
                         task,
                         report_writer,
-                        report,
+                        upload_request,
                     )
                     .await
                 })
@@ -1418,7 +1431,7 @@ impl VdafOps {
                         metrics,
                         task,
                         report_writer,
-                        report,
+                        upload_request,
                     )
                     .await
                 })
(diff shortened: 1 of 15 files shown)


tgeoghegan:
This PR does a lot of the necessary work. I think what's missing is:

- it doesn't build, because among other things it pulls in a local copy of crate `prio`
- there might be a more efficient way to handle writes of batches of reports. Then again, `report_writer.rs` already does this.
- we don't handle chunked encodings and I'm not sure if we handle HTTP/2 streams, which I suspect is what we'd want if we made real use of this feature. However that's not strictly necessary if this PR's goal is just conformity with draft-ietf-ppm-dap-16 rather than an efficient batch upload implementation.

messages/src/lib.rs
```diff
@@ -1540,6 +1540,130 @@ impl Decode for Report {
...
+}
+
+impl Encode for UploadResponse {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
+        encode_fixlen_items(bytes, &self.status)
+    }
+
+    fn encoded_len(&self) -> Option<usize> {
+        Some(ReportId::LEN + 1)
+    }
```
jcjones: Just as soon as I `push` I see this obvious bug. Pretty sure that the response isn't constant size anymore, oops....

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -408,56 +478,218 @@ async fn upload_handler() {
         "access-control-allow-origin" => "https://example.com/"
     );
 
-    // Reports that aren't aligned to the task time precision must be rejected
```
jcjones: There's no way to alert about time precision now, so I took this test out. Time precision misalignments are going away wholesale in https://github.com/divviup/janus/issues/4019.

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -218,17 +283,18 @@ async fn upload_handler() {
         report.helper_encrypted_input_share().clone(),
     );
     let mut test_conn = post(task.report_upload_uri().unwrap().path())
-        .with_request_header(KnownHeaderName::ContentType, Report::MEDIA_TYPE)
-        .with_request_body(bad_report.get_encoded().unwrap())
+        .with_request_header(KnownHeaderName::ContentType, UploadRequest::MEDIA_TYPE)
+        .with_request_body(
+            UploadRequest::from_slice(std::slice::from_ref(&bad_report))
```
tgeoghegan: TIL `std::slice::from_ref`. Is this different from `&[&bad_report]`?

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -612,11 +595,33 @@ async fn hpke_config_cors_preflight(mut conn: Conn) -> Conn {
 async fn upload<C: Clock>(
     conn: &mut Conn,
     (State(aggregator), BodyBytes(body)): (State<Arc<Aggregator<C>>>, BodyBytes),
```
tgeoghegan: We'll still read the entire body off the wire before we start processing the first report, right? I think that's fine just to reach -16 conformity and prove that the thing works, but we should at least note that more work is needed here to stream requests.

jcjones: That's right, `impl TryFromConn for BodyBytes` reads the whole body. I've filed https://github.com/divviup/janus/issues/4149 for this.

aggregator/src/aggregator/error.rs
```diff
@@ -238,21 +238,17 @@ pub enum ReportRejectionReason {
...
-            ReportRejectionReason::OutdatedHpkeConfig(_) => {
-                "Report is using an outdated HPKE configuration."
-            }
-            ReportRejectionReason::TaskNotStarted => "Task has not started.",
-            ReportRejectionReason::DuplicateExtension => "Report contains duplicate extensions.",
+            ReportRejectionReason::IntervalCollected => ReportError::ReportReplayed,
+            ReportRejectionReason::DecryptFailure => ReportError::HpkeDecryptError,
+            ReportRejectionReason::DecodeFailure => ReportError::InvalidMessage,
+            ReportRejectionReason::TaskEnded => ReportError::TaskExpired,
+            ReportRejectionReason::Expired => ReportError::TaskExpired,
```
divergentdave: This should use `report_dropped`, see https://www.ietf.org/archive/id/draft-ietf-ppm-dap-16.html#section-4.6.2.4-2.7.1

messages/src/tests/upload.rs
```diff
@@ -314,3 +316,12 @@ fn roundtrip_input_share_aad() {
         ),
     )])
 }
+
+#[test]
+fn upload_response_size() {
```
divergentdave: We should add `roundtrip_encoding()` tests for `UploadRequest` and `UploadResponse` for completeness. That helper function will check the `encoded_len()` hint correctness as well.

aggregator/src/aggregator.rs
```diff
@@ -372,9 +373,16 @@ impl<C: Clock> Aggregator<C> {
         Ok((encoded_hpke_config_list, signature))
     }
 
-    async fn handle_upload(&self, task_id: &TaskId, report_bytes: &[u8]) -> Result<(), Arc<Error>> {
-        let report =
-            Report::get_decoded(report_bytes).map_err(|err| Arc::new(Error::MessageDecode(err)))?;
+    async fn handle_upload(
+        &self,
+        task_id: &TaskId,
+        report_bytes: &[u8],
```
divergentdave: We should change this argument's name as well.
```suggestion
        request_bytes: &[u8],
```

client/src/lib.rs
```diff
@@ -550,14 +551,17 @@ impl<V: vdaf::Client<16>> Client<V> {
...
+    /// Upload one or more measurements to the leader, per the [DAP specification][1]. The provided
+    /// measurements are sharded into two shares and then uploaded to the leader.
     ///
     /// [1]: https://www.ietf.org/archive/id/draft-ietf-ppm-dap-07.html#name-uploading-reports
-    #[tracing::instrument(skip(measurement), err)]
-    pub async fn upload(&self, measurement: &V::Measurement) -> Result<(), Error> {
-        self.upload_with_time(measurement, Clock::now(&RealClock::default()))
-            .await
+    #[tracing::instrument(skip(measurements), err)]
+    pub async fn upload(&self, measurements: &[V::Measurement]) -> Result<(), Error> {
```
divergentdave: I think it would be preferable if `Client::upload()` still handled just one report, and we added another method for sharding and submitting multiple reports. (or just relied on `upload_with_time()` as the most general API) Since many measurement types are already vectors, this would make the common case easier to understand. Besides, I think the primary use case for submitting multiple reports per request will be intermediaries collecting reports from multiple systems. Sharding and submitting multiple reports on the same system seems like it will mainly be needed for testing purposes and not practical applications.

messages/src/tests/upload.rs
jcjones: yeaaah, I didn't do it because `roundtrip_encoding` doesn't handle the new ParameterizedDecode types. I'll figure it out...

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -44,26 +45,34 @@ use trillium_tokio::Stopper;
...
+        let body = &take_response_body(test_conn).await;
+        let expected_content_len = format!("{}", body.len());
+        let len_str = expected_content_len.as_str();
+        assert_headers!(&test_conn, "content-length" => len_str);
+        let upload_response = UploadResponse::get_decoded_with_param(&body.len(), body).unwrap();
+
+        for status in upload_response.status() {
+            assert_eq!(status.report_id(), *desired_report_id);
+            assert_eq!(status.error(), desired_report_error);
         }
```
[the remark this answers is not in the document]
divergentdave: I want to make sure that the decoded upload response doesn't have an empty list of report statuses here.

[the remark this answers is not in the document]
jcjones: (`Some(17)` actually)

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -62,26 +62,9 @@ async fn run_error_handler(error: &Error, mut conn: Conn) -> Conn {
...
-                    .with_detail(rejection.reason().detail()),
-            ),
-            _ => conn.with_problem_document(
-                &ProblemDocument::new_dap(DapProblemType::ReportRejected)
-                    .with_task_id(rejection.task_id())
-                    .with_detail(rejection.reason().detail()),
-            ),
-        },
+        Error::ReportRejected(_) => {
+            panic!("no report rejected error should make it to the error handler")
```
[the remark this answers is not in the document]
jcjones: https://github.com/divviup/janus/issues/4150.

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -408,56 +478,218 @@ async fn upload_handler() {
...
+        upload_response.status()[0].report_id(),
+        *expired_report.metadata().id()
+    );
+    assert_eq!(
+        upload_response.status()[0].error(),
+        ReportError::TaskExpired
+    );
+
+    // Verify that the two valid reports were actually written to the datastore
+    // by checking they can't be uploaded again
```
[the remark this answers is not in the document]
jcjones: Fixed in 4ab57fe0 by adding a new helper, `count_client_reports_for_task`

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -44,26 +45,35 @@ use trillium_tokio::Stopper;
...
+        let expected_content_len = format!("{}", body.len());
+        let len_str = expected_content_len.as_str();
+        assert_headers!(&test_conn, "content-length" => len_str);
+        let upload_response = UploadResponse::get_decoded_with_param(&body.len(), body).unwrap();
+
+        for status in upload_response.status() {
+            assert_eq!(status.report_id(), *desired_report_id);
+            assert_eq!(status.error(), desired_report_error);
+            assert_eq!(status.encoded_len(), Some(17)); // TKTK
         }
```
divergentdave: Here's what I was thinking
```suggestion
        assert_eq!(upload_response.status().len(), 1);
        for status in upload_response.status() {
            assert_eq!(status.report_id(), *desired_report_id);
            assert_eq!(status.error(), desired_report_error);
        }
```

client/src/lib.rs
jcjones: I totally agree -- I inherited this from Tim's original PR, and thinking about it this way, well, I reverted in 701c9239.

aggregator/src/aggregator.rs
```diff
@@ -1727,7 +1740,49 @@ impl VdafOps {
...
+            match Self::handle_uploaded_report::<SEED_SIZE, B, A, C>(
+                Arc::clone(&vdaf),
+                clock,
+                hpke_keypairs,
+                metrics,
+                task,
+                report_writer,
+                report,
+            )
+            .await
```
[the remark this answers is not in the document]
jcjones: Good call. I had to use `join_all` because we don't want to fail on a `ReportRejected` error.

in 2d1f7d28

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -408,56 +479,231 @@ async fn upload_handler() {
...
+    let valid_report_2 = create_report(
+        &leader_task,
         &hpke_keypair,
-        /* public */ Vec::from([Extension::new(ExtensionType::Tbd, Vec::new())]),
-        /* leader */ Vec::from([Extension::new(ExtensionType::Tbd, Vec::new())]),
-        /* helper */ Vec::new(),
+        clock.now_aligned_to_precision(task.time_precision()),
     );
+
+    // Upload all three reports in a single batch
```
divergentdave:
```suggestion
    // Upload all four reports in a single batch
```

Issue #3827, opened 2025-05-21
Remove Interval::from_time


jcjones:
After PR #3820 it's only used in one place, and that'll easily factor out. Let's get rid of it.

PR #3835, opened 2025-05-23
Add missing fields to manual PartialEq implementations

### LINKED ISSUE #3834: `impl PartialEq for BatchAggregation` does not check `client_timestamp_interval`
> We do not currently check `client_timestamp_interval` in the manual `PartialEq` implementation.

_Originally posted by @divergentdave in https://github.com/divviup/janus/pull/3829#discussion_r2105221527_

We could also skim other manual trait implementations for similarly missing fields.

divergentdave:
This fixes #3834, and fixes test breakages by moving off of `Interval::from_time()` (see #3827). It also changes various manual implementations so that they start by destructuring their arguments. This uncovered one more missing field in the collector crate.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ constcat = "0.5"
 deadpool = "0.12.2"
 deadpool-postgres = "0.14.1"
 divviup-client = "0.4"
-educe = { version = "0.6.0", default-features = false, features = ["Debug"] }
+educe = { version = "0.6.0", default-features = false, features = ["Debug", "PartialEq", "Eq"] }
 fixed = "1.29"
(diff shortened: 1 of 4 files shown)


tgeoghegan (approved):
I like that the compiler will now complain if we add fields to these structures and fail to compare them. But could we derive the `PartialEq` implementations? Can `educe` help with special cases such as deliberately ignored fields?

divergentdave:
We can indeed derive these, and `educe`'s defaults for trait bounds just work even in cases where the standard library's macro fails.

tgeoghegan:
image
That's the good stuff

PR #3844, opened 2025-05-29
Add histograms for clock skew of report timestamps

### LINKED ISSUE #2729: Metrics: emit a histogram of clock skew in client timestamps in reports
Recently in production we observed huge numbers of aggregation jobs being created with just one report in them, causing a large backlog of jobs. @divergentdave was able to figure this out by spelunking into the database. It'd be nice to have a metric emitted by Janus that would maintain a histogram of how far off client reports are.

divergentdave:
This adds a pair of histograms to track the positive or negative clock skew of report timestamps relative to the aggregator's clock. Closes #2729.

### DIFF
--- a/aggregator/src/metrics.rs
+++ b/aggregator/src/metrics.rs
@@ -341,3 +341,27 @@ pub(crate) fn aggregate_step_failure_counter(meter: &Meter) -> Counter<u64> {
 
     aggregate_step_failure_counter
 }
+
+const TIME_SKEW_HISTOGRAM_VALUES: &[f64] = &[
+    30.0, 60.0, 120.0, 300.0, 600.0, 1800.0, 3600.0, 7200.0, 14400.0, 28800.0, 86400.0,
+];
+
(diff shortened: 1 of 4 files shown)


aggregator/src/metrics.rs
```diff
@@ -341,3 +341,27 @@ pub(crate) fn aggregate_step_failure_counter(meter: &Meter) -> Counter<u64> {
 
     aggregate_step_failure_counter
 }
+
+const TIME_SKEW_HISTOGRAM_VALUES: &[f64] = &[
+    30.0, 60.0, 120.0, 300.0, 600.0, 1800.0, 3600.0, 7200.0, 14400.0, 28800.0, 86400.0,
```
divergentdave, line 346: Here's a plot of these boundaries on a logarithmic scale:

image

I like that all the ratios between successive boundaries are 2, 2.5, or 3.

tgeoghegan (approved):

tgeoghegan:
This needs a backport to release/0.7 so we can deploy it, right?

PR #3859, opened 2025-06-04
Update wire version and Readme to DAP-14

jcjones:
Taking from the [DAP 14 changelog](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#name-change-log):

- [x] Enforce VDAF aggregation parameter validity. This is not relevant for Prio3, which requires only that reports be aggregated at most once. It is relevant for VDAFs for which validity depends on how many times a report might be aggregated (at most once in DAP). (\*)

  PR: #3768, but we only use Prio3

- [x] Require all timestamps to be truncated by `time_precision`. (\*)

  PRs: #3837, #3836, #3820, #3796, #3818, #3792, #3847

- [x] Clarify conditions for rejecting reports based on the report metadata, including the timestamp and public and private extensions.

  PR: #3630

- [x] Clarify that the Helper responds with 202 Accepted to an aggregation continuation request.

  PR: #3522

- [x] Bump version tag from "dap-13" to "dap-14".

  This PR: #3859

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,6 @@ Note that not every DAP draft has been implemented.
 [dap-04]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/04/
 [dap-07]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/07/
 [dap-09]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/09/
-[dap-13]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/13/
 [dap-14]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/14/
 [dap-gh]: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap
 


README.md
```diff
@@ -44,7 +44,7 @@ branch.
 | `release/0.5` | [`draft-ietf-ppm-dap-04`][dap-04] | Yes | Unmaintained as of June 24, 2024 |
 | `release/0.6` | [`draft-ietf-ppm-dap-07`][dap-07] | Yes, [with errata](#draft-ietf-ppm-dap-07-errata) | Unmaintained as of June 24, 2024 |
 | `release/0.7` | [`draft-ietf-ppm-dap-09`][dap-09] | Yes, [with errata](#draft-ietf-ppm-dap-09-errata) | Supported |
-| `main` | [`draft-ietf-ppm-dap-13`][dap-13] | Not yet (under development) | Supported |
+| `main` | [`draft-ietf-ppm-dap-14`][dap-14] | Not yet (under development) | Supported |
```
tgeoghegan, line 47: If we're not aware of any remaining discrepancies from the draft, then we should just say "Yes". If we find any bugs, we'll either fix them or document errata.
```suggestion
| `main` | [`draft-ietf-ppm-dap-14`][dap-14] | Yes | Supported |
```

README.md
```diff
@@ -55,6 +55,7 @@ Note that not every DAP draft has been implemented.
 [dap-07]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/07/
 [dap-09]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/09/
 [dap-13]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/13/
```
tgeoghegan, line 57: Is this used anywhere anymore?

jcjones:
```suggestion
```

Bye

Issue #3857, opened 2025-06-03
Streaming processing of collection jobs


divergentdave:
Currently, when we step a collection job, we read in a vector of all relevant batch aggregations, construct empty batch aggregations for any rows that don't exist yet, and update or put each batch aggregation into the database. We then exit the first transaction, compute an aggregate share from the batch aggregations, and send an aggregate share request to the helper. We then enter the second transaction, where we check the collection job lease, write the collection job back, and scrub each batch aggregation.

This process is creating memory pressure in some situations. Since we don't currently have upper limits on collection job sizes/durations or a lower limit on time precision, both the number of batch aggregations we read and the number of empty batch aggregations we write can be large. As a first step, we may want to limit the parallelism of the various `try_join_all()` calls that loop over the batch aggregations. Constructing all the `put_batch_aggregation()` and `update_batch_aggregation()` futures at the same time probably adds significantly more overhead than the batch aggregation data structure itself. They'll all pipeline their database queries, and wait to get responses back over oneshot channels. By using some different future combinators, we could limit the number of futures in flight at the same time.

Beyond the futures, we could consider eliminating the `Vec<BatchAggregation>` vectors in various places with further use of streaming processing. We could chunk up the collection batch interval into smaller segments, make windowed requests for existing batch aggregations, and create empty batch aggregations one window at a time. This would add some round trips to the transaction for very large collection jobs. We would also have to move the `compute_aggregate_share()` call up into the `step_collection_job_1` transaction, and refactor it to allow accumulating batch aggregations in multiple steps.

I'm currently working on a test harness to run the collection job driver inside DHAT. This should help identify what is contributing to peak memory usage, and help us validate fixes.

FWIW the helper similarly constructs a big vector of batch aggregations, and writes empty batch aggregations, when processing aggregate share requests. We will need to make similar changes there. Note that the helper does all its work in a single transaction already, which simplifies things somewhat.

PR #3863, opened 2025-06-04
`release/0.7`: Stream `BatchAggregation` rows to DB in groups

tgeoghegan:
If time precision is small, or the task's time interval is large, then a large number of `BatchAggregation`s may be needed to satisfy a collection job, whether that's existing ones that contain some prepared report shares or empty ones that we write to force serialization conflicts with other database transactions. Doing `try_join_all` on hundreds of thousands of futures all at once causes a huge number of allocations, which can cause significant memory pressure due to (1) allocating all the objects that the futures need and (2) the fact that many of these futures will race to populate the prepare SQL statement cache, allocating rather large strings along the way.

A first step toward alleviating that is to use
`StreamExt::buffer_unordered` so that we only handle a limited number of futures at once. We do this in both of the transactions in `step_collection_job_generic` which put or update `BatchAggregation`s.

Part of #3857


aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -76,6 +81,7 @@ where
             metrics: CollectionJobDriverMetrics::new(meter),
             batch_aggregation_shard_count,
             collection_retry_strategy,
+            max_future_concurrency: 1000,
```
tgeoghegan, line 84: I will wire this up in the configuration file once we agree that this is the right way to go.

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -302,18 +309,31 @@ where
...
+                    // Rather than awaiting all the futures at once, which can cause heap growth
+                    // proportional to the number of batch aggregations to write, put them into an
+                    // unordered, buffered stream so that they get run in groups of bounded size.
+                    // #3857
+                    {
+                        // This scope is needed so the borrows of batch_aggregations and
+                        // empty_batch_aggregations get dropped
+                        let mut futures = Vec::with_capacity(
+                            batch_aggregations.len() + empty_batch_aggregations.len(),
+                        );
```
tgeoghegan, line 321: I find it unfortunate to have to allocate a `Vec` up front here. I'd rather do:

```rust
let futures = batch_aggregations
    .iter()
    .map(|ba| tx.update_batch_aggregation(ba).boxed())
    .chain(
        empty_batch_aggregations
            .iter()
            .map(|ba| tx.put_batch_aggregation(ba).boxed()),
    );
let mut stream =
    futures::stream::iter(futures).buffer_unordered(max_future_concurrency);
while let Some(r) = stream.next().await {
    r?;
}
```

But that gets me an error message about having the wrong kind of `FnOnce` and needing to implement `Iterator<_, ...>` instead of `Iterator<'0, ...>` and I'm not sure how to address that. I think it's to do with the lifetime of the `BatchAggregation`s.

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -302,18 +309,31 @@ where
...
+                            futures.push(tx.put_batch_aggregation(ba).boxed());
+                        }
+
+                        // Drain the stream. This will short circuit if any of the futures return
+                        // an error.
+                        let mut stream =
+                            futures::stream::iter(futures).buffer_unordered(max_future_concurrency);
+                        while let Some(r) = stream.next().await {
+                            r?;
+                        }
```
tgeoghegan, line 335: This loop looks like we're running the futures one at a time. But from [the docs](https://docs.rs/futures/0.3.31/futures/stream/trait.StreamExt.html#method.buffer_unordered):

> If this stream’s item can be converted into a future, then this adaptor will buffer up to n futures and then return the outputs in the order in which they complete.

So my understanding is that each group of `max_future_concurrency` futures will get `await`ed for us with the desired concurrency by `buffer_unordered` and this loop just serializes checking for errors.

tgeoghegan:
I also still need to make the equivalent fix on the helper side, but wanted to get review going in case there's a better way to hold a `Stream` someone wants to point out to me.

aggregator/src/aggregator/collection_job_driver.rs
divergentdave: Yeah, I'm concerned that pushing all the futures onto the vector up front will leave the future part of the peak memory consumption as-is, and only reduce the peak via the statement preparation and tracing span allocations. Since the futures themselves act like enums, we'll be reserving space for the heaviest await points, even though the `Unresumed` variant may be smaller.

tgeoghegan:
I've attached some before-and-after DHAT profiles of the collection job driver servicing collection requests, labeled by the commit they were built from. They're not perfect comparisons, because the collection jobs they serviced weren't of exactly equal size, but nonetheless we see that:

- The peak memory footprint (`t-gmax`) declines from 1,128,887,464 bytes to 501,280,698 bytes
- We used to have about 40% of `t-gmax` under `try_join_all` in the second `step_collection_job` transaction and another 45% in the other txn.
- Now instead we have 89-90% of allocations under `FutureExt::boxed`. 

I used `FutureExt::boxed` as a way to get the compiler to stop complaining about me about two different `impl Future`s or two different anonymous async blocks having a different type. But the boxing seems rather expensive, and sorta undermines what we're trying to do by streaming the futures in groups. So I think I need to try again, without `boxed()`.

[WITH this change, collecting 36329 reports: dhat.out.847f1e7f220e35d1a8ead1de3b26d5f967ee1a29.json](https://github.com/user-attachments/files/20599173/dhat.out.847f1e7f220e35d1a8ead1de3b26d5f967ee1a29.json)
[BEFORE this change, collecting 13839 reports: dhat.out.4c60491f44555d4d4314ff9486a029125fa6126a.json](https://github.com/user-attachments/files/20599175/dhat.out.4c60491f44555d4d4314ff9486a029125fa6126a.json)

aggregator/src/aggregator/collection_job_driver.rs
tgeoghegan: I think this might work:

```rust
futures::stream::iter(                                  
    batch_aggregations                                  
        .iter()                                         
        .map(|ba| (true, ba))                           
        .chain(empty_batch_aggregations.iter().map(|ba| (false, ba))),
)                                                       
.map(Result::Ok)                                        
.try_for_each_concurrent(max_future_concurrency, async |(is_update, ba)| {
    if is_update {                                      
        tx.update_batch_aggregation(ba).await           
    } else {                                            
        tx.put_batch_aggregation(ba).await              
    }                                                   
})                                                      
.await?;
```

This way, there's a single closure with a single type so you can create a stream from it. But I'm concerned there's still some sneaky copy in there so I have to try it out under DHAT, which is taking a while.

tgeoghegan:
This looks a lot better to me at 139c46a50f558c837e019a0fe68dc9a79d4149be. Attached a fresh DHAT analysis: the peak working set is now down to less than 70 MB and we don't get nearly so many allocations at t-gmax preparing queries or copying futures around.

[dhat.out.139c46a50f558c837e019a0fe68dc9a79d4149be.json](https://github.com/user-attachments/files/20601140/dhat.out.139c46a50f558c837e019a0fe68dc9a79d4149be.json)

aggregator/src/aggregator.rs
```diff
@@ -3348,15 +3362,31 @@ impl VdafOps {
...
+                    futures::stream::iter(
+                        batch_aggregations
+                            .into_iter()
+                            // existing batch aggregatons must be updated so zip with true
+                            .zip(repeat(true))
+                            // empty batch aggregations are new so zip with false
+                            .chain(empty_batch_aggregations.into_iter().zip(repeat(false)))
+                            .map(|(ba, is_update)| (ba.scrubbed(), is_update)),
+                    )
+                    .map(Result::Ok)
```
divergentdave, line 3380: We could push this `Ok(...)` into the `map()` above. I suspect that iterators may optimize away better than streams.

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -302,18 +310,26 @@ where
...
+                    // buffered stream so that they get run in groups of bounded size. #3857
+                    futures::stream::iter(
+                        batch_aggregations
+                            .iter()
+                            // existing batch aggregatons must be updated so zip with true
+                            .zip(repeat(true))
+                            // empty batch aggregations are new so zip with false
+                            .chain(empty_batch_aggregations.iter().zip(repeat(false))),
+                    )
+                    .map(Result::Ok)
```
divergentdave, line 324: Similarly, we could replace this stream combinator with an iterator combinator
```suggestion
                            .chain(empty_batch_aggregations.iter().zip(repeat(false)))
                            .map(Result::Ok),
                    )
```

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -426,13 +443,18 @@ where
...
-                                    tx.update_batch_aggregation(ba).await
-                                })),
-                                tx.release_collection_job(&lease, None),
-                            )?;
+                            tx.update_collection_job::<SEED_SIZE, Q, A>(&collection_job).await?;
+
+                            // Put all the futures into a buffered stream so that we don't incur the
+                            // combined memory footprint of all the futures at once. #3857
+                            futures::stream::iter(batch_aggregations.iter())
+                                .map(Result::Ok)
```
divergentdave, line 451: Same here
```suggestion
                            futures::stream::iter(
                                batch_aggregations
                                    .iter()
                                    .map(Result::Ok),
                            )
```

divergentdave (approved):

aggregator/src/binaries/aggregator.rs
```diff
@@ -421,6 +426,10 @@ fn default_task_counter_shard_count() -> u64 {
     32
 }
 
+fn default_max_future_concurrency() -> usize {
+    10000
```
divergentdave, line 430: FYI this is different from the other defaults

PR #3872, opened 2025-06-06
Store ping pong transition when polling agg jobs

tgeoghegan:
`prio::vdaf::topology::ping_pong::Transition` is designed so that users can serialize it instead of `PingPongState`, leading to substantial storage savings in some scenarios. This is why we have

```rust
ReportAggregationState::LeaderContinue {
    transition: PingPongTransition,
}
```

When we did the async aggregation job stuff, we introduced `ReportAggregationState::LeaderPoll { state: PingPongState }`, but this makes it impossible to take advantage of `PingPongTransition`. This commit splits `LeaderPoll` into `LeaderPollInit` and `LeaderPollContinue`, which correspond to `LeaderInit` and `LeaderContinue`. This allows us to enjoy the same storage wins as the synchronous case does.

I'm working toward landing #3826 on main, and this is a somewhat complex part of that I've broken out to make it more digestible.

[sizes]: https://github.com/divviup/libprio-rs/pull/683/#issuecomment-1687210371

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_driver.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver.rs
@@ -868,15 +868,21 @@ where
         aggregation_job: AggregationJob<SEED_SIZE, B, A>,
         report_aggregations: Vec<ReportAggregation<SEED_SIZE, A>>,
     ) -> Result<(), Error> {
+        // The leader previously initiated or continued this job, and the helper deferred processing
+        // it. The leader is now are polling the helper to check if it has stepped the job.
         // Only process non-failed report aggregations; convert non-failed report aggregations into
         // stepped aggregations to be compatible with `process_response_from_helper`.
         let stepped_aggregations: Vec<_> = report_aggregations
             .into_iter()
             .filter_map(|report_aggregation| {
                 let leader_state = match report_aggregation.state() {
+                    // Leader was in the init state, so re-hydrate the prepare state into
+                    // PingPongState::Continued.
                     ReportAggregationState::LeaderPollInit { prepare_state } => {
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -76,7 +76,7 @@ postgres-types = "0.2.9"
 pretty_assertions = "1.4.1"
 # Disable default features so that individual workspace crates can choose to
 # re-enable them
-prio = { version = "0.17.0", default-features = false, features = ["experimental"] }
+prio = { version = "0.18.0-alpha.0", default-features = false, features = ["experimental"], git = "https://github.com/divviup/libprio-rs.git", rev = "663390d2d85c4bc173af9bc8fa29ee109faff789" }
```
tgeoghegan, line 79: We'll want to release a `prio 0.18.0-alpha.0` and fix this before merging.

jcjones (commented):
I can say now I've gone through all this and understand not only this change, but... well, I don't want to overstate it, but I better understand now what's happening with the ping-pong VDAF pattern.

I am reasonably confident now I'll be able to make actual review comments on this tomorrow.

tgeoghegan:
I'm glad it's been helpful in that regard, @jcjones. Your perspective here is particularly valuable: please call out places where we could better describe state transitions in comments.

jcjones (approved):
The state machine updates took me a while to properly grok, but the requirement for state coverage in `match` is a beautiful thing. Still, I made sure that wildcards still made sense after the adjustments.

I would like to see a sentence in all the more complex state matches that rather than describe the matcher, says _now we are waiting on the helper to finish processing_ -kind of remark.

..but approved as-is. Do as you think best.

PR #3890, opened 2025-06-18
ci-build.yml: make it clear which tests failed

tgeoghegan:
When you see that step "test status" failed, you have to go check which of "test" or "test-docker" actually failed, which can be annoying because the output from each of those steps is looooooong. With two different test status steps, it's obvious which one failed.


divergentdave (approved):
This looks good, but we could alternately solve this problem by getting rid of `continue-on-error: true` and the test status steps, then making "Upload container logs" run on failure by overriding the default condition of `if: success()` with `if: steps.test.outcome == 'failure' || steps.test-docker.outcome == 'failure'`

tgeoghegan:
I like your solution better. I'll need to thrash around here and try a few things to make sure it works as desired, though.

tgeoghegan:
This works as we want if when either the `test` (https://github.com/divviup/janus/actions/runs/15745157975/job/44380164009) or `test-docker` (https://github.com/divviup/janus/actions/runs/15745349606/job/44380733889) step fails: the upload container logs step runs.

If both actions fail, we skip upload container logs: https://github.com/divviup/janus/actions/runs/15745524031/job/44381091896?pr=3890

And if some other step fails, we skip upload container logs: https://github.com/divviup/janus/actions/runs/15745645526/job/44381430135?pr=3890

One gotcha is that you have to check `failure()` before you check `steps.*.outcome` 🤷🏻‍♂️ 
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#failure-with-conditions

PR #3892, opened 2025-06-23
Lazily instantiate empty batch aggregations

tgeoghegan:
When handing a collection job or aggregate share, we want to write lots of batch aggregations rows. Often, the vast majority of these will be empty batch aggregations that exist only to force conflicts with other SQL transactions. Rather than allocating all these empty BAs at once, we now create a `BatchAggregationsIterator` which lazily constructs them. In order to decide whether an empty BA is needed for a given `(batch_identifier, ord)` tuple, the `BatchAggregationsIterator` needs to know what real batch aggregations exist, so it takes ownership of the `Vec<BatchAggregation<_>>` returned by
`CollectableBatchMode::get_batch_aggregations_for_collection_identifier` and yields a unified stream of both real and empty/synthetic BAs.

In order to avoid traversing this iterator multiple times (which would require cloning the potentially large vector of real BAs), this also introduces `AggregateShareComputer`, which expands the previous `compute_aggregate_share` oneshot function into a structure implementing the initialize/update/finalize pattern familiar from hashers. Now, we can incrementally compute aggregate shares while consuming the BA iterator.

Part of #3857

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -3304,8 +3304,7 @@ impl VdafOps {
                         batch_aggregations,
                     );
 
-                    let aggregate_share =
-                        Arc::new(Mutex::new(AggregateShareComputer::new(&task)));
+                    let aggregate_share = Arc::new(Mutex::new(AggregateShareComputer::new(&task)));
 
                     // Rather than awaiting all the futures at once, which can cause heap growth
                     // proportional to the number of batch aggregations to write, put them into an
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -3,7 +3,7 @@
 pub use crate::aggregator::error::Error;
 use crate::{
     aggregator::{
-        aggregate_share::compute_aggregate_share,
+        aggregate_share::{AggregateShareComputer, BatchAggregationsIterator},
         aggregation_job_init::compute_helper_aggregate_init,
         aggregation_job_writer::{
             AggregationJobWriter, AggregationJobWriterMetrics, InitialWrite,
@@ -43,15 +43,14 @@ use fixed::{
 };
 use futures::{future::try_join_all, stream::TryStreamExt};
 use http::{header::CONTENT_TYPE, Method};
-use itertools::iproduct;
 use janus_aggregator_core::{
     batch_mode::AccumulableBatchMode,
     datastore::{
         self,
         models::{
-            AggregateShareJob, AggregationJob, AggregationJobState, BatchAggregation,
-            BatchAggregationState, CollectionJob, CollectionJobState, LeaderStoredReport,
-            ReportAggregation, ReportAggregationState, TaskAggregationCounter,
+            AggregateShareJob, AggregationJob, AggregationJobState, CollectionJob,
+            CollectionJobState, LeaderStoredReport, ReportAggregation, ReportAggregationState,
+            TaskAggregationCounter,
         },
         Datastore, Error as DatastoreError, Transaction,
     },
@@ -79,7 +78,7 @@ use janus_messages::{
     AggregationJobId, AggregationJobInitializeReq, AggregationJobResp, AggregationJobStep,
     BatchSelector, CollectionJobId, CollectionJobReq, CollectionJobResp, Duration, HpkeConfig,
     HpkeConfigList, InputShareAad, Interval, PartialBatchSelector, PlaintextInputShare,
-    PrepareResp, Report, ReportError, ReportIdChecksum, Role, TaskId,
+    PrepareResp, Report, ReportError, Role, TaskId,
 };
 use opentelemetry::{
     metrics::{Counter, Histogram, Meter},
@@ -102,13 +101,13 @@ use std::{
     borrow::Cow,
     collections::HashSet,
     fmt::Debug,
-    iter::repeat,
+    iter::repeat_with,
     panic,
     path::PathBuf,
     sync::{Arc, Mutex as SyncMutex},
     time::{Duration as StdDuration, Instant},
 };
-use tokio::try_join;
+use tokio::{sync::Mutex, try_join};
 use tracing::{debug, error, info, warn, Level};
 use url::Url;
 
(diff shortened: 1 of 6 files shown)


aggregator/src/aggregator/aggregate_share.rs
```diff
@@ -74,46 +101,146 @@ pub(crate) async fn compute_aggregate_share<
...
+        let key = self.expected_batch_aggregations.next()?;
+
+        // If we have a real BA for that key, yield it, removing it from the HashMap. This lets us
+        // return a value without cloning, and since this iterator gets consumed, the value can't be
+        // yielded again anyway.
+        self.real_batch_aggregations
+            .remove(&key)
+            .map(|real_ba| (real_ba, true))
+            .or_else(|| {
+                // If there was no real BA, synthesize an empty one. This is why we have to yield a
```
divergentdave, line 224: nit: clarity
```suggestion
                // If there was no real BA, synthesize an empty one. This is why we have to yield an owned
```

aggregator/src/aggregator.rs
```diff
@@ -3297,24 +3296,54 @@ impl VdafOps {
...
+                            .map(Result::Ok),
+                    )
+                    .try_for_each_concurrent(
+                        max_future_concurrency,
+                        async |((ba, is_update), aggregate_share)| {
+                            aggregate_share
+                                .lock()
+                                .await
+                                .update(&ba)
+                                .map_err(|e| datastore::Error::User(e.into()))?;
```
divergentdave, line 3325: As in the collection job driver, we could move this to the `is_update` arm (after inlining `ba = ba.scrubbed()` below).

Moreover, we could do the entire aggregate share computation in a separate pass over the `batch_aggregations` vector, before we pass it into the new iterator. This would avoid having to introduce a mutex at all, and would be faster both due to not needing synchronization, and due to better icache locality. (I think the new `AggregateShareComputer` abstraction may still be useful, if we later split up fetching batch aggregations over multiple database queries)

Note that while the existing `compute_aggregate_share()` was an async function, it never awaited in its existing state.

aggregator/src/aggregator.rs
```diff
@@ -3297,24 +3296,54 @@ impl VdafOps {
...
-                            .await
-                            .map_err(|e| datastore::Error::User(e.into()))?;
+                    let aggregate_share = Arc::new(Mutex::new(AggregateShareComputer::new(&task)));
+
+                    // Rather than awaiting all the futures at once, which can cause heap growth
+                    // proportional to the number of batch aggregations to write, put them into an
+                    // buffered stream so that they get run in groups of bounded size. #3857
+                    futures::stream::iter(
+                        batch_aggregations_iter
+                            .clone()
```
divergentdave, line 3314: This clone appears to be unnecessary. This could be expensive, since the iterator contains a `HashMap`.

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -268,17 +267,16 @@ where
...
+                    let batch_aggregations = B::get_batch_aggregations_for_collection_identifier(
+                        tx,
+                        lease.leased().task_id(),
+                        lease.leased().time_precision(),
+                        vdaf.as_ref(),
+                        &collection_identifier,
+                        &aggregation_param,
+                    )
+                    .await?
+                    .into_iter()
```
divergentdave, line 279: We could split this into two statements to avoid formatting strangeness from comments between chained method calls.
```suggestion
                    .await?;
```

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -288,68 +286,76 @@ where
                     // for any that have not already been written to storage. We do this
                     // transactionally to avoid the possibility of overwriting other transactions'
                     // updates to batch aggregations.
-                    batch_aggregations = batch_aggregations
-                        .into_iter()
-                        .map(|ba| ba.collected())
-                        .collect::<Result<Vec<_>, _>>()?;
+                    .map(|ba| ba.collected())
+                    .collect::<Result<Vec<_>, _>>()?
+                    .into_iter();
```
divergentdave, line 291: cont'd
```suggestion
                    let batch_aggregations = batch_aggregations
                        .into_iter()
                        .map(|ba| ba.collected())
                        .collect::<Result<Vec<_>, _>>()?;
```

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -390,28 +396,22 @@ where
...
-                .collect::<Vec<_>>(),
-        );
 
         datastore
             .run_tx("step_collection_job_2", |tx| {
                 let vdaf = Arc::clone(&vdaf);
                 let lease = Arc::clone(&lease);
                 let collection_job = Arc::clone(&collection_job);
-                let batch_aggregations = Arc::clone(&batch_aggregations);
+                let batch_aggregations = all_bas_iter.clone();
```
divergentdave, line 414: I think we should wrap this in an `Arc` before passing it into this transaction, to avoid additional expensive copies.

divergentdave: Whoops, I take it back! The interior mutability of a mutex is a problem because the transaction may need a retry, and thus the second attempt would see an empty iterator.

One other idea I had was to scrub the batch aggregations without having any corresponding list of `BatchAggregation`s in memory, either constructing them on the fly from the same Cartesian product, or just writing a separate UPDATE statement to NULL out the appropriate columns in all applicable rows at once.

tgeoghegan: I think we can make this work while avoiding the cost of copying the BA values around by using `std::borrow::Cow<BatchAggregation<_>>` for the real BAs, and reconstructing the empty ones on the fly. That way we can work with references to the real BAs but still have a single type for the `BatchAggregationsIterator` to yield. The other thing I had to do was re-create the `BatchAggregationsIterator` in either `step_collection_job` closure.

aggregator/src/aggregator/aggregate_share.rs
```diff
@@ -74,46 +120,170 @@ pub(crate) async fn compute_aggregate_share<
...
+        // If we have a real BA for that key, yield it, removing it from the HashMap. This lets us
+        // return a value without cloning, and since this iterator gets consumed, the value can't be
+        // yielded again anyway.
+        self.real_batch_aggregations
+            .remove(&key)
+            .map(|real_ba| (real_ba, true))
+            .or_else(|| {
+                // If there was no real BA, synthesize an empty one. This is why we have to yield an
+                // owned value in the other case: we can't instantiate a BA here and return a
+                // reference.
```
divergentdave, line 269: nit: This comment should be updated/trimmed down to reflect that we're now using a Cow.

aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -436,13 +436,22 @@ where
...
+                            // Scrub all the batch aggregations, including the empty ones.
                             // Put all the futures into a buffered stream so that we don't incur the
                             // combined memory footprint of all the futures at once. #3857
-                            futures::stream::iter(batch_aggregations.iter().map(Result::Ok))
-                                .try_for_each_concurrent(max_future_concurrency, |ba| {
-                                    tx.update_batch_aggregation(ba)
-                                })
-                                .await?;
+                            futures::stream::iter(batch_aggregations_iter.map(|(v, _)| {
+                                Result::Ok(BatchAggregation::scrubbed(v.into_owned()))
```
divergentdave, line 451: nit: more idiomatic
```suggestion
                                Ok(BatchAggregation::scrubbed(v.into_owned()))
```

divergentdave (approved):

PR #3901, opened 2025-06-26
Adopt Rust 2024 edition and MSRV 1.85

tgeoghegan:
This PR is rather noisy because `cargo fmt`'s opinions about how to order `use` statements and some other syntax points have changed: whether to use semicolons in some places and I think formatting inside macros? I don't love how much history is tainted by this change, but we'll have to go to the 2024 edition someday, so we may as well bite the bullet. To highlight some non-perfunctory changes:

- clippy: replace a few cases of `Iterator.repeat().take()` with `Iterator.repeat_n`
- Dockerfiles: move to Rust 1.88 and update `cargo-chef` to 1.71. We were on 1.60, which for one reason or another doesn't recognize the 2024 edition.
- I had to use the new `use<..>` syntax in a couple places where the compiler directed me to ([RPIT lifetime capture rules](https://doc.rust-lang.org/edition-guide/rust-2024/rpit-lifetime-capture.html))
- aggregator_api: changed the function signature of `auth_check` to return `Option<..>` instead of `impl Handler` due to changes to [RPIT lifetime capture rules](https://doc.rust-lang.org/edition-guide/rust-2024/rpit-lifetime-capture.html)
- `gen` keyword is now reserved, so we have to escape a call to a method of that name as `r#gen`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,11 +15,11 @@ members = [
 resolver = "2"
 
 [workspace.package]
-edition = "2021"
+edition = "2024"
 homepage = "https://divviup.org"
 license = "MPL-2.0"
 repository = "https://github.com/divviup/janus"
-rust-version = "1.81.0"
+rust-version = "1.85.0"
 version = "0.8.0-prerelease-1"
 
 [workspace.dependencies]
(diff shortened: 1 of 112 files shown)


tgeoghegan:
I took a look at what we might get to clean up if we use [let chains](https://doc.rust-lang.org/edition-guide/rust-2024/let-chains.html) and frankly it's fairly modest. I was hoping it'd help move some of the trickier code in aggregation and collection job handling to the left, but it doesn't. I'm not sure if it's worth all the churn introduced by this PR.

PR #3902, opened 2025-06-27
use let chaining where possible

tgeoghegan:
# Stacked on #3901 

Rust 2024 allows [`let` chains in `if` and `while`][1] which lets us remove some nested `if` statements. Nothing as dramatic as I'd hoped, but some nice cleanups nonetheless.

[1]: https://doc.rust-lang.org/edition-guide/rust-2024/let-chains.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,11 +15,11 @@ members = [
 resolver = "2"
 
 [workspace.package]
-edition = "2021"
+edition = "2024"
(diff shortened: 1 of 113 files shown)


tgeoghegan:
Ah, I guess we can't do this until Rust stable minus two reaches 1.88. And I guess we'd want to set MSRV to at least 1.88, too.

Issue #3159, opened 2024-05-21
`janus_client`: Cache HPKE configs based on `cache-control` header


tgeoghegan:
Unless configs are provided to the builder, `janus_client` will fetch leader and helper HPKE configurations during construction and use them indefinitely, ignoring whatever `cache-control` HTTP header the aggregator may have provided. This is fine, since `janus_client` is generally used ephemerally and so whatever keys it sees will almost certainly remain valid for the lifetime in memory of the structure. But it'd be nice if we supported caching keys and automatically refetching them when they expire.

This is only particularly useful if keys are being stored somewhere besides memory, so we would want to introduce some kind of storage/cache abstraction for `janus_client` to use.

tgeoghegan:
If we do this for DAP HPKE configs, we should also do it for OHTTP keys.

PR #3903, opened 2025-06-30
`janus_client`: cache HPKE configurations

tgeoghegan:
Adds a `janus_core::http::cached_resource` module which GETs an HTTP resource and then caches it, honoring the `no-cache` and `max-age` `cache-control` directives. We then use this in `janus_client` to cache HPKE configurations.

Part of #3159

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -49,7 +49,7 @@ use http::{header::CONTENT_TYPE, StatusCode};
 use itertools::Itertools;
 use janus_core::{
     hpke::{self, is_hpke_config_supported, HpkeApplicationInfo, Label},
-    http::HttpErrorResponse,
+    http::{cached_resource::CachedResource, HttpErrorResponse},
     retries::{http_request_exponential_backoff, retry_http_request},
     time::{Clock, RealClock, TimeExt},
     url_ensure_trailing_slash,
@@ -60,14 +60,12 @@ use janus_messages::{
 };
 #[cfg(feature = "ohttp")]
 use ohttp::{ClientRequest, KeyConfig};
-use prio::{
-    codec::{Decode, Encode},
-    vdaf,
-};
+use prio::{codec::Encode, vdaf};
 use rand::random;
 #[cfg(feature = "ohttp")]
 use std::io::Cursor;
-use std::{convert::Infallible, fmt::Debug, time::SystemTimeError};
+use std::{convert::Infallible, fmt::Debug, sync::Arc, time::SystemTimeError};
+use tokio::{sync::Mutex, try_join};
 use url::Url;
 
 #[cfg(test)]
@@ -89,6 +87,8 @@ pub enum Error {
     Vdaf(#[from] prio::vdaf::VdafError),
     #[error("HPKE error: {0}")]
     Hpke(#[from] janus_core::hpke::Error),
+    #[error("Cached resource error: {0}")]
+    CachedResource(#[from] janus_core::http::cached_resource::Error),
     #[error("unexpected server response {0}")]
     UnexpectedServerResponse(&'static str),
     #[error("time conversion error: {0}")]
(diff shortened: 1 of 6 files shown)


client/src/lib.rs
```diff
@@ -494,15 +452,18 @@ impl<V: vdaf::Client<16>> ClientBuilder<V> {
 }
 
 /// A DAP client.
+pub type Client<V> = ClientWithClock<RealClock, V>;
```
tgeoghegan, line 455: Since this is on `release/0.7`, I'm trying to avoid breaking the exported API. Other crates in the workspace that pull in `janus_client` still build.

core/src/http/cached_resource.rs
```diff
@@ -0,0 +1,188 @@
...
+
+/// Caches an HTTP resource based on the cache-control header provided by the server.
+#[derive(Debug, Clone)]
+pub struct Refresher<C, Resource> {
+    clock: C,
+    resource: Resource,
+    expires_at: Option<Time>,
+    http_client: reqwest::Client,
+    resource_url: Url,
+    http_request_retry_parameters: ExponentialBackoff<SystemClock>,
```
tgeoghegan, line 91: These fields could be references, but then the struct would need a generic lifetime parameter, and that would propagate up to `janus_client::Client`, and that would break semver. When I forward-port this to `main`, I'll borrow these fields.

tgeoghegan:
See https://github.com/divviup/sre-tools/pull/1109 for the other side of this in the load generator. I think this needs a tweak so we don't impose `Send + Sync` on `InputShare` and `PublicShare`, but I'm not sure what's going on there.

core/src/http/cached_resource_tests.rs
```diff
@@ -0,0 +1,162 @@
...
+    .unwrap_err();
+
+    mock.assert_async().await;
+}
+
+#[tokio::test]
+async fn static_resource() {
+    install_test_trace_subscriber();
+
+    // When a static resouroce is provided, the refresher shouldn't make any network requests
```
jcjones, line 120:
```suggestion
    // When a static resource is provided, the refresher shouldn't make any network requests
```

core/src/http/cached_resource.rs
```diff
@@ -0,0 +1,188 @@
...
+//! [1]: https://datatracker.ietf.org/doc/html/rfc9111#section-5.2
+
+use crate::{
+    http::HttpErrorResponse,
+    retries::retry_http_request,
+    time::{Clock, TimeExt},
+};
+use backoff::{exponential::ExponentialBackoff, SystemClock};
+use http::{header::CACHE_CONTROL, HeaderValue};
+use janus_messages::{Duration, Time};
```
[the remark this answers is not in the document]
tgeoghegan: Not only does this work, it's *better* because now we don't need to pass a `Clock` around or add a generic parameter to various types. Nice!

client/src/lib.rs
```diff
@@ -799,3 +771,68 @@ impl<V: vdaf::Client<16>> Client<V> {
...
+
+impl<C: Clock> HpkeConfiguration<C> {
+    pub(crate) async fn new(
+        hpke_configuration: Option<HpkeConfig>,
+        clock: C,
+        client_parameters: &ClientParameters,
+        aggregator_role: &Role,
+        http_client: &reqwest::Client,
+    ) -> Result<Self, Error> {
+        if let Some(hpke_configuration) = hpke_configuration {
```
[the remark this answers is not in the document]
jcjones: I like your update here!

jcjones (approved):
I'm ready to approve when tests pass. I like it.

client/src/tests/mod.rs
```diff
@@ -304,15 +272,14 @@ async fn unsupported_hpke_algorithms() {
         .with_status(200)
         .with_header(CONTENT_TYPE.as_str(), HpkeConfigList::MEDIA_TYPE)
         .with_body(encoded_hpke_config_list)
-        .expect(1)
+        .expect(2)
```
divergentdave, line 275: Fetching the key twice when there's no Cache-Control header is a bit unfortunate. (The first fetch happens during `new()`, then the second happens during `get()`) What do you think about providing a default TTL (on the order of one minute to one hour) when there's no `Cache-Control` header? We could alternately remove the fetch during `new()`, and change the type of `resource` to be `Option<Resource>`.

client/src/lib.rs
```diff
@@ -670,7 +625,12 @@ impl<V: vdaf::Client<16>> Client<V> {
         Error: From<<T as TryInto<Time>>::Error>,
     {
         let report = self
-            .prepare_report(measurement, &time.try_into()?)?
+            .prepare_report(
+                measurement,
+                &time.try_into()?,
+                self.leader_hpke_config.lock().await.get().await?,
+                self.helper_hpke_config.lock().await.get().await?,
```
divergentdave, line 632: We could `try_join` these two for good measure.

core/src/http/cached_resource.rs
```diff
@@ -0,0 +1,188 @@
...
+        http_request_retry_parameters: &ExponentialBackoff<SystemClock>,
+        resource_url: &Url,
+    ) -> Result<(Resource, Option<Time>), Error> {
+        let response = retry_http_request(http_request_retry_parameters.clone(), || async {
+            http_client.get(resource_url.clone()).send().await
+        })
+        .await?;
+        let status = response.status();
+        if !status.is_success() {
+            return Err(Error::Http(Box::new(HttpErrorResponse::from(status))));
```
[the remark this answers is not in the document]
divergentdave: Ah good point. This is actually taken care of in `retry_http_request()` already.

divergentdave (approved):

client/src/lib.rs
tgeoghegan: We can't acquire the locks within `try_join!` because the lock guard gets dropped, but we can parallelize the HTTP GET requests.

client/src/tests/mod.rs
tgeoghegan: I agree that two fetches are unfortunate, but a default TTL strikes me as surprising. I think defaulting to either always or never refreshing the resource is the more obvious behaviour. I looked to RFC 9111 for guidance and [it says](https://datatracker.ietf.org/doc/html/rfc9111#section-2):

> Although caching is an entirely OPTIONAL feature of HTTP, it can be assumed that reusing a cached response is desirable and that such reuse is the default behavior when no requirement or local configuration prevents it. Therefore, HTTP cache requirements are focused on preventing a cache from either storing a non-reusable response or reusing a stored response inappropriately, rather than mandating that caches always store and reuse particular responses.

So I think defaulting to using a resource forever if there's no cache-control header seems appropriate.

PR #3906, opened 2025-07-03
`janus_client`: cache HPKE and OHTTP keys

tgeoghegan:
Adds a `janus_core::http::cached_resource` module which GETs an HTTP resource and then caches it, honoring the `no-cache` and `max-age` `cache-control` directives. We then use this in `janus_client` to cache HPKE configurations and OHTTP keys.

This is a forward-port of #3903, which went into `release/0.7`. The major differences:

- use the `janus_messages::MediaType` trait to figure out the desired media type instead of passing it around as an argument
- cache OHTTP keys as well as HPKE keys
- introduce a `FromBytes` trait and use that in `CachedResource` instead of `prio::codec::Decode` to gracefully handle `Vec<ohttp::KeyConfig`>
- avoid cloning values where possible

Part of #3159

### DIFF
--- a/core/src/http/cached_resource_tests.rs
+++ b/core/src/http/cached_resource_tests.rs
@@ -1,0 +1,292 @@
+use crate::{
+    http::cached_resource::{CachedResource, expires_at},
+    initialize_rustls,
+    retries::test_util::test_http_request_exponential_backoff,
+    test_util::install_test_trace_subscriber,
+};
+use http::{
+    HeaderValue,
+    header::{CACHE_CONTROL, CONTENT_TYPE},
+};
+use janus_messages::MediaType;
+use prio::codec::{Decode, Encode};
+use std::time::Duration;
+use tokio::time::Instant;
+use url::Url;
+
+#[derive(Debug, Eq)]
+struct TestResource {}
+
+impl Encode for TestResource {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), prio::codec::CodecError> {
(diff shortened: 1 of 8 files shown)


messages/src/lib.rs
```diff
@@ -39,7 +39,7 @@ pub mod taskprov;
 mod tests;
 
 /// Messages which have an HTTP media type associated with them.
-pub trait MediaType: Encode {
+pub trait MediaType {
```
tgeoghegan, line 42: We don't actually need `MediaType` to depend on `Encode`, and in fact by removing this bound it's easier to use `EncodeKeyConfig` in `CachedResource`.

jcjones (approved):

core/src/http/cached_resource_tests.rs
```diff
@@ -0,0 +1,292 @@
...
+        .mock("GET", "/resource")
+        .with_status(200)
+        .with_body(TestResource {}.get_encoded().unwrap())
+        .with_header(CONTENT_TYPE.as_str(), TestResource::MEDIA_TYPE)
+        .expect(1)
+        .create_async()
+        .await;
+
+    // new() will fetch the resource from server. Because no cache-control is provided, resource()
+    // should refetch.
```
divergentdave, line 67: This comment doesn't reflect the `expect(1)` above
```suggestion
    // new() will fetch the resource from server. Because no cache-control is provided, resource()
    // should not refetch.
```

divergentdave (approved):

Issue #3879, opened 2025-06-11
DAP-15: rework structure of CollectionJobResp and AggregationJobResp


jcjones:
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/674 reworks the message structure of `AggregationJobResp` and `CollectionJobResp` to remove their enums

(See also https://github.com/divviup/janus/issues/3451)

jcjones:
Resolved in https://github.com/divviup/janus/pull/3919 and https://github.com/divviup/janus/pull/3927

PR #3919, opened 2025-07-16
Convert AggregationJobResp from enum to struct

jcjones:
This is one half of #3879.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -191,8 +191,8 @@ impl AggregatorMetrics {
 /// Represents the result of a Job Continue operation, which varies whether that was synchronous
 /// or async.
 enum AggregationJobContinueResult {
-    Sync { resp: AggregationJobResp },
-    Async { step: AggregationJobStep },
+    Sync(AggregationJobResp),
+    Async(AggregationJobStep),
 }
 
 /// Config represents a configuration for an Aggregator.
@@ -2363,17 +2363,16 @@ impl VdafOps {
 
                         let resp = match task.aggregation_mode() {
                             Some(AggregationMode::Synchronous) => {
-                                AggregationJobContinueResult::Sync { resp:  AggregationJobResp {
+                                AggregationJobContinueResult::Sync(AggregationJobResp {
                                     prepare_resps: report_aggregations
                                         .iter()
                                         .filter_map(ReportAggregation::last_prep_resp)
                                         .cloned()
                                         .collect(),
-                                }}
-                            }
-                            Some(AggregationMode::Asynchronous) => {
-                                AggregationJobContinueResult::Async { step: aggregation_job.step() }
-                            }
+                                })
+                            },
+                            Some(AggregationMode::Asynchronous) =>
+                                AggregationJobContinueResult::Async(aggregation_job.step()),
                             None => {
                                 return Err(datastore::Error::User(
                                     Error::Internal("task has no aggregation mode".into())
@@ -2557,9 +2556,7 @@ impl VdafOps {
         )
         .await?;
         Ok((
-            AggregationJobContinueResult::Sync {
-                resp: AggregationJobResp { prepare_resps },
-            },
+            AggregationJobContinueResult::Sync(AggregationJobResp { prepare_resps }),
             counters,
         ))
     }
@@ -2598,7 +2595,7 @@ impl VdafOps {
         )
         .await?;
 
-        Ok((AggregationJobContinueResult::Async { step }, counters))
+        Ok((AggregationJobContinueResult::Async(step), counters))
     }
 
     async fn handle_aggregate_continue_generic_write<
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -188,6 +188,13 @@ impl AggregatorMetrics {
     }
 }
 
+/// Represents the result of a Job Continue operation, which varies whether that was synchronous
+/// or async.
+enum AggregationJobContinueResult {
+    Sync { resp: AggregationJobResp },
+    Async { step: AggregationJobStep },
+}
+
 /// Config represents a configuration for an Aggregator.
 #[derive(Debug)]
 pub struct Config {
@@ -456,7 +463,7 @@ impl<C: Clock> Aggregator<C> {
         req_bytes: &[u8],
         auth_token: Option<AuthenticationToken>,
         taskprov_task_config: Option<&TaskConfig>,
-    ) -> Result<Option<AggregationJobResp>, Error> {
+    ) -> Result<AggregationJobContinueResult, Error> {
         let task_aggregator = self
             .task_aggregators
             .get(task_id)
@@ -1048,7 +1055,7 @@ impl<C: Clock> TaskAggregator<C> {
         aggregation_job_id: &AggregationJobId,
         req: AggregationJobContinueReq,
         request_hash: [u8; 32],
-    ) -> Result<Option<AggregationJobResp>, Error> {
+    ) -> Result<AggregationJobContinueResult, Error> {
         self.vdaf_ops
             .handle_aggregate_continue(
                 datastore,
@@ -1518,7 +1525,7 @@ impl VdafOps {
         aggregation_job_id: &AggregationJobId,
         req: AggregationJobContinueReq,
         request_hash: [u8; 32],
-    ) -> Result<Option<AggregationJobResp>, Error> {
+    ) -> Result<AggregationJobContinueResult, Error> {
         match task.batch_mode() {
             task::BatchMode::TimeInterval => {
                 vdaf_ops_dispatch!(self, (vdaf, VdafType, VERIFY_KEY_LENGTH) => {
@@ -2279,7 +2286,7 @@ impl VdafOps {
         aggregation_job_id: &AggregationJobId,
         req: AggregationJobContinueReq,
         request_hash: [u8; 32],
-    ) -> Result<Option<AggregationJobResp>, Error> {
+    ) -> Result<AggregationJobContinueResult, Error> {
         if req.step() == AggregationJobStep::from(0) {
             return Err(Error::InvalidMessage(
                 Some(*task.id()),
@@ -2356,16 +2363,16 @@ impl VdafOps {
 
                         let resp = match task.aggregation_mode() {
                             Some(AggregationMode::Synchronous) => {
-                                Some(AggregationJobResp {
+                                AggregationJobContinueResult::Sync { resp:  AggregationJobResp {
                                     prepare_resps: report_aggregations
                                         .iter()
                                         .filter_map(ReportAggregation::last_prep_resp)
                                         .cloned()
                                         .collect(),
-                                })
+                                }}
                             }
                             Some(AggregationMode::Asynchronous) => {
-                                None
+                                AggregationJobContinueResult::Async { step: aggregation_job.step() }
                             }
                             None => {
                                 return Err(datastore::Error::User(
@@ -2523,7 +2530,7 @@ impl VdafOps {
         mut report_aggregations_to_write: Vec<WritableReportAggregation<SEED_SIZE, A>>,
         aggregation_job: AggregationJob<SEED_SIZE, B, A>,
         report_aggregations: Vec<ReportAggregation<SEED_SIZE, A>>,
-    ) -> Result<(Option<AggregationJobResp>, TaskAggregationCounter), Error> {
+    ) -> Result<(AggregationJobContinueResult, TaskAggregationCounter), Error> {
         // Compute the next aggregation step.
         // TODO(#224): don't hold DB transaction open while computing VDAF updates?
         let aggregation_job = Arc::new(aggregation_job);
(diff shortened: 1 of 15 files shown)


aggregator/src/aggregator/http_handlers.rs
```diff
@@ -266,6 +266,34 @@ where
     }
 }
 
+/// A Trillium handler that returns an empty body.
+#[derive(Clone, Copy)]
+struct EmptyBody {
+    media_type: &'static str,
+    retry_after: &'static str, // TODO: Make this editable? TKTK
+    status: Status,            // TODO: Make this editable? TKTK
```
jcjones, line 274: I have two lines marked `TODO` in here that I'd like some feedback on: They are the `Status` and `Retry-After` fields returned with an empty body in replacement of the `Processing` enumeration.

Right now I'm hard-coding these to 2 seconds and 200 OK, so there's no reason for them to be called out here into the struct. 

As far as I can tell this is fine for the CollectionJobResp as well, so it seems to me that Status should be just hardcoded, problem solved.

But what about the Retry-After? Should we make that variable at all here, now, or should I simplify this struct because YAGNI?

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -641,18 +644,31 @@ where
...
-            let resp = AggregationJobResp::get_decoded(http_response.body())
-                .map_err(Error::MessageDecode)?;
+
+            let length = http_response
+                .headers()
+                .get(CONTENT_LENGTH)
+                .map(parse_content_length)
+                .transpose()?;
+
+            let resp = if length == Some(0) {
```
jcjones, line 654: The semantics here are a little wonky. This treats responses without a content length as finished, which I'm not happy with.

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -266,6 +266,34 @@ where
...
+            retry_after: "2",
+            status: Status::Ok,
+        }
+    }
+}
+
+#[async_trait]
+impl Handler for EmptyBody {
+    async fn run(&self, conn: Conn) -> Conn {
+        conn.with_response_header(KnownHeaderName::ContentType, self.media_type)
```
tgeoghegan, line 290: Content-Type for an empty body seems weird?

aggregator/src/aggregator/http_handlers.rs
tgeoghegan: I think we could just hardcode 200 and it'd be fine. Unless I made a mistake when writing the spec (totally possible!) then you shouldn't ever need a more specific HTTP status to make this unambiguous.

I don't think we're particularly clever about the retry-after. I'm not sure what we would use to estimate this, is the thing. I think a reasonable first try would be to use the interval at which the aggregation job driver polls for runnable aggregation jobs, and maybe throw 20-30% on top of that to account for the processing time.

aggregator/src/aggregator/http_handlers.rs
jcjones: It certainly does. I think I got my wires crossed in a Slack conversation earlier, as I had dropped it in the first pass, but then I thought I saw something... Anyway, yeah, I'll go strip that.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -1929,6 +1970,17 @@ fn parse_retry_after(header_value: &HeaderValue) -> Result<RetryAfter, Error> {
         .map_err(|err| Error::BadRequest(err.into()))
 }
 
+fn parse_content_length(header_value: &HeaderValue) -> Result<i32, Error> {
```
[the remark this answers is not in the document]
jcjones: As I'll need it in the collector too, I'll go into `janus_core` with it as `usize`.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -641,18 +644,31 @@ where
                 .get(RETRY_AFTER)
                 .map(parse_retry_after)
                 .transpose()?;
-            let resp = AggregationJobResp::get_decoded(http_response.body())
-                .map_err(Error::MessageDecode)?;
+
+            let length = http_response
```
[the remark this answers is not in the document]
jcjones: Looks like we never did, but we will now.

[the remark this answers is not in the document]
jcjones: Added `check_content_type` and using it now in these methods.

aggregator/src/aggregator/aggregation_job_continue.rs
```diff
@@ -236,12 +236,30 @@ pub mod test_util {
...
-        decode_response_body::<AggregationJobResp>(&mut test_conn).await
+        let length: Option<usize> = test_conn
+            .response_headers()
+            .get_str(KnownHeaderName::ContentLength)
+            .map(|s| s.parse())
+            .transpose()
+            .unwrap();
+
+        if length.is_some_and(|l| l > 0) {
+            assert_headers!(&test_conn, "content-type" => (AggregationJobResp::MEDIA_TYPE));
```
tgeoghegan, line 250: Why does the value on the right of `=>` have to be a tuple?

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -641,18 +645,34 @@ where
...
                 .map(parse_retry_after)
                 .transpose()?;
-            let resp = AggregationJobResp::get_decoded(http_response.body())
-                .map_err(Error::MessageDecode)?;
+            let length = http_response
+                .headers()
+                .get(CONTENT_LENGTH)
+                .map(parse_content_length)
+                .transpose()
+                .map_err(|e| Error::BadRequest(e.into()))?;
```
tgeoghegan, line 653: Is `Error::BadRequest` appropriate here, since this would in fact be an invalid response from the helper? I'm not certain what errors are available in this case, but if we don't have `BadResponse` already then something generic is probably fine.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -848,8 +868,23 @@ where
...
             .map(parse_retry_after)
             .transpose()?;
-        let resp =
-            AggregationJobResp::get_decoded(http_response.body()).map_err(Error::MessageDecode)?;
+        let length = http_response
+            .headers()
+            .get(CONTENT_LENGTH)
+            .map(parse_content_length)
+            .transpose()
+            .map_err(|e| Error::BadRequest(e.into()))?;
```
tgeoghegan, line 876: As above, `BadRequest` seems wrong for a badly constructed response.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -943,8 +978,23 @@ where
...
```
tgeoghegan, line 986: As above, BadRequest seems inappropriate for a malformed response.

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -266,6 +267,32 @@ where
...
+struct EmptyBody {
+    location: String,
+}
+
+impl EmptyBody {
+    /// Return an EmptyBody with the location set to the relative path to the
+    /// aggregation job for the task.
+    fn for_task_and_job(task_id: &TaskId, aggregation_job_id: &AggregationJobId) -> Self {
+        Self {
+            location: format!("/tasks/{task_id}/aggregation_jobs/{aggregation_job_id}?step=0"),
```
tgeoghegan, line 281: Hmm, is it right to hardcode step=0? If we were running a VDAF with more than one preparation step, then you'd eventually need to go to step=1. In practice, we only actually run Prio3 VDAFs, which finish at step 1 and thus will never have a query besides step=0, but I think Janus supports VDAFs with more steps than that and it'd be nice to maintain that. For instance, I think we have some tests that instantiate the dummy VDAF with multiple rounds so that we can exercise AggregationJobContinue.

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -266,6 +267,32 @@ where
...
+        }
+    }
+}
+
+#[async_trait]
+impl Handler for EmptyBody {
+    async fn run(&self, conn: Conn) -> Conn {
+        conn.with_response_header(KnownHeaderName::RetryAfter, "2")
```
tgeoghegan, line 289: I think that hard coding 2 seconds is fine if our goal right now is to get the message format changes in. But let's log an issue about computing a better value here, based on aggregation job driver poll interval?

tgeoghegan (approved):

aggregator/src/aggregator/http_handlers.rs
divergentdave: We could change this signature to `fn for_aggregation_job(task_id: &TaskId, aggregation_job_id: &AggregationJobId, step: u16) -> Self`. Note that we'll have to create a new enum for the return value of `handle_aggregate_continue()` to pass the aggregation job step out (instead of just `None`). The test expectations in `aggregation_job_continue.rs` will have to be updated as well.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -641,18 +645,34 @@ where
...
-            let resp = AggregationJobResp::get_decoded(http_response.body())
-                .map_err(Error::MessageDecode)?;
+            let length = http_response
+                .headers()
+                .get(CONTENT_LENGTH)
+                .map(parse_content_length)
+                .transpose()
+                .map_err(|e| Error::BadRequest(e.into()))?;
+
+            let resp = if length.is_some_and(|l| l > 0) {
```
divergentdave, line 655: This won't work if we are talking to a helper that streams its responses, and doesn't set `Content-Length`. We could instead call `http_response.body()` sooner, and check its length.

aggregator/src/aggregator/aggregation_job_driver.rs
jcjones: Well, the main options are:
- Error::Internal
- Error::HttpClient
- Error::Http

I suppose Internal isn't bad, in that it's probably what the resulting status code will be.

aggregator/src/aggregator/aggregation_job_driver.rs
jcjones: I've changed this to check the body.

aggregator/src/aggregator/http_handlers.rs
jcjones:
```suggestion
        // To be fixed in issue #3921
        conn.with_response_header(KnownHeaderName::RetryAfter, "2")
```

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -641,18 +642,26 @@ where
                 .get(RETRY_AFTER)
                 .map(parse_retry_after)
                 .transpose()?;
-            let resp = AggregationJobResp::get_decoded(http_response.body())
-                .map_err(Error::MessageDecode)?;
+
```
jcjones, line 645:
```suggestion
```

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -848,8 +857,15 @@ where
             .get(RETRY_AFTER)
             .map(parse_retry_after)
             .transpose()?;
-        let resp =
-            AggregationJobResp::get_decoded(http_response.body()).map_err(Error::MessageDecode)?;
+
```
jcjones, line 860:
```suggestion
```

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -943,8 +959,15 @@ where
...
```
jcjones, line 962:
```suggestion
```

divergentdave (approved):

aggregator/src/aggregator.rs
```diff
@@ -188,6 +188,13 @@ impl AggregatorMetrics {
     }
 }
 
+/// Represents the result of a Job Continue operation, which varies whether that was synchronous
+/// or async.
+enum AggregationJobContinueResult {
+    Sync { resp: AggregationJobResp },
+    Async { step: AggregationJobStep },
```
tgeoghegan, line 195: nit: I'm fine with this as-is, but in a case like this where each enum variant only contains a single value, I think using the struct notation is a bit too verbose. In tuple notation you'd have just:
```suggestion
    Sync(AggregationJobResp),
    Async(AggregationJobStep),
```

And it's a bit more compact when instantiating the values too.

aggregator/src/aggregator.rs
```diff
@@ -2356,16 +2363,16 @@ impl VdafOps {
 
                         let resp = match task.aggregation_mode() {
                             Some(AggregationMode::Synchronous) => {
-                                AggregationJobResp::Finished {
+                                AggregationJobContinueResult::Sync { resp:  AggregationJobResp {
```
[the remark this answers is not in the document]
jcjones: Well, gone now. :)

PR #3934, opened 2025-07-25
`janus_cli`: get aggregation and collection jobs

tgeoghegan:
Adds new subcommands to `janus_cli` to enable easily getting information about aggregation and collection jobs and any leases on them.


jcjones (approved):
I don't feel like I'm able to review the SQL here, but the Rust all looks fine except for my kinda indirect question about the `required_unless_present` that probably I just don't understand.

jcjones (approved):

PR #3944, opened 2025-07-30
Tests for database queries on `MaybeLease`s

tgeoghegan:
Adds test coverage for database accessors added in #3934

### DIFF
--- a/aggregator_core/src/datastore/tests.rs
+++ b/aggregator_core/src/datastore/tests.rs
@@ -1878,14 +1878,13 @@ async fn aggregation_job_acquire_release(ephemeral_datastore: EphemeralDatastore
                         .helper_view()
                         .unwrap();
                 tx.put_aggregator_task(&helper_task).await.unwrap();
-                let helper_aggregation_job_id = random();
                 tx.put_aggregation_job(&AggregationJob::<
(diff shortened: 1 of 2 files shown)


aggregator_core/src/datastore/tests.rs
```diff
@@ -1794,95 +1795,189 @@ async fn aggregation_job_acquire_release(ephemeral_datastore: EphemeralDatastore
...
-                &AggregationJob::<VERIFY_KEY_LENGTH, TimeInterval, Prio3Count>::new(
+                // Write an aggregation job for a task that we are taking on the helper role for.
+                // We don't want to retrieve this one, either.
+                let helper_task =
+                    TaskBuilder::new(task::QueryType::TimeInterval, VdafInstance::Prio3Count)
+                        .build()
+                        .helper_view()
+                        .unwrap();
+                tx.put_aggregator_task(&helper_task).await.unwrap();
+                let helper_aggregation_job_id = random();
```
jcjones, line 1881: nit: This wasn't necessary to return / is otherwise unused, so this could have remained an inline `random()` call.

jcjones (approved):
Quite the test!

Issue #3798, opened 2025-05-05
Wire aggregation job failures into per-task counters


tgeoghegan:
Janus tracks a variety of per-task metrics to count the occurrence of things like report uploads being rejected due to HPKE decryption failures. But if a report gets rejected by the Helper, the Leader's per-task counters won't reflect these failures. In particular, if Clients are configured to use the wrong HPKE keys for the Helper, then reports will fail to be aggregated but no information about this will be available via the Leader-side task counters. This can make it harder for someone deploying a task and clients to debug, as they may not have access to the Helper's per-task counters, Prometheus metrics, or logs.

We could expose some new per-task counters, clearly labeled by aggregator, and increment them as the Leader walks the `prepare_resps` in an `AggregationJobResp`.

PR #3958, opened 2025-08-05
Per-task counters for helper prep errors

tgeoghegan:
Expands `TaskAggregationCounter` so that the leader can report the various `PrepareError`s that a helper might indicate during handling of an aggregation job.

Part of #3798

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_driver.rs
+++ b/aggregator/src/aggregator/aggregation_job_driver.rs
@@ -23,7 +23,7 @@ use janus_aggregator_core::{
         self,
         models::{
             AcquiredAggregationJob, AggregationJob, AggregationJobState, Lease, ReportAggregation,
-            ReportAggregationState,
+            ReportAggregationState, TaskAggregationCounter,
         },
         Datastore,
     },
@@ -51,7 +51,12 @@ use prio::{
 };
 use rayon::iter::{IndexedParallelIterator as _, IntoParallelIterator as _, ParallelIterator as _};
 use reqwest::Method;
-use std::{collections::HashSet, panic, sync::Arc, time::Duration};
+use std::{
+    collections::HashSet,
+    panic,
+    sync::{Arc, Mutex},
+    time::Duration,
+};
 use tokio::{join, sync::mpsc, try_join};
 use tracing::{debug, error, info, info_span, trace_span, warn, Span};
 
@@ -309,6 +314,7 @@ where
         A::PublicShare: PartialEq + Send + Sync,
     {
         let aggregation_job = Arc::new(aggregation_job);
+        let task_aggregation_counter = TaskAggregationCounter::default();
 
         // Only process non-failed report aggregations.
         let report_aggregations: Vec<_> = report_aggregations
@@ -546,6 +552,7 @@ where
             stepped_aggregations,
             report_aggregations_to_write,
             resp,
+            task_aggregation_counter,
         )
         .await
     }
@@ -585,6 +592,8 @@ where
             })
             .collect();
         let report_aggregation_count = report_aggregations.len();
+
+        let task_aggregation_counter = TaskAggregationCounter::default();
 
         // Visit the report aggregations, ignoring any that have already failed; compute our own
         // next step & transitions to send to the helper.
@@ -729,6 +738,7 @@ where
             stepped_aggregations,
             report_aggregations_to_write,
             resp,
+            task_aggregation_counter,
         )
         .await
     }
@@ -749,6 +759,7 @@ where
         stepped_aggregations: Vec<SteppedAggregation<SEED_SIZE, A>>,
         mut report_aggregations_to_write: Vec<WritableReportAggregation<SEED_SIZE, A>>,
         helper_resp: AggregationJobResp,
+        task_aggregation_counters: TaskAggregationCounter,
     ) -> Result<(), Error>
     where
         A::AggregationParam: Send + Sync + Eq + PartialEq,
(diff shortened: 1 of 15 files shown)


tgeoghegan:
This only wires up helper side prep failures into counters because those are all handled in a single place (`AggregationJobDriver::process_response_from_helper`). We don't yet record counters for failures encountered by the leader or things like the helper rejecting the entire aggregation job for some reason, because this PR is already quite large. Follow-on PRs will add counters and instrumentation for these other failure cases as well as what I suspect will be fairly complex tests to induce these failure modes.

Another thing I refrained from doing here is significantly refactoring `TaskAggregationCounter`: given my longstanding opinion that `aggregator_core/src/datastore.rs` is too big, I would like to move methods `Transaction::{get_task_aggregation_counter,increment_task_aggregation_counter}` onto methods on `TaskAggregationCounter`. But I don't think a change that disruptive is a good idea on this release branch. I will probably give it a try when I forward-port this change to `main`, however.

aggregator_api/src/tests.rs
```diff
@@ -78,7 +78,7 @@ async fn get_config() {
             r#""protocol":"DAP-09","dap_url":"https://dap.url/","role":"Either","vdafs":"#,
             r#"["Prio3Count","Prio3Sum","Prio3Histogram","Prio3SumVec"],"#,
             r#""query_types":["TimeInterval","FixedSize"],"#,
-            r#""features":["TokenHash","UploadMetrics","TimeBucketedFixedSize","PureDpDiscreteLaplace"],"#,
+            r#""features":["TokenHash","UploadMetrics","TimeBucketedFixedSize","PureDpDiscreteLaplace","AggregationJobMetrics"],"#,
```
tgeoghegan, line 81: divviup-api uses the `AggregationJobMetrics` feature to govern whether it expects aggregators to expose these counters. In order to avoid needing features like `AggregationJobMetrics-2` or `-3`, I think that while I may not wire up the probes for all aggregation job counters in this PR, it would be wise to add every field we expect to report to `TaskAggregationCounter` and into the database table. That also means fewer DB migrations.

aggregator/src/aggregator/aggregation_job_driver.rs
```diff
@@ -309,6 +314,7 @@ where
         A::PublicShare: PartialEq + Send + Sync,
     {
         let aggregation_job = Arc::new(aggregation_job);
+        let task_aggregation_counter = TaskAggregationCounter::default();
```
jcjones, line 317: Here, and below with `step_aggregation_job_aggregate_continue`, we're constructing the TaskAggregationCounter to pass into the `process_response_from_helper` method, as a separation of concerns principal I imagine, keeping the _models_ away from the processing.

I expected to see then that we'd then be using this separation to exercise the Counter in the internals of `process_response_from_helper`, but now I realize we have no direct tests of that method. It's covered via `step_time_interval_aggregation_job_continue` and `step_fixed_size_aggregation_job_continue` and certainly the comprehensive integration tests, and all the functionality is covered by your `assert_task_aggregation_counter` update, but `process_response_from_helper` method in isolation is a thing we're lacking. Maybe that should be a todo.

aggregator/src/aggregator/aggregation_job_driver/tests.rs
```diff
@@ -84,49 +84,95 @@ async fn aggregation_job_driver() {
...
+        PrepareError::HpkeDecryptError,
+        PrepareError::VdafPrepError,
+        PrepareError::BatchSaturated,
+        PrepareError::TaskExpired,
+        PrepareError::InvalidMessage,
+        PrepareError::ReportTooEarly,
+    ];
+    let mut rejected_reports: Vec<_> = prepare_errors
+        .into_iter()
+        .map(|prepare_error| {
```
jcjones, line 125: every time I see this, I feel bad about writing `for prepare_error in ...` and being far too C.

jcjones (approved):

aggregator/src/aggregator/aggregation_job_driver.rs
tgeoghegan: Yeah, testing that function is isolation is interesting. In the past I think we've avoided doing it because we chose to prioritize higher level tests that exercise the HTTP interfaces or simulate running the aggregation job driver. The other thing is that as it's written today, `process_response_from_helper` meddles with the data store a bunch, which means that to exercise it, you have to do a bunch of prep work to seed the datastore with appropriate data. That's nearly as much work as exercising the aggregation job driver, so to avoid duplicating code, we just do the one.

That said, it probably would be nice to refactor some of Janus' business logic internals so that methods like `process_response_from_helper` could operate purely on objects in memory, and any database transactions would be handled in a calling context. Then it'd be vastly easier to write unit tests for the internals that exercise obtuse corner cases, and we might be able to remove a lot of the code that excruciatingly carefully constructs database records.

aggregator/src/aggregator/aggregation_job_driver/tests.rs
tgeoghegan: I certainly have a bias toward iterators, but you can only go so far with them. The language kinda nudges you toward using for loops by having some of the ceremony around borrowing elided. And more significantly, I often find myself starting out by chaining iterators, but then realizing I need to do something `async` in a closure and glumly resorting to a `for` loop. Point being: both patterns have their place. Use whatever works and don't beat yourself up.

db/00000000000009_task_aggregation_counters_failures.up.sql
```diff
@@ -0,0 +1,24 @@
+ALTER TABLE task_aggregation_counters
+    -- errors encountered by this aggregator
+    ADD COLUMN duplicate_extension BIGINT NOT NULL DEFAULT 0,
+    ADD COLUMN public_share_encode_failure BIGINT NOT NULL DEFAULT 0,
```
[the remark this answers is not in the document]
jcjones: oh and the one above it, `duplicate_extension`

PR #3961, opened 2025-08-11
Extension traits for fluent request auth

tgeoghegan:
Adds extension traits on `reqwest::Builder`, `mockito::Mock` and `trillium_testing::TestConn` so that we can insert or expect an `Authorization: Bearer <blah>` header more fluently when working with a `janus_core::auth_token::AuthenticationToken`.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -63,6 +63,7 @@ use janus_core::{
     Runtime,
     auth_tokens::AuthenticationToken,
     hpke::{self, HpkeApplicationInfo, Label},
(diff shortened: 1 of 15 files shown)


jcjones (approved):
I can't find anything important to remark on. I looked over at how [Trillium sets this all up](https://docs.trillium.rs/src/trillium_conn_id/lib.rs#132-135) which was nicely instructive. I also went on a little sojurn reading about the [#[must_use]](https://rustwiki.org/en/reference/attributes/diagnostics.html#the-must_use-attribute) attribute, which I suppose we could use here, but since Trillium doesn't, it'd just be for the lols.

Issue #3962, opened 2025-08-11
DAP-15: Allow DELETE of Aggregate Shares


jcjones:
Per https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-15#section-4.7.4:
>The Leader can send a DELETE request to the aggregate share, which indicates to the Helper that it can abandon the aggregate share and discard all state related to it.

PR #3968, opened 2025-08-12
Support asynchronous aggregate share collection via GET requests

### LINKED ISSUE #3963: DAP-15: Helper should support GET polling for asynchronous aggregate shares
> The Helper MAY defer handling the aggregate share request. In this case, it indicates that the aggregate share is not yet ready by immediately sending an empty response body. The Helper SHOULD include a Retry-After header field ([[RFC9110](https://www.rfc-editor.org/rfc/rfc9110)], [Section 10.2.3](https://rfc-editor.org/rfc/rfc9110#section-10.2.3)) to suggest a polling interval to the Leader. The Leader then polls the state of the job by sending GET requests to the aggregate share. The Leader SHOULD use each response's Retry-After header field to decide when to try again. The Helper responds the same way until either the share is ready, from which point it responds with the AggregateShare (defined below), or the job fails, from which point it MUST abort with the error that caused the failure.

jcjones:
Support asynchronous Aggregate State collection workflows for compatible helpers. Janus now will use the GET Aggregate Share endpoint during retries after a succesful PUT.

Note: This does not implement asynchronous Aggregation State collection in Janus; Janus will always complete synchronously, but you can request the share idempotently via the GET endpoint nonetheless. This allows for retries.

Changes:
- Add CollectionJobState::AwaitingHelper for tracking async operations.
- Implement GET request handling at aggregate_shares endpoint for helpers.
- Add handle_aggregate_share_get() method for helper aggregate share retrieval.
- Update database schema to support AwaitingHelper state transitions.
- Update collection job driver to change how it interacts with the Helper based on the current state:
    * "Start" state: PUT request with AggregateShareReq body.
    * "AwaitingHelper" state: GET request for polling completion.
    * "Deletion" will be supported in #3962, for now it sends another GET.

> [!NOTE]
> I need to create the documentation for https://docs.divviup.org/references/janus-errors#aggregate-share-id-unrecognized, which I will do after we settle on the problem doc.

Fixes #3963. Is stacked on top of https://github.com/divviup/janus/pull/3947.


aggregator/src/aggregator.rs
```diff
@@ -672,25 +672,17 @@ impl<C: Clock> Aggregator<C> {
         Ok(())
     }
 
-    /// Handle an aggregate share request. Only supported by the helper. `req_bytes` is an encoded
-    /// [`AggregateShareReq`]. Returns an [`AggregateShare`].
-    async fn handle_aggregate_share(
+    /// Ensure an incoming Aggregate Share request is valid and authorized. Returns
+    /// a tuple of a [`TaskAggregator`] and a [`Option<&TaskConfig>`].
```
jcjones, line 676:
```suggestion
    /// Ensure an incoming Aggregate Share request is valid and authorized. Returns
    /// a  [`HpkeConfig`].
```

jcjones:
> So the short version: for consistency, it would be nice to rename to `COLLECTION_JOB_STATE::POLL` but I don't think it's a dealbreaker, because `AWAITING_HELPER` is perfectly clear.

I'm already mucking about in here for the merge, so I've made a commit to rename it. :)

PR #3974, opened 2025-08-13
Support aggregate share job deletion

### LINKED ISSUE #3962: DAP-15: Allow DELETE of Aggregate Shares

jcjones:
Changes:
- Add DELETE request support at aggregate_shares endpoint for cleanup operations, to comply with DAP-15 section 4.7.4.
- Change the collection job driver to send DELETE to the helper on the next step after the collection job state transitions to Deleted.

> [!NOTE]
> I didn't bother doing the `vdaf_ops_dispatch` business for the DELETE calls, as all the generics were unused, so it saves a bunch of boilerplate (about 80 lines -- I started with them). 

Fixes #3962

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -785,6 +785,9 @@ impl<C: Clock> Aggregator<C> {
             .get(task_id)
             .await?
             .ok_or(Error::UnrecognizedTask(*task_id))?;
+        if task_aggregator.task.role() != &Role::Helper {
+            return Err(Error::UnrecognizedTask(*task_id));
+        }
         let _ = self
             .validate_and_authorize_aggregate_share_request(
                 &task_aggregator.task,
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -772,6 +772,31 @@ impl<C: Clock> Aggregator<C> {
             .await
     }
 
+    /// Handle an aggregate share deletion request. Only supported by the helper.
+    async fn handle_delete_aggregate_share(
+        &self,
+        task_id: &TaskId,
+        aggregate_share_id: &AggregateShareId,
+        auth_token: Option<AuthenticationToken>,
+        taskprov_task_config: Option<&TaskConfig>,
+    ) -> Result<(), Error> {
+        let task_aggregator = self
+            .task_aggregators
+            .get(task_id)
+            .await?
+            .ok_or(Error::UnrecognizedTask(*task_id))?;
+        let _ = self
+            .validate_and_authorize_aggregate_share_request(
+                &task_aggregator.task,
+                auth_token,
+                taskprov_task_config,
+            )
+            .await?;
+        task_aggregator
+            .handle_delete_aggregate_share(&self.datastore, aggregate_share_id)
+            .await
+    }
+
     /// Opts in or out of a taskprov task.
     #[tracing::instrument(skip(self, aggregator_auth_token), err(level = Level::DEBUG))]
     async fn taskprov_opt_in(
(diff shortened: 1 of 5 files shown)


aggregator/src/aggregator/collection_job_driver.rs
```diff
@@ -376,9 +376,7 @@ impl CollectionJobDriver {
                 Ok((Method::PUT, Some(request_body)))
             }
             CollectionJobState::Poll => Ok((Method::GET, None)),
-            // TODO: This needs to send DELETE, not GET, but our Helper isn't able to do this
-            // yet. See Issue #3962
-            CollectionJobState::Deleted => Ok((Method::GET, None)),
+            CollectionJobState::Deleted => Ok((Method::DELETE, None)),
```
tgeoghegan, line 379: What happens if the DELETE the leader sends to the helper fails? Presumably we'll do the usual retry dance, but if we give up on that, what state is the collection job left in on the leader side?

aggregator_core/src/datastore.rs
```diff
@@ -4630,6 +4630,38 @@ ON CONFLICT(task_id, batch_identifier, aggregation_param) DO UPDATE
         )
     }
 
+    /// Delete an `aggregate_share_job` row from the datastore.
+    #[tracing::instrument(skip(self), err(level = Level::DEBUG))]
+    pub async fn delete_aggregate_share_job(
```
tgeoghegan, line 4635: What's the story for deleting things like batch aggregations or report aggregations? Do we expect that garbage collection will eventually clean them up?

tgeoghegan (approved):

aggregator_core/src/datastore.rs
jcjones: `delete_expired_collection_artifacts` will definitely still get them, as it's keyed off the TaskID instead.

This is where I have a weak understanding still of the data model. The `batch_aggregations` table has `batch_identifier` onto which we can cascade a delete from the aggregate share jobs.

The `report_aggregations` table connects to the `aggregation_jobs` table which has `batch_id`. Would you expect the delete to cascade through all three tables, @tgeoghegan ?
- batch_aggregations, via batch_aggregations.batch_identifier
- aggregation_jobs, via aggregation_jobs.batch_id
- report_aggregations, via aggregation_jobs.aggregation_job_id

aggregator/src/aggregator/collection_job_driver.rs
jcjones: @tgeoghegan Take a look at the update in 438bd575

aggregator/src/aggregator.rs
```diff
@@ -772,6 +772,31 @@ impl<C: Clock> Aggregator<C> {
...
 
+    /// Handle an aggregate share deletion request. Only supported by the helper.
+    async fn handle_delete_aggregate_share(
+        &self,
+        task_id: &TaskId,
+        aggregate_share_id: &AggregateShareId,
+        auth_token: Option<AuthenticationToken>,
+        taskprov_task_config: Option<&TaskConfig>,
+    ) -> Result<(), Error> {
+        let task_aggregator = self
```
tgeoghegan, line 783: This should check that we are acting as the helper in this task, akin to the check in [`handle_delete_collection_job`](https://github.com/divviup/janus/blob/f0b9435b764042254252607796861e6125924705/aggregator/src/aggregator.rs#L659)

tgeoghegan (approved):

PR #3993, opened 2025-08-19
Helper must check for reports whose timestamps are after the task expired

### LINKED ISSUE #3989: Missing task expiration check in helper aggregation initialization
We don't currently check report timestamps against the task end time on the helper side. This should produce `ReportError.task_expired`. We represent this as `ReportError::TaskExpired` internally, but that is not yet used. See DAP draft-15 section 4.6.2.4, item 5.

jcjones:
This check is specified from DAP-15 section 4.6.2.4, item 5.

Fixes #3989

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_init.rs
+++ b/aggregator/src/aggregator/aggregation_job_init.rs
@@ -364,7 +364,7 @@ where
                                     .report_share()
                                     .metadata()
                                     .time()
-                                    .is_after(task_end)
+                                    .ge(task_end)
                                 {
                                     return Err(ReportError::TaskExpired);
                                 }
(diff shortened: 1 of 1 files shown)


aggregator/src/aggregator/aggregation_job_init.rs
```diff
@@ -341,6 +341,21 @@ where
...
                         });
 
+                        // Reject reports after the task has expired. (§4.6.2.4, item 5)
+                        let shares = shares.and_then(|shares| {
+                            if let Some(task_end) = task.task_end() {
+                                if prepare_init
+                                    .report_share()
+                                    .metadata()
+                                    .time()
+                                    .is_after(task_end)
```
divergentdave, line 351: The requirement says "Check if the report's timestamp is after the task's task_interval", and DAP time intervals are half-open, so I think we ought to treat reports as expired if their timestamp is equal to the task end time.

divergentdave (approved):

PR #3999, opened 2025-08-22
Leader must reject reports with duplicate extensions

### LINKED ISSUE #3997: DAP-15: Resolve input share extension processing differences between Helper and Leader
Per section [4.6.2.4, Input Share Validation](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-15#name-input-share-validation), steps 6-7, the Aggregator needs to confirm:
- There are no duplicate extensions
- There are no unrecognized extensions

The Helper does this in two pieces:
- Duplication is checked here:  https://github.com/divviup/janus/blob/main/aggregator/src/aggregator/aggregation_job_init.rs#L220-L236
- Unrecognized extensions are caught earlier, during the Decode operation: https://github.com/divviup/janus/blob/main/messages/src/lib.rs#L1027-L1034

The Leader, by nature of also using the Extension type, gets the benefit of checking for unrecognized extensions the same way. It is missing an equivalent duplicate extensions check.

jcjones:
Note there's an error response inconsistency between Helper and Leader -- Leader returns ReportRejectionReason::DuplicateExtension while Helper returns ReportError::InvalidMessage (line 235 in aggregation_job_init.rs).

Also note that this adds the telemetry to TaskUploadCounter, but we already have -- completely unused -- DuplicateExtension in TaskAggregationCounter.

Maybe this is putting the rejection in the wrong place for the Leader?

Fixes #3997

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1886,6 +1886,7 @@ impl VdafOps {
         // Check if any two extensions have the same extension type across public
         // and private extension fields. If so, the Aggregator MUST mark the input
         // share as invalid with error invalid_message. (§4.6.2.4 step 7)
+        //
         // Note at this point we can't provide Error::InvalidMessage here because
         // we're packaging this into a report rejection list.
         let mut extensions = HashMap::new();
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -99,7 +99,7 @@ use rand::{Rng, random, rng};
 use reqwest::Client;
 use std::{
     borrow::Cow,
-    collections::HashSet,
+    collections::{HashMap, HashSet},
     fmt::Debug,
     panic,
     path::PathBuf,
@@ -1882,6 +1882,32 @@ impl VdafOps {
                 return Err(reject_report(ReportRejectionReason::DecodeFailure).await?);
             }
         };
+
+        // Check if any two extensions have the same extension type across public
+        // and private extension fields. If so, the Aggregator MUST mark the input
+        // share as invalid with error invalid_message. (§4.6.2.4 step 7)
+        // Note at this point we can't provide Error::InvalidMessage here because
+        // we're packaging this into a report rejection list.
+        let mut extensions = HashMap::new();
+        if !leader_private_extensions
+            .iter()
+            .chain(report.metadata().public_extensions())
+            .all(|extension| {
+                extensions
+                    .insert(*extension.extension_type(), extension.extension_data())
+                    .is_none()
+            })
+        {
+            debug!(
+                task_id = %task.id(),
+                report_id = ?report.metadata().id(),
+                "Received report share with duplicate extensions",
+            );
+            metrics
+                .upload_decode_failure_counter
+                .add(1, &[KeyValue::new("type", "duplicate_extension")]);
+            return Err(reject_report(ReportRejectionReason::DuplicateExtension).await?);
+        }
 
         let report = LeaderStoredReport::new(
             *task.id(),
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -315,6 +315,7 @@ impl<C: Clock> Aggregator<C> {
             .with_unit("{error}")
             .build();
         upload_decode_failure_counter.add(0, &[]);
+        upload_decode_failure_counter.add(0, &[KeyValue::new("type", "duplicate_extension")]);
 
         let report_aggregation_success_counter = report_aggregation_success_counter(meter);
         let aggregate_step_failure_counter = aggregate_step_failure_counter(meter);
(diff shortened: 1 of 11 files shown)


aggregator/src/aggregator.rs
```diff
@@ -1883,6 +1883,33 @@ impl VdafOps {
...
+            })
+        {
+            debug!(
+                task_id = %task.id(),
+                report_id = ?report.metadata().id(),
+                "Received report share with duplicate extensions",
+            );
+            metrics
+                .upload_decode_failure_counter
+                .add(1, &[KeyValue::new("type", "duplicate_extension")]);
```
divergentdave, line 1909: We could add a `upload_decode_failure_counter.add(0, &[Keyvalue::new("type", "duplicate_extension")]);` in `Aggregator::new()` to eagerly populate this time series at startup.

divergentdave (approved):

db/00000000000001_initial_schema.up.sql
```diff
@@ -169,6 +169,7 @@ CREATE TABLE task_upload_counters(
     report_too_early       BIGINT NOT NULL DEFAULT 0, -- reports whose timestamp is too far in the future.
     task_not_started       BIGINT NOT NULL DEFAULT 0, -- reports sent to the task before it started.
     task_ended             BIGINT NOT NULL DEFAULT 0, -- reports sent to the task after it ended.
+    duplicate_extension    BIGINT NOT NULL DEFAULT 0, -- reports with duplicate extensions
```
tgeoghegan, line 172: `divviup-api` will need matching changes: https://github.com/divviup/divviup-api/issues/1884

aggregator/src/aggregator.rs
```diff
@@ -1883,6 +1883,33 @@ impl VdafOps {
...
         };
 
+        // Check if any two extensions have the same extension type across public
+        // and private extension fields. If so, the Aggregator MUST mark the input
+        // share as invalid with error invalid_message. (§4.6.2.4 step 7)
+        //
+        // Note at this point we can't provide Error::InvalidMessage here because
+        // we're packaging this into a report rejection list.
+        let mut extensions = HashMap::new();
+        if !leader_private_extensions
```
tgeoghegan, line 1893: I don't suppose that the existing code that handles this on the helper side could be factored out into a method used by both? Then again it's so little code that it may not be worth it.

tgeoghegan (commented):
> Note there's an error response inconsistency between Helper and Leader -- Leader returns ReportRejectionReason::DuplicateExtension while Helper returns ReportError::InvalidMessage (line 235 in aggregation_job_init.rs).

`ReportRejectionReason::DuplicateExtension` doesn't seem right to me. [The pertinent text from draft-ietf-ppm-dap-15](
https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-15#section-4.5.2-26):

> If the same extension type appears more than once among the public extensions and the private extensions in the Leader's input share, then the Leader MUST ignore the report and MAY abort with error invalidMessage.

(See discussion of that MAY below) It seems fine for us to track this rejection reason internally to Janus (in particular so we can increment a counter) but what error does the client get? I'm not sure if the test you added lets us examine the HTTP response code and problem document a client would get. FWIW I think it'd be fine to add a test proving that `ReportRejectionReason::DuplicateExtension` gets rendered into the appropriate HTTP response. We don't necessarily need to rewrite the test to exercise the HTTP layer.

> Also note that this adds the telemetry to TaskUploadCounter, but we already have -- completely unused -- DuplicateExtension in TaskAggregationCounter.
> 
> Maybe this is putting the rejection in the wrong place for the Leader?

The MAYs in the various upload path checks are intended to allow leaders to defer checks from the upload path (which is high-volume and keeps a client waiting), but Janus opts to do these checks as soon as possible. Or rather, we chose to do it in the more obvious place and figured we'd move it to the aggregation job path if it ever proved to be a performance problem. It never has -- what has caused us perf issues during upload is writing reports to the database (IO), not validating incoming messages (compute).

Anyway, we need that other country so that the helper can record such occurrences, since it never gets to see reports until it handles aggregation jobs. Admittedly the scenario of a helper getting duplicate extensions is quite unlikely -- you'd need a client to have correctly constructed the leader report share but mangled the helper's -- but such a bug being improbably makes tools for detecting it that much more valuable, I think.

aggregator/src/aggregator.rs
jcjones: Not really. The inputs are different, and so are the metrics. The only part that is generic is this:
```
.all(|extension| {
                                extensions
                                    .insert(*extension.extension_type(), extension.extension_data())
                                    .is_none()
                            }) {
                                debug!(
                                    task_id = %task.id(),
                                    report_id = ?prepare_init.report_share().metadata().id(),
                                    "Received report share with duplicate extensions",
                                );
```

which isn't nothin', but basically just the lambda inside `all` is actually generic.

tgeoghegan (approved):

PR #4014, opened 2025-08-26
Move to 0.9.0-prerelease-1 targeting DAP-16 (dap-gh currently)

jcjones:


.github/dependabot.yml
```diff
@@ -16,13 +16,13 @@ updates:
       # opentelemetry-prometheus, and will add it back after the 1.0 release.
       - dependency-name: opentelemetry
         versions:
-        - ">= 0.25, < 1.0"
+          - ">= 0.25, < 1.0"
```
tgeoghegan, line 19: Let's remove these indentation changes

.github/dependabot.yml
```diff
@@ -159,3 +159,81 @@ updates:
     target-branch: "release/0.7"
     ignore:
       - dependency-name: "postgres"
+
+  # Manage dependencies on the release/0.8 branch
```
tgeoghegan, line 163: I'm not sure we need to do this: we don't intend to deploy `release/0.8` anywhere, which means we don't really care if its dependencies contain vulnerabilities. And the thing is that running dependabot on an extra release branch does consume GitHub Actions runner time (which we pay for) and in some cases adds human toil for wrangling dependabot PRs.

So instead of adding this stanza, we could just not, and mark the branch as unsupported in the top-level README.md.

tgeoghegan (approved):

PR #4032, opened 2025-09-08
Add timeout support to LIFO request queue (release/0.7)

jcjones:
> [!note]
> This is against `release/0.7`

Restructures the LIFO request queue to support configurable timeouts for requests that are taking too long to be serviced.

Changes:
- Add RequestTimeout error variant with HTTP 429 responses
- Implement an optional `timeout_ms` configuration option for `helper_aggregation_request_queue`
  When set to None, it should be as if this commit wasn't made.
- When `timeout_ms` is set for a request queue, the queue future is wrapped in a `tokio::time::timeout` of that duration.
- Adds timeout testing to `quickcheck_full` and a new timeout-specific, non-quickcheck queue test.

### DIFF
--- a/aggregator/src/aggregator/http_handlers.rs
+++ b/aggregator/src/aggregator/http_handlers.rs
@@ -172,7 +172,7 @@ async fn run_error_handler(error: &Error, mut conn: Conn) -> Conn {
             &ProblemDocument::new(
                 "https://docs.divviup.org/references/janus-errors#request-timeout",
                 "Request timed out waiting in queue.",
-                Status::RequestTimeout,
+                Status::TooManyRequests,
             )
             .with_detail("The request spent too long waiting to be processed."),
         ),
--- a/aggregator/src/aggregator/queue.rs
+++ b/aggregator/src/aggregator/queue.rs
@@ -151,9 +151,8 @@ impl LIFORequestQueue {
                                     },
                                     DispatcherMessage::Cancel(id) => {
                                         debug!(?id, "removing request from the queue");
-                                        if stack.remove(&id).is_some() {
-                                            metrics.requests_cancelled.add(1, &[]);
-                                        }
+                                        stack.remove(&id);
+                                        metrics.requests_cancelled.add(1, &[]);
                                     },
                                 }
                             }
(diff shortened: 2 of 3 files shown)


aggregator/src/aggregator/error.rs
```diff
@@ -147,6 +147,10 @@ pub enum Error {
     ClientDisconnected,
     #[error("too many requests")]
     TooManyRequests,
+    #[error("request timed out waiting in queue")]
+    RequestTimeout,
+    #[error("request timed out during service")]
+    ServiceTimeout,
```
divergentdave, line 153: This variant is no longer constructed

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -299,6 +307,9 @@ pub(crate) static AGGREGATE_SHARES_ROUTE: &str = "tasks/:task_id/aggregate_share
 pub struct HelperAggregationRequestQueue {
     pub depth: usize,
     pub concurrency: u32,
+    /// Maximum lifespan, in milliseconds, of requests in the Request Queue.
+    #[serde(default)]
+    pub timeout_ms: Option<u64>,
```
divergentdave, line 312: We should also document this in `docs/samples/advanced_config/aggregator.yaml`

aggregator/src/aggregator/queue.rs
```diff
@@ -863,4 +1046,91 @@ mod tests {
...
+            let mut requests = Vec::new();
+            for _ in 0..concurrency {
+                let handler = Arc::clone(&handler);
+                requests.push(tokio::spawn(
+                    async move { get("/").run_async(&handler).await },
+                ));
+            }
+
+            // Wait for all active slots to be filled
+            tokio::time::sleep(Duration::from_millis(10)).await;
```
jcjones, line 1091: when I forward-port this, I want to make this use `wait_for`. I gave up this round.

aggregator/src/aggregator/queue.rs
```diff
@@ -125,22 +133,28 @@ impl LIFORequestQueue {
...
                                             permit_tx.send(Err(Error::TooManyRequests));
+                                            metrics.requests_rejected.add(1, &[]);
                                         }
                                     },
                                     DispatcherMessage::Cancel(id) => {
-                                        debug!(?id, "removing from the queue");
-                                        stack.remove(&id);
-                                     },
+                                        debug!(?id, "removing request from the queue");
+                                        if stack.remove(&id).is_some() {
```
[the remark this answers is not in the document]
jcjones: I removed the conditional.

PR #4057, opened 2025-09-22
make CI fail if clippy finds something

tgeoghegan:
Run `clippy` with `-Dwarnings` so that the process exits non-zero if it finds a lint, which will cause GitHub to fail the build.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -132,9 +132,9 @@ jobs:
     - name: Format
       run: cargo fmt --message-format human -- --check
     - name: Clippy
-      run: cargo clippy --profile ci --workspace --all-targets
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-build.yml
```diff
@@ -132,9 +132,9 @@ jobs:
     - name: Format
       run: cargo fmt --message-format human -- --check
     - name: Clippy
-      run: cargo clippy --profile ci --workspace --all-targets
+      run: cargo clippy --profile ci --workspace --all-targets -- -Dwarnings
     - name: Clippy (all features)
-      run: cargo clippy --profile ci --workspace --all-targets --all-features
+      run: cargo clippy --profile ci --workspace --all-targets --all-features -- -Dwarnings
```
divergentdave, line 137: This is redundant with our RUSTFLAGS environment variable above

Issue #4019, opened 2025-08-28
DAP-16: Represent Times and Durations as Multiples of Precision


jcjones:
Times and Durations now represented as "number of time_precisions since the Epoch" rather than seconds

PR #4181, opened 2025-11-24
Use chrono::DateTime<Utc> for trait Clock

jcjones:
This is again part of #4019, as now we can't return `Time` from `Clock` once `Time` becomes defined as a multiple of `TaskPrecision`.

This adds a somewhat unfortunate number of `to_time()` calls sprinkled around the codebase, where things are still expecting a `janus_messages::Time` type. I haven't audited all of those sites yet to be able to tell which will change to itself move to a `chrono` type... I figure this change is already plenty big, again.

I've gone back and forth with the API for this and finally settled on everything possible using the `add_duration` `sub_timeprecision` and such methods instead of direct math, because while it ends up as more lines, it feels safer. That has made some of the test changes a bit more verbose/awkward than I'd prefer, but I think it's more readable.

### DIFF
--- a/core/src/time.rs
+++ b/core/src/time.rs
@@ -370,6 +370,7 @@ impl DateTimeExt for DateTime<Utc> {
 }
 
 /// Extension methods on [`Time`].
+///
 /// Deprecation notice: These methods will be substanially revised as part of #4019.
 pub trait TimeExt: Sized {
     /// Compute the start of the batch interval containing this Time, given the task time precision.


core/src/time.rs
```diff
@@ -249,7 +249,128 @@ impl TimeDeltaExt for TimeDelta {
...
+    fn saturating_difference(&self, other: &Time) -> Duration {
+        Duration::from_seconds(
+            self.as_seconds_since_epoch()
+                .saturating_sub(other.as_seconds_since_epoch()),
+        )
+    }
+}
+
 /// Extension methods on [`Time`].
+/// Deprecation notice: These methods will be substanially revised as part of #4019.
```
divergentdave: nit: new paragraph
```suggestion
///
/// Deprecation notice: These methods will be substanially revised as part of #4019.
```

divergentdave (approved):

Issue #4218, opened 2025-12-15
CLI binaries need updates for Times and Durations unit changes


jcjones:
In order to keep things moving, I didn't update all the binaries to support the new #4019 changes, but they do need to be updated.

Issue #4217, opened 2025-12-15
BatchIdentifiers need to be interconnected with TimePrecisions for proper garbage collection


jcjones:
`batch_mode.rs`'s `is_batch_garbage_collected` makes an incorrect assumption about the Time Precision of the incoming `batch_identifier`, because plumbing through precision on that trait is nontrivial. It still needs doing.

Issue #4216, opened 2025-12-15
Key Rotator: Use chrono types


jcjones:
As part of #4019, `key_rotator` also needs a revamp, but it's working as-is and that change is big enough. Come back to it.

PR #4219, opened 2025-12-16
Change janus_messages::Duration and Time to be in units of TimePrecision

jcjones:
Resolves https://github.com/divviup/janus/issues/4019

The rest of the owl:
```
    /\_/\
   ( o o )
   (  V  )
  /--m-m--\
```

This small change consists of a few commits which refactor `janus_messages::Time` and `::Duration` to be in `janus_messages::TimePrecision` units. After the change to the `janus_messages` module, the first series of commits were done with regexes to get everything to compile. Then I progressively fixed tests, until they all passed. Then I've been performing cleanup, resolving issues from self-review and otherwise. 

I have filed follow-on issues for things that seemed bigger but not intrinsic to this change:
- [Key Rotator: Use chrono types #4216](https://github.com/divviup/janus/issues/4216)
- [BatchIdentifiers need to be interconnected with TimePrecisions for proper garbage collection #4217](https://github.com/divviup/janus/issues/4217)
- [CLI binaries need updates for Times and Durations unit changes #4218](https://github.com/divviup/janus/issues/4218)

I did include two other notable fixes/refactors in this PR which arguably could have gone into separate ones, and still could:
- [Fix Interval::EMPTY round-trip asymmetry](https://github.com/divviup/janus/commit/bbfbffa72697c8596202ffca8dedc75e7cad4c37), which came up while I was fixing tests but seems like a non-issue in production
- [Use IntervalExt::merged_with instead of manual interval construction](https://github.com/divviup/janus/commit/4070d0ad2d1a041809c6d95bea4205deb0dc6ebc), which is mostly that we had a slick way to make `client_timestamp_interval` one place using `fold` and `IntervalExt::merged_with` which _didn't_ produce zero-width intervals, and then we had the other way that needed explicit `add_duration(Duration::from_seconds(1), &time_precision)`. This seemed worthwhile to take here.

### DIFF
--- a/aggregator_core/src/datastore/tests.rs
+++ b/aggregator_core/src/datastore/tests.rs
@@ -97,12 +97,6 @@ const START_TIMESTAMP: u64 = START_TIME_UNITS * TIME_PRECISION_SECONDS;
 const REPORT_TIME_AT_EXPIRY_EDGE: Time =
     Time::from_time_precision_units(START_TIME_UNITS - REPORT_EXPIRY_AGE_UNITS);
 
-// Report time for most tests - at START_TIME, reports with this timestamp have age 0.
-// This preserves the semantics from before the time precision changes.
-const OLDEST_ALLOWED_REPORT_TIME: Time = START_TIME;
-const OLDEST_ALLOWED_REPORT_TIMESTAMP: u64 =
-    OLDEST_ALLOWED_REPORT_TIME.as_seconds_since_epoch(&TIME_PRECISION);
-
 #[test]
 fn check_supported_versions() {
     if SUPPORTED_SCHEMA_VERSIONS[0] != *SUPPORTED_SCHEMA_VERSIONS.iter().max().unwrap() {
@@ -725,10 +719,6 @@ async fn get_unaggregated_client_reports_for_task(ephemeral_datastore: Ephemeral
 
     let clock = MockClock::new(START_TIMESTAMP);
     let ds = ephemeral_datastore.datastore(clock.clone()).await;
-    let unexpired_time = OLDEST_ALLOWED_REPORT_TIME
-        .add_duration(&Duration::ONE)
-        .unwrap();
-    let report_interval = Interval::minimal(unexpired_time).unwrap();
     let task = TaskBuilder::new(
         task::BatchMode::TimeInterval,
         AggregationMode::Synchronous,
@@ -749,16 +739,13 @@ async fn get_unaggregated_client_reports_for_task(ephemeral_datastore: Ephemeral
     .leader_view()
     .unwrap();
 
-    let first_unaggregated_report = LeaderStoredReport::new_dummy(*task.id(), unexpired_time);
-    let second_unaggregated_report = LeaderStoredReport::new_dummy(*task.id(), unexpired_time);
-    let expired_report = LeaderStoredReport::new_dummy(
-        *task.id(),
-        OLDEST_ALLOWED_REPORT_TIME
-            .sub_duration(&Duration::ONE)
-            .unwrap(),
-    );
-    let aggregated_report = LeaderStoredReport::new_dummy(*task.id(), unexpired_time);
-    let unrelated_report = LeaderStoredReport::new_dummy(*unrelated_task.id(), unexpired_time);
+    let report_interval = Interval::minimal(START_TIME).unwrap();
+    let first_unaggregated_report = LeaderStoredReport::new_dummy(*task.id(), START_TIME);
+    let second_unaggregated_report = LeaderStoredReport::new_dummy(*task.id(), START_TIME);
+    let expired_report =
+        LeaderStoredReport::new_dummy(*task.id(), START_TIME.sub_duration(&Duration::ONE).unwrap());
+    let aggregated_report = LeaderStoredReport::new_dummy(*task.id(), START_TIME);
+    let unrelated_report = LeaderStoredReport::new_dummy(*unrelated_task.id(), START_TIME);
 
     // Set up state.
     ds.run_tx("test-unaggregated-reports", |tx| {
@@ -795,7 +782,7 @@ async fn get_unaggregated_client_reports_for_task(ephemeral_datastore: Ephemeral
     .unwrap();
 
     // Advance the clock to "enable" report expiry.
-    clock.advance(REPORT_EXPIRY_AGE_PLUS_ONE);
+    clock.advance(REPORT_EXPIRY_AGE);
 
     // Verify that we can acquire both unaggregated reports.
     let got_reports_ids: HashSet<_> = ds
@@ -907,9 +894,7 @@ async fn get_unaggregated_client_reports_for_task(ephemeral_datastore: Ephemeral
             tx.check_timestamp_columns_at_create_time(
                 "client_reports",
                 "test-unaggregated-reports",
-                OLDEST_ALLOWED_REPORT_TIME
-                    .as_naive_date_time(&TIME_PRECISION)
-                    .unwrap(),
+                START_TIME.as_naive_date_time(&TIME_PRECISION).unwrap(),
                 false,
             )
             .await;
@@ -928,9 +913,7 @@ async fn get_unaggregated_client_reports_for_task(ephemeral_datastore: Ephemeral
                     assert_eq!(tx.clock.now().naive_utc(), updated_at, "{report_id:?}");
                 } else {
                     assert_eq!(
-                        OLDEST_ALLOWED_REPORT_TIME
-                            .as_naive_date_time(&TIME_PRECISION)
-                            .unwrap(),
+                        START_TIME.as_naive_date_time(&TIME_PRECISION).unwrap(),
                         updated_at
                     );
                 }
@@ -1299,37 +1282,18 @@ async fn count_client_reports_for_interval(ephemeral_datastore: EphemeralDatasto
     .leader_view()
     .unwrap();
 
-    let interval_start = Time::from_time_precision_units(START_TIME_UNITS + 1);
-    let expired_report_in_interval = LeaderStoredReport::new_dummy(
-        *task.id(),
-        interval_start
-            .sub_timedelta(
-                &TimeDelta::seconds(TIME_PRECISION_SECONDS as i64),
-                &TIME_PRECISION,
-            )
-            .unwrap(),
-    );
-    let first_report_in_interval = LeaderStoredReport::new_dummy(*task.id(), interval_start);
-    let second_report_in_interval = LeaderStoredReport::new_dummy(
-        *task.id(),
-        interval_start
-            .add_timedelta(
-                &TimeDelta::seconds(TIME_PRECISION_SECONDS as i64),
-                &TIME_PRECISION,
-            )
-            .unwrap(),
-    );
+    let expired_report_in_interval =
+        LeaderStoredReport::new_dummy(*task.id(), START_TIME.sub_duration(&Duration::ONE).unwrap());
+    let first_report_in_interval = LeaderStoredReport::new_dummy(*task.id(), START_TIME);
+    let second_report_in_interval =
+        LeaderStoredReport::new_dummy(*task.id(), START_TIME.add_duration(&Duration::ONE).unwrap());
     let report_outside_interval = LeaderStoredReport::new_dummy(
         *task.id(),
-        interval_start
-            .add_timedelta(
-                &TimeDelta::seconds(10 * TIME_PRECISION_SECONDS as i64),
-                &TIME_PRECISION,
-            )
+        START_TIME
+            .add_duration(&Duration::from_time_precision_units(10))
             .unwrap(),
     );
-    let report_for_other_task =
-        LeaderStoredReport::new_dummy(*unrelated_task.id(), OLDEST_ALLOWED_REPORT_TIME);
+    let report_for_other_task = LeaderStoredReport::new_dummy(*unrelated_task.id(), START_TIME);
 
     // Set up state.
     ds.run_unnamed_tx(|tx| {
@@ -1368,7 +1332,7 @@ async fn count_client_reports_for_interval(ephemeral_datastore: EphemeralDatasto
     .unwrap();
 
     // Advance the clock to "enable" report expiry.
-    clock.advance(REPORT_EXPIRY_AGE_PLUS_ONE);
+    clock.advance(REPORT_EXPIRY_AGE);
 
     let (report_count, no_reports_task_report_count) = ds
         .run_unnamed_tx(|tx| {
@@ -1379,12 +1343,7 @@ async fn count_client_reports_for_interval(ephemeral_datastore: EphemeralDatasto
                     .count_client_reports_for_interval(
                         task.id(),
                         &Interval::new(
-                            interval_start
-                                .sub_timedelta(
-                                    &TimeDelta::seconds(TIME_PRECISION_SECONDS as i64),
-                                    &TIME_PRECISION,
-                                )
-                                .unwrap(),
+                            interval_start.sub_duration(&Duration::ONE).unwrap(),
                             Duration::from_time_precision_units(3),
                         )
                         .unwrap(),
@@ -1396,12 +1355,7 @@ async fn count_client_reports_for_interval(ephemeral_datastore: EphemeralDatasto
                     .count_client_reports_for_interval(
                         no_reports_task.id(),
                         &Interval::new(
-                            interval_start
-                                .sub_timedelta(
-                                    &TimeDelta::seconds(TIME_PRECISION_SECONDS as i64),
-                                    &TIME_PRECISION,
-                                )
-                                .unwrap(),
+                            interval_start.sub_duration(&Duration::ONE).unwrap(),
                             Duration::from_time_precision_units(3),
                         )
                         .unwrap(),
@@ -1464,17 +1418,12 @@ async fn count_client_reports_for_batch_id(ephemeral_datastore: EphemeralDatasto
                 let batch_id = random();
                 let expired_report = LeaderStoredReport::new_dummy(
                     *task.id(),
-                    OLDEST_ALLOWED_REPORT_TIME
-                        .sub_duration(&Duration::ONE)
-                        .unwrap(),
+                    START_TIME.sub_duration(&Duration::ONE).unwrap(),
                 );
-                let report_0 =
-                    LeaderStoredReport::new_dummy(*task.id(), OLDEST_ALLOWED_REPORT_TIME);
+                let report_0 = LeaderStoredReport::new_dummy(*task.id(), START_TIME);
                 let report_1 = LeaderStoredReport::new_dummy(
                     *task.id(),
-                    OLDEST_ALLOWED_REPORT_TIME
-                        .add_duration(&Duration::ONE)
-                        .unwrap(),
+                    START_TIME.add_duration(&Duration::ONE).unwrap(),
                 );
 
                 let expired_aggregation_job = AggregationJob::<0, LeaderSelected, dummy::Vdaf>::new(
@@ -1482,12 +1431,7 @@ async fn count_client_reports_for_batch_id(ephemeral_datastore: EphemeralDatasto
                     random(),
                     dummy::AggregationParam(22),
                     batch_id,
-                    Interval::minimal(
-                        OLDEST_ALLOWED_REPORT_TIME
-                            .sub_duration(&Duration::ONE)
-                            .unwrap(),
-                    )
-                    .unwrap(),
+                    Interval::minimal(START_TIME.sub_duration(&Duration::ONE).unwrap()).unwrap(),
                     AggregationJobState::Active,
                     AggregationJobStep::from(0),
                 );
@@ -1499,7 +1443,7 @@ async fn count_client_reports_for_batch_id(ephemeral_datastore: EphemeralDatasto
                     random(),
                     dummy::AggregationParam(22),
                     batch_id,
-                    Interval::minimal(OLDEST_ALLOWED_REPORT_TIME).unwrap(),
+                    Interval::minimal(START_TIME).unwrap(),
                     AggregationJobState::Active,
                     AggregationJobStep::from(0),
                 );
@@ -1754,7 +1698,7 @@ async fn roundtrip_aggregation_job(ephemeral_datastore: EphemeralDatastore) {
         random(),
         dummy::AggregationParam(23),
         batch_id,
-        Interval::minimal(OLDEST_ALLOWED_REPORT_TIME).unwrap(),
+        Interval::minimal(START_TIME).unwrap(),
         AggregationJobState::Active,
         AggregationJobStep::from(0),
     );
@@ -1763,7 +1707,7 @@ async fn roundtrip_aggregation_job(ephemeral_datastore: EphemeralDatastore) {
         random(),
         dummy::AggregationParam(23),
         random(),
-        Interval::minimal(OLDEST_ALLOWED_REPORT_TIME).unwrap(),
+        Interval::minimal(START_TIME).unwrap(),
         AggregationJobState::Active,
         AggregationJobStep::from(0),
     );
@@ -1849,9 +1793,7 @@ async fn roundtrip_aggregation_job(ephemeral_datastore: EphemeralDatastore) {
             tx.check_timestamp_columns_at_create_time(
                 "aggregation_jobs",
                 "test-update-aggregation-jobs",
-                OLDEST_ALLOWED_REPORT_TIME
-                    .as_naive_date_time(&TIME_PRECISION)
-                    .unwrap(),
+                START_TIME.as_naive_date_time(&TIME_PRECISION).unwrap(),
                 true,
             )
             .await;
@@ -1992,7 +1934,10 @@ async fn aggregation_job_acquire_release(ephemeral_datastore: EphemeralDatastore
     // Setup: insert a few aggregation jobs.
     install_test_trace_subscriber();
 
-    const LEASE_DURATION: StdDuration = StdDuration::from_secs(300);
+    let lease_duration_sec = 300;
+    let lease_duration_std = StdDuration::from_secs(lease_duration_sec);
+    let lease_duration = Duration::from_seconds(lease_duration_sec, &TIME_PRECISION);
+
     let clock = MockClock::new(START_TIMESTAMP);
     let ds = Arc::new(ephemeral_datastore.datastore(clock.clone()).await);
 
@@ -2054,20 +1999,10 @@ async fn aggregation_job_acquire_release(ephemeral_datastore: EphemeralDatastore
                             (),
                             (),
                             Interval::minimal(
-                                OLDEST_ALLOWED_REPORT_TIME
-                                    .add_timedelta(
-                                        &TimeDelta::seconds(
-                                            LEASE_DURATION.as_secs().try_into().unwrap(),
-                                        ),
-                                        &TIME_PRECISION,
-                                    )
+                                START_TIME
+                                    .add_duration(&lease_duration)
                                     .unwrap()
-                                    .add_timedelta(
-                                        &TimeDelta::seconds(
-                                            LEASE_DURATION.as_secs().try_into().unwrap(),
-                                        ),
-                                        &TIME_PRECISION,
-                                    )
+                                    .add_duration(&lease_duration)
                                     .unwrap(),
                             )
                             .unwrap(),
@@ -2091,7 +2026,7 @@ async fn aggregation_job_acquire_release(ephemeral_datastore: EphemeralDatastore
                     finished_aggregation_job_id,
                     (),
                     (),
-                    Interval::minimal(OLDEST_ALLOWED_REPORT_TIME).unwrap(),
+                    Interval::minimal(START_TIME).unwrap(),
                     AggregationJobState::Finished,
                     AggregationJobStep::from(1),
                 ))
@@ -2246,7 +2181,7 @@ async fn aggregation_job_acquire_release(ephemeral_datastore: EphemeralDatastore
     }
 
     let want_expiry_time =
-        clock.now().naive_utc() + chrono::Duration::from_std(LEASE_DURATION).unwrap();
+        clock.now().naive_utc() + chrono::Duration::from_std(lease_duration_std).unwrap();
     let want_aggregation_jobs: Vec<_> = task_and_aggregation_job_ids
         .iter()
         .map(|(task_id, aggregation_job_id)| {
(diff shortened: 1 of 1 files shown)


integration_tests/src/janus.rs
```diff
@@ -189,7 +189,7 @@ impl JanusInProcess {
             min_aggregation_job_size: 1,
             max_aggregation_job_size: 100,
             aggregation_job_creation_report_window: 5000,
-            late_report_grace_period_s: 3600,
+            late_report_grace_period_units: 10,
```
tgeoghegan: Why 10? Though I don't think the 3600 we had previously was particularly carefully considered.

aggregator_api/src/routes.rs
tgeoghegan: This has some interesting implications for the aggregator API and especially interactions with divviup-api. At a minimum, we'll need to teach divviup-api to handle 0.7 Januses differently than 0.9 ones, so I think that means we'll want to add a feature to 0.9 aggregators like `TimePrecision` (in `fn get_config`, `aggregator_api/src/routes.rs`). Then we'll have to decide how this works in divviup-api task provisioning: do we still let users express task expiration, etc., in "real" times and convert to time precision multiples in divviup-api, or do we make users of divviup-api (e.g. Mozilla) do the math?

I filed https://github.com/divviup/divviup-api/issues/2034 for some of these questions. For now we should at least add an aggregator feature, but maybe we should make the aggregator API deal in old-style times and durations?

integration_tests/tests/integration/simulation/bad_client.rs
```diff
@@ -60,9 +60,7 @@ pub(super) async fn upload_replay_report(
         &measurement,
         report_id.as_ref(),
     )?;
-    let rounded_time = report_time
-        .to_batch_interval_start(task.time_precision())
-        .unwrap();
+    let rounded_time = *report_time;
```
tgeoghegan: Delete this line and pass `*report_time` to `prepare_report` below

messages/src/lib.rs
```diff
@@ -299,31 +391,25 @@ pub struct Interval {
...
-        let duration = Duration::from_seconds(time_precision.as_seconds());
-        start
-            .0
-            .checked_add(time_precision.as_seconds())
-            .ok_or(Error::IllegalTimeArithmetic("duration overflows time"))?;
-
-        Ok(Self { start, duration })
+    /// This is the preferred constructor for batch bucket intervals.
+    /// For intervals with arbitrary durations, use [`Interval::new`].
+    pub fn single(start: Time) -> Result<Self, Error> {
```
tgeoghegan: I don't love this method name because my first thought is that it refers to the fact that the method returns a single interval and not a list, but so does `::new`. How about `minimal`, which I think better describes what sets this apart from other constructors?

messages/src/lib.rs
```diff
@@ -140,39 +138,93 @@ impl TryFrom<&Url> for url::Url {
...
+    ///
+    /// The duration will be rounded down to the nearest multiple of time_precision.
+    ///
+    /// # Arguments
+    /// * `seconds` - Duration in seconds
+    /// * `time_precision` - The task's time precision
+    ///
+    /// # Panics
+    /// Panics if `time_precision.as_seconds()` is 0.
+    ///
```
tgeoghegan: nit: this is the style we use elsewhere. 
```suggestion
    /// # Arguments
    ///
    /// * `seconds` - Duration in seconds
    /// * `time_precision` - The task's time precision
    ///
    /// # Panics
    ///
    /// Panics if `time_precision.as_seconds()` is 0.
```

messages/src/lib.rs
```diff
@@ -140,39 +138,93 @@ impl TryFrom<&Url> for url::Url {
     }
 }
 
-/// DAP protocol message representing a duration with a resolution of seconds.
+/// DAP protocol message representing a duration as a multiple of the task's time precision.
```
tgeoghegan: This isn't quite right. If we have a time precision 5, then 0, 5, 10, 15, ... are multiples, but we'd represent the time as 0, 1, 2, 3.

messages/src/lib.rs
```diff
@@ -216,25 +276,83 @@ impl Decode for Duration {
 
 impl Display for Duration {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{} seconds", self.0)
+        write!(f, "{} time_precision units", self.0)
     }
 }
 
-/// DAP protocol message representing an instant in time with a resolution of seconds.
+/// DAP protocol message representing an instant in time as a multiple of the task's time
```
tgeoghegan: As with `Duration`, it's not quite right to call this a multiple.

messages/src/lib.rs
```diff
@@ -216,25 +276,83 @@ impl Decode for Duration {
...
+    /// The time will be rounded down to the nearest multiple of time_precision.
+    ///
+    /// # Arguments
+    /// * `time` - SystemTime
+    /// * `time_precision` - The task's time precision
+    ///
+    /// # Panics
+    /// Panics if `time_precision.as_seconds()` is 0, or `time` is before the UNIX Epoch.
+    ///
+    pub fn from_systemtime(time: SystemTime, time_precision: &TimePrecision) -> Self {
```
tgeoghegan: I'm not sure we should have this method. `SystemTime` is just one of the many representations of time that a DAP user could be working with, and all we're really doing here is saving them the trouble of computing seconds since the epoch. It looks like the only place we use it is in an example in the client crate. I think we should avoid committing to this public API until we have a stronger case for maintaining it.

aggregator_core/src/batch_mode.rs
```diff
@@ -418,90 +412,3 @@ impl CollectableBatchMode for LeaderSelected {
...
-        batch_mode::CollectableBatchMode,
-        task::{AggregationMode, BatchMode, test_util::TaskBuilder},
-    };
-    use janus_core::vdaf::VdafInstance;
-    use janus_messages::{
-        Duration, Interval, Time, batch_mode::TimeInterval, taskprov::TimePrecision,
-    };
-
-    #[test]
-    fn validate_collect_identifier() {
```
tgeoghegan: Should we still test that collection intervals smaller than the task's time precision get rejected?

aggregator_core/src/batch_mode.rs
```diff
@@ -346,10 +339,11 @@ impl TimeIntervalBatchIdentifierIter {
         // Sanity check that the given interval is of an appropriate length. We use an assert as
         // this is expected to be checked before this method is used.
         assert_eq!(
-            batch_interval.duration().as_seconds() % time_precision.as_seconds(),
+            batch_interval.duration().as_seconds(time_precision) % time_precision.as_seconds(),
             0
```
tgeoghegan: `batch_interval.duration() == Duration::ZERO`? Or we could define a method `Duration::is_zero` that does the same thing.

aggregator/src/aggregator.rs
```diff
@@ -1838,19 +1837,21 @@ impl VdafOps {
...
         let report_deadline = now
-            .add_duration(task.tolerable_clock_skew())
+            .add_duration(task.tolerable_clock_skew(), task.time_precision())
             .map_err(|err| Arc::new(Error::from(err)))?;
 
         if let Ok(clock_skew) = report
             .metadata()
             .time()
-            .difference_as_time_delta(&now.to_time())
+            .difference_as_time_delta(&now.to_time(task.time_precision()), task.time_precision())
```
tgeoghegan: This looks weird, like we're converting in and out of the time precision domain unnecessarily. Could we just subtract `metadata().time().0 - now.to_time().0` (or some equivalent that doesn't require destructuring the tuple)?

aggregator/src/aggregator.rs
```diff
@@ -3586,7 +3576,11 @@ impl VdafOps {
             {
                 let aggregate_share_expiry_time =
                     batch_interval.end().add_duration(report_expiry_age)?;
-                if clock.now().is_after(&aggregate_share_expiry_time) {
+                if clock
+                    .now()
+                    .to_time(task.time_precision())
+                    .is_after(&aggregate_share_expiry_time)
```
tgeoghegan: I wonder if we could make this (and several other places where we compare times) more ergonomic by defining `is_after`/`is_before` methods on `TimeExt` that deal with `DateTime<Utc>`.

aggregator/src/aggregator/batch_creator.rs
```diff
@@ -113,17 +109,18 @@ where
...
+        let time_bucket_start_opt =
+            self.properties
+                .task_batch_time_window_size
+                .map(|batch_time_window_size| {
+                    // While everything is in units of the time precision, we still have to
+                    // bucket things by the batch_time_window_size.
+                    let batch_window_units = batch_time_window_size.as_time_precision_units();
+                    Time::from_time_precision_units(
+                        (report.client_timestamp().as_time_precision_units() / batch_window_units)
+                            * batch_window_units,
```
tgeoghegan: surely `/ batch_window_units * batch_window_units` cancels out? Then `Time::from_time_precision_units(report.client_timestamp().as_time_precision_units()) is just `report.client_timestamp()`?

jcjones: Yeah, that's the issue exactly, the possibility for them being a factor of each other. I'm going to rework that in https://github.com/divviup/janus/issues/4225.

messages/src/lib.rs
```diff
@@ -140,39 +138,93 @@ impl TryFrom<&Url> for url::Url {
...
+    ///
+    /// # Panics
+    /// Panics if `time_precision.as_seconds()` is 0.
+    ///
+    pub fn as_seconds(&self, time_precision: &TimePrecision) -> u64 {
+        self.0.saturating_mul(time_precision.as_seconds())
+    }
+
+    /// Construct a [`Duration`] from the raw number of time_precision units.
+    /// This is primarily for testing and internal use.
```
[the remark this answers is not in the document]
jcjones: hmmm, I removed the comments on these two instead, because while I intended them for testing/internal use, I use them even inside things like `TimeExt` for `add_timedelta` which is important all over.

aggregator_core/src/task.rs
```diff
@@ -275,7 +275,7 @@ impl AggregatorTask {
                 return Err(Error::InvalidParameter(
                     "batch_time_window_size is not supported for taskprov",
                 ));
-            } else if batch_time_window_size.as_seconds()
+            } else if batch_time_window_size.as_seconds(&common_parameters.time_precision)
                 % common_parameters.time_precision.as_seconds()
```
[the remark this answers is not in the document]
jcjones: I can't explain how I got here, but I assure you it was multiple steps that made sense at the time. Fixed.

messages/src/lib.rs
```diff
@@ -140,39 +138,93 @@ impl TryFrom<&Url> for url::Url {
...
 
 impl Duration {
-    pub const ZERO: Duration = Duration::from_seconds(0);
+    pub const ZERO: Duration = Duration::from_time_precision_units(0);
+    pub const ONE: Duration = Duration::from_time_precision_units(1);
 
-    /// Create a duration representing the provided number of seconds.
-    pub const fn from_seconds(seconds: u64) -> Self {
-        Self(seconds)
+    /// Create a duration representing the provided number of seconds, given the task's time precision.
```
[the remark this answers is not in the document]
jcjones: I didn't know that was our style. Will do.

messages/src/lib.rs
```diff
@@ -261,32 +379,6 @@ impl Decode for Time {
...
-
-impl PartialOrd<DateTime<Utc>> for Time {
-    fn partial_cmp(&self, other: &DateTime<Utc>) -> Option<std::cmp::Ordering> {
-        let other_timestamp =
-            u64::try_from(other.timestamp()).expect("timestamps must be non-negative");
-        self.as_seconds_since_epoch().partial_cmp(&other_timestamp)
-    }
-}
-
 /// DAP protocol message representing a half-open interval of time with a resolution of seconds;
```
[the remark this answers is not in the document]
jcjones: Thanks! Went for:
```
/// DAP protocol message representing a half-open interval of time in terms of the task's time
//  precision. The start of the interval is included while the end of the interval is excluded.
```

messages/src/lib.rs
jcjones: Good idea. Removed.

aggregator_core/src/batch_mode.rs
jcjones: This whole bit is wack. The assert goes away, and this becomes:
```
impl TimeIntervalBatchIdentifierIter {
    fn new(time_precision: &TimePrecision, batch_interval: &Interval) -> Self {
        Self {
            step: 0,
            total_step_count: batch_interval.duration().as_time_precision_units(),
            start: *batch_interval.start(),
            time_precision: *time_precision,
        }
    }
}
```

jcjones: in dd359ca08df24fbbfabe59bb509f892e646aab53

messages/src/lib.rs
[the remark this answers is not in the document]
jcjones: in dd359ca08df24fbbfabe59bb509f892e646aab53

messages/src/lib.rs
jcjones: in dd359ca08df24fbbfabe59bb509f892e646aab53

aggregator_core/src/batch_mode.rs
```diff
@@ -71,15 +73,11 @@ pub trait AccumulableBatchMode: BatchMode {
 #[async_trait]
 impl AccumulableBatchMode for TimeInterval {
     fn to_batch_identifier(
-        task: &AggregatorTask,
```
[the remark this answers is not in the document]
jcjones: in dd359ca08df24fbbfabe59bb509f892e646aab53

integration_tests/src/janus.rs
jcjones: in dd359ca08df24fbbfabe59bb509f892e646aab53

aggregator/src/aggregator/upload_tests.rs
```diff
@@ -758,9 +704,12 @@ async fn upload_report_report_expired() {
...
 
     // Advance the clock to expire the report.
-    clock.advance(TimeDelta::seconds(61));
+    // We need to advance past the expiry age (60s) plus enough to move to the next time precision
+    // bucket (100s precision means we need >100s total time).
+    clock.advance(TimeDelta::seconds(200));
 
-    // Try to upload the report, verify that we get the expected error.
+    // Try to upload the report. For expired reports, the upload may succeed but the report
+    // won't be stored (it's dropped asynchronously during batch write).
```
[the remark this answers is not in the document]
jcjones: me fail english? that's unpossible!

I meant to say the upload OUGHT TO pass and the upload status SHOULD CONSIDER containing a ReportDropped, compliant with RFC 6919.

aggregator/src/aggregator/upload_tests.rs
```diff
@@ -758,9 +704,12 @@ async fn upload_report_report_expired() {
     );
 
     // Advance the clock to expire the report.
-    clock.advance(TimeDelta::seconds(61));
+    // We need to advance past the expiry age (60s) plus enough to move to the next time precision
+    // bucket (100s precision means we need >100s total time).
+    clock.advance(TimeDelta::seconds(200));
```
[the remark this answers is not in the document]
jcjones: in 79e201b8a4d35837d1bfb32ff838e3c2ada9de01

aggregator/src/aggregator/upload_tests.rs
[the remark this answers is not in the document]
jcjones: in 79e201b8a4d35837d1bfb32ff838e3c2ada9de01

aggregator_core/src/datastore/models.rs
```diff
@@ -2130,6 +2145,44 @@ impl SqlInterval {
     pub fn as_interval(&self) -> Interval {
         self.0
     }
+
+    /// Convert an interval from task time precision to SQL time precision.
+    pub fn from_task_interval(
```
tgeoghegan: `task_interval` doesn't sound right. `from_dap_time_interval`?

core/src/time.rs
```diff
@@ -361,47 +348,42 @@ impl DateTimeExt for DateTime<Utc> {
...
 
-    /// Add the provided TimePrecision to this time.
-    fn add_time_precision(&self, duration: &TimePrecision) -> Result<Self, Error>;
+    /// Add 1 time_precision unit to this time.
+    fn add_time_precision(&self) -> Result<Self, Error>;
 
-    /// Subtract the provided TimePrecision from this time.
-    fn sub_time_precision(&self, duration: &TimePrecision) -> Result<Self, Error>;
+    /// Subtract 1 time_precision unit from this time.
+    fn sub_time_precision(&self) -> Result<Self, Error>;
```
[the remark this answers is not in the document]
jcjones: Done in 50b25d04.

aggregator_core/src/datastore/models.rs
```diff
@@ -2229,10 +2299,18 @@ impl ToSql for SqlInterval {
...
+            empty_range_to_sql(out);
+            return Ok(postgres_types::IsNull::No);
+        }
+
+        // Note: The interval should already be in SQL_UNIT_TIME_PRECISION (1 second) units,
+        // and if it isn't, badness ensues. This will stop being an problem as part of
+        // Issue #4206.
+        let start_sql_usec = time_to_sql_timestamp(*self.0.start(), &SQL_UNIT_TIME_PRECISION)
+            .map_err(|_| "microsecond timestamp of Interval start overflowed")?;
+        let end_sql_usec = time_to_sql_timestamp(self.0.end(), &SQL_UNIT_TIME_PRECISION)
```
tgeoghegan: Hmm. I dislike that we have `start() -> &Time` `end() -> Time`, especially since `Time` is `Copy` and it's just a `u64`, so we shouldn't bother with references anyway. Ah well, that's not a new issue.

aggregator/src/aggregator.rs
jcjones: I did this a little differently than that, but see in 705d0622.

jcjones:
> More comments on the datastore stuff. Please re-request review when you're done addressing things and I will do another pass.

Will do, but at this point, I'm out! I've resolved everything that I've ... resolved with commits. The unresolved comments will be there for me when we return! See you in 2026!

aggregator/src/aggregator/taskprov_tests.rs
```diff
@@ -180,17 +180,17 @@ where
...
+                .to_time(&time_precision)
+                .add_duration(&task_duration)
                 .unwrap(),
         ))
         .with_report_expiry_age(peer_aggregator.report_expiry_age().copied())
         .with_min_batch_size(min_batch_size as u64)
-        .with_time_precision(TimePrecision::from_seconds(1))
-        .with_tolerable_clock_skew(Duration::from_seconds(1))
+        .with_time_precision(time_precision)
+        .with_tolerable_clock_skew(Duration::from_seconds(1, &time_precision))
```
divergentdave: FYI this will round to zero, but I don't think it's important for these tests.

integration_tests/tests/integration/simulation/arbitrary.rs
```diff
@@ -264,24 +276,27 @@ fn arbitrary_report_time(g: &mut Gen, context: &Context) -> Time {
...
 
     let duration_fn = g
         .choose(&[
-            (|_g: &mut Gen, context: &Context| -> TimePrecision { context.time_precision })
-                as fn(&mut Gen, &Context) -> TimePrecision,
-            (|g: &mut Gen, context: &Context| -> TimePrecision {
-                TimePrecision::from_seconds(
-                    context.time_precision.as_seconds() * (1 + u64::from(u8::arbitrary(g) & 0x1f)),
+            (|_g: &mut Gen, context: &Context| -> Duration {
+                Duration::from_seconds(1, &context.time_precision)
```
divergentdave: I think this would get rounded down, which would lead to `Interval::new()` returning an error below. Returning one unit would better preserve the intent of this.
```suggestion
                Duration::from_time_precision_units(1)
```

messages/src/taskprov.rs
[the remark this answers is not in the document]
jcjones: In https://github.com/divviup/janus/pull/4219/changes/705d06227cb686a64d8a1d78ecb9848b5e306c35..1ce2c36b62e585804c1eabb295ad37607945bab4. (I missed the header divider in 705d06227cb686a64d8a1d78ecb9848b5e306c35, whoops)

aggregator_core/src/task.rs
```diff
@@ -1354,8 +1351,8 @@ pub mod test_util {
         pub fn build(self) -> Task {
             // If the tolerable clock skew is unset, copy the time_precision
             if *self.0.tolerable_clock_skew() == Duration::ZERO {
-                let time_precision = *self.0.time_precision();
-                return self.with_tolerable_clock_skew(time_precision.into()).0;
+                let _time_precision = *self.0.time_precision();
```
divergentdave: nit: this line is unnecessary now
```suggestion
```

aggregator/src/aggregator.rs
jcjones: There's no way to be that clean, but I got rid of the TimeDelta conversions in 8112241d, so it's clean_er_.

aggregator_core/src/datastore/models.rs
```diff
@@ -619,12 +624,20 @@ impl<T> Lease<T> {
     /// Returns how long remains until the expiry time, with an optional clock skew allowance.
     /// The math saturates, because we want to timeout immediately if any of these
     /// subtractions would underflow.
-    pub fn remaining_lease_duration(&self, current_time: &Time, skew_seconds: u64) -> StdDuration {
+    pub fn remaining_lease_duration(
+        &self,
+        current_time: &DateTime<Utc>,
```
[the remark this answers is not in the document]
jcjones: Since `Clock::now` returns `DateTime<Utc>`, this change pushes the conversion from inside the function to the callers which just moves the complexity around.

I think the bigger question is whether we should use `NaiveDateTime` from `Clock::now`, which I was going out of my way to avoid since then we have no timezone tracking.

We could change the database to use Utc though, but I don't want to add that here. I'll open an issue.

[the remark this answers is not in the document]
jcjones: https://github.com/divviup/janus/issues/4261

aggregator_core/src/datastore/models.rs
jcjones: I reworked those in 973bbf0605fb9ac260bc83a11ad8139917b2121f to utilize `Copy` and not return a borrow.

aggregator/src/aggregator/aggregation_job_init.rs
```diff
@@ -316,14 +316,14 @@ where
                             .report_share()
                             .metadata()
                             .time()
-                            .difference_as_time_delta(&now.to_time())
+                            .difference_as_time_delta(&now.to_time(task.time_precision()), task.time_precision())
```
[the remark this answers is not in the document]
jcjones: Fixed in e65f9c42ae33b712ae6162fd47023a59dae83808

aggregator/src/aggregator/taskprov_tests.rs
jcjones: Removed the 0-duration clock skew.

aggregator/src/aggregator.rs
```diff
@@ -1838,30 +1837,37 @@ impl VdafOps {
...
+            if skew_seconds > 0 {
+                // Early report: report time > now
+                metrics
+                    .early_report_clock_skew_histogram
+                    .record(skew_seconds as u64, &[]);
+            } else if skew_seconds < 0 {
+                // Past report: now > report time
+                metrics
+                    .past_report_clock_skew_histogram
+                    .record((skew_seconds) as u64, &[]);
```
divergentdave: This is missing a negation compared to the leader side
```suggestion
                    .record((-skew_seconds) as u64, &[]);
```

aggregator_core/src/batch_mode.rs
```diff
@@ -370,14 +352,14 @@ impl Iterator for TimeIntervalBatchIdentifierIter {
...
+        let interval = Interval::minimal(
             self.start
                 .add_timedelta(
                     &TimeDelta::try_seconds_unsigned(self.step * self.time_precision.as_seconds())
                         .unwrap(),
+                    &self.time_precision,
                 )
                 .unwrap(),
-            self.time_precision,
         )
```
divergentdave: This could be simplified to just get the number of time precision units from self.start, add self.step, and return `Interval::minimal(Time::from_time_precision_units(...))`.

tgeoghegan (approved):

aggregator_core/src/batch_mode.rs
jcjones: Whoops. I fixed a lot of these, but obviously missed this one. Thanks! This cleans up a bunch of boilerplate, too -- (+37, -83)!

aggregator_core/src/datastore.rs
```diff
@@ -872,17 +858,16 @@ UPDATE tasks SET task_end = $1, updated_at = $2, updated_by = $3
             .await?
             .ok_or(Error::MutationTargetNotFound)?;
 
-        if let Some(end) = task_end {
-            end.validate_precision(&task_info.time_precision)
-                .map_err(|e| Self::unaligned_time_error(task_id, &task_info.time_precision, e))?;
-        }
+        if let Some(_end) = task_end {}
```
divergentdave: nit: this can be cleaned up
```suggestion
```

aggregator_core/src/batch_mode.rs
jcjones: note that one is a bigger than usual update, worth re-checking it!

aggregator_core/src/datastore/tests.rs
```diff
@@ -499,20 +521,16 @@ async fn get_task_ids(ephemeral_datastore: EphemeralDatastore) {
 #[tokio::test]
 async fn roundtrip_report(ephemeral_datastore: EphemeralDatastore) {
```
divergentdave: I think the changes to this test are changing the semantics of it. The change in the baseline time from `MockClock::default()`'s 'UNIX epoch plus one billion seconds' is complicating things though. Note that previously, the clock started out at one fixed time, and the report expiry age was backed out from the difference from an earlier time constant, `OLDEST_ALLOWED_REPORT_TIMESTAMP`. (Though, the difference between epoch+1e9 and epoch+1e3 is pretty huge, and I understand that might lead to range issues in newly-added conversions) In the existing code, the first report was written with time `OLDEST_ALLOWED_REPORT_TIMESTAMP`, before `clock.now()`, whereas in the new code, the report time and the current time are equal. This means we're no longer checking some edge cases in the first half of this test. In particular, `OLDEST_ALLOWED_REPORT_TIMESTAMP` is no longer the oldest allowed report timestamp, since reports with that timestamp have an age of zero.

Near the end of the test, we advance the clock by `REPORT_EXPIRY_AGE_PLUS_ONE`, whereas we used to advance the clock by one second, in order to cause the report to become expired, and invisible to our queries. This change offsets the above difference-in-differences between the current time and the report time.

We could fix this by defining a separate `START_TIME` constant, starting the mock clock at that, setting `OLDEST_ALLOWED_REPORT_TIME` as `START_TIME - REPORT_EXPIRY_AGE`, using that in the report, and then changing the `clock.advance()` call to step by just one unit. Since the constants are also used elsewhere, this might have further implications in other tests, so we may want to use separate constants/variables, or make other test changes if they have similar issues related to the meaning of `OLDEST_ALLOWED_REPORT_TIMESTAMP`.

aggregator_core/src/datastore.rs
jcjones: lol

aggregator_core/src/datastore/tests.rs
jcjones: Okay, I reworked this in 926ec0fe:
    - Introduced your `START_TIME` constant to establish a proper baseline for clock initialization
    - Created `REPORT_TIME_AT_EXPIRY_EDGE` constant for reports at the edge of expiry
    - roundtrip_report now starts with a report that has `age = REPORT_EXPIRY_AGE`
    - Advancing clock by `ONE_UNIT` (100 seconds) properly expires the report, yay!

Doing this, I moved most tests to start at `START_TIME` instead, as it seems clearer, and changed a bunch of `Interval::minimal(Time::from_seconds_since_epoch(1100...` to be instead `START_TIMESTAMP + 1`.

aggregator_core/src/datastore/tests.rs
```diff
@@ -689,15 +723,18 @@ async fn get_unaggregated_client_reports_for_task(ephemeral_datastore: Ephemeral
 
     install_test_trace_subscriber();
 
-    let clock = MockClock::new(OLDEST_ALLOWED_REPORT_TIMESTAMP.as_seconds_since_epoch());
+    let clock = MockClock::new(START_TIMESTAMP);
     let ds = ephemeral_datastore.datastore(clock.clone()).await;
-    let report_interval = Interval::new(OLDEST_ALLOWED_REPORT_TIMESTAMP, TIME_PRECISION).unwrap();
+    let unexpired_time = OLDEST_ALLOWED_REPORT_TIME
+        .add_duration(&Duration::ONE)
```
divergentdave: I think we could get rid of this addition here (using just `START_TIMESTAMP` instead), and change `clock.advance(REPORT_EXPIRY_AGE_PLUS_ONE);` back to `clock.advance(REPORT_EXPIRY_AGE);` below, to keep this closer to the original test.

aggregator_core/src/datastore/tests.rs
```diff
@@ -3450,27 +3493,31 @@ async fn crypter() {
...
-    let first_batch_interval =
-        Interval::new_with_duration(OLDEST_ALLOWED_REPORT_TIMESTAMP, Duration::from_seconds(100))
-            .unwrap();
-    let second_batch_interval = Interval::new_with_duration(
-        OLDEST_ALLOWED_REPORT_TIMESTAMP
-            .add_timedelta(&TimeDelta::seconds(100))
+    let first_batch_interval = Interval::minimal(
+        OLDEST_ALLOWED_REPORT_TIME
+            .add_duration(&Duration::ONE)
+            .unwrap(),
```
divergentdave: We could anchor this at `START_TIME`, peel off one of the additions for `second_batch_interval`, and change the `clock_advance()` calls to use `REPORT_EXPIRY_AGE`.

aggregator_core/src/datastore/tests.rs
```diff
@@ -1405,16 +1464,16 @@ async fn count_client_reports_for_batch_id(ephemeral_datastore: EphemeralDatasto
                 let batch_id = random();
                 let expired_report = LeaderStoredReport::new_dummy(
                     *task.id(),
-                    OLDEST_ALLOWED_REPORT_TIMESTAMP
-                        .sub_timedelta(&TimeDelta::seconds(2))
+                    OLDEST_ALLOWED_REPORT_TIME
+                        .sub_duration(&Duration::ONE)
```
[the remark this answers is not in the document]
jcjones: I don't quite understand what you mean on this one, but I moved it to START_TIME

divergentdave (approved):

PR #4222, opened 2025-12-16
Migrate off of rustls-pemfile to rustls-pki-types directly

### LINKED ISSUE #4221: Migrate off of rustls-pemfile
RUSTSEC-2025-0134

Advisory: https://rustsec.org/advisories/RUSTSEC-2025-0134
The rustls-pemfile crate is no longer maintained. The repository has been archived since August
2025, and users are encouraged to depend directly on the underlying PEM parsing code included
in rustls-pki-types since 1.9.0. The latest version of rustls-pemfile is in fact a thin wrapper
around the same code used in rustls-pki-types, so migrating should be straightforward.

The new API is represented by the [`PemObject`][PemObject] trait, which provides methods for
reading a single or multiple PEM objects from a file or byte slice.

[PemObject]: https://docs.rs/rustls-pki-types/latest/rustls_pki_types/pem/trait.PemObject.html
Announcement: https://github.com/rustls/pemfile/issues/61

jcjones:
Advisory: https://rustsec.org/advisories/RUSTSEC-2025-0134

Resolves #4221

### DIFF
--- a/aggregator/src/binary_utils.rs
+++ b/aggregator/src/binary_utils.rs
@@ -27,7 +27,6 @@ use std::{
     fmt::{self, Debug, Formatter},
     fs,
     future::Future,
-    io,
     net::SocketAddr,
     panic,
     path::{Path, PathBuf},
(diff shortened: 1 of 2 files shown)


aggregator/src/binary_utils.rs
```diff
@@ -189,8 +190,15 @@ pub async fn datastore<C: Clock>(
 
 /// Loads a series of certificates from a PEM file into a rustls [`RootCertStore`].
 fn load_pem_trust_store(path: impl AsRef<Path>) -> Result<RootCertStore, io::Error> {
-    let mut buf_read = BufReader::new(File::open(path)?);
-    let der_certs = rustls_pemfile::certs(&mut buf_read).collect::<Result<Vec<_>, _>>()?;
+    let pem_data = fs::read(path)?;
+    let der_certs = CertificateDer::pem_slice_iter(&pem_data)
+        .collect::<Result<Vec<_>, _>>()
+        .map_err(|err| {
+            io::Error::new(
```
divergentdave: We could just change the signature of this function to return `rustls_pki_types::pem::Error` instead. We already apply an anyhow "context" string to the error at the sole callsite.

PR #4263, opened 2026-01-08
Wrap all comments to 100 characters. Add a lint for long comments.

### LINKED ISSUE #4262: Enforce 100-character wrap
In my editor I noticed that this line is just over 100 columns. Can you please go over the whole change and make sure everything wraps at 100?

_Originally posted by @tgeoghegan in https://github.com/divviup/janus/pull/4219#discussion_r2625146454_

`rustfmt` has `max_width` which does some of this but only for non-comment lines. There's also new settings in nightly that does everything we want.

But I can use ripgrep easily here too, and write a lint. Do that.

jcjones:
Also configures `rustfmt` to 100 character `max_width`, which only impacts code reflows, and isn't a hard limit. But here in this change, I'm only worried about comments.

Fix #4262

NOTE: This will conflict all over the place with #4219, which I totally know and understand, but _most_ of it _won't_, and I am gonna rebase it afterward. But I just had a little more steam left tonight and wanted to bang this out.

### DIFF
--- a/aggregator/src/aggregator/garbage_collector.rs
+++ b/aggregator/src/aggregator/garbage_collector.rs
@@ -267,6 +267,7 @@ mod tests {
                     .unwrap();
 
                     // Collection artifacts.
+                    //
                     // unrealistic, but induces GC
                     let batch_identifier = Interval::minimal(client_timestamp).unwrap();
                     tx.put_batch_aggregation(
@@ -461,6 +462,7 @@ mod tests {
                     .unwrap();
 
                     // Collection artifacts.
+                    //
                     // unrealistic, but induces GC
                     let batch_identifier = Interval::minimal(client_timestamp).unwrap();
                     tx.put_batch_aggregation(
--- a/aggregator/src/aggregator/http_handlers/tests/aggregation_job_init.rs
+++ b/aggregator/src/aggregator/http_handlers/tests/aggregation_job_init.rs
@@ -1,5 +1,5 @@
-#![allow(clippy::unit_arg)] /* allow reference to dummy::Vdaf's public share, which has the unit
-type */
+// allow reference to dummy::Vdaf's public share, which has the unit type
+#![allow(clippy::unit_arg)]
 
 use crate::aggregator::{
     BatchAggregationsIterator,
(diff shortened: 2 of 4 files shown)


.github/workflows/ci-build.yml
```diff
@@ -121,8 +121,18 @@ jobs:
...
         toolchain: ${{ matrix.rust-toolchain }}
         components: clippy, rustfmt
+    - name: Install ripgrep
+      run: sudo apt-get update && sudo apt-get install -y ripgrep
     - name: Format
       run: cargo fmt --message-format human -- --check
+    - name: Check comment line length
+      run: |
+        # Find comments exceeding 100 characters (excluding URLs)
+        if rg --line-number --max-columns 150 '^.{101,}$' --type rust | rg ' //' | grep -v "://"; then
```
tgeoghegan: will this catch `/* */` comments?

rustfmt.toml
```diff
@@ -0,0 +1,6 @@
+# Enforce consistent code formatting across the project
+max_width = 100
```
tgeoghegan: This causes no reformatting of existing code?

aggregator/src/aggregator/garbage_collector.rs
```diff
@@ -266,7 +267,8 @@ mod tests {
                     .unwrap();
 
                     // Collection artifacts.
-                    let batch_identifier = Interval::minimal(client_timestamp).unwrap(); // unrealistic, but induces GC
+                    // unrealistic, but induces GC
```
divergentdave: An extra blank line here would help
```suggestion
                    //
                    // unrealistic, but induces GC
```

aggregator/src/aggregator/http_handlers/tests/aggregation_job_init.rs
```diff
@@ -1,4 +1,5 @@
-#![allow(clippy::unit_arg)] // allow reference to dummy::Vdaf's public share, which has the unit type
+#![allow(clippy::unit_arg)] /* allow reference to dummy::Vdaf's public share, which has the unit
+type */
```
divergentdave: For consistency with the other places:
```suggestion
// allow reference to dummy::Vdaf's public share, which has the unit type
#![allow(clippy::unit_arg)]
```

Issue #4149, opened 2025-11-05
Implement Streaming Uploads


jcjones:
`impl TryFromConn for BodyBytes` in `http_handlers.rs` reads the whole body, despite [RecievedBody supporting streaming implementations](https://docs.rs/trillium-http/0.3.17/trillium_http/struct.ReceivedBody.html). 

We need to rework the upload reports step to support streaming the results.

Issue #4148, opened 2025-11-05
Support streaming Measurement uploads in the client


jcjones:
From https://github.com/divviup/janus/pull/3831#discussion_r2495021717:

> It'd be cool (future enhancement) if there was an upload interface that took something like a measurement source instead of an explicit slice, so that a caller could initiate an upload session, then have the client start uploading reports, but then the caller could continue to feed new measurements into the upload session. I think maybe just taking `Iterator<Item = V::Measurement>` could work, but I'm not sure if it'd be possible to have upload dequeue items from the iterator while the caller is also enqueuing.

PR #4272, opened 2026-01-15
Implement streaming report uploads

jcjones:
This change plumbs through asynchronous stream processing for the `Report` upload path, so that bytes incoming via Trillium's methods are decoded without buffering more than a few reports at a time, and are then processed as before, in whatever order, as fast as the executor can go. The resulting errors are collected and returned at the conclusion.

I've implemented an HTTP/1.1 test of this using raw chunked encoding.

Implements #4149.


aggregator/src/aggregator.rs
```diff
@@ -1767,41 +1757,83 @@ impl VdafOps {
...
+
+        loop {
+            // If stream is exhausted and no futures remain, we're done
+            if stream_eof && futures.is_empty() {
+                break;
+            }
+
+            select! {
+                // Poll for new reports from the stream (only if not eof)
+                stream_result = report_stream.next(), if !stream_eof => {
```
tgeoghegan: Does something bad happen if we call `report_stream.next()` when it's already yielded `None`? What I'm getting at is, can we avoid tracking `stream_eof`?

aggregator/src/aggregator.rs
```diff
@@ -1767,41 +1757,83 @@ impl VdafOps {
...
+                    match stream_result {
+                        Some(Ok(report)) => {
+                            // Got a report - wrap in Arc and move it into the async task
+                            let report_arc = Arc::new(report);
```
tgeoghegan: Why do we need to wrap the report in `Arc`? Can't we just move the value into `handle_uploaded_report`?

aggregator/src/aggregator.rs
```diff
@@ -1767,41 +1757,83 @@ impl VdafOps {
...
+                                    metrics,
+                                    task,
+                                    report_writer,
+                                    &report_arc,
+                                ).await
+                            });
+                        }
+                        Some(Err(e)) => {
+                            // Stream error (decode failure, client disconnect) - fail fast
+                            return Err(Arc::new(e));
```
tgeoghegan: If we get an error here, then we'll abort handling the entire stream of `Report`, right?

1. What if this happens partway through the stream and we've already called `Self::handle_uploaded_report` a bunch of times?  Those will still go through, right? I think that's OK, because report uploading is meant to be idempotent. So if the client tries again and re-sends the same reports, we'll gracefully ignore the ones previously successfully handled and resume handling the stream of reports. So this is fine, but it'd be good to have a test of idempotence that checks this.
1. `Self::handle_upload_report` does most of the per-report validation described in [4.4.2.2](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.4.2.2). I think the exception is decoding the `struct Report`, though, because evidently we do that before even entering this `select`. The question is: if an individual report gets rejected because it can't be decoded, do we abort the entire stream or do we just have an individual `ReportUploadStatus { error = invalid_message, id = id_of_the_report }` in the response? I think that the latter is what we want. So maybe we should lower the decoding of individual `Report` structures into `handle_upload_report`?

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -591,40 +596,78 @@ async fn hpke_config_cors_preflight(mut conn: Conn) -> Conn {
...
 
+/// Streams reports decoded from an async reader. This function reads the body in chunks and
+/// yields reports as soon as they're fully decoded. When a chunk boundary falls in the middle
+/// of a report, the incomplete bytes are buffered until the next chunk arrives.
+fn decode_reports_stream<R>(mut body: R) -> impl Stream<Item = Result<Report, Error>>
+where
+    R: AsyncRead + Unpin,
+{
+    async_stream::try_stream! {
+        const CHUNK_SIZE: usize = 64 * 1024;
```
tgeoghegan: Should we make this configurable? Must/should this chunk size match a chunk size configured on load balancers?

aggregator/src/aggregator.rs
divergentdave: FWIW the private type inside `async-stream-impl` does implement `FusedStream` already, but this isn't documented.

jcjones: Oooh, I think I see!

aggregator/src/aggregator.rs
divergentdave: Agreed, I think `Result<Result<Report, CodecError>, aggregator::Error>` would make sense to separate these concerns.

aggregator/src/aggregator.rs
```diff
@@ -1767,41 +1757,83 @@ impl VdafOps {
...
+                                }
+                            },
+                        }
                     }
                 },
-                Ok(_) => {
-                    // Report succeeded, no status entry needed
+
+                // If both arms are disabled (stream exhausted + no futures), yield to prevent
+                // busy loop. The check at the top of the loop will break us out.
```
divergentdave: FWIW the docs say that if all branches are disabled by their preconditions, and there is no `else` expression, then it panics. I think that the else expression is just evaluated, not evaluated and awaited, so we could get rid of the if block at the top, and break here instead of awaiting `yield_now()`.

aggregator/src/aggregator.rs
```diff
@@ -1767,41 +1757,83 @@ impl VdafOps {
...
+                            stream_eof = true;
+                        }
                     }
-                    _ => {
-                        // Non-rejection errors are fatal
-                        return Err(e);
+                },
+
+                // Poll for completed report processing
+                report_result = futures.next(), if !futures.is_empty() => {
```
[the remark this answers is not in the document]
jcjones: Thanks for this, I figured out the above `eof` thing due to your comment here! :) c4b0be1a

aggregator/src/aggregator.rs
jcjones: This was fun, but I think I got it in efa84bb4.

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -1171,3 +1171,99 @@ async fn upload_client_early_disconnect() {
...
+        .without_signals()
+        .with_stopper(stopper.clone())
+        .with_prebound_server(server)
+        .spawn(handler);
+
+    let mut client_socket = TcpStream::connect(local_addr).await.unwrap();
+    let request_line_and_headers = format!(
+        "POST /tasks/{task_id}/reports HTTP/1.1\r\n\
+        Content-Type: application/dap-upload-req\r\n\
+        Transfer-Encoding: chunked\r\n\r\n"
```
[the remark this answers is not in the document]
jcjones: We do already have a test that looks just like this above -- I think raw testing of this feature is not _unreasonable_. Getting it exactly right in a higher abstraction -- using `flush` commands, for example -- is notoriously tough, because often implementations are real careful about their buffering, as they get more complex.

Firefox has more than a few of these even with a robust HTTP lib just because streaming semantics are otherwise somewhat ... malleable.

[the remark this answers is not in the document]
jcjones: As a note, we will teach `janus_client` these ways in https://github.com/divviup/janus/issues/4148

aggregator/src/aggregator/http_handlers.rs
jcjones: This is a great idea. In 93697351.

aggregator/src/aggregator/http_handlers/tests/report.rs
```diff
@@ -1171,3 +1171,489 @@ async fn upload_client_early_disconnect() {
...
+            .write_all(chunk_length_line.as_bytes())
+            .await
+            .unwrap();
+        client_socket.write_all(&encoded).await.unwrap();
+        client_socket.write_all(b"\r\n").await.unwrap(); // Chunk terminator
+        client_socket.flush().await.unwrap();
+        if i % 10 == 9 {
+            info!(i, report_count, "sleep start");
+            sleep(StdDuration::from_millis(500)).await;
+        }
```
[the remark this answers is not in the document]
jcjones: Removed in a7e91e12

aggregator/src/aggregator.rs
```diff
@@ -1767,42 +1757,71 @@ impl VdafOps {
...
+
+                // Poll for completed report processing
+                Some(result) = futures.next() => {
+                    match result {
+                        Ok(_) => {
+                            // Report succeeded, no status entry needed
+                        }
+                        Err(e) => match e.borrow() {
+                            Error::ReportRejected(rejection) => {
+                                status.push(ReportUploadStatus::new(
```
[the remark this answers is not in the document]
jcjones: Good catch! 01af4b4

PR #4552, opened 2026-04-15
Resolve RUSTSEC-2026-0099

### LINKED ISSUE #3744: Remove workaround for `rustls-webpki` feature enablement once trillium-rs/trillium#694 lands
We put a workaround into `integration_tests/Cargo.toml` to force enabling feature `rustls-webpki/std` in #3733 (release/0.7) and #3734 (main). Once https://github.com/trillium-rs/trillium/pull/694 lands, we can take the new `trillium-rustls` release and remove the workaround.

jcjones:


divergentdave (approved):

tgeoghegan:
If I'm reading the history in #3744 right, we had enabled feature `std` to work around having duplicate copies of `rustls-webpki` in the tree. Can we go back to default features for this dep? Or maybe even remove the `rustls-webpki` dep altogether since we no longer pull in `trillium-rustls`?

jcjones:
@tgeoghegan I should have read the issue closer. It sure 'nuff works without it. I'll try this on the backport too.

PR #4593, opened 2026-05-07
test partially replayed aggregation job init

tgeoghegan:
The other day we did some analysis and concluded that it should be OK for aggregation job init requests to contain a mix of replayed and new reports. This commit adds a test to make sure of that.


tgeoghegan:
#4592 for the equivalent change on release/0.7.

jcjones (approved):

PR #4594, opened 2026-05-12
[DAP-17] Rename VDAF "Preparation" to "Verification"

jcjones:
> [!WARNING] 
> This renames all metrics, too, so we'll need to revisit all the dashboards. I believe that is the correct move here.

- Renames the database columns, but we're assuming we're starting from clean anyway for the DAP-16 changes.
- The public `AggregationJobResp.prepare_resps` rename to `verify_resps` is a public API break for the external Rust consumers of `janus_messages`.


tgeoghegan:
You'll want to throw label `allow-changed-migrations` on this and re-run CI (which happens on its own if you push new commits). Otherwise this is still a draft so I will wait before I review.

aggregator/src/metrics.rs
```diff
@@ -327,29 +327,29 @@ pub(crate) fn aggregate_step_failure_counter(meter: &Meter) -> Counter<u64> {
...
     for failure_type in [
-        "missing_prepare_message",
+        "missing_verify_message",
         "missing_leader_input_share",
         "missing_helper_input_share",
-        "prepare_init_failure",
-        "prepare_step_failure",
-        "prepare_message_failure",
+        "verify_init_failure",
+        "verify_step_failure",
```
divergentdave: This is an existing problem, but this is out of sync with `error.rs`.
```suggestion
        "verify_next_failure",
```

aggregator/src/metrics.rs
```diff
@@ -327,29 +327,29 @@ pub(crate) fn aggregate_step_failure_counter(meter: &Meter) -> Counter<u64> {
...
+        "helper_verify_share_decode_failure",
         "continue_mismatch",
         "accumulate_failure",
         "finish_mismatch",
         "helper_step_failure",
         "plaintext_input_share_decode_failure",
         "duplicate_extension",
         "missing_client_report",
-        "missing_prepare_message",
+        "missing_verify_message",
```
divergentdave: Another existing problem: we already have this at the beginning of the array.
```suggestion
```

aggregator/src/aggregator/error.rs
divergentdave:
```suggestion
            format!("{peer_role}_verify_share_decode_failure"),
```

divergentdave:
```suggestion
            format!("{peer_role}_verify_message_decode_failure"),
```

aggregator/src/metrics.rs
```diff
@@ -327,29 +327,29 @@ pub(crate) fn aggregate_step_failure_counter(meter: &Meter) -> Counter<u64> {
...
         "unknown_hpke_config_id",
         "decrypt_failure",
         "input_share_decode_failure",
         "input_share_aad_encode_failure",
         "public_share_decode_failure",
         "public_share_encode_failure",
-        "prepare_message_decode_failure",
-        "leader_prep_share_decode_failure",
-        "helper_prep_share_decode_failure",
+        "verify_message_decode_failure",
```
divergentdave: Another existing problem: this is out of step with `error.rs`, which templates in the aggregator role.
```suggestion
        "leader_verify_message_decode_failure",
        "helper_verify_message_decode_failure",
```

db/00000000000001_initial_schema.up.sql
```diff
@@ -322,20 +322,20 @@ CREATE TABLE report_aggregations(
...
+    verify_init   BYTEA,                  -- the verification initialization message received from the Leader (opaque DAP message)
     require_taskbind_extension  BOOLEAN,  -- is the taskprov extension required?
 
     -- Additional data for state HelperContinue & HelperContinueProcessing.
-    helper_prep_state  BYTEA,  -- the current VDAF prepare state (opaque VDAF message)
+    helper_verify_state  BYTEA,  -- the current VDAF verify state (opaque VDAF message)
 
     -- Additional data for state HelperContinueProcessing.
-    prepare_continue  BYTEA,  -- the preparation continuation message received from the Leader (opaque VDAF message)
+    verify_continue   BYTEA,  -- the verification continuation message received from the Leader (opaque DAP message)
```
tgeoghegan, line 338: nit: "opaque VDAF message" was correct, because `Vdaf.VerifyState` `.VerifierShare` and `.VerifierMessage` are defined by each VDAF and opaque to the DAP layer.

https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-19#section-5

aggregator/src/metrics.rs
```diff
@@ -327,29 +327,29 @@ pub(crate) fn aggregate_step_failure_counter(meter: &Meter) -> Counter<u64> {
     // Initialize counters with desired status labels. This causes Prometheus to see the first
     // non-zero value we record.
     for failure_type in [
-        "missing_prepare_message",
+        "missing_verify_message",
```
tgeoghegan: Do we have an issue anywhere for Divvi Up to adapt to new metric names in its Janus deployments?

tgeoghegan (approved):
This PR does what it threatens to do. 👍🏻

divergentdave (approved):

db/00000000000001_initial_schema.up.sql
jcjones: oh, hmm. Thanks!

aggregator/src/metrics.rs
jcjones: https://github.com/divviup/janus-ops/issues/3312

db/00000000000001_initial_schema.up.sql
jcjones: I see my confusion. OK, I'll double-check all these comments are accurate.

aggregator/src/metrics.rs
jcjones: [a44f80f](https://github.com/divviup/janus/pull/4594/commits/a44f80f2ab836ee2b55e0ca02b3fed2de1d2db4a) I reconciled the list and found these edits:

- Added leader_ping_pong_message_mismatch, helper_ping_pong_message_mismatch, which were missing.

- Removed missing_verify_message, missing_leader_input_share, missing_helper_input_share,
      missing_client_report, which were only used here.

aggregator/src/metrics.rs
jcjones: Thank you!

aggregator/src/metrics.rs
divergentdave: SGTM. `missing_prepare_message` went away in #1811. `missing_leader_input_share` and `missing_helper_input_share` went away in #709. `missing_client_report` went away in #2516. It looks like some of these removals fell out of changes to load more database fields in a single query, as opposed to executing multiple queries and doing software joins/null checks afterward with this sort of error handling.

PR #4602, opened 2026-05-20
Use constants for label values of counter metric

divergentdave:
This introduces constants for the values of the `type` label on `janus_step_failures`. This should make it easier to keep initialization and other code in sync during future changes, by inspecting references of each of these constants throughout the codebase. I also eliminated some runtime string formatting by adding the peer aggregator's role to a match block.


tgeoghegan (approved):

jcjones (approved):
I like it

PR #4624, opened 2026-06-05
[DAP-17/18] Move TaskConfig/VdafConfig from taskprov into base DAP as TaskConfiguration

jcjones:
DAP-17+ defines TaskConfiguration and VDAF config encodings in the base protocol rather than the taskprov extension. This PR:

- Moves TaskProv stuff from `messages/src/taskprov.rs` to  `messages/src/task.rs` with minor renames
- `min_batch_size` is now `u64` (was `u32`)
- VDAF `max_measurement`/`max_weight` fields are now `u64` (was `u32`)
- `task_start`/`task_duration` are removed as direct fields, replaced by a `task_interval` `TaskExtension`
- `batch_config` is now an explicit opaque field
- `TaskExtensionType` gains a `TaskInterval` variant
- `VdafConfig` and `TaskExtensionType` preserve unknown codepoints via `Unknown` variants, so a peer advertising a newer revision's VDAF type or task extension stays interoperable


messages/src/task.rs
```diff
@@ -0,0 +1,645 @@
+//! Task configuration data structures as defined in DAP-18[1].
+//! [1]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/18/
```
divergentdave, line 3: This needs a blank line for it to be recognized as a reference link, and not one continuous paragraph with two broken intra-doc links.
```suggestion
//! Task configuration data structures as defined in DAP-18[1].
//!
//! [1]: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/18/
```

core/src/vdaf.rs
jcjones: I wonder if this TODO should be resolved.

divergentdave: Yes, this comment was added when the related DAP changes were still under discussion, and we settled on a 64-bit field for max_measurement.

messages/src/task.rs
```diff
@@ -0,0 +1,646 @@
...
+        time_precision: TimePrecision,
+        min_batch_size: u64,
+        batch_mode: batch_mode::Code,
+        batch_config: Vec<u8>,
+        task_start: Time,
+        task_duration: Duration,
+        vdaf_config: VdafConfig,
+        mut extensions: Vec<TaskExtension>,
+    ) -> Result<Self, Error> {
+        Self::validate_extensions(&extensions)?;
```
jcjones:
```suggestion
```
This is a duplicate validation with what I do in `new` up on line 66, so no need to do it here.

messages/src/task.rs
```diff
@@ -0,0 +1,646 @@
...
+    time_precision: TimePrecision,
+    min_batch_size: u64,
+    batch_mode: batch_mode::Code,
+    batch_config: Vec<u8>,
+    vdaf_config: VdafConfig,
+    extensions: Vec<TaskExtension>,
+}
+
+impl TaskConfiguration {
+    /// Validates that extension types are in strictly increasing order (per DAP-17 spec).
```
jcjones:
```suggestion
    /// Validates that extension types are in strictly increasing order (per DAP-18 §4.2.2).
```

aggregator/src/aggregator/taskprov_tests.rs
```diff
@@ -406,7 +444,75 @@ async fn taskprov_aggregate_init() {
     );
     let got_task = got_task.unwrap();
     assert_eq!(test.task.taskprov_helper_view().unwrap(), got_task);
-    assert_eq!(got_task.taskprov_task_info(), Some(b"foobar".as_slice()));
+    assert_eq!(got_task.task_info(), Some(b"foobar".as_slice()));
+}
+
+/// A taskprov `TaskConfiguration` may omit the optional `task_interval` extension (DAP-18
```
tgeoghegan:
```suggestion
/// A `TaskConfiguration` may omit the optional `task_interval` extension (DAP-18
```

aggregator/src/aggregator.rs
```diff
@@ -861,9 +860,16 @@ impl<C: Clock> Aggregator<C> {
 
         let vdaf_verify_key = peer_aggregator.derive_vdaf_verify_key(task_id, &vdaf_instance);
 
-        let task_end = task_config
-            .task_start()
-            .add_duration(task_config.task_duration())?;
+        let (task_start, task_end) = match task_config.task_interval()? {
```
tgeoghegan: It's kinda funky that we transform an `Interval` into two `Time`s like this. It looks like we do this because `AggregatorTask::new` expects task start and end as separate values, but I would think it's no longer possible to have a task have a start but no end, and vice versa. Should we revisit the data model here so that these values are an `Interval` all the way through?

aggregator_core/src/task.rs
```diff
@@ -120,11 +120,10 @@ struct CommonTaskParameters {
     /// How much clock skew to allow between client and aggregator. Reports from
     /// farther than this duration into the future will be rejected.
     tolerable_clock_skew: Duration,
-    /// The `task_info` byte string from a Taskprov `TaskConfig` struct. This is only present for
-    /// tasks created via Taskprov.
-    ///
-    /// This field is used to distinguish tasks with otherwise equivalent DAP task parameters.
-    taskprov_task_info: Option<Vec<u8>>,
+    /// The `task_info` byte string from a `TaskConfiguration` struct. `None` for API-provisioned
+    /// tasks, which have no `TaskConfiguration`; used to distinguish tasks with otherwise
```
tgeoghegan, line 140: Separately from this PR, we should consider allowing API-provisioned tasks to set a task info string. That would require plumbing this up into divviup-api, so it's a minor undertaking, but I don't see why we wouldn't allow this.

messages/src/task.rs
```diff
@@ -0,0 +1,645 @@
...
+impl TaskConfiguration {
+    /// Validates that extension types are in strictly increasing order (per DAP-18 §4.2.2).
+    fn validate_extensions(extensions: &[TaskExtension]) -> Result<(), Error> {
+        for window in extensions.windows(2) {
+            if window[0].extension_type() >= window[1].extension_type() {
+                return Err(Error::InvalidParameter(
+                    "task extensions must be in strictly increasing order of extension_type",
+                ));
+            }
+        }
```
tgeoghegan: Would [`&[T]::is_sorted`](https://doc.rust-lang.org/std/primitive.slice.html#method.is_sorted) work?

messages/src/task.rs
```diff
@@ -0,0 +1,645 @@
...
+        time_precision: TimePrecision,
+        min_batch_size: u64,
+        batch_mode: batch_mode::Code,
+        batch_config: Vec<u8>,
+        vdaf_config: VdafConfig,
+        extensions: Vec<TaskExtension>,
+    ) -> Result<Self, Error> {
+        if task_info.is_empty() {
+            return Err(Error::InvalidParameter("task_info must not be empty"));
+        }
```
tgeoghegan, line 62: Is this right? We observe elsewhere that tasks provisioned via aggregator API could have empty task info. I just went and looked and sure enough, draft-ietf-ppm-dap-18 has `opaque task_info<1..2^8-1>;` which means empty task_info is illegal.

messages/src/task.rs
```diff
@@ -0,0 +1,645 @@
...
+    pub fn new_with_task_interval(
+        task_info: Vec<u8>,
+        leader_aggregator_endpoint: Url,
+        helper_aggregator_endpoint: Url,
+        time_precision: TimePrecision,
+        min_batch_size: u64,
+        batch_mode: batch_mode::Code,
+        batch_config: Vec<u8>,
+        task_start: Time,
+        task_duration: Duration,
```
tgeoghegan: I think this echoes a comment I left elsewhere, but why not have a single `Interval` parameter?

messages/src/task.rs
```diff
@@ -0,0 +1,645 @@
...
+}
+
+impl VdafConfig {
+    const RESERVED: u32 = 0x00000000;
+    const PRIO3_COUNT: u32 = 0x00000001;
+    const PRIO3_SUM: u32 = 0x00000002;
+    const PRIO3_SUM_VEC: u32 = 0x00000003;
+    const PRIO3_HISTOGRAM: u32 = 0x00000004;
+    const PRIO3_MULTIHOT_COUNT_VEC: u32 = 0x00000005;
+    const POPLAR1: u32 = 0x00000006;
```
tgeoghegan: I think we should consider dropping everything to do with Poplar1. We have no plans to deploy it in Janus, and in any case, that VDAF isn't supported by DAP because you'd have to do multiple collections against a single batch.

messages/src/task.rs
```diff
@@ -0,0 +1,645 @@
...
+    }
+}
+
+/// DAP message indicating the type of a task extension.
+///
+/// Equality, ordering, and hashing are all defined in terms of the underlying codepoint, so
+/// `Unknown(0x0001)` compares and hashes equal to `TaskInterval`.
+#[derive(Clone, Copy, Debug)]
+#[non_exhaustive]
+pub enum TaskExtensionType {
```
tgeoghegan, line 662: Elsewhere, like [`HpkeKemId`](https://github.com/divviup/janus/blob/230e2d2efc403ff52b2238a8c3c823e63788ba94/messages/src/lib.rs#L702), we used `#[repr(u16)]`. I think we should be consistent here.

messages/src/task.rs
```diff
@@ -0,0 +1,645 @@
...
+    }
+
+    fn to_u16(self) -> u16 {
+        match self {
+            Self::Reserved => Self::RESERVED,
+            Self::TaskInterval => Self::TASK_INTERVAL,
+            Self::Unknown(val) => val,
+        }
+    }
+}
```
tgeoghegan: Is this better as `std::convert::From` implementations?

messages/src/task.rs
```diff
@@ -0,0 +1,645 @@
...
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn new(
+        task_info: Vec<u8>,
+        leader_aggregator_endpoint: Url,
+        helper_aggregator_endpoint: Url,
+        time_precision: TimePrecision,
+        min_batch_size: u64,
+        batch_mode: batch_mode::Code,
+        batch_config: Vec<u8>,
```
divergentdave: Since `batch_config` is intended to model extra parameters that may be required by a future batch mode, it might be nicer to model these two as a single Rust enum (similar to `VdafConfig`). If we ever support a batch mode with a non-empty batch_config, we could decode the embedded opaque data at the same time we decode the whole `TaskConfiguration`. For the current state of the code base, this means we wouldn't have to sprinkle empty byte arrays into our initialization (which is a bit of a layering violation).

messages/src/task.rs
```diff
@@ -0,0 +1,645 @@
...
+            vdaf_config,
+            extensions,
+        })
+    }
+
+    /// Convenience constructor that creates a `task_interval` extension from the provided
+    /// `task_start` and `task_duration`, inserting it at the correct sorted position in the
+    /// given extensions.
+    #[allow(clippy::too_many_arguments)]
+    pub fn new_with_task_interval(
```
divergentdave: I think introducing a builder for `TaskConfiguration` may be a bit nicer for our test code. The two constructors we have now share a lot of their signatures.

messages/src/task.rs
jcjones: /shrug

messages/src/task.rs
jcjones: On second thought, I went with the [ExtensionType](https://github.com/divviup/janus/blob/230e2d2efc403ff52b2238a8c3c823e63788ba94/messages/src/lib.rs#L859) strategy here rather than the HpkeKemId one to be forward-compatibile -- `repr` will hard-fail decoding anything it doesn't reocgnize, but the task-extension TLV framing is supposed to let receivers round-trip extensions they don't understand, right? So I went with this pattern.

aggregator/src/aggregator.rs
jcjones: That's a pretty sizable refactor down into the database, but I agree with you. How about I take that on as a follow-up?

messages/src/task.rs
jcjones: I wanted to match what the database had. See above. I'm opening a follow-on:  https://github.com/divviup/janus/issues/4625

aggregator_core/src/task.rs
jcjones: https://github.com/divviup/janus/issues/4626

messages/src/task.rs
```diff
@@ -0,0 +1,635 @@
...
+    pub fn vdaf_config(&self) -> &VdafConfig {
+        &self.vdaf_config
+    }
+
+    pub fn extensions(&self) -> &[TaskExtension] {
+        &self.extensions
+    }
+
+    /// Searches extensions for a `task_interval` extension and decodes it.
+    pub fn task_interval(&self) -> Result<Option<Interval>, Error> {
```
divergentdave: Similarly, we may want to consider representing task extensions with an enum, and only storing the raw encoded form of an extension if we don't have support for the extension.

messages/src/lib.rs
```diff
@@ -52,6 +53,9 @@ pub enum Error {
     /// An invalid parameter was passed.
     #[error("{0}")]
     InvalidParameter(&'static str),
+    /// A codec error occurred while processing a message.
+    #[error("codec error: {0}")]
+    Codec(#[from] prio::codec::CodecError),
```
divergentdave: If we get rid of encoded byte arrays for inner fields of `TaskConfiguration`, we could potentially drop this error variant.

messages/src/task.rs
divergentdave: On `HpkeKemId` we're using the `num_enum` derive macros, and those seem to care about having a `#[repr(u_)]` macro on the enum. I think we can achieve forwards-compatibility either way, the difference is whether we write the match blocks or the macro does.

jcjones: I refactored this in 73b7c1dc499c3aa990eeed07a80de2b465d02605, alongside the other comment about making TaskExtension an enum.

messages/src/task.rs
jcjones: I like this version a lot better instead of the opaque vector. That's in 73b7c1dc499c3aa990eeed07a80de2b465d02605 too.

messages/src/task.rs
jcjones: I did this in 73b7c1dc499c3aa990eeed07a80de2b465d02605, and marked it as `test-util`. 

Honestly it's a much nicer interface, and I would be open to a follow-on refactor to use it instead of `new`.

messages/src/task.rs
jcjones: Yup, now `TaskExtension::TaskInterval`, hope that reflects your ideal!

messages/src/tests/task.rs
```diff
@@ -0,0 +1,1021 @@
...
+        TaskExtension::Unknown { extension_type, extension_data } => {
+            assert_eq!(*extension_type, TaskExtensionType::Unknown(0xBEEF));
+            assert_eq!(extension_data, &[0xDE, 0xAD, 0xBE, 0xEF]);
+        }
+    );
+}
+
+#[test]
+fn task_extension_type_equality_by_codepoint() {
+    use std::collections::HashSet;
```
tgeoghegan: move this `use` to the top of the file

messages/src/task.rs
tgeoghegan: We took this change in DAP (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/787), which will appear in draft-ietf-ppm-dap-19, so I think we can go straight to that here. However I think we'll want a check that taskprov tasks have a nonempty task_info, though I don't think that spec's authors have updated yet.

messages/src/task.rs
```diff
@@ -0,0 +1,714 @@
...
+    }
+}
+
+impl Encode for TaskExtensionType {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
+        u16::from(*self).encode(bytes)
+    }
+
+    fn encoded_len(&self) -> Option<usize> {
+        Some(2)
```
tgeoghegan, line 701: super nit: `std::mem::size_of(u16)` would be more future proof... against... the size of u16 changing... or if we end up on an architecture with bytes that aren't 8 bits... ok fine I'll shut up.

messages/src/task.rs
```diff
@@ -0,0 +1,714 @@
...
+    }
+}
+
+impl Encode for TaskExtension {
+    fn encode(&self, bytes: &mut Vec<u8>) -> Result<(), CodecError> {
+        self.extension_type().encode(bytes)?;
+        match self {
+            Self::TaskInterval(interval) => {
+                // The interval is a fixed 16 bytes; encode it as the length-prefixed payload.
+                let mut interval_bytes = Vec::new();
```
tgeoghegan, line 621: Hmm, it's kind of a shame to have to do this allocation. Not for this PR, but maybe we could add an affordance to `prio::codec` to encode with a length prefix.

messages/src/task.rs
```diff
@@ -0,0 +1,714 @@
...
+            }
+        };
+
+        if sub.position() as usize != vdaf_config_len {
+            return Err(CodecError::Other(
+                anyhow!(
+                    "vdaf_configuration has {} trailing bytes",
+                    vdaf_config_len - sub.position() as usize
+                )
+                .into(),
```
tgeoghegan: `CodecError::BytesLeftOver` seems appropriate here.

aggregator/src/aggregator.rs
tgeoghegan: #4625

tgeoghegan (approved):

messages/src/task.rs
jcjones: I'd rather leave this as-is for now, and remove it when we have the next DAP, just to reduce points of confusion.

messages/src/task.rs
jcjones: doh

messages/src/task.rs
jcjones: I don't think it's that hot a path.

tgeoghegan: I agree. I got obsessive about this kind of thing while working on zk-cred-longfellow and it's hard to stop being fussy about it.

divergentdave (approved):

PR #4648, opened 2026-06-23
[DAP-17] Require upload extensions in strictly increasing order

jcjones:
Per DAP-18 §4.4.3, a report's public and private extensions must each appear in strictly increasing order of extension_type, which also rules out duplicates within a list.

Private extensions are validated in PlaintextInputShare::decode, since the input share is decrypted and decoded per-report. Public extensions live in plaintext ReportMetadata inside each batched ReportShare, so a decode-time rejection would fail the whole AggregationJobInitReq; they are instead validated per-report by the aggregator alongside the existing duplicate/unrecognized-extension checks, surfaced via a new unsorted_extension metric label.

ExtensionType now derives equality, hashing, and ordering from its codepoint so ordering matches the wire value.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_init.rs
+++ b/aggregator/src/aggregator/aggregation_job_init.rs
@@ -16,7 +16,7 @@ use janus_core::{
 };
 use janus_messages::{
     ExtensionType, InputShareAad, PlaintextInputShare, ReportError, Role, VerifyResp,
-    VerifyStepResult,
+    VerifyStepResult, extensions_are_strictly_increasing,
 };
 use opentelemetry::{
     KeyValue,
@@ -41,6 +41,7 @@ use crate::{
         INPUT_SHARE_DECODE_FAILURE, MISSING_OR_MALFORMED_TASKBIND_EXTENSION,
         PLAINTEXT_INPUT_SHARE_DECODE_FAILURE, PUBLIC_SHARE_DECODE_FAILURE,
         UNEXPECTED_TASKBIND_EXTENSION, UNKNOWN_HPKE_CONFIG_ID, UNRECOGNIZED_EXTENSION,
+        UNSORTED_EXTENSION,
     },
 };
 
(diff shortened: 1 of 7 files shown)


aggregator/src/aggregator/aggregation_job_init.rs
```diff
@@ -247,6 +248,23 @@ where
                                     ReportError::InvalidMessage
                                 })?;
 
+                            // Public extensions must be in strictly increasing order of extension
+                            // type (DAP-18 §4.4.3); private extensions are validated when the
```
jcjones, line 252: I chose this over  4.5.3.4. Input Share Validation because it's specifically about extensions and links straight to the validation. But both are legit options, I can change these comments.

tgeoghegan (approved):

aggregator/src/aggregator/upload_tests.rs
```diff
@@ -1046,6 +1046,83 @@ async fn upload_report_duplicate_extensions() {
...
+        .run_unnamed_tx(|tx| {
+            let task_id = *task.id();
+            Box::pin(async move { TaskUploadCounter::load(tx, &task_id).await })
+        })
+        .await
+        .unwrap();
+    assert_eq!(
+        got_counters,
+        Some(TaskUploadCounter::new_with_values(
+            0, 1, 0, 0, 0, 0, 0, 0, 0, 0
```
divergentdave, line 1121: I confirmed that this test is rejecting the report for the right reason via checking the log output, but checking `TaskUploadCounter` could be a bit of a blunt instrument. Ideally we should also treat the reserved extension as an unsupported extension type, which could also bump the report decoding counter. (see #4671) Checking our OTel metrics could be more precise, though the harness setup code in this file doesn't use `InMemoryMetricInfrastructure` yet. Until we have support for two different extension types, there's no way to make a request that has only this one problem with it, but on the flip side, it's also impossible for order to matter until then too.

I think this test is probably fine as-is, but at least this journey led to me opening two issues.

divergentdave (approved):

aggregator/src/aggregator/upload_tests.rs
jcjones: I see https://github.com/divviup/janus/issues/4671, did you need me to file one for `InMemoryMetricInfrastructure` in upload_tests.rs?

divergentdave: No, we can do that as needed. The second issue I was referring to was on the DAP spec.

PR #4666, opened 2026-06-24
[DAP-18] Add verification key ID field to AggregationJobInitReq

jcjones:
DAP-18 adds a leading `uint8 verification_key_id` field to `AggregationJobInitReq`, identifying which prearranged VDAF verification key the leader nominated.

- Add `verification_key_id: u8` to `AggregationJobInitializeReq`, encoded and decoded first per the new wire order
- The leader always nominates key 0, since Janus uses a single verification key per task
- The helper decodes and ignores the field for now

The `extensions` field and strictly-increasing-order rules from the same draft revision are [tracked separately](https://github.com/divviup/janus/issues/4402).

### DIFF
--- a/messages/src/lib.rs
+++ b/messages/src/lib.rs
@@ -2351,6 +2351,7 @@ impl Distribution<AggregationJobId> for StandardUniform {
 #[derive(Clone, Educe, PartialEq, Eq)]
 #[educe(Debug)]
 pub struct AggregationJobInitializeReq<B: BatchMode> {
+    verification_key_id: u8,
     #[educe(Debug(ignore))]
     aggregation_parameter: Vec<u8>,
     partial_batch_selector: PartialBatchSelector<B>,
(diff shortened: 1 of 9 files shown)


divergentdave (approved):
Regarding the title and description: `verification_key_id` is actually added in draft-ietf-ppm-dap-18. ([diff](https://author-tools.ietf.org/iddiff?url2=draft-ietf-ppm-dap-18))

jcjones:
> Regarding the title and description: `verification_key_id` is actually added in draft-ietf-ppm-dap-18. ([diff](https://author-tools.ietf.org/iddiff?url2=draft-ietf-ppm-dap-18))

uhh oops

jcjones:
I guess https://github.com/divviup/janus/issues/4402 _does_ say "17+".  All these NC-17 commits...

messages/src/lib.rs
```diff
@@ -2360,17 +2361,24 @@ pub struct AggregationJobInitializeReq<B: BatchMode> {
 impl<B: BatchMode> AggregationJobInitializeReq<B> {
     /// Constructs an aggregate initialization request from its components.
     pub fn new(
+        verification_key_id: u8,
```
tgeoghegan, line 2364: We don't expect verification key ID to be used until a future draft defines an extension and some semantics for it. In the meantime, I wonder if instead of adding `0` to 8,000 `AggregationJobInitializeReq::new` calls, we could just always set the key ID to 0 here in `new`. We'd still be able to parse non-zero values off the wire, and could change the constructor once verification key ID becomes real.

But then again you've already fixed all the callsites so this is fine as is.

tgeoghegan (approved):

PR #4668, opened 2026-06-25
Use the new Github Actions Background/Parallel functions

jcjones:
From https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -80,7 +80,7 @@ jobs:
       run: cargo test --profile ci --locked --all-targets --features=prometheus --no-run
     - name: Test
       id: test
-      wait: kind # ensure the backgrounded `kind` install is on PATH before the k8s tests
+      wait: [kind] # ensure the backgrounded `kind` install is on PATH before the k8s tests
       env:
         RUST_LOG: info
         JANUS_E2E_LOGS_PATH: ${{ github.workspace }}/test-logs
@@ -152,7 +152,7 @@ jobs:
       background: true
       run: cargo +nightly fmt --message-format human -- --check
     - name: Check Cargo.toml sorting
-      wait: install-sort
+      wait: [install-sort]
       run: cargo sort --workspace --check
     # Clippy, Clippy (all features), and Document all share the cargo build
     # directory and serialize on its lock, so keep them sequential.
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -47,6 +47,12 @@ jobs:
       # We should always make sure that the `kind` CLI we install is from the
       # same release as the node image version used by
       # `janus_core::test_util::kubernetes::EphemeralCluster`
+      #
+      # kind isn't needed until the Test step and is unrelated to the cargo
+      # build directory, so install it in the background while the Rust
+      # toolchain installs and the early cargo builds run.
+      id: kind
+      background: true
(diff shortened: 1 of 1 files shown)


jcjones:
This seems to line up with https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsparallel but errored during parsing; must not be fully rolled-out yet. Will check it again later.

.github/workflows/ci-build.yml
```diff
@@ -125,22 +145,26 @@ jobs:
...
       run: cargo clippy --profile ci --workspace --all-targets --all-features
     - name: Document
       run: cargo doc --profile ci --workspace --all-features
-    - name: cargo-deny
-      uses: EmbarkStudios/cargo-deny-action@v2.0.20
-      with:
-        command: check bans licenses sources -A unmatched-organization
+    # Surface failures from the backgrounded Format and cargo-deny steps.
+    - name: Collect background lints
+      wait-all: true
```
divergentdave: I don't think this is necessary, the docs say "An implicit `wait-all` runs before any post-job cleanup."

divergentdave (approved):

.github/workflows/ci-build.yml
```diff
@@ -74,6 +80,7 @@ jobs:
       run: cargo test --profile ci --locked --all-targets --features=prometheus --no-run
     - name: Test
       id: test
+      wait: kind # ensure the backgrounded `kind` install is on PATH before the k8s tests
```
divergentdave: Maybe the issue is that `wait` expects an array?
```suggestion
      wait: [kind] # ensure the backgrounded `kind` install is on PATH before the k8s tests
```

.github/workflows/ci-build.yml
```diff
@@ -125,22 +145,26 @@ jobs:
...
+    # `cargo fmt` only parses sources and doesn't take the build-directory lock,
+    # so run it in the background alongside the clippy compile.
     - name: Format
+      id: fmt
+      background: true
       run: cargo +nightly fmt --message-format human -- --check
-    - name: Install cargo-sort
-      uses: taiki-e/install-action@cargo-sort
     - name: Check Cargo.toml sorting
+      wait: install-sort
```
divergentdave:
```suggestion
      wait: [install-sort]
```

.github/workflows/ci-build.yml
jcjones: I wondered about that, but I wanted to see what happened. :)

.github/workflows/ci-build.yml
jcjones: Maybe... [docs say it can be a string:](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepswait) 

> Pauses the job until one or more background steps complete. A wait step performs no work itself, it only blocks until the referenced background steps finish. Provide a single step id as a string, or multiple step ids as an array.

jcjones:
@divergentdave Nope, same lint failure, all of them indicating aborting on the new syntax: 
```
Invalid workflow file: .github/workflows/ci-build.yml#L1
(Line: 84, Col: 7): Unexpected value 'env', (Line: 88, Col: 7): Unexpected value 'run', (Line: 156, Col: 7): Unexpected value 'run'
```

jcjones:
Worked!

divergentdave (approved):

PR #4670, opened 2026-06-25
Ignore trillium* major/minor updates on release/0.7

jcjones:
Replaces https://github.com/divviup/janus/pull/4665


divergentdave (approved):
We should probably do the same for other related crates, like `trillium` itself and `trillium-rustls`.

divergentdave (approved):

Issue #4625, opened 2026-06-05
Change Task Start/Duration to Interval representation


jcjones:
In https://github.com/divviup/janus/pull/4624 we discussed simplifying the representation of Tasks to use Interval, which I didn't take on immediately because it changes the data model all the way to the DB, and is a sizable change in its own right.

But it seems like a good change.

PR #4672, opened 2026-06-26
[DAP-17] TaskConf AAD Part 1: Add canonical TaskConfiguration builder and conversions

jcjones:
Foundation for binding TaskConfiguration into HPKE AADs:

- `VdafInstance::to_vdaf_config` and `BatchMode::to_batch_config` forward conversions (inherent methods, since orphan rules block `From` impls).
- `janus_core::task_config::build_task_configuration`, the single (prod, at least) path that touches endpoint URLs and the `task_interval` extension, with an encoded test vector. We'll be using this elsewhere.

Part plan:
- Part 2 is database changes, including switching database to the un-normalized URLs  (https://github.com/divviup/janus/issues/4625)
- Part 3 is message definitions
- Part 4 is aggregator
- Part 5 is client
- Part 6 is collector

This is part of https://github.com/divviup/janus/issues/4402

### DIFF
--- a/aggregator_core/src/task.rs
+++ b/aggregator_core/src/task.rs
@@ -60,9 +60,8 @@ impl BatchMode {
     ///
     /// This is the inverse of [`BatchMode::try_from(&BatchConfig)`](TryFrom). The
     /// `batch_time_window_size` of [`BatchMode::LeaderSelected`] is a Janus-specific parameter that
-    /// is not part of the DAP batch configuration, and is silently dropped; both aggregators must
-    /// drop it identically for the synthesized task configuration (and therefore the HPKE AAD it is
-    /// bound into) to match.
+    /// is not part of the DAP batch configuration, and is silently dropped. Therefore, the HPKE AAD
+    /// does not force the aggregators to agree on the `batch_time_window_size`.
     pub fn to_batch_config(&self) -> BatchConfig {
         match self {
             BatchMode::TimeInterval => BatchConfig::TimeInterval,
--- a/aggregator_core/src/task.rs
+++ b/aggregator_core/src/task.rs
@@ -54,6 +54,23 @@ pub enum BatchMode {
     },
 }
 
+impl BatchMode {
+    /// Returns the [`BatchConfig`] message representation of this batch mode, for inclusion in a
+    /// [`TaskConfiguration`](janus_messages::TaskConfiguration).
+    ///
+    /// This is the inverse of [`BatchMode::try_from(&BatchConfig)`](TryFrom). The
+    /// `batch_time_window_size` of [`BatchMode::LeaderSelected`] is a Janus-specific parameter that
+    /// is not part of the DAP batch configuration, and is silently dropped; both aggregators must
+    /// drop it identically for the synthesized task configuration (and therefore the HPKE AAD it is
+    /// bound into) to match.
+    pub fn to_batch_config(&self) -> BatchConfig {
+        match self {
+            BatchMode::TimeInterval => BatchConfig::TimeInterval,
+            BatchMode::LeaderSelected { .. } => BatchConfig::LeaderSelected,
+        }
+    }
+}
+
 impl TryFrom<batch_mode::Code> for BatchMode {
     type Error = Error;
 
@@ -1392,8 +1409,8 @@ mod tests {
         vdaf::vdaf_dp_strategies,
     };
     use janus_messages::{
-        Duration, HpkeAeadId, HpkeConfig, HpkeConfigId, HpkeKdfId, HpkeKemId, HpkePublicKey,
-        TaskId, Time, TimePrecision,
+        BatchConfig, Duration, HpkeAeadId, HpkeConfig, HpkeConfigId, HpkeKdfId, HpkeKemId,
+        HpkePublicKey, TaskId, Time, TimePrecision,
     };
     use rand::random;
     use serde_json::json;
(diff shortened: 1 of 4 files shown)


core/src/task_config.rs
```diff
@@ -0,0 +1,209 @@
...
+        time_precision,
+        min_batch_size,
+        batch_config,
+        vdaf_config,
+        extensions,
+    )
+}
+
+/// Normalize an aggregator endpoint and convert it to the DAP wire [`DapUrl`] representation.
+fn normalize_endpoint(endpoint: Url) -> Result<DapUrl, Error> {
```
divergentdave: Mention of normalizing URLs has me concerned. Section 4.1 says "Protocol participants MUST NOT re-encode or otherwise alter a Url when including it in a ... cryptographic digest." The `url::Url` type does make some changes on its own, before our trailing slash normalization here. I ran into this previously in Hickory DNS, because we used to use `url::Url` to store part of the RDATA for a CAA record with an iodef value. Ideally we should only do this sort of normalization upstream of `Url::join()` calls, and not AAD computation, to avoid incompatibilities with other implementations that pick URLs that aren't preserved by this. We should either store the unchanged form of the endpoint and normalize it on-demand, or store both the raw input and the normalized URL for the two different purposes.

The same section also says "The encoding of any particular URL into ASCII is agreed upon by participants out of band from the protocol." This means it's up to us to figure out how to produce a `janus_messages::Url` from whatever provisioning protocol. Taskprov has this taken care of, since it uses the same record format for provisioning as we use for the AAD. For our aggregator API, `PostTaskReq` does use `url::Url`, which is a bit unfortunate. Maybe we should change this to store a UTF-8 string or byte string, and separately validate during deserialization that it can be successfully parsed into a URL (without making that parsed form the source of truth). Looking at `janus_cli` provisioning of YAML task files, it also uses a `url::Url` field, though it uses `AggregatorTask` directly. Fixing this would fall out naturally from making the datastore layer preserve URLs without normalization.

jcjones: OK, I've updated the PR description, and reworked to not normalize. I'm also now planning to, in Part 2 when I touch the DB, change the DB representations to the `janus_messages::Url` form so we don't normalize there, either. This is ready for a re-review.

aggregator_core/src/task.rs
```diff
@@ -54,6 +54,23 @@ pub enum BatchMode {
...
 
+impl BatchMode {
+    /// Returns the [`BatchConfig`] message representation of this batch mode, for inclusion in a
+    /// [`TaskConfiguration`](janus_messages::TaskConfiguration).
+    ///
+    /// This is the inverse of [`BatchMode::try_from(&BatchConfig)`](TryFrom). The
+    /// `batch_time_window_size` of [`BatchMode::LeaderSelected`] is a Janus-specific parameter that
+    /// is not part of the DAP batch configuration, and is silently dropped; both aggregators must
+    /// drop it identically for the synthesized task configuration (and therefore the HPKE AAD it is
+    /// bound into) to match.
```
divergentdave: Suggested rephrasing: saying that "both aggregators must drop it identically" is a bit confusing, as the other aggregator may not support this parameter in the first place.
```suggestion
    /// is not part of the DAP batch configuration, and is silently dropped. Therefore, the HPKE AAD
    /// does not force the aggregators to agree on the `batch_time_window_size`.
```

tgeoghegan (approved):

core/src/task_config.rs
```diff
@@ -0,0 +1,200 @@
...
+                100,
+                BatchConfig::TimeInterval,
+                VdafConfig::Prio3Count,
+                None,
+            ),
+            Err(Error::InvalidParameter(_))
+        );
+    }
+
+    /// Pins the exact encoded bytes produced by the canonical builder. Because taskprov clients can
```
divergentdave: nit: this is now part of the core DAP spec
```suggestion
    /// Pins the exact encoded bytes produced by the canonical builder. Because DAP implementations can
```

core/src/vdaf.rs
```diff
@@ -133,6 +133,67 @@ impl VdafInstance {
...
+    /// Returns the [`VdafConfig`] message representation of this VDAF instance, for inclusion in a
+    /// [`TaskConfiguration`](janus_messages::TaskConfiguration).
+    ///
+    /// This is the inverse of [`VdafInstance::try_from(&VdafConfig)`](TryFrom). It is an inherent
+    /// method rather than a `From`/`TryFrom` impl because the orphan rules forbid implementing a
+    /// foreign trait to produce the foreign [`VdafConfig`] type from this crate.
+    ///
+    /// The differential privacy strategy is **not** representable in [`VdafConfig`] and is silently
+    /// dropped. Both aggregators must drop it identically for the synthesized task configuration
+    /// (and therefore the HPKE AAD it is bound into) to match.
```
divergentdave: Another rephrasing: I think aggregators have no choice but to drop differential privacy information here, as the VDAF configuration struct doesn't include it. We can say something about the way forward here, though.
```suggestion
    /// dropped. It is anticipated that future task extensions will allow for agreement on differential
    /// privacy mechanisms.
```

core/src/vdaf.rs
```diff
@@ -133,6 +133,67 @@ impl VdafInstance {
...
+                length: to_u32(*length)?,
+                chunk_length: to_u32(*chunk_length)?,
+            },
+
+            #[cfg(feature = "test-util")]
+            VdafInstance::Fake { rounds } => VdafConfig::Fake { rounds: *rounds },
+            #[cfg(feature = "test-util")]
+            VdafInstance::FakeFailsVerifyInit | VdafInstance::FakeFailsVerifyStep => {
+                return Err("VDAF instance has no VdafConfig representation");
+            }
```
divergentdave, line 194: Thinking ahead, we might need to just return `VdafConfig::Fake` with a fixed number of rounds, or do something else, in order for the relevant tests to go through once we start calling this in the aggregator.

divergentdave (approved):

core/src/vdaf.rs
jcjones: Yeah, honestly not sure yet :)

PR #4673, opened 2026-06-26
Reject reserved report extension type

### LINKED ISSUE #4671: Reject reports using the reserved extension type
`handle_uploaded_report()` and `compute_helper_aggregate_init()` reject reports with unrecognized extension types. We should additionally reject reports that have the reserved extension type anywhere in their extension lists, since we don't actually support that extension in any meaningful sense.

divergentdave:
This closes #4671. The second commit fixes a place where rustfmt was getting stuck, so I recommend reviewing this commit-by-commit with whitespace changes ignored.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1954,7 +1954,7 @@ impl VdafOps {
             .iter()
             .chain(report.metadata().public_extensions())
         {
-            if matches!(extension.extension_type(), ExtensionType::Unknown(_)) {
+            if !extension.extension_type().is_supported() {
                 debug!(
                     task_id = %task.id(),
                     report_id = ?report.metadata().id(),
@@ -3928,6 +3928,17 @@ async fn send_request_to_helper(
             timer.finish_attempt("error");
             Err(error.into())
         }
+    }
+}
+
+trait ExtensionTypeExt {
+    /// Determines whether a report extension type is supported by this implementation.
+    fn is_supported(&self) -> bool;
+}
+
+impl ExtensionTypeExt for ExtensionType {
+    fn is_supported(&self) -> bool {
+        matches!(self, ExtensionType::Taskbind)
     }
 }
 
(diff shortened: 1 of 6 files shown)


jcjones (changes requested):
This looks good. I like this positive-list use of `matches!` for being structurally immune to the `Unknown` codepoint issue.

I don't think we have a _deliberate_ test that a lone, correctly-sorted `Reserved` extension gets rejected. 

The one for `upload_report_unsorted_extensions` gets close, but it's getting rejected by the strictly-increasing check I added before it'd reach `extension_type_is_supported`. 

There are some other tests that used `Reserved` to exercise duplicate detection, which now get short-circuited before reaching the duplicate branch, so they  are testing this behavior, but they _aren't_ testing their old behavior anymore. There's probably more than these, but this is a quick ripgrep --
- http_handlers/tests/report.rs:592–624 will bail before the dupe check
- http_handlers/tests/aggregation_job_init.rs 's "verify_init_8" and 9

I think we should adjust those tests' expectations somehow  before landing this.

aggregator/src/aggregator.rs
```diff
@@ -3931,6 +3931,11 @@ async fn send_request_to_helper(
     }
 }
 
+/// Determines whether a report extension type is supported by this implementation.
+fn extension_type_is_supported(extension_type: ExtensionType) -> bool {
```
tgeoghegan: nit: would this be more natural as `ExtensionType::is_supported`?

divergentdave: I considered that, but that would have to live in `janus_messages`, and whether extension types are supported is moreso a property of the aggregator itself.

tgeoghegan: That argument makes sense. We could then make this an extension trait on `ExtensionType`, which we've done before for other `janus_messages` item. But I'm being fussy.

jcjones (approved):
I didn't realize what a pain this was going to be. Sorry about that, but I appreciate the detail work here to cover that masking issue.

PR #4674, opened 2026-06-26
Run documentation tests in CI

divergentdave:
This adds doc tests to CI and fixes one broken test. Cargo's documentation notes that `cargo test --doc` cannot be combined with other types of targets. As a result, `cargo test --all-targets` does not include documentation tests.


jcjones (approved):
Thank you!

PR #4676, opened 2026-06-29
Add tracing spans with request parameters

divergentdave:
This adds some more tracing spans annotated with selected request parameters. Where necessary, I split up async methods so I could do so with the `instrument` attribute macro. This should provide better context when we log errors.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -1545,7 +1545,7 @@ impl VdafOps {
 
     #[tracing::instrument(
         skip(self, datastore, metrics, task, req, request_hash),
-        fields(task_id = ?task.id()),
+        fields(task_id = ?task.id(), step = %req.step()),
         err(level = Level::DEBUG)
     )]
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):
The new spans log at DEBUG by default, which I think is sensible. It'd be interesting if we could selectively enable them at runtime. I believe this can be done by adding the right value to the end of `RUST_LOG`, but I wonder if there's some option to `tracing::instrument` where we can put in a span label so that it's easy to step up the log level to INFO or ERROR. Maybe `name` or `target`?

divergentdave:
The span has INFO level, which is the default. Only the event emitted when an error is returned is at the DEBUG level. The span gets a name by default from the method's name, and a target from the containing module. I don't think we can dynamically change the severity of the span or event, as those get handled at compile time, but we can change our filters at runtime, and we have an admin API to do so.

jcjones (approved):
Smooth, nice improvement! My big thing was to make sure these wouldn't spam, and no, they won't.

PR #4684, opened 2026-06-30
[DAP-17] TaskConf AAD Part 2A: Mandatory task_info and Interval task validity

### LINKED ISSUE #4625: Change Task Start/Duration to Interval representation

jcjones:
Provisions the task data the HPKE-AAD TaskConfiguration will need (Part 3), with no AAD changes yet.

- task_info is now a mandatory, persisted `Vec<u8>` (non-empty, <=255 bytes) rather than an `Option` that serde silently dropped. In the DB it's `BYTEA NOT NULL`; serde/API forms carry it as an unpadded base64url; taskprov passes the real value from its `TaskConfiguration`.

- Task validity is represented as a single `Option<Interval>` all the way to the DB: the `task_end` column becomes `task_duration,` and the API/serde forms are `task_start + task_duration`. Half-open intervals are now unrepresentable, as we all wanted.  Report-admission checks derive start/end from the interval, and just for safety I added that the `PATCH` task-end path returns 400 when end is `< task_start`.

Part 2B will change `url::Url` to `janus_messages::Url` for our in-DB representations, to keep from causing AAD errors with URL normalization business. That turns out to be a nice standalone task.

Closes #4625.

### DIFF
--- a/aggregator/src/aggregator/aggregation_job_init.rs
+++ b/aggregator/src/aggregator/aggregation_job_init.rs
@@ -428,6 +428,18 @@ where
                         Ok(shares)
                     });
 
+                    // Reject reports once the task has been deactivated. This is Janus-specific
+                    // (not part of the task's TaskConfiguration/AAD) and is compared against
+                    // theaggregator's current clock, not the report timestamp.
+                    let shares = shares.and_then(|shares| {
+                        if let Some(deactivate_at) = task.deactivate_at() {
+                            if now >= deactivate_at {
+                                return Err(ReportError::TaskExpired);
+                            }
+                        }
+                        Ok(shares)
+                    });
+
                     // Next, the aggregator runs the verification-state initialization algorithm
                     // for the VDAF associated with the task and computes the first state
                     // transition. [...] If either step fails, then the aggregator MUST fail
--- a/aggregator/src/aggregator/aggregation_job_init.rs
+++ b/aggregator/src/aggregator/aggregation_job_init.rs
@@ -430,7 +430,7 @@ where
 
                     // Reject reports once the task has been deactivated. This is Janus-specific
                     // (not part of the task's TaskConfiguration/AAD) and is compared against
-                    // theaggregator's current clock, not the report timestamp.
+                    // the aggregator's current clock, not the report timestamp.
                     let shares = shares.and_then(|shares| {
                         if let Some(deactivate_at) = task.deactivate_at() {
                             if now >= deactivate_at {
--- a/aggregator/src/aggregator/aggregation_job_init.rs
+++ b/aggregator/src/aggregator/aggregation_job_init.rs
@@ -409,7 +409,7 @@ where
                                 .report_share()
                                 .metadata()
                                 .time()
-                                .is_before(task_start)
+                                .is_before(&task_start)
                             {
                                 return Err(ReportError::TaskNotStarted);
                             }
@@ -421,7 +421,7 @@ where
                     // of the task_end time. (§4.6.2.4 step 4 [dap-16])
                     let shares = shares.and_then(|shares| {
                         if let Some(task_end) = task.task_end() {
-                            if verify_init.report_share().metadata().time().ge(task_end) {
+                            if verify_init.report_share().metadata().time().ge(&task_end) {
                                 return Err(ReportError::TaskExpired);
                             }
                         }
@@ -723,8 +723,8 @@ mod tests {
     };
     use janus_messages::{
         AggregationJobId, AggregationJobInitializeReq, AggregationJobResp, Duration, Extension,
-        ExtensionType, MediaType, PartialBatchSelector, ReportError, ReportMetadata, TimePrecision,
-        VerifyResp, VerifyStepResult, batch_mode::TimeInterval,
+        ExtensionType, Interval, MediaType, PartialBatchSelector, ReportError, ReportMetadata,
+        TimePrecision, VerifyResp, VerifyStepResult, batch_mode::TimeInterval,
     };
     use prio::{
         codec::Encode,
@@ -1225,7 +1225,15 @@ mod tests {
         .with_time_precision(time_precision)
         .with_tolerable_clock_skew(Duration::from_seconds(500, &time_precision))
         .with_aggregator_auth_token(AuthenticationToken::Bearer(random()))
-        .with_task_end(Some(task_end_time))
+        .with_task_interval(Some(
+            Interval::new(
+                task_end_time
+                    .sub_duration(&Duration::from_seconds(1000, &time_precision))
+                    .unwrap(),
+                Duration::from_seconds(1000, &time_precision),
+            )
+            .unwrap(),
+        ))
         .build();
         let helper_task = task.helper_view().unwrap();
         let ephemeral_datastore = ephemeral_datastore().await;
@@ -1367,7 +1375,13 @@ mod tests {
         .with_time_precision(time_precision)
         .with_tolerable_clock_skew(Duration::from_seconds(500, &time_precision))
         .with_aggregator_auth_token(AuthenticationToken::Bearer(random()))
-        .with_task_start(Some(task_start_time))
+        .with_task_interval(Some(
+            Interval::new(
+                task_start_time,
+                Duration::from_seconds(2000, &time_precision),
+            )
+            .unwrap(),
+        ))
         .build();
         let helper_task = task.helper_view().unwrap();
         let ephemeral_datastore = ephemeral_datastore().await;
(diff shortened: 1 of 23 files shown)


jcjones:
Whoops, how'd I miss that test. Lemme work this out.

db/00000000000001_initial_schema.up.sql
```diff
@@ -107,15 +107,18 @@ CREATE TABLE tasks(
...
     report_expiry_age           BIGINT,                    -- the maximum age of a report before it is considered expired (and acceptable for garbage collection), in seconds. NULL means that GC is disabled.
     min_batch_size              BIGINT NOT NULL,           -- the minimum number of reports in a batch to allow it to be collected
     time_precision              BIGINT NOT NULL,           -- the duration to which clients are expected to round their report timestamps, in seconds
     tolerable_clock_skew        BIGINT NOT NULL,           -- the maximum acceptable clock skew to allow between client and aggregator, in seconds
     collector_hpke_config       BYTEA,                     -- the HPKE config of the collector (encoded HpkeConfig message)
     vdaf_verify_key             BYTEA NOT NULL,            -- the VDAF verification key (encrypted)
-    task_info                   BYTEA,                     -- the task_info field of a TaskConfiguration structure; NULL for API-provisioned tasks, which have no TaskConfiguration
+    task_info                   BYTEA NOT NULL,            -- the task_info field of a TaskConfiguration structure; non-empty and at most 255 bytes
```
divergentdave: This is out of date with some of the other changes.
```suggestion
    task_info                   BYTEA NOT NULL,            -- the task_info field of a TaskConfiguration structure; at most 255 bytes
```

aggregator_core/src/task.rs
```diff
@@ -152,10 +152,9 @@ struct CommonTaskParameters {
     /// How much clock skew to allow between client and aggregator. Reports from
     /// farther than this duration into the future will be rejected.
     tolerable_clock_skew: Duration,
-    /// The `task_info` byte string from a `TaskConfiguration` struct. `None` for API-provisioned
-    /// tasks, which have no `TaskConfiguration`; used to distinguish tasks with otherwise
-    /// equivalent DAP task parameters.
-    task_info: Option<Vec<u8>>,
+    /// The `task_info` byte string from a `TaskConfiguration` struct, used to distinguish tasks
+    /// with otherwise equivalent DAP task parameters. Must be non-empty and at most 255 bytes.
```
divergentdave:
```suggestion
    /// with otherwise equivalent DAP task parameters. Must be at most 255 bytes.
```

aggregator_core/src/datastore.rs
```diff
@@ -837,6 +837,14 @@ DELETE FROM tasks WHERE task_id = $1",
...
+    ///
+    /// Because the task's validity interval is stored as a (start, duration) pair and a half-open
+    /// interval is not representable, the requested end is translated into a new `task_duration`:
+    ///
+    /// * `Some(end)`: `task_duration` is recomputed as `end - task_start`, leaving `task_start`
+    ///   unchanged. If the task has no `task_start`, both columns remain NULL (the interval cannot
+    ///   be set without a start), which is a no-op.
+    /// * `None`: both `task_start` and `task_duration` are cleared, removing the interval entirely.
     #[tracing::instrument(skip(self), err(level = Level::DEBUG))]
     pub async fn update_task_end(
```
divergentdave: This API actually poses a problem now, because it gets incorporated into the task's AAD. Changing the duration midstream would change the AAD, and break decryption of reports. We use this API as the backend for disabling or deleting a task from the control plane currently (not entirely sure which). I think we need to get rid of `update_task_end()` and introduce some new implementation-specific task field that we can modify to reimplement this feature.

aggregator_api/src/routes.rs
```diff
@@ -139,6 +139,11 @@ pub(super) async fn post_task<C: Clock>(
 
     let vdaf_verify_key = SecretBytes::new(vdaf_verify_key_bytes);
 
+    let task_info = URL_SAFE_NO_PAD
+        .decode(&req.task_info)
+        .context("Invalid base64 value for task_info")
```
divergentdave: For clarity:
```suggestion
        .context("Invalid base64url value for task_info")
```

aggregator_api/src/routes.rs
```diff
@@ -307,6 +317,23 @@ pub(super) async fn patch_task<C: Clock>(
         .run_tx("patch_task", |tx| {
             Box::pin(async move {
                 if let Some(task_end) = req.task_end {
```
divergentdave: As per the other comment, changing the task end time is no longer safe now that it's part of the AAD.

aggregator_core/src/datastore.rs
jcjones: I'm playing with adding a `deactivate_at` time outside the task configuration, as an Option<DateTime> that we can set via API and it just ... deactivates then. If we set it to now or the past, it's instant.

jcjones: @divergentdave please take a look at [aacea12](https://github.com/divviup/janus/pull/4684/commits/aacea1292a8bad731428b28b86542b6f28001a44) and see what you think!

aggregator/src/aggregator/aggregation_job_init.rs
```diff
@@ -421,7 +421,19 @@ where
...
+                            if verify_init.report_share().metadata().time().ge(&task_end) {
+                                return Err(ReportError::TaskExpired);
+                            }
+                        }
+                        Ok(shares)
+                    });
+
+                    // Reject reports once the task has been deactivated. This is Janus-specific
+                    // (not part of the task's TaskConfiguration/AAD) and is compared against
+                    // theaggregator's current clock, not the report timestamp.
```
divergentdave:
```suggestion
                    // the aggregator's current clock, not the report timestamp.
```

aggregator_core/src/datastore.rs
jcjones: Opened https://github.com/divviup/divviup-api/issues/2343

aggregator/src/aggregator/aggregation_job_init.rs
divergentdave: FWIW this will be a behavior change from the perspective of control plane users: previously disabling or deleting a task set Janus's task end time to now, which meant previously sharded reports could continue to make their way through the system, and newer reports would be rejected. This provides a sharper cutoff, stopping any in-process aggregation. I think this is a reasonable change for that UI, but just wanted to note it explicitly.

db/00000000000001_initial_schema.up.sql
```diff
@@ -107,15 +107,19 @@ CREATE TABLE tasks(
...
     min_batch_size              BIGINT NOT NULL,           -- the minimum number of reports in a batch to allow it to be collected
     time_precision              BIGINT NOT NULL,           -- the duration to which clients are expected to round their report timestamps, in seconds
     tolerable_clock_skew        BIGINT NOT NULL,           -- the maximum acceptable clock skew to allow between client and aggregator, in seconds
     collector_hpke_config       BYTEA,                     -- the HPKE config of the collector (encoded HpkeConfig message)
     vdaf_verify_key             BYTEA NOT NULL,            -- the VDAF verification key (encrypted)
-    task_info                   BYTEA,                     -- the task_info field of a TaskConfiguration structure; NULL for API-provisioned tasks, which have no TaskConfiguration
+    task_info                   BYTEA NOT NULL,            -- the task_info field of a TaskConfiguration structure; at most 255 bytes
+    deactivate_at               TIMESTAMP WITH TIME ZONE,  -- stop accepting reports at this time (or null). (Janus-specific)
```
divergentdave: We have previously used just `TIMESTAMP` a.k.a. `TIMESTAMP WITHOUT TIME ZONE`. Looking back on our previous discussions, both types occupy eight bytes. The [docs](https://www.postgresql.org/docs/current/datatype-datetime.html) say that both are represented internally by the database as a UTC timestamp. The database treats `WITH TIME ZONE` differently when parsing a textual input without an explicit time zone, assuming that it's in local time.

I think we should make this column a `TIMESTAMP` in keeping with past practice, use the `NaiveDateTime` implementations of `FromSql`/`ToSql`, and immediately convert to/from a `DateTime<Utc>` using `and_utc()`/`naive_utc()`. This will ensure that the database's conception of its own local timezone never comes into play.

divergentdave (approved):

db/00000000000001_initial_schema.up.sql
jcjones: Did so in [d561b42](https://github.com/divviup/janus/pull/4684/commits/d561b425691c9f5fc600a1d22a19e40ba30a73d1)

aggregator/src/aggregator/aggregation_job_init.rs
jcjones: I noted that in https://github.com/divviup/divviup-api/issues/2343

PR #4692, opened 2026-07-06
[DAP-17] TaskConf AAD Part 2B: Use janus_messages::Url 

jcjones:
This set of commits changes the endpoint URLs to a byte-preserving `janus_messages::Url` form instead of `url::Url`. This is largely mechanical, and I left it in four commits for ease of review... though it's not that bad to go through as one.

Note that one of the final commits, https://github.com/divviup/janus/pull/4692/commits/c2d58781d62ca04be428c659543a925eb1c02ff1, adds the "fast fail" functionality, for purely code-compilation reasons. So when you see by commit 2 that we'd store a bad URL, uh, I do fix that.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -56,7 +56,7 @@ use janus_messages::{
     AggregationJobStep, BatchSelector, CollectionJobId, CollectionJobReq, CollectionJobResp,
     Duration, ExtensionType, HpkeConfig, HpkeConfigList, InputShareAad, Interval,
     PartialBatchSelector, PlaintextInputShare, Report, ReportError, ReportUploadStatus, Role,
-    TaskConfiguration, TaskId, UploadErrors, VerifyResp,
+    TaskConfiguration, TaskId, UploadErrors, Url as DapUrl, VerifyResp,
     batch_mode::{LeaderSelected, TimeInterval},
     extensions_are_strictly_increasing,
 };
@@ -933,7 +933,7 @@ impl<C: Clock> Aggregator<C> {
         task_id: &TaskId,
         task_config: &TaskConfiguration,
         aggregator_auth_token: Option<&AuthenticationToken>,
-    ) -> Result<(&PeerAggregator, Url, Url), Error> {
+    ) -> Result<(&PeerAggregator, DapUrl, DapUrl), Error> {
         let peer_aggregator_url = match peer_role {
             Role::Leader => task_config.leader_aggregator_endpoint(),
             Role::Helper => task_config.helper_aggregator_endpoint(),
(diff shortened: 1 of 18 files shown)


core/src/lib.rs
```diff
@@ -62,6 +62,16 @@ pub fn url_ensure_trailing_slash(mut url: Url) -> Url {
     url
 }
 
+/// Converts a [`janus_messages::Url`] into a [`url::Url`] ready for [`Url::join`], ensuring a
+/// trailing slash so `join` does not drop the last path component.
+///
+/// The trailing-slash fixup applies only to this routing copy, never to the stored
+/// [`janus_messages::Url`] — its exact bytes are bound into HPKE AADs and must not be re-encoded
+/// (DAP-18 §4.1).
+pub fn url_for_join(url: &janus_messages::Url) -> Result<Url, url::ParseError> {
```
tgeoghegan, line 76: This might be nicer as an extension trait on `url::Url`. Also, it looks like we only ever use `url_ensure_trailing_slash` to make leader or helper URLs joinable. Do we ever use the one without the other? Can they become a single function?

aggregator/src/aggregator.rs
```diff
@@ -964,8 +964,9 @@ impl<C: Clock> Aggregator<C> {
         );
         Ok((
             peer_aggregator,
-            task_config.leader_aggregator_endpoint().try_into()?,
-            task_config.helper_aggregator_endpoint().try_into()?,
+            // Byte-preserving clones of the wire endpoints (no re-encoding, per DAP-18 §4.1).
+            task_config.leader_aggregator_endpoint().clone(),
+            task_config.helper_aggregator_endpoint().clone(),
```
tgeoghegan, line 969: If it doesn't turn into a whole lifetime adventure, could we avoid the clones and make this return `(&PeerAggregator, &DapUrl, &DapUrl)`?

tgeoghegan (approved):

aggregator/src/aggregator.rs
jcjones: It's a mess of lifetime issues I'm afraid.

core/src/lib.rs
jcjones: I considered both, too, but it felt like I'd need it more than I guess I do. Lemme go get that into a commit.

jcjones: Note the two callers operate on different types (url::Url, infallible vs. janus_messages::Url, fallible), so this doesn't really eliminate anything, but I did move to a `UrlExt` trait. Take a look at 21453c49. I think it's cleaner.

The deeper merge isn't safe, at least not without some more substantial refactoring -- the builder joins its `url::Url` endpoints without ever going through `url_for_join`, so that's now a problem for the AAD.

aggregator/src/aggregator.rs
tgeoghegan: Not worth the hassle then.

core/src/lib.rs
tgeoghegan: OK. It doesn't really matter a ton since these aren't crates that we export for usage outside Janus, but I like to have fewer loose `pub` items hanging around. Last thought is that it'd be nice to refactor `url_for_join` into a method on `janus_messages::Url`. I don't think you could put it in `UrlExt` because it returns a `Self` value so you'd run into annoying trait object stuff. Purely optional, btw, do it or don't at your discretion.

divergentdave (approved):

PR #4693, opened 2026-07-07
[DAP-17] Leader-side Collection Job Extensions support

jcjones:
This implements leader-side job extensions. Helper-side needs these extensions embedded in AggregateShareReq, which is part of the AAD work.

Since there _aren't_ any, I don't plumb this down to the database. Rather, the leader makes sure there aren't any, for now.

I'm really satisfied with the CollectionRequestBuilder, fwiw. I like those ergonomics.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -53,10 +53,10 @@ use janus_core::{
 use janus_messages::{
     AggregateShare, AggregateShareAad, AggregateShareId, AggregateShareReq,
     AggregationJobContinueReq, AggregationJobId, AggregationJobInitializeReq, AggregationJobResp,
-    AggregationJobStep, BatchSelector, CollectionJobId, CollectionJobReq, CollectionJobResp,
-    Duration, ExtensionType, HpkeConfig, HpkeConfigList, InputShareAad, Interval,
-    PartialBatchSelector, PlaintextInputShare, Report, ReportError, ReportUploadStatus, Role,
-    TaskConfiguration, TaskId, UploadErrors, Url as DapUrl, VerifyResp,
+    AggregationJobStep, BatchSelector, CollectionJobExtensionType, CollectionJobId,
+    CollectionJobReq, CollectionJobResp, Duration, ExtensionType, HpkeConfig, HpkeConfigList,
+    InputShareAad, Interval, PartialBatchSelector, PlaintextInputShare, Report, ReportError,
+    ReportUploadStatus, Role, TaskConfiguration, TaskId, UploadErrors, Url as DapUrl, VerifyResp,
     batch_mode::{LeaderSelected, TimeInterval},
     extensions_are_strictly_increasing,
 };
@@ -3169,6 +3169,28 @@ impl VdafOps {
         collection_job_id: &CollectionJobId,
         req: Arc<CollectionJobReq<B>>,
     ) -> Result<Vec<u8>, Error> {
+        // Collection job extensions must be in strictly increasing order of extension type, and
+        // every extension type must be supported (DAP-19 §4.2.2).
+        if !req
+            .extensions()
+            .is_sorted_by(|a, b| a.extension_type() < b.extension_type())
+        {
+            return Err(Error::InvalidExtension(
+                Some(*task.id()),
+                "collection job extensions are not in strictly increasing order of extension type",
+            ));
+        }
+        if req
+            .extensions()
+            .iter()
+            .any(|extension| !extension.extension_type().is_supported())
+        {
+            return Err(Error::UnsupportedExtension(
+                Some(*task.id()),
+                "collection job contains an unsupported extension type",
+            ));
+        }
+
         let aggregation_param = Arc::new(
             A::AggregationParam::get_decoded(req.aggregation_parameter())
                 .map_err(Error::MessageDecode)?,
@@ -4197,6 +4219,18 @@ trait ExtensionTypeExt {
 impl ExtensionTypeExt for ExtensionType {
     fn is_supported(&self) -> bool {
         matches!(self, ExtensionType::Taskbind)
+    }
+}
+
+trait CollectionJobExtensionTypeExt {
+    /// Determines whether a collection job extension type is supported by this implementation.
+    fn is_supported(&self) -> bool;
+}
+
+impl CollectionJobExtensionTypeExt for CollectionJobExtensionType {
+    fn is_supported(&self) -> bool {
+        // No collection job extension types are defined yet, so none are supported.
+        false
     }
 }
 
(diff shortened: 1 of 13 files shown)


messages/src/lib.rs
```diff
@@ -1716,6 +1716,116 @@ impl<B: BatchMode> Decode for Query<B> {
...
+
+/// DAP protocol message representing the type of a collection job extension.
+///
+/// No extension types are defined yet, so every codepoint other than the reserved `0x0000` decodes
+/// to `Unknown`. Equality, ordering, and hashing are all defined in terms of the underlying
+/// codepoint.
+#[derive(Clone, Copy, Debug, FromPrimitive, IntoPrimitive)]
+#[repr(u16)]
+#[non_exhaustive]
+pub enum CollectionJobExtensionType {
```
jcjones, line 1781: we use `enum` elsewhere, but I'm tempted to make this 
```rust
#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
pub struct CollectionJobExtensionType(u16);
``` 

so the boilerplate becomes more compiler-generated.

aggregator/src/aggregator.rs
```diff
@@ -4200,6 +4222,18 @@ impl ExtensionTypeExt for ExtensionType {
...
 }
 
+trait CollectionJobExtensionTypeExt {
+    /// Determines whether a collection job extension type is supported by this implementation.
+    fn is_supported(&self) -> bool;
+}
+
+impl CollectionJobExtensionTypeExt for CollectionJobExtensionType {
+    fn is_supported(&self) -> bool {
+        // No collection job extension types are defined yet, so none are supported.
```
jcjones: I'm worried about YAGNI here; like, this could just be "not empty? reject!" but I went for the more verbose, more ready for extensions path here. But I can't really justify it.

aggregator/src/aggregator/http_handlers/tests/collection_job.rs
tgeoghegan, line 45: It's unfortunate to burden the majority of callsites with a `Vec::new()` call, which unfortunately doesn't even tell the reader what it's an empty list of.

We could do a builder-style thing here, but without a whole separate type. Call it a renovator: add `fn with_extensions(mut self, extensions: Vec<...>) -> Self` to `CollectionJobReq`. Then the rare places where we do want to set extensions can still do so fluently with `CollectionJobReq::new(...).with_extensions(vec![this, that, the_other])`.

aggregator/src/aggregator.rs
```diff
@@ -3169,6 +3169,28 @@ impl VdafOps {
         collection_job_id: &CollectionJobId,
         req: Arc<CollectionJobReq<B>>,
     ) -> Result<Vec<u8>, Error> {
+        // Collection job extensions must be in strictly increasing order of extension type, and
+        // every extension type must be supported (DAP-19 §4.2.2).
```
tgeoghegan: 4.2.2. discusses task extensions. You want 4.6.2.

Also: this is implied by "strictly increasing" but for dummies like me, why not also mention in the comment that this check rejects duplicate extensions?

aggregator/src/aggregator.rs
tgeoghegan: I think that if this were public API we were eventually going to flesh out, it might make sense to set it up now, but since we can extend our internal interfaces at will whenever we want, I think we should delete this.

messages/src/lib.rs
tgeoghegan: I'm all for making the compiler generate more boilerplate, but I don't think I understand the implications of using `enum` there?

In any case, if you want to pitch a better way to handle enums in `janus_messages`, and use that consistently across the ones we define, I'm open to hearing it. (and I don't mean that as a roundabout way of telling you to go away, I'm sincere: I'm curious what the tradeoffs are of different ways of handling integer backed enums, and we should document our preference and rationale in `docs/DEVELOPMENT.md`) But I think for this PR, we should stick to the established way of handling it.

messages/src/problem_type.rs
```diff
@@ -72,7 +74,12 @@ impl DapProblemType {
             DapProblemType::BatchOverlap => {
                 "The queried batch overlaps with a previously queried batch."
             }
-            DapProblemType::UnsupportedExtension => "The report includes an unsupported extension.",
+            DapProblemType::UnsupportedExtension => {
+                "The message includes an unsupported extension."
+            }
+            DapProblemType::InvalidExtension => {
+                "An extensions list is out of order or contains an invalid extension encoding."
```
tgeoghegan, line 81: What are the cases of "invalid extension encoding"? Stuff like the extension type and the length prefix of the encoded extension having the wrong width?

tgeoghegan (changes requested):

messages/src/problem_type.rs
jcjones: That's verbatim from https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-19#table-1.

But yeah, in Janus, invalidExtension is currently emitted only for out-of-order. Malformed encoding fails earlier at decode and won't make it to this error type right now.

I can change that text, but I figured it should match the doc?

aggregator/src/aggregator/http_handlers/tests/collection_job.rs
jcjones: I like it. See 5ecb114d.

aggregator/src/aggregator.rs
jcjones: in 5ecb114d

aggregator/src/aggregator.rs
jcjones: As you wish! 5ecb114d

messages/src/lib.rs
jcjones: https://github.com/divviup/janus/issues/4696

collector/src/lib.rs
```diff
@@ -453,49 +504,40 @@ impl<V: vdaf::Collector> Collector<V> {
...
-        query: Query<B>,
-        aggregation_parameter: &V::AggregationParam,
-    ) -> Result<Collection<V::AggregateResult, B>, Error> {
-        let job = self.start_collection(query, aggregation_parameter).await?;
-        self.poll_until_complete(&job).await
-    }
-
-    /// Send a collection request to the leader aggregator, using a randomly generated
-    /// [`CollectionJobId`].
+    /// Begin a collection request to the leader aggregator.
```
divergentdave: This is moreso a builder method, we don't actually begin anything until we send the request.
```suggestion
    /// Construct a collection request to send to the leader aggregator.
```

divergentdave (approved):

messages/src/problem_type.rs
tgeoghegan: Yeah, I guess the hair I'm splitting is that many of those cases would be captured by `invalidMessage`. But this text is fine.

tgeoghegan (approved):

PR #4695, opened 2026-07-08
more fluent problem document construction

tgeoghegan:
Add `ProblemDocument::with_task_id_maybe` so that callers can fluently add `Option<TaskId>` to a problem document, and make the existing `ProblemDocument::with_detail` silently ignore empty strings. This makes some callsites more fluent.

### DIFF
--- a/aggregator/src/aggregator/problem_details.rs
+++ b/aggregator/src/aggregator/problem_details.rs
@@ -28,8 +28,8 @@ pub struct ProblemDocument<'a> {
     title: &'static str,
     #[serde(serialize_with = "serialize_status")]
     status: StatusCode,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    taskid: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none", rename = "taskid")]
+    task_id: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     detail: Option<&'a str>,
     #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


tgeoghegan:
I noticed this could be a little neater while reviewing #4693 and decided to tidy it up in passing. Once #4693 lands, I'll rebase this to use the new method in the `Error::UnsupportedExtensions` and `Error::InvalidExtension` cases it adds.

aggregator/src/aggregator/problem_details.rs
```diff
@@ -29,7 +29,7 @@ pub struct ProblemDocument<'a> {
     #[serde(serialize_with = "serialize_status")]
     status: StatusCode,
     #[serde(skip_serializing_if = "Option::is_none")]
-    taskid: Option<String>,
+    task_id: Option<String>,
```
divergentdave, line 32: This is incorrect, see https://www.ietf.org/archive/id/draft-ietf-ppm-dap-19.html#section-3.6-7
```suggestion
    taskid: Option<String>,
```

tgeoghegan: Oh, whoops, I was trying to make this more idiomatically Rusty, but forgot the variable name is used in serialization. Thank you for catching this. I opted to stick with the field name `task_id` and use `serde(rename="taskid")`. I also added a test to check that the JSON serialization uses the right keys, not just that our implementation can parse its own output.

jcjones (approved):
👍🏼

divergentdave (approved):

PR #4698, opened 2026-07-08
[DAP-18] TaskConf AAD Part 3: Get all the bits and bobs in place for the TaskConfiguration at all endpoints

jcjones:
DAP-18 binds the task's TaskConfiguration -- both aggregator endpoints, task_info, the VDAF/batch config, and the validity interval -- into the HPKE AADs for input and aggregate shares, which requires every party (leader, helper, client, collector, candlestick maker) to independently reconstruct identical `TaskConfiguration` bytes. 

This PR builds the machinery to do that, without yet touching the wire format, building atop https://github.com/divviup/janus/pull/4707.

AggregatorTask gains an `own_aggregator_endpoint` beside the existing peer endpoint, stamped at creation from the configured `public_dap_url` for API tasks, and from the received config for taskprov tasks (which `taskprov_opt_in` previously discarded, can't save that memory now!) 

It also gains a `task_configuration()` adapter that synthesizes the config for API tasks and returns the received wire bytes verbatim for taskprov. 

Endpoints move to `janus_messages::Url` so nothing is re-encoded.

The Client and Collector builders grow the parameters needed to construct a matching config (the collector previously had no helper endpoint at all), validated fail-fast at `build()`, and the integration-test harness threads the real `task_info`/`interval` through so client- and collector-built configs match those at the aggregators.

Nothing is bound into an AAD yet: these values are threaded but not load-bearing, so the change is behavior-preserving. So, uh, the tests still pass.

The actual AAD turn-on --  the InputShareAad / AggregateShareAad / AggregateShareReq struct changes -- is in another castle, and I have to admit I am not sure I have all the byte-identical test changes in here yet, because that changeset still doesn't pass. But look, _this one_ passes, and it is going in the right direction while only being +800/-100.

I just mean to say, while the AAD turn-on is theoretically ~150 lines, it might be ... more... still.

### DIFF
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -713,7 +713,8 @@ WHERE success = TRUE ORDER BY version DESC LIMIT(1)",
                 "-- put_aggregator_task()
 INSERT INTO tasks (
     task_id, aggregator_role, aggregation_mode, peer_aggregator_endpoint,
-    batch_mode, vdaf, task_start, task_duration, report_expiry_age, min_batch_size,
+    own_aggregator_endpoint, batch_mode, vdaf, task_start, task_duration,
+    report_expiry_age, min_batch_size,
     time_precision, tolerable_clock_skew, collector_hpke_config,
     vdaf_verify_key, task_info, deactivate_at, aggregator_auth_token_type,
     aggregator_auth_token, aggregator_auth_token_hash,
@@ -721,7 +722,7 @@ INSERT INTO tasks (
     updated_at, updated_by)
 VALUES (
     $1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16, $17, $18,
-    $19, $20, $21, $22, $23, $24
+    $19, $20, $21, $22, $23, $24, $25
 )
 ON CONFLICT DO NOTHING",
             )
@@ -735,6 +736,8 @@ ON CONFLICT DO NOTHING",
                     /* aggregation_mode */ &task.aggregation_mode().copied(),
                     /* peer_aggregator_endpoint */
                     &task.peer_aggregator_endpoint().as_str(),
+                    /* own_aggregator_endpoint */
+                    &task.own_aggregator_endpoint().as_str(),
                     /* batch_mode */ &Json(task.batch_mode()),
                     /* vdaf */ &Json(task.vdaf()),
                     /* task_start */
@@ -889,7 +892,8 @@ UPDATE tasks SET
             .prepare_cached(
                 "-- get_aggregator_task()
 SELECT
-    aggregator_role, aggregation_mode, peer_aggregator_endpoint, batch_mode,
+    aggregator_role, aggregation_mode, peer_aggregator_endpoint,
+    own_aggregator_endpoint, batch_mode,
     vdaf, task_start, task_duration, report_expiry_age, min_batch_size,
     time_precision, tolerable_clock_skew, collector_hpke_config,
     vdaf_verify_key, task_info, deactivate_at, aggregator_auth_token_type,
@@ -913,6 +917,7 @@ FROM tasks WHERE task_id = $1",
                 "-- get_aggregator_tasks()
 SELECT
     task_id, aggregator_role, aggregation_mode, peer_aggregator_endpoint,
+    own_aggregator_endpoint,
     batch_mode, vdaf, task_start, task_duration, report_expiry_age, min_batch_size,
     time_precision, tolerable_clock_skew, collector_hpke_config,
     vdaf_verify_key, task_info, deactivate_at, aggregator_auth_token_type,
@@ -934,6 +939,7 @@ FROM tasks",
         // Scalar task parameters.
         let aggregator_role: AggregatorRole = row.get("aggregator_role");
         let peer_aggregator_endpoint = row.get::<_, String>("peer_aggregator_endpoint").parse()?;
+        let own_aggregator_endpoint = row.get::<_, String>("own_aggregator_endpoint").parse()?;
         let batch_mode = row.try_get::<_, Json<task::BatchMode>>("batch_mode")?.0;
         let aggregation_mode: Option<AggregationMode> = row.get("aggregation_mode");
         let vdaf = row.try_get::<_, Json<VdafInstance>>("vdaf")?.0;
@@ -1048,6 +1054,7 @@ FROM tasks",
         let task = AggregatorTask::new(
             *task_id,
             peer_aggregator_endpoint,
+            own_aggregator_endpoint,
             batch_mode,
             vdaf,
             vdaf_verify_key,
--- a/aggregator_core/src/datastore.rs
+++ b/aggregator_core/src/datastore.rs
@@ -28,7 +28,8 @@ use janus_core::{
 use janus_messages::{
     AggregateShareId, AggregationJobId, BatchId, CollectionJobId, Duration, Extension,
     HpkeCiphertext, HpkeConfig, HpkeConfigId, Interval, Query, ReportId, ReportIdChecksum,
-    ReportMetadata, Role, TaskId, Time, TimePrecision, VerifyContinue, VerifyInit, VerifyResp,
+    ReportMetadata, Role, TaskConfiguration, TaskId, Time, TimePrecision, VerifyContinue,
+    VerifyInit, VerifyResp,
     batch_mode::{BatchMode, LeaderSelected, TimeInterval},
 };
 use leases::{acquired_aggregation_job_from_row, acquired_collection_job_from_row};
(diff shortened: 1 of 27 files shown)


aggregator_core/src/datastore.rs
```diff
@@ -934,6 +945,11 @@ FROM tasks",
         // Scalar task parameters.
         let aggregator_role: AggregatorRole = row.get("aggregator_role");
         let peer_aggregator_endpoint = row.get::<_, String>("peer_aggregator_endpoint").parse()?;
+        let own_aggregator_endpoint = row.get::<_, String>("own_aggregator_endpoint").parse()?;
+        let taskprov_task_config = row
```
tgeoghegan, line 949: nit: I think this would be a bit tidier as `let taskprov_task_config: Option<Vec<u8>> = row.get(...)...` but I acknowledge that the function and the rest of this module already liberally mix type annotations and the dread turbofish, so go with whatever you prefer and perhaps we'll normalize it someday.

db/00000000000001_initial_schema.up.sql
```diff
@@ -104,6 +104,8 @@ CREATE TABLE tasks(
     task_id                     BYTEA UNIQUE NOT NULL,     -- 32-byte TaskID as defined by the DAP specification
     aggregator_role             AGGREGATOR_ROLE NOT NULL,  -- the role of this aggregator for this task
     peer_aggregator_endpoint    TEXT NOT NULL,             -- peer aggregator's API endpoint
+    own_aggregator_endpoint     TEXT NOT NULL,             -- this aggregator's own API endpoint
```
tgeoghegan, line 107: I'm 99.9% sure `TEXT` is fine here. But if we want this to be an opaque byte string that never gets messed with by anything so that AADs always work out, is `BYTEA` safer? Does `TEXT` impose any restrictions on stored data, or could it possibly do any surprising normalizations?

tgeoghegan (approved):

db/00000000000001_initial_schema.up.sql
jcjones: We also want this to be readable without shenanigans, so `TEXT` is the right idea here. Postgres `TEXT` doesn't normalize, it only enforces UTF-8 validity with no embedded `\0`s -- which is good, actually! Our input type is ASCII-validated anyway.  Our `own_endpoint_stored_verbatim` test checks that the result is byte-exact, for future's sake.

(Note that `peer_aggregator_endpoint` is equally AAD-bound and also `TEXT`.)

The truly binary column here -- `taskprov_task_config` -- is `BYTEA`.

aggregator_core/src/datastore.rs
jcjones: Gonna leave it as-is then.

db/00000000000001_initial_schema.up.sql
tgeoghegan: Cool, I appreciate that thorough analysis. Rigorous consideration was all I wanted here.

tools/src/bin/collect.rs
```diff
@@ -623,6 +623,14 @@ fn new_collector<V: vdaf::Collector>(
         vdaf,
         time_precision,
     )
+    // The helper endpoint and remaining TaskConfiguration parameters are not yet exposed as CLI
```
divergentdave, line 626: I wonder if the `TaskConfiguration` being in the AADs will now motivate passing around `TaskConfiguration` blobs as a means to communicate DAP task parameters. Maybe we could replace most of the existing command line flags, plus the new ones we'd need otherwise, with another base64url blob. (for a later PR)

db/00000000000001_initial_schema.up.sql
```diff
@@ -104,6 +104,8 @@ CREATE TABLE tasks(
...
+    taskprov_task_config        BYTEA,                     -- verbatim wire TaskConfiguration for taskprov tasks (bound into HPKE AADs); NULL for API-provisioned tasks
```
divergentdave, line 108: I saw in other comments that the `TaskConfiguration` for Taskprov tasks "is not byte-reconstructible from the stored parameters", necessitating this column, because "it would drop DP strategies and unknown extensions". It seems like we have two different sources of truth for the `TaskConfiguration`, and we switch between them depending on provisioning mechanism. This seems potentially fraught -- could the same issues that keep us from reproducing a Taskprov task's config also affect other tasks? I think it would be preferable if we stick to one mechanism for producing the `TaskConfiguration`, either storing it or reconstructing it in a content-preserving way. Doing both seems like it's the worst of both worlds, as we both have to be careful about our data model mapping 1-1 to and from the `TaskConfiguration` bytes, and we have an extra set of code paths on top of that.

Always saving and using the serialized `TaskConfiguration` seems like it would be the easiest to maintain. If we were to go the other route, I think I'd want us to have a fuzzer that deserializes arbitrary TaskConfiguration objects, constructs a task from it as we would do in Taskprov, calls `task_configuration()`, encodes the result, and checks that it is equal to the original fuzzer input.

To address the specific reasons given, first, the differential privacy configuration used to be in older versions of Taskprov, but it has been removed from more recent ones. Thus, our differential privacy strategy field is an implementation-specific task parameter for the time being, until an updated differential privacy draft gets published to set out how to track this as a task extension. At that point, we could re-base our differential privacy support on that extension, and ensure agreement on the differential privacy mechanism between all parties. In the meantime, I think it should be fine for us to drop the dp_strategy field when serializing a `TaskConfiguration` struct. Second, I think unknown extensions in a Taskprov HTTP header ought to cause us to opt out of a task. The editor's copy says we MUST opt out if "the DAP batch mode or VDAF is not implemented" -- lack of support for a task extension should logically be a reason to opt out as well, since DAP says "extensions are mandatory to support." We ought to refuse to aggregate, rather than carry two conflicting representations of the task's extensions.

integration_tests/src/lib.rs
```diff
@@ -22,6 +22,11 @@ pub struct TaskParameters {
     pub vdaf: VdafInstance,
     pub min_batch_size: u64,
     pub time_precision: TimePrecision,
+    /// The task's `task_info`, sourced from the provisioned task so the client and collector bind
+    /// a byte-identical `TaskConfiguration` into their HPKE AADs.
+    pub task_info: Vec<u8>,
+    /// The task's optional collection interval, for the same byte-identity reason.
```
divergentdave: I think "collection interval" could be potentially confusing; better to stick with "task interval".
```suggestion
    /// The optional task interval, for the same byte-identity reason.
```

client/src/lib.rs
```diff
@@ -23,13 +23,20 @@
...
 //!         leader_url,
 //!         helper_url,
 //!         TimePrecision::from_seconds(300),
-//!         vdaf
+//!         vdaf,
 //!     )
+//!     .with_task_info(b"[task info]".to_vec())
+//!     .with_min_batch_size(1000)
+//!     .with_batch_config(BatchConfig::TimeInterval)
+//!     .with_vdaf_config(VdafConfig::Prio3Histogram { length: 12, chunk_length: 4 })
```
divergentdave, line 38: This duplication between the `VdafConfig` and the `Prio3::new_histogram()` arguments is a little unfortunate. After we get things working, maybe we should come back here and add convenience constructors/builder methods for well-known VDAF types?

divergentdave (commented):
Reviewing these changes has gotten me ruminating about task provisioning. Now that we have a standardized `TaskConfiguration` serialization format in DAP, I wonder if it would make more sense to use that when communicating task configuration between different components, outside of the interactions defined in DAP. For example, the next version of our JavaScript library could take a `TaskId` and a `TaskConfiguration`, and it wouldn't need anything else. This could also provide a more convenient path forward for supporting future extensions (i.e. one defining differential privacy mechanisms) because we could piggy back on each extension's defined serialization, and we wouldn't have to adapt each configuration parameter to every library and language binding. At any rate, these thoughts are beyond the scope of this PR.

client/src/lib.rs
jcjones: https://github.com/divviup/janus/issues/4711

db/00000000000001_initial_schema.up.sql
jcjones: https://github.com/divviup/janus/issues/4712

tools/src/bin/collect.rs
jcjones: Yeah... `TaskConfiguration` is task-shape only; the collector still needs the things that aren't in it: `task_id`, collector auth token, collector HPKE keypair/credential, and the query itself (batch interval / leader-selected). 

So the shape would be, like: "one `--task-config <blob>` replaces the DAP-task-parameter flags, existing credential/query flags stay."

`collect.rs` is the natural first consumer. I'll make note of this to think through, too!

jcjones: https://github.com/divviup/janus/issues/4713

divergentdave (approved):

PR #4707, opened 2026-07-10
[DAP-18] TaskConf AAD: Provision in-process test task after aggregator startup

jcjones:
Restructure `JanusInProcess` to provision its task via a new `provision_task()` method once its ephemeral port is known, so `JanusInProcessPair` builds one task with consistent real endpoints and provisions both aggregator views. A starting point for the AAD binding work going on over in https://github.com/divviup/janus/pull/4698.

Doesn't do anything else.


tgeoghegan (approved):
I was sure it'd be more complicated than this, but if it works, it works.

jcjones:
When I started copying things out, I was a little surprised too that this is all that was needed. But I am sorry to say that the rest of this won't be so small. :/

PR #4709, opened 2026-07-10
Add a test name filter ability to xtask

jcjones:
I kept wanting this, so here.

### DIFF
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -21,8 +21,7 @@ struct CargoArgs {
     #[clap(long, action)]
     locked: bool,
 
-    /// Arguments forwarded to `cargo test`, e.g. a test-name filter. Everything after a `--`
-    /// separator is passed through verbatim.
+    /// Arguments forwarded to the test binary. Pass them after a `--` separator.
     #[clap(last = true)]
     test_args: Vec<String>,
 }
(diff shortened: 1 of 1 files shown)


xtask/src/main.rs
```diff
@@ -157,6 +162,7 @@ fn run_docker_tests(images: ContainerImages, cargo_args: CargoArgs) -> Result<()
         "--package=janus_integration_tests",
         "--features=testcontainer",
     ]);
+    command.args(&cargo_args.test_args);
```
divergentdave: I think we should add a `--` argument here first. Skipping it might goof up `--test-threads=1`, `--include-ignored`, `--show-output`, etc.

xtask/src/main.rs
```diff
@@ -20,6 +20,11 @@ struct CargoArgs {
     /// Require Cargo.lock is up to date
     #[clap(long, action)]
     locked: bool,
+
+    /// Arguments forwarded to `cargo test`, e.g. a test-name filter. Everything after a `--`
+    /// separator is passed through verbatim.
+    #[clap(last = true)]
```
divergentdave, line 25: FWIW, here's what `cargo test`'s configuration looks like: https://github.com/rust-lang/cargo/blob/0a28f7930c7b559c37fc221347114f9c6434f2ae/src/bin/cargo/commands/test.rs#L9 I think `last(true)` isn't necessary based on this, but it's small potatoes. If the test arguments don't themselves start with a hyphen, they get recognized and passed on, and `cargo test` only requires `--` as a delimiter in cases where the arguments could be confused as flags for `cargo` itself.

divergentdave (approved):

xtask/src/main.rs
jcjones: Good idea. In cc56e2ea

xtask/src/main.rs
jcjones: Mmm. I tested it, and while you're right, I also like being explicit here, which we ensure by the `last = true`.

tgeoghegan (approved):

PR #4708, opened 2026-07-10
fix lints for clippy 1.97

jcjones:

PR #4710, opened 2026-07-14
clippy 1.97 lints

tgeoghegan:
Effectively a backport of #4708, but I let `clippy --fix` do the work.

### DIFF
--- a/core/src/http/cached_resource.rs
+++ b/core/src/http/cached_resource.rs
@@ -171,20 +171,17 @@ pub(crate) fn expires_at<'a, I: IntoIterator<Item = &'a HeaderValue>>(
             return Some(Instant::now());
         }
 
-        if let Some(max_age) = directive.strip_prefix("max-age=") {
-            let parsed = match max_age.parse() {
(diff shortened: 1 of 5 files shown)


core/src/http/cached_resource.rs
```diff
@@ -171,7 +171,8 @@ pub(crate) fn expires_at<'a, I: IntoIterator<Item = &'a HeaderValue>>(
             return Some(Instant::now());
         }
 
-        if let Some(max_age) = directive.strip_prefix("max-age=") {
+        {
```
divergentdave: We could drop these braces

divergentdave (approved):

core/src/http/cached_resource.rs
tgeoghegan: Indeed. I'm surprised clippy let that one go.

divergentdave: I think leaving a block behind is the conservative approach when it comes to preserving drop order, but that's not often a concern for us.

Issue #4713, opened 2026-07-14
Add `--task-config <blob>`?


jcjones:
Yeah... `TaskConfiguration` is task-shape only; the collector still needs the things that aren't in it: `task_id`, collector auth token, collector HPKE keypair/credential, and the query itself (batch interval / leader-selected). 

So the shape would be, like: "one `--task-config <blob>` replaces the DAP-task-parameter flags, existing credential/query flags stay."

`collect.rs` is the natural first consumer. I'll make note of this to think through, too!

_Originally posted by @jcjones in https://github.com/divviup/janus/pull/4698#discussion_r3580977284_

jcjones:
Whatever we do here should also adjust the `interop_binaries/src/commands/janus_interop_collector.rs` too.

PR #4714, opened 2026-07-15
[DAP-17] TaskConf AAD Part 4: bind TaskConfiguration into HPKE AADs 

jcjones:
I'm afraid this is another _draw the rest of the owl_ that can't split up worth a darn and have green tests. Except for the ones I'm marking to ignore!

This is broken into sub-commits which deal with these specific headings... but I feel like this one is just as easy to review as a big blob of ~crazy~ _high quality code_

## Bind each task's TaskConfiguration into the input-share and aggregate-share HPKE AADs.

Both ends recompute each AAD independently, so byte-identical encoding across every party is mandatory or everything errors out.

- `InputShareAad` gains `task_configuration` as field 2
- `AggregateShareAad` becomes `{task_id, task_configuration, collection_job_req}`,  dropping `aggregation_parameter + batch_selector`, now matching DAP-18
    - This requires the Helper to do more validations (they became MUST) which I add before the datastore transaction.
- `AggregateShareReq` becomes `{collection_job_req, batch_selector, report_count, checksum}`:
    - `collection_job_req` replaces the standalone `aggregation_parameter`
    - The Leader forwards its stored `CollectionJobReq` verbatim and the Helper binds it directly (no reconstruction)
    - the aggregation parameter is now reached via `req.collection_job_req().aggregation_parameter()`

> [!IMPORTANT]
> §4.6.4 defines consistency per batch mode, and for time-interval permits any selector within the queried interval (containment):
> `all batch buckets whose batch bucket identifiers are contained within the batch interval specified in the CollectionJobReq's query`
> Here I implement the stricter requirement that the selector reconstruct the query exactly (equality). The reason is that the `PUT` path stores `batch_selector().batch_identifier()`, and the poll/`GET` path rebuilds the AAD's query from that stored identifier alone via `query_for_collection_identifier`. That reconstruction is lossy for anything narrower than the query, so a strict sub-interval would decrypt on the `PUT` path and silently fail on the poll path. Equality fails with `batchInvalid` at `PUT`; containment would accept and fail silently later. (`invalidMessage` vs `batchInvalid` is a judgment call imo, I'm open to changing. I've already changed my mind once.)
> The consequence is that Janus rejects some spec-conformant sub-interval requests -- if we want true containment, the Helper needs to persist the query rather than re-derive it, which yet another big expansion of scope for this already big change. I'm erroring on YAGNI here, but if we think we want that, I want to file an issue and take it as a follow-on. I have a test for this, noting the choice.
> For leader-selected the check is trivially satisfied, since the query body is empty. (phew)

> [!NOTE]
> `core/src/vdaf.rs`: `to_vdaf_config()` (which is `#[cfg(test-util)]`) maps `FakeFailsVerifyInit` and `FakeFailsVerifyStep` to `VdafConfig::Fake`, so the Helper can build an `InputShareAad` before the intentional verify failure.

## Update the interop binaries to build a byte-identical `TaskConfiguration`.

- Flip AggregatorAddTaskRequest.leader/.helper to `janus_messages::Url` so endpoint bytes survive interop provisioning verbatim (no `url::Url` re-encoding); the interop aggregator binds them directly.
- Add a shared `INTEROP_TASK_INFO` placeholder (the interop test API has no `task_info` concept) and an `interop_batch_config()` helper, used identically by the interop aggregator, client, and collector.
- Carry `min_batch_size + batch_mode` in the interop upload request, and the helper `endpoint + min_batch_size` in the collector `add_task` request; new fields are serde-optional so a non-Janus runner still parses.
- Test harness: source real `task_info/task_interval/helper` endpoint into the collector (collect_aggregate_result_generic now takes a helper_port) and send  `min_batch_size + batch_mode` from the container client.
- Ignore the `janus_janus_*` container tests: a follow-up will re-enable them.... somehow.

## A Stake in `tools/collect` (see #4713)

`new_collector` fabricates its `TaskConfiguration` values to make things work. But now they're _actively_ wrong. I've made the two AAD-consuming paths (`run_collection` and `run_poll_job`) now fail fast with an error naming the missing parameters. #4713 (David's suggested `--task-config <blob>`) is probably The Way.

`new-job` is unaffected and still works as it never computes an AAD. Success through laziness!

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -53,10 +53,10 @@ use janus_core::{
 use janus_messages::{
     AggregateShare, AggregateShareAad, AggregateShareId, AggregateShareReq,
     AggregationJobContinueReq, AggregationJobId, AggregationJobInitializeReq, AggregationJobResp,
-    AggregationJobStep, CollectionJobId, CollectionJobReq, CollectionJobResp, Duration,
-    ExtensionType, HpkeConfig, HpkeConfigList, InputShareAad, Interval, PartialBatchSelector,
-    PlaintextInputShare, Report, ReportError, ReportUploadStatus, Role, TaskConfiguration, TaskId,
-    UploadErrors, Url as DapUrl, VerifyResp,
+    AggregationJobStep, CollectionJobExtension, CollectionJobId, CollectionJobReq,
+    CollectionJobResp, Duration, ExtensionType, HpkeConfig, HpkeConfigList, InputShareAad,
+    Interval, PartialBatchSelector, PlaintextInputShare, Report, ReportError, ReportUploadStatus,
+    Role, TaskConfiguration, TaskId, UploadErrors, Url as DapUrl, VerifyResp,
     batch_mode::{LeaderSelected, TimeInterval},
     extensions_are_strictly_increasing,
 };
@@ -3178,24 +3178,7 @@ impl VdafOps {
         collection_job_id: &CollectionJobId,
         req: Arc<CollectionJobReq<B>>,
     ) -> Result<Vec<u8>, Error> {
-        // Collection job extensions must be in strictly increasing order of extension type, which
-        // also rejects duplicates, and every extension type must be supported (DAP-19 §4.6.2). No
-        // collection job extension types are defined yet, so any extension is unsupported.
-        if !req
-            .extensions()
-            .is_sorted_by(|a, b| a.extension_type() < b.extension_type())
-        {
-            return Err(Error::InvalidExtension(
-                Some(*task.id()),
-                "collection job extensions are not in strictly increasing order of extension type",
-            ));
-        }
-        if !req.extensions().is_empty() {
-            return Err(Error::UnsupportedExtension(
-                Some(*task.id()),
-                "collection job contains an unsupported extension type",
-            ));
-        }
+        validate_collection_job_extensions(task.id(), req.extensions())?;
 
         let aggregation_param = Arc::new(
             A::AggregationParam::get_decoded(req.aggregation_parameter())
@@ -3819,6 +3802,28 @@ impl VdafOps {
         aggregate_share_id: &AggregateShareId,
         dp_strategy: Arc<S>,
     ) -> Result<AggregateShare, Error> {
+        // DAP-19 §4.6.4: the Helper validates the Collector's extensions independently of the
+        // Leader.
+        validate_collection_job_extensions(
+            task.id(),
+            aggregate_share_req.collection_job_req().extensions(),
+        )?;
+
+        // DAP-19 §4.6.4: the Helper verifies the Leader-chosen batch selector against the query in
+        // the CollectionJobReq, which the AAD binds. The spec defines consistency per batch mode
+        // and would permit any time-interval selector within the queried interval, but we require
+        // exact round-tripping: the poll path rebuilds the AAD's query from the stored selector
+        // alone, so a narrower selector would decrypt here and fail there.
+        if B::query_for_collection_identifier(
+            aggregate_share_req.batch_selector().batch_identifier(),
+        ) != *aggregate_share_req.collection_job_req().query()
+        {
+            return Err(Error::InvalidMessage(
+                Some(*task.id()),
+                "batch selector is inconsistent with the collection job request's query",
+            ));
+        }
+
         // §4.4.4.3: check that the batch interval meets the requirements from §4.6
         if !B::validate_collection_identifier(
             aggregate_share_req.batch_selector().batch_identifier(),
@@ -4054,6 +4059,30 @@ impl VdafOps {
             })?;
         Ok(())
     }
+}
+
+/// Validates a collection job's extensions, per DAP-19 §4.6.2. Both the Leader (on collection job
+/// creation) and the Helper (on aggregate share requests, independently of the Leader) must apply
+/// this.
+fn validate_collection_job_extensions(
+    task_id: &TaskId,
+    extensions: &[CollectionJobExtension],
+) -> Result<(), Error> {
+    // Strictly increasing order also rejects duplicates.
+    if !extensions.is_sorted_by(|a, b| a.extension_type() < b.extension_type()) {
+        return Err(Error::InvalidExtension(
+            Some(*task_id),
+            "collection job extensions are not in strictly increasing order of extension type",
+        ));
+    }
+    // No collection job extension types are defined yet, so any extension is unsupported.
+    if !extensions.is_empty() {
+        return Err(Error::UnsupportedExtension(
+            Some(*task_id),
+            "collection job contains an unsupported extension type",
+        ));
+    }
+    Ok(())
 }
 
 fn write_task_aggregation_counter<C: Clock>(
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -53,10 +53,10 @@ use janus_core::{
 use janus_messages::{
     AggregateShare, AggregateShareAad, AggregateShareId, AggregateShareReq,
     AggregationJobContinueReq, AggregationJobId, AggregationJobInitializeReq, AggregationJobResp,
-    AggregationJobStep, BatchSelector, CollectionJobId, CollectionJobReq, CollectionJobResp,
-    Duration, ExtensionType, HpkeConfig, HpkeConfigList, InputShareAad, Interval,
-    PartialBatchSelector, PlaintextInputShare, Report, ReportError, ReportUploadStatus, Role,
-    TaskConfiguration, TaskId, UploadErrors, Url as DapUrl, VerifyResp,
+    AggregationJobStep, CollectionJobId, CollectionJobReq, CollectionJobResp, Duration,
+    ExtensionType, HpkeConfig, HpkeConfigList, InputShareAad, Interval, PartialBatchSelector,
+    PlaintextInputShare, Report, ReportError, ReportUploadStatus, Role, TaskConfiguration, TaskId,
+    UploadErrors, Url as DapUrl, VerifyResp,
     batch_mode::{LeaderSelected, TimeInterval},
     extensions_are_strictly_increasing,
 };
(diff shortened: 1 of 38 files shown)


divergentdave:
> §4.6.4 defines consistency per batch mode, and for time-interval permits any selector within the queried interval (containment):
> `all batch buckets whose batch bucket identifiers are contained within the batch interval specified in the CollectionJobReq's query`

I interpret this spec text as describing how the query relates to internal state, not the batch selector. Later on, 5.1.2 says that for the time interval mode, "The payload of BatchSelector.config is `struct { Interval batch_interval; } TimeIntervalBatchSelectorConfig;` where batch_interval is the batch interval requested by the Collector." This is a straightforward requirement that the same batch interval appear inside the batch_selector field and the nested query field.

Edit: Nevermind, the next paragraph gives the more flexible rules for when the batch selector is consistent query, so I'm now getting mixed messages from this section.

aggregator/src/aggregator/http_handlers/tests/aggregate_share.rs
```diff
@@ -125,6 +131,17 @@ async fn aggregate_share_request_invalid_batch_interval() {
     datastore.put_aggregator_task(&helper_task).await.unwrap();
 
     let request = AggregateShareReq::new(
+        CollectionJobReq::new(
+            Query::new_time_interval(
+                Interval::new(
+                    clock.now().to_time(task.time_precision()),
+                    // Collect request will be rejected because batch interval is too small
+                    Duration::from_seconds(task.time_precision().as_seconds() - 1, &time_precision),
```
divergentdave: nit: `Duration::ZERO` would be simpler here

aggregator/src/aggregator/http_handlers/tests/aggregate_share.rs
```diff
@@ -500,8 +571,8 @@ async fn aggregate_share_request() {
                 aggregate_share_resp.encrypted_aggregate_share(),
                 &AggregateShareAad::new(
                     *task.id(),
-                    dummy::AggregationParam(0).get_encoded().unwrap(),
-                    request.batch_selector().clone(),
+                    task.helper_view().unwrap().task_configuration().unwrap(),
```
divergentdave: It might be nice to add a method to `test_util::Task` that returns the task configuration, since we call this chain of methods frequently.

tgeoghegan:
I think we should correct the spec text to resolve some ambiguities here about whether the query and (full) batch selector need to match exactly. IMO no, they do not, but David is right that the text is unclear at best, if not contradictory here. I wrote up github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/791 to track draft-ietf-ppm-dap fixes.

Returning to this PR: I'm not comfortable with the decision to require the aggregate share job's batch selector to exactly match the collection job request's query in the time interval case. It seems reasonable to me to allow leaders to emit batch selectors that are narrower than the query and I think Janus should handle this since we're the reference implementation of DAP. We went to the trouble of making this possible in the spec by including the collection job req in the aggregate share req, so I think we should wire this up properly.

I'm sensitive to the point that persisting the collection job req on the helper side expands the scope of this PR, though, so I'd be OK leaving that as a TODO and doing it in a later PR. Or perhaps resequencing the stack of changes so that you *first* make a change to persist the collection job req on the helper side and *then* rebase this change to start folding that into the AAD. I think the latter would be better, but you're doing the actual typing, J.C., so I defer to your preference.

aggregator/src/aggregator.rs
```diff
@@ -3651,11 +3635,19 @@ impl VdafOps {
                 .map_err(Error::MessageEncode)?,
             &AggregateShareAad::new(
                 *task.id(),
-                aggregate_share_job
-                    .aggregation_parameter()
-                    .get_encoded()
-                    .map_err(Error::MessageEncode)?,
-                BatchSelector::<B>::new(aggregate_share_job.batch_identifier().clone()),
+                task.task_configuration()?,
+                // On the poll path we no longer have the Collector's original CollectionJobReq
```
tgeoghegan, line 3639: As discussed in another comment, we should either take a TODO to persist the collection job req and use it in the AAD here, or rebase this change on another that persists the collect req.

tgeoghegan (commented):
The code changes in here look fine, though I think it needs re-review after rebasing on #4707 and potentially the change to persist collection job req on the Helper side.

messages/src/tests/mod.rs
```diff
@@ -5,3 +5,67 @@ mod hpke;
...
+        Duration::from_time_precision_units(28),
+    )
+    .unwrap()
+    .build()
+    .unwrap()
+}
+
+/// The wire encoding of [`test_task_configuration`], mirroring
+/// [`task::roundtrip_task_configuration`].
+pub(crate) const TASK_CONFIGURATION_HEX: &str = concat!(
```
divergentdave: FYI, if you'd like to go back to using `concat!()`: I think you could replace the `TASK_CONFIGURATION_HEX` const with a declarative macro, with no arguments, that stamps out the same `concat()` expression.

interop_binaries/src/lib.rs
```diff
@@ -30,10 +30,25 @@ use testcontainers::{ContainerAsync, Image};
 use tokio::sync::Mutex;
 use tracing_log::LogTracer;
 use tracing_subscriber::{EnvFilter, Registry, prelude::*};
-use url::Url;
 
 pub mod commands;
 
+/// The interop test API has no `task_info` concept, so every Janus interop binary (aggregator,
```
divergentdave, line 36: This is fine for now. divergentdave/draft-dcook-ppm-dap-interop-test-design#49 tracks updating the API definition for more recent drafts.

aggregator/src/aggregator/collection_job_tests.rs
```diff
@@ -465,8 +468,16 @@ async fn collection_job_success_leader_selected() {
...
             &AggregateShareAad::new(
                 *test_case.task.id(),
-                aggregation_param.get_encoded().unwrap(),
-                BatchSelector::new_leader_selected(batch_id),
+                test_case
+                    .task
+                    .leader_view()
+                    .unwrap()
+                    .task_configuration()
+                    .unwrap(),
```
divergentdave: Call sites like these could benefit from a helper method on `test_util::Task` as well. (flagging it because it's wrapped across multiple lines)

interop_binaries/src/commands/janus_interop_collector.rs
```diff
@@ -46,13 +47,24 @@ struct AddTaskRequest {
     task_id: String,
     #[educe(Debug(method(std::fmt::Display::fmt)))]
     leader: DapUrl,
+    // The collector never connects to the helper, but its endpoint is bound into the
+    // AggregateShareAad, so it must match the aggregators' byte-for-byte. Optional (serde default)
+    // so a non-Janus runner that omits it still parses. (See issue #4713)
+    #[educe(Debug(method(std::fmt::Display::fmt)))]
+    #[serde(default = "unused_helper_endpoint")]
```
divergentdave: We could probably get rid of the default method, and require that a helper URL be present in the request. If another client did use the fallback path, they'd still need to use the same URL in order to decrypt the result. I think it's fine for these endpoints to drift from the API as we're implementing new DAP versions, until we settle on what the new interop test API will be. (same for min_batch_size)

aggregator/src/aggregator.rs
```diff
@@ -3651,11 +3635,19 @@ impl VdafOps {
...
+                // (there is no incoming AggregateShareReq), so reconstruct it from the stored job.
+                // This must be byte-identical to the request the Collector sent; see
+                // `query_for_collection_identifier`. When we support extensions (Issue #4715)
+                // this will need to change.
```
jcjones:
```suggestion
                // This must be byte-identical to the request the Collector sent; see
                // `query_for_collection_identifier` (and Issue #4743). When we support
                // extensions (Issue #4715) this will need to change.
```

aggregator/src/aggregator.rs
jcjones: I've opened #4743 for this.

jcjones:
> The code changes in here look fine, though I think it needs re-review after rebasing on #4707 and potentially the change to persist collection job req on the Helper side.

This is already based on #4707. I've opened #4743 for the `CollectionJobReq` persistence, which I want to take as a follow-on for this rather than make this big change even bigger.

aggregator/src/aggregator/http_handlers/tests/aggregate_share.rs
jcjones: Ooo, this is a great one, but there are callsites way outside this PR. I'm going to do this in this PR and open a follow-on to get the rest of them.

tgeoghegan (approved):

aggregator/src/aggregator/collection_job_tests.rs
jcjones: There's actually a lot of these. I'm going to do some that are adjacent to the AAD stuff now, right within these diffs, and open a follow-on for the rest.

aggregator/src/aggregator/http_handlers/tests/aggregate_share.rs
jcjones: https://github.com/divviup/janus/issues/4744

aggregator/src/aggregator/collection_job_tests.rs
jcjones: https://github.com/divviup/janus/issues/4744

interop_binaries/src/commands/janus_interop_collector.rs
jcjones: in 125d738

aggregator/src/aggregator/collection_job_tests.rs
jcjones: Also 3a64daa

messages/src/tests/mod.rs
jcjones: went back in 3a64daa

aggregator/src/aggregator/http_handlers/tests/aggregate_share.rs
jcjones: in 3a64daa

divergentdave (approved):

PR #4741, opened 2026-07-23
Replace git-version with a custom build script

divergentdave:
This replaces our use of `git-version` with a custom build script. Per m-ou-se/rust-git-version#26, spurious rebuilds can be triggered when the `.git/index` file is changed, and this seems to explain the rebuilds we've sometimes seen between running `cargo test --no-run` and `cargo test`. We can't eliminate this dependency entirely without introducing a build tool other than Cargo, but we can make it conditional. We already use the `GIT_REVISION` environment variable as a fallback source for the truncated commit hash, for use in container build environments with no git executable or .git directory. The main idea of this PR is to switch the order in which we check these, and omit the dependency on `.git/index` when the environment variable is in use. Then we can set the environment variable in non-Docker CI builds, and hopefully avoid spurious rebuilds there.

A second benefit of this PR is that it prints a warning message when it uses the fallback "unknown" value. This directed me to add the `GIT_REVISION` environment variable to our interop test container builds.

### DIFF
--- a/aggregator_api/build.rs
+++ b/aggregator_api/build.rs
@@ -1,0 +1,107 @@
+use std::{
+    borrow::Cow,
+    env::{self, VarError},
+    error::Error,
+    path::{Path, PathBuf},
+    process::{Command, Stdio},
+};
+
+/// Get the current git commit, and pass it to rustc via an environment variable.
+///
+/// This build script replaces our previous use of `git-version`, in order to conditionally reduce
+/// spurious rebuilds when internal git files change.
+///
+/// The environment variable GIT_REVISION can be used to supply a revision when git is not
+/// available, i.e. in container builds. The same environment variable name is used both for
+/// communication from build tooling to this build script and from this build script to rustc.
+fn main() {
+    // Check for an environment variable input first. This is provided by a Dockerfile or CI build
+    // script. This allows us to short-circuit running `git describe`, and lets us avoid file
+    // dependencies.
+    println!("cargo::rerun-if-env-changed=GIT_REVISION");
+    match env::var("GIT_REVISION") {
+        Ok(git_revision) => {
+            println!("cargo::rustc-env=GIT_REVISION={git_revision}");
+            return;
+        }
+        Err(VarError::NotPresent) => {}
+        Err(VarError::NotUnicode(_)) => {
+            panic!("invalid value of GIT_REVISION environment variable")
+        }
+    }
+
+    // Get the path of the current package.
+    //
+    // We don't need to print "cargo::rerun-if-env-changed=CARGO_MANIFEST_DIR" because "it is not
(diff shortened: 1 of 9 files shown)


aggregator_api/src/lib.rs
```diff
@@ -235,14 +234,10 @@ fn parse_hpke_config_id_param(config_id: &str) -> Result<HpkeConfigId, Error> {
     )?))
 }
 
-/// Returns the git revision used to build this crate, using `git describe` if available, or the
-/// environment variable `GIT_REVISION`. Returns `"unknown"` instead if neither is available.
+/// Returns the git revision used to build this crate.
+///
+/// This is set by the build script using `git describe`, if available. Otherwise, the compile-time
+/// environment variable `GIT_REVISION` is used, or a fallback value of "unknown".
```
tgeoghegan: IIUC all the behavior described here is in the build script, so I find it confusing to discuss the logic here when the body of the function is just `env!("GIT_REVISION")`. I think this doccomment should say something like "Environment variable `GIT_REVISION` is guaranteed to be set by `aggregator_api/build.rs`", and then let that program explain its own fallback logic.

aggregator_api/build.rs
```diff
@@ -0,0 +1,100 @@
...
+    let manifest_dir = PathBuf::from(
+        env::var_os("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR must be set by Cargo"),
+    );
+    let workspace_dir = manifest_dir
+        .parent()
+        .expect("unexpected directory structure");
+    let workspace_dir_string = workspace_dir
+        .to_str()
+        .expect("workspace directory is not valid UTF-8");
+    let git_dir = workspace_dir.join(".git");
```
tgeoghegan: I'm wondering if there could be perverse cases where `.git` is not alongside `Cargo.toml` as we expect. Does git guarantee that? Perhaps the right thing to do would be to frst check `$CARGO_MANIFEST_DIR/.git`, then go up into parent directories until we find a `.git`? But TBH I think this is good enough if it works for our CI and dev workflows, and it'll work for 99% of cases where people build Janus.

tgeoghegan (approved):
This LGTM. To make sure I understand: if either `git(1)` is not installed or `.git` is not present (e.g., compiling from a [source tarball](https://github.com/divviup/janus/releases)), then the build script falls back to setting `GIT_VERSION=unknown`, correct?

aggregator_api/build.rs
divergentdave: Yeah, we could get this from `git rev-parse --git-dir` instead. I had skipped that step from `git-version`. This assumption indeed breaks when using worktrees, as `.git` becomes a text file.

tgeoghegan (approved):
Perfect, I use worktrees pretty often so I'm glad this is handled.

PR #4745, opened 2026-07-28
Make the integration tests use matching host:ports for all entities

jcjones:
While each container aggregator is assigned a random port by Docker, we can use reqwest shenanigans to capture and proxy all the traffic from the client and collector to reach the right places, all using the same host:port strings, ensuring that all parties end up with the same AAD bytes.

### DIFF
--- a/integration_tests/src/lib.rs
+++ b/integration_tests/src/lib.rs
@@ -36,18 +36,12 @@ pub struct TaskParameters {
 /// Components of one aggregator's DAP endpoint.
 #[derive(Debug)]
 pub enum AggregatorEndpointFragments {
-    /// The aggregator is in a Kind cluster.
-    ///
-    /// It is reached from the host via a port forward on localhost (ephemeral port, supplied
-    /// later) and from within the cluster at its service name on port 8080. The scheme is always
-    /// `http`.
-    VirtualNetwork { host: String, path: String },
-    /// The aggregator is in a Docker network.
+    /// The aggregator is in a virtual network: a Docker network or a Kind cluster.
     ///
     /// It serves on port 8080 at `host`. The single `http://{host}:8080{path}` URL is
-    /// byte-identical for all parties; the host reaches it through a per-aggregator forwarding
-    /// proxy (Docker assigns the host port dynamically). The scheme is always `http`.
-    DockerNetwork { host: String, path: String },
+    /// identical for all parties; the host reaches it through a per-aggregator forwarding
+    /// proxy that bridges to an ephemeral host port, and the scheme is always `http`.
+    VirtualNetwork { host: String, path: String },
     /// The aggregator is running on localhost.
     ///
     /// No port forwarding is involved, so the same URL is used in all circumstances. The port
@@ -61,19 +55,18 @@ pub enum AggregatorEndpointFragments {
 }
 
 impl AggregatorEndpointFragments {
-    /// Provides the URL for the aggregator's endpoint from the perspective of the host. If the
-    /// aggregator is in a virtual network, this will use a port forward, with the provided port
-    /// number. If the aggregator itself is running on the host, it will use the provided port
-    /// number as well. If the aggregator is running remotely, the port number is ignored and the
-    /// URL is returned.
+    /// Provides the URL for the aggregator's endpoint from the perspective of the host. A
+    /// virtual-network aggregator uses its byte-identical `http://{host}:8080` URL (the host
+    /// reaches it through a forwarding proxy, so the port number is ignored). An aggregator running
+    /// on the host uses the provided port number. A remote aggregator ignores the port number and
+    /// returns its URL.
     pub fn endpoint_for_host(&self, port: u16) -> Url {
         match self {
-            AggregatorEndpointFragments::VirtualNetwork { path, .. }
-            | AggregatorEndpointFragments::Localhost { path } => {
-                Url::parse(&format!("http://127.0.0.1:{port}{path}")).unwrap()
-            }
-            AggregatorEndpointFragments::DockerNetwork { host, path } => {
+            AggregatorEndpointFragments::VirtualNetwork { host, path } => {
                 Url::parse(&format!("http://{host}:8080{path}")).unwrap()
+            }
+            AggregatorEndpointFragments::Localhost { path } => {
+                Url::parse(&format!("http://127.0.0.1:{port}{path}")).unwrap()
             }
             AggregatorEndpointFragments::Remote { url } => url.clone(),
         }
@@ -86,9 +79,6 @@ impl AggregatorEndpointFragments {
     pub fn endpoint_for_virtual_network(&self) -> Url {
         match self {
             AggregatorEndpointFragments::VirtualNetwork { host, path } => {
-                Url::parse(&format!("http://{host}:8080{path}")).unwrap()
-            }
-            AggregatorEndpointFragments::DockerNetwork { host, path } => {
                 Url::parse(&format!("http://{host}:8080{path}")).unwrap()
             }
             AggregatorEndpointFragments::Localhost { .. } => panic!(
@@ -107,9 +97,6 @@ impl AggregatorEndpointFragments {
             AggregatorEndpointFragments::VirtualNetwork {
                 path: self_path, ..
             }
-            | AggregatorEndpointFragments::DockerNetwork {
-                path: self_path, ..
-            }
             | AggregatorEndpointFragments::Localhost { path: self_path } => *self_path = path,
             AggregatorEndpointFragments::Remote { .. } => {
                 panic!("cannot set path for remote aggregator")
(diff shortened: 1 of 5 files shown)


divergentdave:
Third idea: we could rewrite both the host and port of requests if we configured reqwest clients with a proxy listening on an ephemeral port. Then we could forward proxied requests to Docker's ephemeral port, and we could use port 8080 in the task configuration. This would remove the need for the re-binding retry loop. I don't know how much it would take to set up a proxy, maybe there's an existing crate we could drop in, or maybe it would be a whole production.

jcjones:
> maybe there's an existing crate we could drop in, or maybe it would be a whole production.

I've been giving this a try this morning, and wow this is way easier than I thought it would be. The whole proxy is this:

```rust
 async fn spawn_splice_proxy(upstream_port: u16) -> u16 {
      let listener = TcpListener::bind((Ipv6Addr::LOCALHOST, 0)).await.unwrap();
      let proxy_port = listener.local_addr().unwrap().port();
      tokio::spawn(async move {
          loop {
              let (mut inbound, _) = listener.accept().await.unwrap();
              tokio::spawn(async move {
                  let mut upstream =
                      TcpStream::connect((Ipv6Addr::LOCALHOST, upstream_port)).await.unwrap();
                  let _ = tokio::io::copy_bidirectional(&mut inbound, &mut upstream).await;
              });
          }
      });
      proxy_port
  }
```

divergentdave:
If an HTTP proxy for our purposes is this simple, could we skip the extra Rust TCP listener and Tokio task by having our `Proxy::custom()` object just return `http://127.0.0.1:{docker ephemeral port number}/` instead of `http://127.0.0.1:{splice proxy port number}/`?

integration_tests/src/lib.rs
```diff
@@ -36,12 +36,14 @@ pub struct TaskParameters {
...
 #[derive(Debug)]
 pub enum AggregatorEndpointFragments {
-    /// The aggregator is in a virtual network, (a Docker network or a Kind cluster) so different
-    /// URLs must be used depending on whether it is accessed from within the same virtual network
-    /// or via a port forward on localhost. It is assumed that the port will always be 8080 within
-    /// the virtual network. The port number of forwarded ports will be supplied later. The scheme
-    /// is assumed to always be 'http:'.
+    /// The aggregator is in a Kind cluster, reached from the host via a port forward on localhost
+    /// (ephemeral port, supplied later) and from within the cluster at its service name on port
+    /// 8080. The scheme is always 'http:'.
```
tgeoghegan: nit: this was already wrong, but isn't the scheme just "http"? The ":" is a detail of URL structure.

tgeoghegan: nit: Rust convention is for doccomments to start with a single, short summary sentence and then more detail in following paragraphs. If we're touching comments anyway, then I think it's nice to bring them in line with language conventions.

same observation for the other enum variant.

https://rust-lang.github.io/rfcs/1574-more-api-documentation-conventions.html#summary-sentence
```suggestion
    /// The aggregator is in a Kind cluster.
    ///
    /// The aggregator is reached from the host via a port forward on localhost
    /// (ephemeral port, supplied later) and from within the cluster at its service name on port
    /// 8080. The scheme is always 'http:'.
```

integration_tests/tests/integration/janus.rs
```diff
@@ -63,14 +77,87 @@ impl JanusContainerPair {
...
+            .host_str()
+            .unwrap()
+            .to_string();
+        let (leader_proxy_url, helper_proxy_url) = (leader_proxy.url(), helper_proxy.url());
+        task_parameters.endpoint_fragments.in_process_http_client = Some(
+            reqwest::Client::builder()
+                .proxy(reqwest::Proxy::custom(move |url| match url.host_str() {
+                    Some(host) if host == leader_host => Some(leader_proxy_url.clone()),
+                    Some(host) if host == helper_host => Some(helper_proxy_url.clone()),
+                    _ => None,
```
tgeoghegan: I wonder if this should panic if `url.host_str()` is `Some` value that's anything but the leader or helper host, to noisily catch misconfiguration? Or does this HTTP client ever need to talk to anything besides the helper and leader?

tgeoghegan (approved):
Slick! `tokio::io::copy_bidirectional` seems magical to me.

integration_tests/src/lib.rs
```diff
@@ -36,12 +36,14 @@ pub struct TaskParameters {
 /// Components of one aggregator's DAP endpoint.
 #[derive(Debug)]
 pub enum AggregatorEndpointFragments {
-    /// The aggregator is in a virtual network, (a Docker network or a Kind cluster) so different
-    /// URLs must be used depending on whether it is accessed from within the same virtual network
-    /// or via a port forward on localhost. It is assumed that the port will always be 8080 within
-    /// the virtual network. The port number of forwarded ports will be supplied later. The scheme
-    /// is assumed to always be 'http:'.
+    /// The aggregator is in a Kind cluster, reached from the host via a port forward on localhost
```
divergentdave: I suspect we'll need to do the same thing for Kind clusters as we do for Docker networks, though this will be hard to test until we start making container images. To that end, I think it would make sense to keep this enum as having only three variants, make all the changes we need to get Docker working with the `VirtualNetwork` variant, and assume those will fix the Kind/staging cases as well until we can actually test them way down the line.

jcjones:
> Slick! `tokio::io::copy_bidirectional` seems magical to me.

It is super magic, but as David said, it's not even necessary! Even more simplification coming as soon as my tests pass -- but I'm confident they will!

integration_tests/tests/integration/common.rs
```diff
@@ -63,6 +67,28 @@ pub fn build_test_task(
...
+                    leader: AggregatorEndpointFragments::DockerNetwork {
+                        host: leader_host,
+                        path: "/".to_string(),
+                    },
+                    helper: AggregatorEndpointFragments::DockerNetwork {
+                        host: helper_host,
+                        path: "/".to_string(),
+                    },
+                    ohttp_config: None,
+                    in_process_http_client: None,
```
divergentdave, line 70: We should add a comment stating that this will be filled in with `Some` later in the process.

integration_tests/tests/integration/janus.rs
jcjones: I added the panic in ce5c30039f60c77728d5c75b5b00b115ff458efd

integration_tests/src/lib.rs
jcjones: Good catch. I did these in ce5c30039f60c77728d5c75b5b00b115ff458efd.

integration_tests/tests/integration/janus.rs
```diff
@@ -44,16 +44,18 @@ struct JanusContainerPair {
 #[cfg(feature = "testcontainer")]
 impl JanusContainerPair {
     /// Set up a new pair of containerized Janus test instances, and set up a new task in each using
-    /// the given VDAF and batch mode.
+    /// the given VDAF and batch mode. Each aggregator serves on port 8080; the host-side in-process
+    /// client and collector reach them through per-aggregator forwarding proxies, so all parties
+    /// use the same endpoint strings.
```
divergentdave: nit: outdated docs
```suggestion
    /// the given VDAF and batch mode. Each aggregator serves on port 8080; the host-side in-process
    /// client and collector reach them by using the Docker port forwards as HTTP proxies, so all parties
    /// use the same endpoint strings.
```

integration_tests/tests/integration/janus.rs
```diff
@@ -63,6 +65,37 @@ impl JanusContainerPair {
         let leader = JanusContainer::new(test_name, &network, &task, Role::Leader).await;
         let helper = JanusContainer::new(test_name, &network, &task, Role::Helper).await;
 
+        // The host reaches each aggregator by pointing reqwest's per-request proxy at the
+        // container's Docker-assigned host port; the aggregator serves the absolute-form request
+        // reqwest sends through a proxy, so no forwarding process of our own is needed.
```
divergentdave: I don't understand what this sentence is saying, particularly "absolute-form request".

integration_tests/tests/integration/common.rs
```diff
@@ -29,9 +29,13 @@ use url::Url;
 /// Different contexts or test harnesses that integration tests may be run against, which require
 /// different configuration customizations.
 pub enum TestContext {
-    /// Aggregators are running in a virtual network, like a Docker network or a Kind cluster.
+    /// Aggregators are running in a Kind cluster, reached from the host via port forwards.
     #[allow(dead_code)]
     VirtualNetwork,
+    /// Aggregators are running in a Docker network, each on a distinct port reachable identically
+    /// from within the network and from the host.
```
divergentdave: I think this needs to be updated as well.

integration_tests/tests/integration/janus.rs
jcjones: uhhh, yeah, ::backspace backspace backspace::

integration_tests/tests/integration/common.rs
jcjones: [4c1a199](https://github.com/divviup/janus/pull/4745/commits/4c1a199eaa6d611d089aee0606fcd894711dbe10)

integration_tests/tests/integration/janus.rs
jcjones: [4c1a199](https://github.com/divviup/janus/pull/4745/commits/4c1a199eaa6d611d089aee0606fcd894711dbe10)

integration_tests/src/lib.rs
jcjones: [808ac32](https://github.com/divviup/janus/pull/4745/commits/808ac32fba2023779441df89ad3dab2b217a6ada)

divergentdave (approved):

Issue #4715, opened 2026-07-15
Support Collection Extensions


jcjones:
DAP-18 added CollectionJobExtension type, which Janus does not currently support. Instead, Janus enforces that no Collection Extensions are provided.

This is a tracking issue for collection extensions.

jcjones:
We're deferring this issue until we have a specification of the Differential Privacy collection extension ready, rather than implementing the feature with nothing to test with.

PR #4766, opened 2026-08-03
Persist CollectionJobReq on the Helper, and allow narrower batch selectors

### LINKED ISSUE #4743: Persist Collection Job Req on the Helper
Similar to #4712 (needing to persist verbatim vs reconstruct from parts), the Helper should persist the Collector's `CollectionJobReq` alongside the `AggregateShareJob` rather than reconstructing it.

Today the aggregate-share poll path (which is a GET) rebuilds the `CollectionJobReq` from the stored batch identifier via the `query_for_collection_identifier`. That reconstruction is lossy in two ways (as of PR #4717, anyway): it drops any extensions (because we don't support any), and it derives the query from that batch selector -- which forces them to be identical for the AADs to match.

Persisting the request verbatim in the Helper's database enables two important things:
- Collection Extensions (#4715)
- Batch selectors narrower than the query ("containment"), rather than being required to match exactly. See https://github.com/divviup/janus/pull/4714#discussion_r3658680341 and https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/791.

jcjones:
This _enables_ collection extensions and batch selectors narrower than the query, but only does the work for the latter. Collection extensions are left to #4715 based on being unnecessary at this time.

Resolves #4743

> [!NOTE]
> I changed the existing `AggregateShareRequestRejected` to yield a problem document rather than a blank page, which affects several tests, but feels more right. That said, there's no DAP Problem Type for this (see my note on `http_handlers.rs:189`) and maybe there should be? But also, then I'd need to break `AggregateShareRequestRejected` apart a bit since we're using it for several purposes and now I distinguish them based on the `detail` field.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -3894,6 +3894,27 @@ impl VdafOps {
                         return Ok(aggregate_share_job);
                     }
 
+                    // No job exists for this batch & aggregation parameter, so any job already
+                    // using this aggregate share ID is for a different batch. Reject it: the poll
+                    // path looks jobs up by ID alone and cannot disambiguate.
+                    if tx
+                        .get_aggregate_share_job_by_id::<SEED_SIZE, B, A>(
+                            vdaf.as_ref(),
+                            task.id(),
+                            aggregate_share_id,
+                        )
+                        .await?
+                        .is_some()
+                    {
+                        return Err(datastore::Error::User(
+                            Error::ForbiddenMutation {
+                                resource_type: "aggregate share job",
+                                identifier: aggregate_share_id.to_string(),
+                            }
+                            .into(),
+                        ));
+                    }
+
                     // This is a new aggregate share request, compute & validate the response.
                     debug!(
                         ?aggregate_share_req,
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -3873,25 +3873,16 @@ impl VdafOps {
                     {
                         // DAP requires duplicate requests to be identical.
                         if aggregate_share_job.aggregate_share_id() != &aggregate_share_id
+                            || aggregate_share_job.collection_job_req()
+                                != aggregate_share_req.collection_job_req()
                         {
                             return Err(datastore::Error::User(
-                                Error::AggregateShareRequestRejected(
-                                    *task.id(),
-                                    "aggregate share request is a duplicate but uses a different aggregate share ID"
+                                Error::ForbiddenMutation {
+                                    resource_type: "aggregate share job",
+                                    identifier: aggregate_share_job
+                                        .aggregate_share_id()
                                         .to_string(),
-                                )
-                                .into(),
-                            ));
-                        }
-                        if aggregate_share_job.collection_job_req()
-                            != aggregate_share_req.collection_job_req()
-                        {
-                            return Err(datastore::Error::User(
-                                Error::AggregateShareRequestRejected(
-                                    *task.id(),
-                                    "aggregate share request is a duplicate but carries a different collection job request"
-                                        .to_string(),
-                                )
+                                }
                                 .into(),
                             ));
                         }
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -3636,18 +3636,7 @@ impl VdafOps {
             &AggregateShareAad::new(
                 *task.id(),
                 task.task_configuration()?,
-                // On the poll path we no longer have the Collector's original CollectionJobReq
-                // (there is no incoming AggregateShareReq), so reconstruct it from the stored job.
-                // This must be byte-identical to the request the Collector sent; see
-                // `query_for_collection_identifier` (and Issue #4743). When we support
-                // extensions (Issue #4715) this will need to change.
-                CollectionJobReq::new(
-                    B::query_for_collection_identifier(aggregate_share_job.batch_identifier()),
-                    aggregate_share_job
-                        .aggregation_parameter()
-                        .get_encoded()
-                        .map_err(Error::MessageEncode)?,
-                ),
+                aggregate_share_job.collection_job_req().clone(),
             )
             .get_encoded()
             .map_err(Error::MessageEncode)?,
@@ -3811,14 +3800,11 @@ impl VdafOps {
         )?;
 
         // DAP-19 §4.6.4: the Helper verifies the Leader-chosen batch selector against the query in
-        // the CollectionJobReq, which the AAD binds. The spec defines consistency per batch mode
-        // and would permit any time-interval selector within the queried interval, but Janus
-        // requires exact round-tripping, otherwise the poll path rebuilds the AAD's query from
-        // the stored selector alone, so a narrower selector would decrypt here and fail there.
-        if B::query_for_collection_identifier(
+        // the CollectionJobReq, which the AAD binds.
+        if !B::is_batch_identifier_consistent_with_query(
             aggregate_share_req.batch_selector().batch_identifier(),
-        ) != *aggregate_share_req.collection_job_req().query()
-        {
+            aggregate_share_req.collection_job_req().query(),
+        ) {
             return Err(Error::BatchInvalid(
                 *task.id(),
                 format!(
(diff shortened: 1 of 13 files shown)


messages/src/tests/collection.rs
```diff
@@ -187,6 +187,66 @@ fn collection_job_req_decode_is_lenient_about_extension_order() {
     assert_eq!(req.extensions().len(), 2);
 }
 
+#[test]
+fn collection_job_req_decode_encode_is_byte_identical() {
```
tgeoghegan: Did `roundtrip_collection_job_req` not already prove this? Or perhaps `roundtrip_encoded` could be extended slightly to check that for all types?

aggregator_core/src/batch_mode.rs
```diff
@@ -474,4 +463,38 @@ mod tests {
...
+        ] {
+            assert!(TimeInterval::is_batch_identifier_consistent_with_query(
+                &selector, &query
+            ));
+        }
+
+        for selector in [
+            interval(99, 10),  // starts before the query
+            interval(105, 10), // ends after the query
+            interval(110, 1),  // entirely after the query
```
tgeoghegan, line 493: Should we have a case for entirely before the query?

aggregator_core/src/datastore/models.rs
```diff
@@ -2038,12 +2038,16 @@ pub struct AggregateShareJob<const SEED_SIZE: usize, B: BatchMode, A: AsyncAggre
     /// Checksum over the aggregated report shares, as described in §4.4.4.3.
     #[educe(Debug(ignore))]
     checksum: ReportIdChecksum,
+    /// The collector's request, as forwarded by the leader. Retained verbatim because the
+    /// aggregate share AAD binds it, and the poll path has to rebuild that AAD.
```
tgeoghegan: nit: per that Rust doccomment RFC this would look better as a summary line and then details in a follow on paragraph:

```suggestion
    /// The collector's request, as forwarded by the leader.
    ///
    /// Retained verbatim because the aggregate share AAD binds it,
    /// and the poll path has to rebuild that AAD.
```

But we don't have a linter for this which means we're not _really_ serious about enforcing it so I leave it up to you whether to deal with it.

aggregator/src/aggregator/http_handlers.rs
```diff
@@ -186,7 +186,16 @@ impl Error {
             | Error::HttpClient(_)
             | Error::Http { .. }
             | Error::TaskParameters(_) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
-            Error::AggregateShareRequestRejected(_, _) => StatusCode::BAD_REQUEST.into_response(),
+            // Note: DAP defines no error for this; `batchOverlap` may kinda fit the dupe
+            // cases, but it does not cover the too-late rejection that shares this variant.
```
tgeoghegan: I believe this was a case where we (DAP editors) felt that HTTP error semantics were sufficient and we didn't need a DAP-specific problem type. We have an existing `Error::ForbiddenMutation` used when Janus detects attempts to mutate aggregation jobs or collection jobs that I think would work here. We then serve an [HTTP 409 Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/409) which seems right to me.

However I do support constructing a problem details. Our precedent here has been to construct a document with a link to [the errors page on docs.divviup.org](https://docs.divviup.org/references/janus-errors). [for example](https://github.com/divviup/janus/blob/7593bd3d1cba5e4ef42277cd71a3b1b6928ec040/aggregator/src/aggregator/http_handlers.rs#L135).  `Error::ForbiddenMutation` contains:

```
{
    resource_type: &'static str,
    identifier: String,
}
```

...so you could use those values to render a nice explanation in the problem details as well as the link.

divergentdave: I think using `about:blank` as a URI would be fine as well, this case seems like it's fine to categorize as just "bad request".

messages/src/tests/collection.rs
jcjones: I'll add a case to roundtrip_encoded, good call.

aggregator/src/aggregator.rs
```diff
@@ -3885,20 +3871,18 @@ impl VdafOps {
                         )
                         .await?
                     {
-                        // Duplicate aggregate share job found - verify the aggregate share ID
-                        // matches. Note, when we add collection extensions, those will need
-                        // to be checked here as well. (Issue #4715)
+                        // DAP requires duplicate requests to be identical.
                         if aggregate_share_job.aggregate_share_id() != &aggregate_share_id
+                            || aggregate_share_job.collection_job_req()
+                                != aggregate_share_req.collection_job_req()
```
divergentdave, line 3877: We should check the batch selector as well, just in case. (The report count and checksum are okay to skip, as those are a diagnostic tool in the first place)

divergentdave (approved):

aggregator/src/aggregator/http_handlers/tests/aggregate_share.rs
```diff
@@ -822,10 +825,23 @@ async fn aggregate_share_request_duplicate_with_different_id() {
...
+    let mut response =
         put_aggregate_share_request(&task, &request, &aggregate_share_id_2, &router).await;
 
-    assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+    assert_eq!(response.status(), StatusCode::CONFLICT);
+    assert_eq!(
+        take_problem_details(&mut response).await,
+        json!({
+            "status": StatusCode::CONFLICT.as_u16(),
+            "type": "https://docs.divviup.org/references/janus-errors#forbidden-mutation",
```
tgeoghegan, line 840: I guess a PR on [divviup/public-docs](https://github.com/divviup/public-docs) is forthcoming to add this? No need to block this PR, FWIW.

tgeoghegan (approved):

aggregator/src/aggregator/http_handlers/tests/aggregate_share.rs
jcjones: Yeah, I have that started.

aggregator/src/aggregator.rs
jcjones: The literal check is a no-op, I think. `BatchSelector<B>` wraps nothing but `B::BatchIdentifier` so comparing the selectors is comparing the identifiers. 

But, uh `aggregate_share_id` doesn't have anything ensuring its uniqueness.  The dedup lookup is keyed on `batch`+`param`, so a leader that PUTs the same aggregate share ID against a different batch identifier doesn't hit our dedup path at all; it creates a second row sharing that ID. 

The poll path then does `WHERE aggregate_share_id = $2` via `query_opt`, which would return error for more than one row. Which becomes a 500.

I guess it's always been this way, but this PR makes the exposure a bit wider. Before, the helper derived the query from the stored batch identifier. A batch identifier had to equal its query interval exactly.  Now that narrower selectors are consistent, one `collection_job_req` legitimately maps to many batch identifiers, so duping IDs is a more ... achievable? error condition.

I feel like I should guard against this on the cache miss scenario and do `ForbiddenMutation`, but I haven't figured out where yet.

divergentdave: I was referring to the `batch_selector` field of `AggregateShareReq`. That may be different than what is wrapped in the collection job request's query.

Ah, some of this may be left over from before aggregate shares had their own IDs. Doing a separate check by ID for an existing aggregate share would make sense I think.

I think we should also ensure that if a leader sends two identical aggregate share requests under different IDs, we send the same ciphertext in response to the second one instead of returning an error for overlapping batches. This could be useful for recovering from operational issues on the leader side without unnecessary data loss.

jcjones: Perhaps a real fix would be to add a `UNIQUE(task_id, aggregate_share_id)` constraint to the `aggregate_share_jobs` table.

jcjones: https://github.com/divviup/janus/issues/4768

jcjones: > I was referring to the `batch_selector` field of `AggregateShareReq`. That may be different than what is wrapped in the collection job request's query.

Reasonably confident here that that would still be doing a comparison against its own clone. The case I think you're guarding against -- same query, different selector -- isn't going to make it here. A different selector is a different key, so it'll go down the cache-miss path.

I'm adding alookup by aggregate share ID there, which rejects binding one ID to two batches (with a 409), and _that_ has turned up three existing tests that are re-using a single ID across different batches, all of which would have caused 500s if they hit the poll path. Which they didn't.

I'm also adding a test `aggregate_share_request_same_id_different_batch` to explicitly check this. Take a look at afefe0e9.

messages/src/tests/collection.rs
jcjones: in afefe0e9

aggregator/src/aggregator/http_handlers/tests/aggregate_share.rs
jcjones: https://github.com/divviup/public-docs/pull/196

Issue #4712, opened 2026-07-14
Rethink TaskConfiguration source-of-truth


jcjones:
I saw in other comments that the `TaskConfiguration` for Taskprov tasks "is not byte-reconstructible from the stored parameters", necessitating this column, because "it would drop DP strategies and unknown extensions". It seems like we have two different sources of truth for the `TaskConfiguration`, and we switch between them depending on provisioning mechanism. This seems potentially fraught -- could the same issues that keep us from reproducing a Taskprov task's config also affect other tasks? I think it would be preferable if we stick to one mechanism for producing the `TaskConfiguration`, either storing it or reconstructing it in a content-preserving way. Doing both seems like it's the worst of both worlds, as we both have to be careful about our data model mapping 1-1 to and from the `TaskConfiguration` bytes, and we have an extra set of code paths on top of that.

Always saving and using the serialized `TaskConfiguration` seems like it would be the easiest to maintain. If we were to go the other route, I think I'd want us to have a fuzzer that deserializes arbitrary TaskConfiguration objects, constructs a task from it as we would do in Taskprov, calls `task_configuration()`, encodes the result, and checks that it is equal to the original fuzzer input.

To address the specific reasons given, first, the differential privacy configuration used to be in older versions of Taskprov, but it has been removed from more recent ones. Thus, our differential privacy strategy field is an implementation-specific task parameter for the time being, until an updated differential privacy draft gets published to set out how to track this as a task extension. At that point, we could re-base our differential privacy support on that extension, and ensure agreement on the differential privacy mechanism between all parties. In the meantime, I think it should be fine for us to drop the dp_strategy field when serializing a `TaskConfiguration` struct. Second, I think unknown extensions in a Taskprov HTTP header ought to cause us to opt out of a task. The editor's copy says we MUST opt out if "the DAP batch mode or VDAF is not implemented" -- lack of support for a task extension should logically be a reason to opt out as well, since DAP says "extensions are mandatory to support." We ought to refuse to aggregate, rather than carry two conflicting representations of the task's extensions.

_Originally posted by @divergentdave in https://github.com/divviup/janus/pull/4698#discussion_r3574601149_

jcjones:
[...]

PR #4781, opened 2026-08-06
[DAP-17] Move TaskConfiguration Source of Truth (Part 1)

jcjones:
This is two independent commits, which independently result in green tests, albeit with a (temporary) production panic while I get the next pieces in place.

First, a small change to opt-out of TaskProv tasks with unsupported extensions. This is needed anyway, I just sorta stumbled on this while I was looking at the rest of the job.

Second, a big mechanical change to make `AggregatorTask::task_configuration()` infallible, because when it comes from the database, it will be. That said, at the conclusion of this it _isn't_, hence the prospect of panics. But the next part will do the database work.

### [DAP-17] Opt out of taskprov tasks with unsupported extensions

DAP task extensions are mandatory, so a TaskConfiguration
carrying an extension we don't implement is a reason to opt out rather
than aggregate under a task we only sort-of understand.

The check goes in taskprov_opt_in rather than taskprov_authorize_request.
Authorize_request runs on every taskprov request, which would reject
tasks already in the datastore, which seems to not be what we want.

Part of https://github.com/divviup/janus/issues/4712.

### [DAP-17] Make AggregatorTask::task_configuration() infallible

Returns TaskConfiguration by value rather than Result. This is part of
changing TaskConfiguration to be a value stored in the database rather
than one generated as needed.

This is not a change intended to stick around. I'm adding `expect()`s to
production code (OH NO!!!), which will get removed in the next PR.

But since this is kinda large and mechanical, let's keep it standalone
and stack the next PR on it.

So, two of the four error arms were already unreachable and are removed:
role() yields only Leader or Helper, and task_info length is validated at
construction, so build_task_configuration cannot fail for a task that
exists.

The remaining two become expect()s: a taskprov task missing its verbatim
config, and a VDAF that to_vdaf_config() rejects. These expects won't be
there when the TaskConfiguration is stored on the task, which is the next
PR.

Part of https://github.com/divviup/janus/issues/4712.

### DIFF
--- a/aggregator/src/aggregator/error.rs
+++ b/aggregator/src/aggregator/error.rs
@@ -303,6 +303,8 @@ pub enum OptOutReason {
     /// Catch-all error for generally invalid parameters.
     #[error("invalid parameter: {0}")]
     InvalidParameter(String),
+    #[error("unsupported task extension: {0:#06x}")]
+    UnsupportedExtension(u16),
 }
 
 impl Error {
--- a/aggregator/src/aggregator/taskprov_tests.rs
+++ b/aggregator/src/aggregator/taskprov_tests.rs
@@ -4,7 +4,10 @@ use assert_matches::assert_matches;
 use axum::body::Body;
 use base64::{Engine, engine::general_purpose::URL_SAFE_NO_PAD};
 use chrono::TimeDelta;
-use http::{Request, StatusCode};
+use http::{
+    Request, StatusCode,
+    header::{AUTHORIZATION, CONTENT_TYPE},
+};
 use janus_aggregator_core::{
     AsyncAggregator,
     datastore::{
@@ -326,9 +329,9 @@ async fn taskprov_aggregate_init() {
                             .unwrap()
                             .path(),
                     )
-                    .header(http::header::AUTHORIZATION, "Bearer invalid_token")
+                    .header(AUTHORIZATION, "Bearer invalid_token")
                     .header(
-                        http::header::CONTENT_TYPE,
+                        CONTENT_TYPE,
                         AggregationJobInitializeReq::<LeaderSelected>::MEDIA_TYPE,
                     )
                     .header(
@@ -356,7 +359,7 @@ async fn taskprov_aggregate_init() {
                     )
                     .with_authentication_token(test.peer_aggregator.primary_aggregator_auth_token())
                     .header(
-                        http::header::CONTENT_TYPE,
+                        CONTENT_TYPE,
                         AggregationJobInitializeReq::<LeaderSelected>::MEDIA_TYPE,
                     )
                     .header(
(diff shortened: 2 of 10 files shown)


aggregator_core/src/task.rs
```diff
@@ -387,37 +387,42 @@ impl AggregatorTask {
...
+            "taskprov task is missing its TaskConfiguration"
+        );
 
         let (leader_endpoint, helper_endpoint) = match self.role() {
             Role::Leader => (
                 &self.own_aggregator_endpoint,
                 &self.peer_aggregator_endpoint,
             ),
-            Role::Helper => (
+            // `role()` yields only Leader or Helper.
```
divergentdave: I'd suggest leaving the `Role::Helper` branch alone and changing the default branch to panic instead of return an error. Getting a non-aggregator role from `AggregatorTask::role()` would be a programming error, so I think it's fine to panic in that situation. This would be preferable for silently continuing with an arbitrary pair of endpoints. We could also add a method that returns `AggregatorRole` to remove fallibility entirely, or change the return type of `role()`, but we don't need to do that here.

divergentdave (approved):

aggregator_core/src/task.rs
jcjones: Now that you mention it, we do it that way elsewhere, too, don't we? OK, good call.

aggregator/src/aggregator/error.rs
```diff
@@ -303,6 +303,8 @@ pub enum OptOutReason {
     /// Catch-all error for generally invalid parameters.
     #[error("invalid parameter: {0}")]
     InvalidParameter(String),
+    #[error("unsupported task extension: {0:#06x}")]
```
tgeoghegan, line 306: huh, TIL the length of `0x` is included in the padding length

aggregator/src/aggregator/taskprov_tests.rs
```diff
@@ -1000,6 +1001,97 @@ async fn taskprov_opt_out_peer_aggregator_does_not_exist() {
...
+        .clone()
+        .oneshot(
+            Request::builder()
+                .method("PUT")
+                .uri(format!(
+                    "/tasks/{another_task_id}/aggregation_jobs/{aggregation_job_id}"
+                ))
+                .with_authentication_token(test.peer_aggregator.primary_aggregator_auth_token())
+                .header(
+                    http::header::CONTENT_TYPE,
```
tgeoghegan: nit: we could `use http::header::CONTENT_TYPE;` above. I think it's clear we're talking about a header.

tgeoghegan (approved):
Love to see a bunch of `.unwrap()` go away!

aggregator/src/aggregator/error.rs
jcjones: I was channeling my pythonisms and it just happened to work.

aggregator/src/aggregator/taskprov_tests.rs
jcjones: I was following the other ~25 sites in the file, but why not just fix it? so, fixing it: f177da36

PR #4782, opened 2026-08-07
[DAP-17] Store and Recall TaskConfiguration from DB (Part 2)

### LINKED ISSUE #4712: Rethink TaskConfiguration source-of-truth

jcjones:
This is the transition from TaskProv TaskConfig to just plain **TaskConfig**. It becomes a not-NULL field in the Tasks table. Accordingly I rework how AggregatorTask construction works to provide a path for "TaskConfig is provided" and one for "TaskConfig is synthesized" -- and once set, it goes into the Tasks table and gets loaded upon future use.

In a lot of ways this is a very easy change at this point: When we're getting a task from the db, we load it. We already did all the hard work.

> [!NOTE]
> I tried to write some comments for us in the future about the different starting points for Task Configs at the relevant places, but I'm not sure if they're actually helpful enough to keep. I'm tempted to strip them back down. Your opinions are solicited! 

Closes #4712.

### DIFF
--- a/aggregator_core/src/task.rs
+++ b/aggregator_core/src/task.rs
@@ -268,11 +268,6 @@ pub struct AggregatorTask {
     /// The task's `TaskConfiguration`, bound verbatim into HPKE AADs. Adopted from the wire for
     /// taskprov tasks and synthesized once at provisioning time for API-provisioned tasks; either
     /// way it is frozen in the database and never re-synthesized on read.
-    ///
-    /// Persisted only via the datastore, not [`SerializedAggregatorTask`] (like `deactivate_at`);
-    /// deserializing that form re-synthesizes it via [`AggregatorTask::new`]. Safe because that
-    /// form accepts only the Leader and Helper roles, so a taskprov task — whose config must stay
-    /// verbatim — can never be provisioned through it.
     task_config: TaskConfiguration,
 }
 
--- a/aggregator_core/src/task.rs
+++ b/aggregator_core/src/task.rs
@@ -265,18 +265,21 @@ pub struct AggregatorTask {
     /// Deactivation instant. When set and reached (per the aggregator's clock), the
     /// aggregator stops accepting reports for this task. This is Janus-specific state.
     deactivate_at: Option<DateTime<Utc>>,
-    /// For taskprov tasks, the `TaskConfiguration` exactly as received on the wire, bound verbatim
-    /// into HPKE AADs. `None` for API-provisioned tasks, whose `TaskConfiguration` is synthesized
-    /// from the stored parameters. Reconstructing a taskprov config from those parameters is not
-    /// byte-safe (it would drop DP strategies and unknown extensions), so the wire bytes are kept.
+    /// The task's `TaskConfiguration`, bound verbatim into HPKE AADs. Adopted from the wire for
+    /// taskprov tasks and synthesized once at provisioning time for API-provisioned tasks; either
+    /// way it is frozen in the database and never re-synthesized on read.
     ///
     /// Persisted only via the datastore, not [`SerializedAggregatorTask`] (like `deactivate_at`);
-    /// taskprov tasks are never provisioned through that serde form.
-    taskprov_task_config: Option<TaskConfiguration>,
+    /// deserializing that form re-synthesizes it via [`AggregatorTask::new`]. Safe because that
+    /// form accepts only the Leader and Helper roles, so a taskprov task — whose config must stay
+    /// verbatim — can never be provisioned through it.
+    task_config: TaskConfiguration,
 }
 
 impl AggregatorTask {
-    /// Create a new [`AggregatorTask`] with the provided values.
+    /// Create a new [`AggregatorTask`] with the provided values, synthesizing its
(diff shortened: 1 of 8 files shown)


aggregator_core/src/task.rs
```diff
@@ -265,18 +265,21 @@ pub struct AggregatorTask {
...
+    /// The task's `TaskConfiguration`, bound verbatim into HPKE AADs. Adopted from the wire for
+    /// taskprov tasks and synthesized once at provisioning time for API-provisioned tasks; either
+    /// way it is frozen in the database and never re-synthesized on read.
     ///
     /// Persisted only via the datastore, not [`SerializedAggregatorTask`] (like `deactivate_at`);
-    /// taskprov tasks are never provisioned through that serde form.
-    taskprov_task_config: Option<TaskConfiguration>,
+    /// deserializing that form re-synthesizes it via [`AggregatorTask::new`]. Safe because that
+    /// form accepts only the Leader and Helper roles, so a taskprov task — whose config must stay
+    /// verbatim — can never be provisioned through it.
```
divergentdave: I think this last sentence is a little backwards. Taskprov's task provisioning is trivially safe because we take in the `TaskConfiguration` as an encoded blob, store it, and always use it in AADs. For our other task provisioning mechanisms, which use `SerializedAggregatorTask` or `PostTaskReq`, it is on us to make sure that the same inputs lead to the same `TaskConfiguration` (or an error) in both aggregators and in what we provide for use on the clients. So, we could say that the `TaskConfiguration` does not appear in a `SerializedAggregatorTask`, but is implicit in its fields and our processing rules.

divergentdave (approved):

aggregator_core/src/task.rs
jcjones: When I was writing this comment, I am sure I was trying to make it the change make sense to myself -- but coming at this fresh, the second paragraph can just drop. For anyone arriving tomorrow, this field gets stored in the DB and gets read from the DB. How it gets there in the first place -- whether it was on the wire or constructed from bits -- is immaterial.

jcjones: e3d56596

aggregator_core/src/task.rs
```diff
@@ -265,18 +265,16 @@ pub struct AggregatorTask {
...
+    /// taskprov tasks and synthesized once at provisioning time for API-provisioned tasks; either
+    /// way it is frozen in the database and never re-synthesized on read.
+    task_config: TaskConfiguration,
 }
 
 impl AggregatorTask {
-    /// Create a new [`AggregatorTask`] with the provided values.
+    /// Create a new [`AggregatorTask`] with the provided values, synthesizing its
+    /// [`TaskConfiguration`]. This is the provisioning constructor; a task adopting a
+    /// caller-supplied configuration uses [`AggregatorTask::new_with_task_config`] instead.
```
tgeoghegan, line 277: As David argued elsewhere, I think we should make the aggregator API deal in `TaskConfiguration` so that there's a single path through all this. Do we have an issue for that?

tgeoghegan (approved):

aggregator_core/src/task.rs
divergentdave: I created #4798 for this.

Issue #4711, opened 2026-07-14
Deduplicate `VdafConfig` and the `Prio3::new_histogram()` into a builder


jcjones:
This duplication between the `VdafConfig` and the `Prio3::new_histogram()` arguments is a little unfortunate. After we get things working, maybe we should come back here and add convenience constructors/builder methods for well-known VDAF types?

_Originally posted by @divergentdave in https://github.com/divviup/janus/pull/4698#discussion_r3575035232_

PR #4793, opened 2026-08-10
Deduplicate VdafConfig and VDAF constructor arguments (#4711)

jcjones:
As David pointed out (see #4711), building a client or collector required stating the VDAF's parameters twice: once to `Prio3::new_histogram(2, 12, 4)` and again as `VdafConfig::Prio3Histogram { length: 12, chunk_length: 4 }`. And a mismatch wasn't not a compile error, just a silent HPKE AAD byte-identity bug that surfaces as a decryption failure at the aggregator.

This adds `ConfiguredVdaf<V>` -- it holds a concrete VDAF alongside the `VdafConfig` describing it, with `prio3_*` constructors that build both from one set of parameters.

It also adds `Client::builder_from_configured_vdaf`, `Collector::builder_from_configured_vdaf`, and `CollectorBuilder::from_configured_vdaf`.

`with_vdaf_config` remains for the configs with no concrete VDAF to derive from.

Note this uncovered two existing mismatches: the client's `setup_client` and the collector's `setup_collector` test helpers both hardcoded `VdafConfig::Prio3Count` while callers passed Prio3Sum, Prio3Histogram, and `dummy::Vdaf`. Oops.

### DIFF
--- a/client/src/lib.rs
+++ b/client/src/lib.rs
@@ -7,35 +7,29 @@
 //! # Examples
 //!
 //! ```no_run
-//! use prio::vdaf::prio3::Prio3Histogram;
-//! use janus_messages::{BatchConfig, TimePrecision, TaskId, Url, VdafConfig};
+//! use janus_core::vdaf::ConfiguredVdaf;
+//! use janus_messages::{BatchConfig, TimePrecision, TaskId, Url};
 //! use std::str::FromStr;
 //!
 //! #[tokio::main]
 //! async fn main() {
 //!     let leader_url = Url::try_from("https://leader.example.com/").unwrap();
 //!     let helper_url = Url::try_from("https://helper.example.com/").unwrap();
-//!     let vdaf = Prio3Histogram::new_histogram(
-//!         2,
-//!         12,
-//!         4
-//!     ).unwrap();
 //!     let taskid = "rc0jgm1MHH6Q7fcI4ZdNUxas9DAYLcJFK5CL7xUl-gU";
 //!     let task = TaskId::from_str(taskid).unwrap();
 //!
 //!     // The task parameters below are bound into HPKE AADs and MUST match those provisioned to the
 //!     // aggregators byte-for-byte.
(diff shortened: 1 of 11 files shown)


client/src/lib.rs
```diff
@@ -549,6 +544,26 @@ impl<V: vdaf::Client<16>> Client<V> {
         )
     }
 
+    /// Creates a [`ClientBuilder`] from the required set of DAP task parameters and a
+    /// [`ConfiguredVdaf`], which supplies both the VDAF and its [`VdafConfig`].
+    pub fn builder_from_configured_vdaf(
```
divergentdave: This should be the main entry point that people use, so I suggest renaming methods so that this is `Client::builder()`, and the one above as `Client::builder_with_custom_vdaf()` or similar.

client/src/lib.rs
divergentdave: We could add a `VdafConfig` to the arguments here, since it is always required, and it should be tightly coupled to the `vdaf: V` argument. Then, we could get rid of the `with_vdaf_config()` method on the builder.

collector/src/lib.rs
divergentdave: Similarly, I suggest adding a `VdafConfig` argument to this, renaming `builder()` to `builder_with_custom_vdaf()` or similar, and `builder_from_configured_vdaf()` to `builder()`.

divergentdave (approved):

tgeoghegan (approved):

client/src/lib.rs
jcjones: Oh, I should have thought of that -- but I am so glad you did!

client/src/lib.rs
jcjones: fd34ddea

client/src/lib.rs
jcjones: fd34ddea

collector/src/lib.rs
jcjones: fd34ddea

divergentdave (approved):

PR #4795, opened 2026-08-10
Ignore smartstring RUSTSEC advisory. (release/0.7 backport)

jcjones:
It's part of Trillium, which we'll soon be rid of.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -29,8 +29,7 @@ ignore = [
     # See <https://github.com/fussybeaver/bollard/issues/612>.
     "RUSTSEC-2025-0134",
 
-    # 'smartstring' is part of Trillium. It goes away once divviup-client is migrated
-    # off Trillium.
+    # 'smartstring' is part of Trillium, and it's unmaintained.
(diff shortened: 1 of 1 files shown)


deny.toml
```diff
@@ -28,6 +28,10 @@ ignore = [
     # We have migrated from `rustls-pemfile` but not everything else has.
     # See <https://github.com/fussybeaver/bollard/issues/612>.
     "RUSTSEC-2025-0134",
+
+    # 'smartstring' is part of Trillium. It goes away once divviup-client is migrated
+    # off Trillium.
```
divergentdave: I don't know that we plan to migrate this release branch from Trillium to Axum. We should instead just note that this dependency is unmaintained, and keep an eye on things.

divergentdave (approved):

deny.toml
jcjones:
```suggestion
    # 'smartstring' is part of Trillium, and it's unmaintained.
```

PR #4799, opened 2026-08-14
Replace DAP task parameter flags with --task-config

### LINKED ISSUE #4713: Add `--task-config <blob>`?

jcjones:
`--task-config` takes the task's base64url TaskConfiguration and supplies the aggregator endpoints, time precision, batch configuration, and VDAF configuration, so the tool can reconstruct the AAD the aggregators bind.

This gets the collector, both CLI and the interop version.

Closes https://github.com/divviup/janus/issues/4713.

### DIFF
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -34,12 +34,12 @@
 //! // aggregators byte-for-byte.
 //! let collector = Collector::builder(
 //!     task_id,
-//!     leader_url,
 //!     collector_credential.authentication_token(),
 //!     collector_credential.hpke_keypair(),
 //!     ConfiguredVdaf::prio3_count().unwrap(),
-//!     time_precision,
 //! )
+//! .with_leader_endpoint(leader_url)
+//! .with_time_precision(time_precision)
 //! .with_helper_endpoint(helper_url)
 //! .with_task_info(b"[task info]".to_vec())
 //! .with_min_batch_size(1000)
@@ -348,16 +348,20 @@ where
 pub struct CollectorBuilder<V: vdaf::Collector> {
     /// Unique identifier for the task.
     task_id: TaskId,
-    /// The base URL of the leader's aggregator API endpoints.
-    leader_endpoint: DapUrl,
     /// The authentication information needed to communicate with the leader aggregator.
     authentication: AuthenticationToken,
     /// HPKE keypair used for decryption of aggregate shares.
     hpke_keypair: HpkeKeypair,
     /// An implementation of the task's VDAF.
     vdaf: V,
-    /// The task's time precision.
-    time_precision: TimePrecision,
+    /// The base URL of the leader's aggregator API endpoints, bound into the task's
+    /// `TaskConfiguration`. Required before [`Self::build`] unless
+    /// [`Self::with_task_configuration`] supplies it; set via [`Self::with_leader_endpoint`].
+    leader_endpoint: Option<DapUrl>,
+    /// The task's time precision, bound into the task's `TaskConfiguration`. Required before
+    /// [`Self::build`] unless [`Self::with_task_configuration`] supplies it; set via
+    /// [`Self::with_time_precision`].
+    time_precision: Option<TimePrecision>,
     /// The base URL of the helper's aggregator API endpoints. Required before [`Self::build`]; set
     /// via [`Self::with_helper_endpoint`].
     helper_endpoint: Option<DapUrl>,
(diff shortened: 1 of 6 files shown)


collector/src/lib.rs
```diff
@@ -542,9 +602,6 @@ pub struct Collector<V: vdaf::Collector> {
     /// The base URL of the leader's aggregator API endpoints.
     #[educe(Debug(method(std::fmt::Display::fmt)))]
     leader_endpoint: DapUrl,
```
divergentdave: Could we remove this field and use `self.task_configuration.leader_aggregator_endpoint()` instead? It looks like we just use the `leader_endpoint` field to build URLs, and the builder field gets validated against the corresponding TaskConfiguration field when available.

interop_binaries/src/commands/janus_interop_collector.rs
```diff
@@ -132,12 +133,10 @@ struct TaskState {
     task_id: TaskId,
     keypair: HpkeKeypair,
     leader_url: DapUrl,
-    helper_url: DapUrl,
     vdaf: VdafObject,
     auth_token: AuthenticationToken,
     time_precision: TimePrecision,
```
divergentdave: I think `leader_url` and `time_precision` could be pruned here as well.

collector/src/lib.rs
divergentdave, line 646: For these builder methods, I think we could drop the `leader_endpoint` and `time_precision` arguments, since we now have two ways we can provide them. We would need to add some new setter methods and add more checks in `CollectorBuilder::build()` for the newly optional fields. This would make setting up a collector easier if you are starting out with the task configuration, because you'd just need to provide the TaskId, TaskConfiguration, authentication material, and collector private key, plus something for the generic type variable. This would be about as simple as we could make it.

tools/src/bin/collect.rs
```diff
@@ -56,34 +57,15 @@ impl From<clap::Error> for Error {
...
+    fn from(error: VdafError) -> Self {
+        Error::Anyhow(error.into())
+    }
 }
 
 // Parsers for command-line arguments:
 
-#[derive(Debug, Clone, Copy, ValueEnum, PartialEq, Eq)]
-#[clap(rename_all = "lower")]
-enum VdafType {
```
divergentdave, line 76: This gets us Prio3MultihotCountVec support for free now!

divergentdave (approved):

collector/src/lib.rs
jcjones: In 960494f. Great idea!

interop_binaries/src/commands/janus_interop_collector.rs
jcjones: Sure enough. In 960494f.

collector/src/lib.rs
jcjones: Yeah, done in 960494f :)

collector/src/lib.rs
```diff
@@ -418,66 +424,110 @@ impl<V: vdaf::Collector> CollectorBuilder<V> {
...
+        })
+    }
+
+    /// Resolves the task's [`TaskConfiguration`], either from whatever was supplied to
+    /// [`Self::with_task_configuration`] or by composing one from the individual setters.
+    fn resolve_task_configuration(&self) -> Result<TaskConfiguration, Error> {
+        let Some(task_configuration) = &self.task_configuration else {
+            return Ok(build_task_configuration(
+                self.task_info
+                    .clone()
```
tgeoghegan, line 466: All the `clone`s are unfortunate since the single calling method takes `self` by value anyway. I think if you inlined this function into the caller, then you'd be able to take the various fields by value instead of cloning references.

tgeoghegan (approved):

collector/src/lib.rs
jcjones: Opened https://github.com/divviup/janus/issues/4800 as a follow-on. Thanks!

PR #4818, opened 2026-08-19
[DAP-18] Remove PartialBatchSelector from CollectionJobResp

### LINKED ISSUE #4783: Remove `PartialBatchSelector` from `CollectionJobResp`
`messages/src/lib.rs:2036` still carries it; the spec struct is now `{report_count, interval, leader_encrypted_agg_share, helper_encrypted_agg_share}`.

https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/778

jcjones:
The DAP spec dropped the field (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/778); the response is now `{report_count, interval, leader_encrypted_agg_share, helper_encrypted_agg_share}`. Cascades into the collector's public Collection type, the collect CLI's batch-id output, and the interop collector, which can no longer observe a leader-selected batch ID from the response.

Fixes #4783

### DIFF
--- a/collector/src/lib.rs
+++ b/collector/src/lib.rs
@@ -74,6 +74,7 @@ mod credential;
 
 use std::{
     convert::TryFrom,
+    marker::PhantomData,
     time::{Duration as StdDuration, SystemTime},
 };
 
@@ -94,8 +95,8 @@ use janus_core::{
 };
 use janus_messages::{
     AggregateShareAad, BatchConfig, CollectionJobExtension, CollectionJobId, CollectionJobReq,
-    CollectionJobResp, Interval, MediaType, PartialBatchSelector, Query, Role, TaskConfiguration,
-    TaskId, TimePrecision, Url as DapUrl, VdafConfig,
+    CollectionJobResp, Interval, MediaType, Query, Role, TaskConfiguration, TaskId, TimePrecision,
+    Url as DapUrl, VdafConfig,
     batch_mode::{BatchMode, TimeInterval},
 };
 use prio::{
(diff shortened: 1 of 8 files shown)


messages/src/lib.rs
```diff
@@ -2033,11 +2034,11 @@ impl Distribution<CollectionJobId> for StandardUniform {
 /// aggregate shares for a given query.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct CollectionJobResp<B: BatchMode> {
-    pub partial_batch_selector: PartialBatchSelector<B>,
     pub report_count: u64,
     pub interval: Interval,
     pub leader_encrypted_agg_share: HpkeCiphertext,
     pub helper_encrypted_agg_share: HpkeCiphertext,
+    pub phantom: PhantomData<B>,
```
divergentdave, line 2041: I think we could drop the batch mode generic parameter and phantom data field. (maybe as a separate commit, since this will require a lot of other straightforward changes) This isn't really used to enforce any invariants, and I think it's unlikely at this point we'll be adding another field that would require the same generic parameter. Getting rid of the generic parameter will slightly shrink the amount of monomorphized code we emit, but I think the effect of this struct would be pretty minor in that regard.

collector/src/lib.rs
```diff
@@ -291,21 +292,16 @@ pub struct Collection<T, B>
 where
     B: BatchMode,
 {
-    partial_batch_selector: PartialBatchSelector<B>,
     report_count: u64,
     interval: (DateTime<Utc>, Duration),
     aggregate_result: T,
+    phantom: PhantomData<B>,
```
divergentdave, line 298: Same here, we no longer need the batch mode generic parameter since we don't have batch mode-specific AAD reconstruction logic anymore.

messages/src/lib.rs
jcjones: Fair. It occurred to me, but there's a lot. I'll open a second PR stacked on this one with that change.

divergentdave (approved):

tgeoghegan (approved):

PR #4819, opened 2026-08-20
Drop BatchMode generic parameter from CollectionJobResp and Collection

jcjones:
Follow-on for #4818.


divergentdave (approved):

tgeoghegan (approved):
Very nice, I never liked how we had to pretend to be dealing with `TimeInterval` in order to get to the constant for media type.

PR #4821, opened 2026-08-20
Add detail fields to invalid message errors

divergentdave:
On the main branch, we are better about adding detail fields to invalid message errors than on the release branch. This PR backfills some detail fields where we use the invalid message problem type. I left the `CodecError` case message fairly generic, even though that's the one I ran into most recently, because I wanted to be conservative about reflecting sensitive information in error messages. The other cases are easier to justify because they use `&'static str` values.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -391,58 +391,64 @@ impl<C: Clock> Aggregator<C> {
         task_id_base64: Option<&[u8]>,
     ) -> Result<(Vec<u8>, Option<Signature>), Error> {
         // Retrieve the appropriate HPKE config list.
-        let hpke_config_list =
-            if self.cfg.taskprov_config.enabled || self.cfg.require_global_hpke_keys {
-                // If we're running in taskprov mode or requiring global keys, unconditionally
-                // provide the global keys and ignore the task_id parameter.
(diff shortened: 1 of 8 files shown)


aggregator/src/aggregator.rs
tgeoghegan, line 1: Is there any real change here, or is this just `rustfmt` making different layout decisions?

tgeoghegan (approved):
The error changes look good, but I'm not sure we need what look like formatting changes in `aggregator/src/aggregator/aggregator.rs`.

aggregator/src/aggregator.rs
divergentdave: Yeah, this makes a couple string literals longer. I tried manually undoing the formatting changes in my first push, but got punished by rustfmt. See https://github.com/divviup/janus/commit/5d8f8e09207393773c2820639fd052cf916080b7. This hunk looks better with "hide whitespace" turned on. The comments in the next match block are still over the 100 character line limit, so that could explain why rustfmt got stuck, but I don't really know how my change got it unstuck. I can try re-wrapping that comment in a separate commit.

tgeoghegan: No, it's not worth the extra effort to fight the formatter. Let's just go ahead with it as is.

PR #4823, opened 2026-08-24
Record size of streaming request/response bodies

### LINKED ISSUE #4822: Fix http.server.request.duration and http.server.response.duration with Transport-Encoding: chunked
Currently when we handle an HTTP request, we check the request and response headers for a Content-Length header, and use that to populate the `http.server.request.duration` and `http.server.response.duration` histograms. If there is no such header, we use 0 instead. This is fine for requests without a body, but it will get requests and responses using Transport-Encoding: chunked wrong. A nearby comment suggests the possibility of switching to `opentelemetry-instrumentation-tower`, but it looks like this has the same limitation. We may be able to address this by wrapping request and response bodies in our own type, but I haven't looked at what the constraints on the body type are.

divergentdave:
Closes #4822. This changes our metrics middleware to wrap request and response bodies, in order to accurately measure request and response sizes when a Content-Length header is not available.


aggregator_core/src/http_server.rs
```diff
@@ -50,13 +58,13 @@ impl HttpMetrics {
                 .with_boundaries(TIME_HISTOGRAM_BOUNDARIES.to_vec())
                 .build(),
             request_body_size: meter
-                .f64_histogram("http.server.request.body_size")
+                .u64_histogram("http.server.request.body_size")
```
jcjones: ............. why was this ever a float? lol

jcjones (approved):
fun.

aggregator_core/src/http_server.rs
divergentdave: Also, I just noticed there's a slight inconsistency in the metric name versus https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpserverrequestbodysize

aggregator_core/src/http_server.rs
```diff
@@ -141,16 +156,90 @@ pub async fn http_metrics_middleware(
...
+            histogram,
+            route,
+        }
+    }
+
+    /// Record an observation of the body size to a histogram metric.
+    ///
+    /// This method keeps track of whether an observation has been recorded, so it is safe to call
+    /// multiple times.
+    fn record_metric(&mut self) {
```
tgeoghegan, line 192: Instead of tracking a state in a boolean, would it be possible for this method to take `self` by value? Perhaps that's impossible given how `http_body::Body` works?

tgeoghegan (approved):

aggregator_core/src/http_server.rs
divergentdave: Both call sites only have mutable access to `self`. I call this from both `poll_frame()` and from `drop()` to guard against the possibility that whatever is consuming the stream calls `is_end_stream()`, gets true, and drops the body without calling `poll_frame()` a final time to get `None`. Plus, it's also possible the consumer could call `poll_frame()` again after getting `None`.

PR #4824, opened 2026-08-25
[DAP-18] Replace PartialBatchSelector with aggregation job extensions

### LINKED ISSUE #4784: Replace `PartialBatchSelector` with aggregation job extensions
https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/762

`AggregationJobInitializeReq` (`messages/src/lib.rs`) needs `AggregationJobExtension extensions<0..2^16-1>` in place of the partial batch
selector. 

Requires a new extension type and registry, a `leader_selected_batch_id` extension carrying the `BatchID`, strictly-increasing ordering, and Helper handling: 

```
unknown type --> fail job unsupportedExtension; out of order -->  invalidMessage
leader-selected batch mode with the extension absent --> invalidMessage.
```

jcjones:
Fixes #4784

Per [DAP PR #762](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/762), replaces the `PartialBatchSelector` in `AggregationJobInitializeReq` with an `AggregationJobExtension` list.

- New `AggregationJobExtension` + `AggregationJobExtensionType` registry (n.b. `leader_selected_batch_id` = 0x0001).
- Leader-selected batch ID now travels as an extension; the message is no longer generic over batch mode.
- Helper enforcement: unknown type becomes `unsupportedExtension`; out-of-order or missing/malformed batch-id becomes `invalidMessage`.

### DIFF
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -2318,15 +2318,11 @@ impl VdafOps {
             }
         }
 
-        // Recover the partial batch identifier from the extensions. In leader-selected batch mode
-        // this requires the `leader_selected_batch_id` extension; its absence is `invalidMessage`.
+        // Recover the partial batch identifier from the extensions. Leader-selected batch mode
+        // requires the `leader_selected_batch_id` extension; time-interval mode forbids it.
         let partial_batch_identifier =
-            B::partial_batch_identifier_from_extensions(req.extensions()).ok_or_else(|| {
-                Error::InvalidMessage(
-                    Some(*task.id()),
-                    "aggregation job is missing a required batch identifier extension",
-                )
-            })?;
+            B::partial_batch_identifier_from_extensions(req.extensions())
+                .map_err(|detail| Error::InvalidMessage(Some(*task.id()), detail))?;
 
         // Build initial aggregation job & report aggregations.
         let client_timestamp_interval = req
--- a/aggregator/src/aggregator.rs
+++ b/aggregator/src/aggregator.rs
@@ -52,11 +52,12 @@ use janus_core::{
 };
 use janus_messages::{
     AggregateShare, AggregateShareAad, AggregateShareId, AggregateShareReq,
-    AggregationJobContinueReq, AggregationJobId, AggregationJobInitializeReq, AggregationJobResp,
-    AggregationJobStep, CollectionJobExtension, CollectionJobId, CollectionJobReq,
-    CollectionJobResp, Duration, ExtensionType, HpkeConfig, HpkeConfigList, InputShareAad,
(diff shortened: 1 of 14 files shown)


aggregator/src/aggregator/http_handlers/tests/aggregation_job_init.rs
```diff
@@ -69,12 +70,7 @@ async fn aggregate_leader() {
         .await
         .unwrap();
 
-    let request = AggregationJobInitializeReq::new(
-        0,
-        Vec::new(),
-        PartialBatchSelector::new_time_interval(),
-        Vec::new(),
-    );
+    let request = AggregationJobInitializeReq::new(0, Vec::new(), Vec::new(), Vec::new());
```
tgeoghegan, line 73: It's kind of unfortunate to have all this `Vec::new(), Vec::new(), Vec::new()`. It's kinda like a C API that takes 11 arguments and 8 of them are NULL. But I don't think we could do better without defining some type like `AggregationJobExtensionList` that would just be an alias for `Vec<Extension>` that's more annoying to use. So I think we should leave this as is.

tgeoghegan (approved):

aggregator/src/aggregator/aggregation_job_init.rs
```diff
@@ -1486,41 +1486,60 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn aggregation_job_init_wrong_query() {
```
divergentdave, line 1489: I think it would still make sense to have a test that sets up a time interval test and tries sending an aggregation job with the leader selected batch ID extension.

aggregator/src/aggregator.rs
```diff
@@ -2300,6 +2298,36 @@ impl VdafOps {
...
+        for extension in req.extensions() {
+            if !extension.extension_type().is_supported() {
+                return Err(Error::UnsupportedExtension(
+                    Some(*task.id()),
+                    "unrecognized aggregation job extension type",
+                ));
+            }
+        }
+
+        // Recover the partial batch identifier from the extensions. In leader-selected batch mode
```
divergentdave: This is missing a check for an extra `leader_selected_batch_id` extension when we don't need it because the task uses the time interval batch mode.

jcjones: In 0496647.

aggregator/src/aggregator/aggregation_job_init.rs
jcjones: Added in 0496647.

divergentdave (approved):

Issue #3779, opened 2025-04-15
Better ordering of `HpkeConfigList`


tgeoghegan:
DAP says that the list served from `/hpke_config` is an "HpkeConfigList structure contains one or more HpkeConfig structures in decreasing order of preference" ([1]). Right now, we have no logic for deciding configuration preference. At a minimum it seems like we should prefer newer HPKE configs over older ones, on the premise that we're trying to get clients to move forward. We could also use a mechanism to indicate if a particular algorithm set is preferred.

[1]: https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-4.7.1

divergentdave:
I think both sorting by creation time and sorting by a custom algorithm order make sense. We could update the SQL to sort by creation time, and then use `sort_by_key()` in the application to sort preferred algorithms first, while otherwise preserving the creation time order.

PR #4847, opened 2026-09-11
Update hpke-dispatch, enable use of X-Wing

divergentdave:
This updates hpke-dispatch to v0.9.0. I added support for X-Wing and SHA-3 KDFs throughout the application as well.


cjpatton:
LFG ⚛️

tgeoghegan (approved):

tgeoghegan:
One question here: which HPKE suite will Janus present first, by default, from `/hpke_config`? DAP says that the first one listed is the server's preferred config. It seems like we should list XWing first for PQ safety, though we'll have to hope that clients that aren't yet ready for hybrid have implemented HPKE config list traversal and don't just take the 0th element.

divergentdave:
It looks like we are currently fetching and presenting HPKE configs in database order, i.e. an unspecified order.

tgeoghegan:
We have an issue logged for this already: https://github.com/divviup/janus/issues/3779

At a minimum, I think we need configuration options so that deployments can assert the order in which configs are presented. I'm OK with doing that separately from this change but I think we'd want both before we deploy a `release/0.7` Janus that supports PQ HPKE.

PR #4850, opened 2026-09-11
Update hpke-dispatch, enable use of X-Wing (backport)

divergentdave:
This backports #4847.


divergentdave:
This is technically a semantic versioning breaking change, because we're updating the minor version of the `hpke` crate in use, and `hpke::HpkeError` is exposed via the `janus_core::hpke::Error` enum. However, I think it's unlikely anyone is destructuring errors this deep, and adding support for hybrid KEMs seems to be worth the breakage.

tgeoghegan (approved):
IMO occasional SemVer breakage is a good idea, to remind people that life is unfair.

